PR #4, opened 2025-05-06
Lazily initialise KotlinAIAgent

nomisRev:
There were 2 usages of the `cs: CoroutineScope` parameter:
  1. Install features, and call event handlers, we can remove the need for this by lazily initialing the KotlinAIAgent
  2. Running the agent sessions in a new coroutine, this is only done for `init` defined in the base class but we can safely use a `coroutineScope` in the run function.

We can also safely remove `isReady` from `FeatureMessageProcessor` while preserving the correct order of `initialise`, but I'd prefer to do that in a seperate PR.


agents/agents-core/src/commonMain/kotlin/ai/grazie/code/agents/core/AIAgent.kt
```diff
@@ -55,7 +57,7 @@ abstract class AIAgent<TStrategy : AIAgentStrategy<TConfig>, TConfig : AIAgentCo
      *
      * @param prompt The initial input string to start the agent execution process.
      */
-    suspend fun run(prompt: String) {
```
fzhinkin: If the goal is to let sub-classes to run something prior the original `run`, then it's better to keep `run` final and add an empty abstract `beforeRun` and let descendants to override it.

agents/agents-local/src/commonMain/kotlin/ai/grazie/code/agents/local/KotlinAIAgent.kt
```diff
@@ -133,9 +125,23 @@ class KotlinAIAgent(
...
+    private var initialised = false
+    private val initialisedMutex = Mutex()
+
+    override suspend fun run(prompt: String) {
+        if (initialised) return
 
+        initialisedMutex.withLock {
+            if (initialised) return
+
+            FeatureContext(this@KotlinAIAgent).installFeatures()
```
sdubov: There is an issue with this approach - some feature interceptors require initialisation stage (please see `ai.grazie.code.agents.local.features.AIAgentPipeline#interceptAgentCreated`). So, if you move feature install logic in a suspend run, we will loose part of handlers. Features, from it's turn, also need `KotlinAiAgent#AIAgentPipeline` that is constructed inside `KotlinAiAgent`. 

I think we need to decouple several things here. In general, I was thinking to extract features (probably with pipeline) out from an agent. This will allow to prepare them in an outer coroutines scope as needed and pass a list of prepared featured into `KotlinAiAgent` constructor.

nomisRev: I'm not sure I understand the issue by reading at the code you mention 🤔 maybe missing some extra context.

> There is an issue with this approach - some feature interceptors require initialisation stage (please see ai.grazie.code.agents.local.features.AIAgentPipeline#interceptAgentCreated). So, if you move feature install logic in a suspend run, we will loose part of handlers. 

That code path is still respected since `onAgentCreated` is also moved to this lazy initialisation step, and it's still triggered immediately after installing the features. It's not really "onAgentCreated" anymore rather "onAgentInitialised"..

> Features, from it's turn, also need KotlinAiAgent#AIAgentPipeline that is constructed inside KotlinAiAgent.

This is also still respected, no? `AIAgentPipeline` is eagerly initialised and always available before feature installation.
It's still guaranteed to all finish before `run` is ever actually triggered, similarly as we did before with manual synchronisation.

> I think we need to decouple several things here. In general, I was thinking to extract features (probably with pipeline) out from an agent. This will allow to prepare them in an outer coroutines scope as needed and pass a list of prepared featured into KotlinAiAgent constructor.

Not entirely sure what you mean? Would this require a `suspend` step initialisation step before the `KotlinAIAgent` creation? I've experimented with such things in the past, but users found it problematic in usage. Example:

```kotlin
suspend fun KotlinAIAgent(
   ...
): KotlinAIAgent = TODO()

class KotlinAIAgent internal constructor (...)
```

This drastically reduces flexibility, and a lot of frameworks do not allow for dependencies (Spring's `@Bean`, Dagger's `@Provides`, [Koin currently doesn't support it](https://github.com/InsertKoinIO/koin/issues/388), Ktor's new DI module doesn't support it, etc).

nomisRev: For extra context, https://github.com/JetBrains/code-engine/pull/428/files#r2048980038

agents/agents-core/src/commonMain/kotlin/ai/grazie/code/agents/core/AIAgent.kt
sdubov: +1 to this comment. I would also keep `run()` API final, and create a separate one, `beforeRun()`, `prepare()`, etc. that can be overridden.

agents/agents-local/src/commonMain/kotlin/ai/grazie/code/agents/local/KotlinAIAgent.kt
sdubov: > That code path is still respected since onAgentCreated is also moved to this lazy initialisation step, and it's still triggered immediately after installing the features. It's not really "onAgentCreated" anymore rather "onAgentInitialised"..

I see. As you mentioned it is no longer an event that describe agent creation in that case :) There are still might be some cases when you want to trace this scenarios, e.g., to draw a graph. You even might not want to call `run()`, but you still want to get the event. With this approach I would rather delete the `onAgentCreate()` event rather than firing it in an unrelated stage since I feel it might produce some misunderstanding for users.

> This is also still respected, no? AIAgentPipeline is eagerly initialised and always available before feature installation.

That's true. Everything besides agents creation can be fetched inside out from the `run()` method.

> Not entirely sure what you mean? Would this require a suspend step initialisation step before the KotlinAIAgent creation? I've experimented with such things in the past, but users found it problematic in usage. Example:

I think (and the main point from [the issue](https://youtrack.jetbrains.com/issue/JBAI-13625/Make-feature-install-non-suspend-and-get-rid-of-cs.launch...-in-init)) is that the initialisation should not be suspend.
We might need to prepare "agent features" outside of the `KotlinAiAgent` init logic. Example: you want to launch a server for a feature to broadcast events from an agent execution. We can change the Feature + Agent API" initialisation logic from: "init everything from KotlinAiAgent constructor" -> to: "prepare features under suspend context if needed before we create an agent, and pass a list of prepared features to the agent". Wdyt?

nomisRev: > I think (and the main point from [the issue](https://youtrack.jetbrains.com/issue/JBAI-13625/Make-feature-install-non-suspend-and-get-rid-of-cs.launch...-in-init)) is that the initialisation should not be suspend.

Yes, making the initialisation non-suspend would make things much easier and less error-prone. Unless I misunderstand your comments here, and from the YT issue. I think there is more of a middle-ground, and how Ktor implements these things.

> to: "prepare features under suspend context if needed before we create an agent, and pass a list of prepared features to the agent".

In practice you'll see that it's almost never necessary, at least that's what Ktor usage has shown. Looking at the current usage it's only used for Grazie's FileSystem API, but kotlinx-io doesn't require suspension. Neither does Ktor's EmbeddedEngine API for starting the server.

> With this approach I would rather delete the onAgentCreate() event rather than firing it in an unrelated stage since I feel it might produce some misunderstanding for users.

If we drop `suspend` from feature installation, I would suggest to also just drop `suspend` here to support cases like drawing a graph like you say.

sdubov: >Looking at the current usage it's only used for Grazie's FileSystem API, but kotlinx-io doesn't require suspension. Neither does Ktor's EmbeddedEngine API for starting the server.

The main reason for suspend is in the class `ai.grazie.code.agents.local.features.common.writer.FeatureMessageRemoteWriter`. The logic is the following. RemoteWriter need to start a local server (that is used to broadcast agent events via SSE to a client, e.g. IDEs). To launch a server we need a suspend context. Launching a server is a part of feature preparation logic (it is related to a feature, not to an agent). Any feature can implement other "writers" that can require some preparation.

Sorry, not sure how ktor solves such problems. As far as we discussed yesterday, their initialisation is not even suspend. I guess, it is just a very simple fields setup or something like that.

So, the whole general question I have here is - "how do we want to initialise features". From my understanding, we can:
1. Move it out of agent, do it before an agent initialisation, and pass a list of prepared "hot features" into an agent;
2. or, combine feature preparation logic with agent's `run` method, but it sounds a bit "messy" from my perspective.

nomisRev: > To launch a server we need a suspend context.

This is a self imposed requirement by `FeatureMessageServer.startServer`, but it's not actually require by Ktor in the implementation. So the `suspend` modifier can be removed there.

> Any feature can implement other "writers" that can require some preparation.

The other writer `FeatureMessageFileWriter` relies on Grazie File API which requires suspension but refactoring to KotlinX IO would remove that `suspend` requirement as well. This refactoring was already suggested by @fzhinkin unrelated to this.

There is also `MutexLock` usage in writers and in some cases within `onAgentCreated` in a `readStages` block. The Mutex could potentially be replaced with [AtomicFU locks](https://github.com/Kotlin/kotlinx-atomicfu/blob/master/atomicfu/src/commonMain/kotlin/kotlinx/atomicfu/locks/Synchronized.common.kt) which _might_ be moved to Kotlin Std soon-ish (@fzhinkin? [KotlinLang message](https://kotlinlang.slack.com/archives/C0B8Q383C/p1745846969927139?thread_ts=1745842697.888779&cid=C0B8Q383C).

This is a very significant design change though.. and I am not familiar with the existing usage or design of this library. So I'm have no idea of the impact is of removing `suspend` from `onAgentCreated` outside of the changes mentioned above.

Ktor indeed relies on simple fields through the config for plugins that require suspension for initialisation, but out of the 1st party plugins I don't think there is any. It's avoided in general, except for very heavy dependencies which get offloaded into `Application.async`. I've seen some 3rd party plugins rely on custom `CoroutineScope` to do suspending (and async) initialisation, i.e. for Kafka. It looks something like this:

```kotlin
suspend fun remoteConfig(): FeatureConfig = TODO()

fun main() = runBlocking {
  val remoteConfig = remoteConfig()
  val agent = KotlinAIAgent(...) {
     install(SomeFeature) {
        config = remoteConfig
     }
  }
  // rest of program
}
```

> or, combine feature preparation logic with agent's run method, but it sounds a bit "messy" from my perspective.

It's a bit messy but only way to sanely get rid of `CoroutineScope`, but with a non-suspending `onAgentCreated` and non-suspending feature installation I am unsure how to achieve it otherwise.

There is also the option of keeping the `CoroutineScope`, and keeping `launch` in `init`. Ktor also uses `launch` inside `init` in some cases. We can expose 4 options to correctly deal with lifecycle:
  1. Provide a proper `CoroutineScope` through the constructor.
  2. Implicitly pass a `CoroutineScope` through a `fun CoroutineScope.KotlinAIAgent(...)` smart-constructor.
  3. Pass a `Job` to a constructor overload.
  4. Instantiate the `KotlinAIAgent` **without** a `CoroutineScope` or `Job` and explicitly call `close()` or `use`.
  
  
  If we keep the `fun CoroutineScope.init` change from this PR, then only suspend initialisation runs on this scope. So if a user ever calls `run` then at that point the scope will already completed, and they don't even need to properly pass a correct `CoroutineScope`..

agents/agents-core/src/commonMain/kotlin/ai/grazie/code/agents/core/AIAgent.kt
nomisRev: Yes, great suggestion!

nomisRev: No longer applicable.

agents/agents-local/src/commonMain/kotlin/ai/grazie/code/agents/local/KotlinAIAgent.kt
```diff
@@ -85,7 +85,7 @@ class KotlinAIAgent(
                 pipeline
             )
         session = activeSession
-        cs.launch(Dispatchers.SuitableForIO) {
```
sdubov: Should not we delete `cs` parameter from `KotlinAiAgent` as well in this commit?

PR #6, opened 2025-05-06
Clean up all examples

tiginamaria:
- Remove composio (as it was python-based), rider/fleet agents like wizard and env setup
- Remove tool-registry module (as we probably don't need it anymore after removing production agents)
- Add examples of agents where missed


build.gradle.kts
```diff
@@ -60,10 +60,10 @@ allprojects {
...
 
 subprojects {
-    tasks.withType<Jar> {
-        val automaticModuleName = archiveBaseName.map { it.replace("-", ".") + ".jvm" }
-        manifest.attributes("Automatic-Module-Name" to automaticModuleName)
-    }
+//    tasks.withType<Jar> {
+//        val automaticModuleName = archiveBaseName.map { it.replace("-", ".") + ".jvm" }
+//        manifest.attributes("Automatic-Module-Name" to automaticModuleName)
+//    }
```
sproshev: it was added for `jpms` module in CE and no longer needed here

tiginamaria: I'll remove it then

PR #21, opened 2025-05-09
JBAI-13675. Extract examples into a top-level module

sdubov:
Main changes in commits:

- Extract `examples` into a top-level  module;
- Refactor tools and move all related classes inside a module;
- Refactor Agent Exception classes.


agents/agents-tools/src/commonMain/kotlin/ai/grazie/code/agents/core/tools/tools/StageTool.kt
```diff
@@ -1,8 +1,8 @@
-package ai.grazie.code.agents.core.tools
+package ai.grazie.code.agents.core.tools.tools
 
-import ai.grazie.code.agents.core.tools.serialization.serializeToolDescriptorsToJsonString
+import ai.grazie.code.agents.core.tools.Tool
 
-class ToolStage(
+class StageTool(
```
Rizzen, line 5: What does it mean? Do we still have stages?

sdubov: I think it was an occasional rename when I moved items between modules. Thank you for catching this. Created a RP with reverting this rename - https://github.com/JetBrains/koan-agents/pull/29

PR #25, opened 2025-05-11
Move LLModel outside of Prompt, also clean up memory feature, and add prompt.withUpdatedParams and prompt.withUpdatedMessages functions

Ololoshechkin:


prompt/prompt-model/src/commonMain/kotlin/ai/jetbrains/code/prompt/dsl/Prompt.kt
```diff
@@ -28,5 +27,35 @@ data class Prompt(
             return PromptBuilder.from(prompt).also(init).build()
         }
     }
-}
 
+    fun withMessages(newMessages: List<Message>): Prompt = copy(messages = newMessages)
+
+    fun withUpdatedMessages(update: MutableList<Message>.() -> Unit): Prompt =
```
EugeneTheDev, line 33: Does it have to be mutable? Seems like it works with just regular list too

prompt/prompt-model/src/commonMain/kotlin/ai/jetbrains/code/prompt/dsl/Prompt.kt
```diff
@@ -28,5 +27,35 @@ data class Prompt(
...
+        this.copy(messages = messages.toMutableList().apply { update() })
+
+    fun withParams(newParams: LLMParams): Prompt = copy(params = newParams)
+
+    class LLMParamsUpdateContext internal constructor(
```
EugeneTheDev, line 38: I see that you are creating a new class with the same fields as in LLMParams, just to make them mutable. I don't have a strong opinion against it, just wanted to confirm, do you think it's necessary? We will now have to maintain two similar sets of fields, and to always keep in mind that we will have to add new fields in two places. Maybe it would be fine to just ask user to provide a new LLMParam object instead of creating this convenience abstraction? If you think it's still worth it, feel free to disregard this comment

agents/agents-core/src/commonMain/kotlin/ai/grazie/code/agents/core/dsl/extension/LocalAgentSubgraphs.kt
```diff
@@ -22,12 +22,12 @@ internal suspend fun LocalAgentStageContext.promptWithTLDR(
...
-                params = params ?: prompt.params,
-            )
+            prompt
+                .withUpdatedMessages { filterNot { it is Message.System } }
+                .withParams(params ?: prompt.params)
         }
+        if (model != null) changeModel(model)
+
         updatePrompt {
             system(systemMessage)
```
EugeneTheDev, line 32: I'm a bit concerned with `system` here. There's only a single system message allowed in the prompt (I already shot myself in the leg with this one). This looks like it can lead to the same problem, since it's not rewriting, but updating the prompt. Or am I missing the point?

agents/agents-core/src/commonMain/kotlin/ai/grazie/code/agents/core/dsl/extension/LocalAgentSubgraphs.kt
```diff
@@ -22,12 +22,12 @@ internal suspend fun LocalAgentStageContext.promptWithTLDR(
...
-            prompt.copy(
-                messages = prompt.messages.filterNot { it is Message.System },
-                model = model ?: prompt.model,
-                params = params ?: prompt.params,
-            )
+            prompt
+                .withUpdatedMessages { filterNot { it is Message.System } }
+                .withParams(params ?: prompt.params)
         }
+        if (model != null) changeModel(model)
```
EugeneTheDev, line 29: This function would change the model and won't restore it back, which is not communicated by the function name. AFAIU it is only an internal utility function for subgraphs, that is used only in one place. So to avoid confusion maybe we can rename it, or, even better, remove it altogether and embed in the usage site?

prompt/prompt-model/src/commonMain/kotlin/ai/jetbrains/code/prompt/dsl/Prompt.kt
Ololoshechkin: dropLast() ?

Ololoshechkin: prompt. withUpdatedMessages { dropLast(2) }

prompt/prompt-model/src/commonMain/kotlin/ai/jetbrains/code/prompt/dsl/Prompt.kt
Ololoshechkin: I wanted to allow writing easily (same as with `prompt. withUpdatedMessages { dropLast(2) }`) :
`prompt. withUpdatedParams { temperature = 100500 }`

agents/agents-core/src/commonMain/kotlin/ai/grazie/code/agents/core/dsl/extension/LocalAgentSubgraphs.kt
Ololoshechkin: I think OpenAI supports it but Anthropic doesn't. But let's address it separately

PR #33, opened 2025-05-12
Cleanup embeddings module

fzhinkin:
- Aligned package names with directories names
- Move tests to commonTest srcset
- Enable explicitApi mode


embeddings/embeddings-llm-remote/src/commonMain/kotlin/ai/jetbrains/embeddings/local/LLMEmbedder.kt
```diff
@@ -1,6 +1,5 @@
 package ai.jetbrains.embeddings.local
```
fzhinkin, line 1: The module is names `embeddings-llm-remote`, but the package is `ai.jetrbrains.embeddings.local`. Should the package be `remote` instead of `local`, @Ololoshechkin ?

Ololoshechkin: Let's also name the module `embeddings-llm` and the package might be embeddings.llm

FYI --  I created a related ticket here:
https://youtrack.jetbrains.com/issue/JBAI-13882/Support-streaming-for-Ollama-and-make-Ollama-inherit-an-LLMClient

PR #36, opened 2025-05-13
Rename [AI|Local]Agent... to just Agent

tiginamaria:
Local should be removed, and to much AI so better also to keep naming things shorter


agents/agents-core/src/commonMain/kotlin/ai/grazie/code/agents/core/agent/entity/AgentStrategy.kt
```diff
@@ -189,7 +189,7 @@ class LocalAgentStrategy(
                     config = config,
                 )
 
-                val context = LocalAgentStageContextImpl(
+                val context = AgentStageContextImpl(
```
EugeneTheDev, line 192: Another suggestion for the next renaming round: let's remove `Impl` from the name, it doesn't look that nice. We can call the base class `...Base` and then this one just `AgentStageContext`

sdubov: `AgentStageContext` is the interface. According to the Java naming convention we need to name it as a very general name and implementations should be named more concrete, e.g. `AgentStageContext` and `<Name>AgentStageContext`. `Base` I would use for an abstract class, but not for an interface. I also do not like `Impl` suffixes though, but it often used in Java.

EugeneTheDev: Valid point, thanks. Can we maybe rename it to `DefaultAgentStageContext`, `SimpleAgentStageContext` or something like this? AFAIK `Impl` is not used in Kotlin libraries that much

tiginamaria: I'll remove AgentStageContext and move/merge with AgentContext so maybe I'll return back with question to you soon

agents/agents-core/src/commonMain/kotlin/ai/grazie/code/agents/core/agent/entity/AgentNode.kt
```diff
@@ -1,25 +1,25 @@
...
+    open fun addEdge(edge: AgentEdge<Output, *>) {
         edges = edges + edge
     }
 
 
-    data class ResolvedEdge(val edge: LocalAgentEdge<*, *>, val output: Any?)
+    data class ResolvedEdge(val edge: AgentEdge<*, *>, val output: Any?)
     suspend fun resolveEdge(
-        context: LocalAgentStageContext,
+        context: AgentStageContext,
```
Ololoshechkin, line 19: If we are getting rid os stages, let's name it "AgentContext"

agents/agents-core/src/commonMain/kotlin/ai/grazie/code/agents/core/agent/AIAgentTool.kt
```diff
@@ -2,13 +2,13 @@ package ai.grazie.code.agents.core.agent
...
 import ai.grazie.code.agents.core.agent.AIAgentTool.AgentToolResult
-import ai.grazie.code.agents.core.api.AIAgent
+import ai.grazie.code.agents.core.api.BaseAgent
 import ai.grazie.code.agents.core.tools.*
 import kotlinx.serialization.KSerializer
 import kotlinx.serialization.Serializable
 import kotlinx.serialization.json.Json
 
-fun AIAgent.asTool(
+fun BaseAgent.asTool(
```
Ololoshechkin, line 11: Didn't we agree that the class would be `class AIAgent : AIAgentBase`
and `interface AIAgentBase`

??

agents/agents-core/src/commonMain/kotlin/ai/grazie/code/agents/core/agent/Agent.kt
```diff
@@ -66,26 +66,26 @@ open class AIAgentBase(
...
-            feature: KotlinAIAgentFeature<Config, Feature>,
+            feature: AgentFeature<Config, Feature>,
             configure: Config.() -> Unit = {}
         ) {
             agent.install(feature, configure)
         }
     }
 
-    private val pipeline = AIAgentPipeline()
+    private val pipeline = AgentPipeline()
```
Ololoshechkin, line 78: AIAgent pipeline?

agents/agents-core/src/commonMain/kotlin/ai/grazie/code/agents/core/agent/AIAgentTool.kt
tiginamaria: These changes were made before the discussion, so I'll redo the renaming based on what we set

PR #41, opened 2025-05-13
Hide declarations annotated as an Internal API

fzhinkin:


agents/agents-core/src/commonMain/kotlin/ai/grazie/code/agents/core/agent/entity/LocalAgentNode.kt
```diff
@@ -33,13 +33,19 @@ abstract class LocalAgentNode<Input, Output> internal constructor() {
         return null
     }
 
+    /**
+     * @suppress
```
Ololoshechkin, line 37: Why is it needed?

fzhinkin: Internal*Api should not be listed in the API Reference (that's why it's internal, right?) and this annotation hides it.

PR #45, opened 2025-05-14
(JBAI-13895) [agents/prompt] add ad-hoc test to verify multiple system messages in prompt

EugeneTheDev:
Wrote ad-hoc test (marked as disabled since we don't need to run it every time). It works. Decided to leave it here, maybe we can repurpose it as a proper test or use it as a baseline to verify with more models later

### DIFF
--- a/prompt/prompt-executor/prompt-executor-llms/build.gradle.kts
+++ b/prompt/prompt-executor/prompt-executor-llms/build.gradle.kts
@@ -23,6 +23,8 @@ kotlin {
         commonTest {
             dependencies {
                 implementation(kotlin("test"))
+                implementation(project(":prompt:prompt-executor:prompt-executor-clients:prompt-executor-openai-client"))
+                implementation(project(":prompt:prompt-executor:prompt-executor-clients:prompt-executor-anthropic-client"))
                 implementation(libs.kotlinx.coroutines.test)
             }
         }
@@ -30,8 +32,6 @@ kotlin {
             dependencies {
                 implementation(kotlin("test-junit5"))
 
-                implementation(project(":prompt:prompt-executor:prompt-executor-clients:prompt-executor-openai-client"))
-                implementation(project(":prompt:prompt-executor:prompt-executor-clients:prompt-executor-anthropic-client"))
                 implementation(libs.ktor.client.cio)
             }
         }


prompt/prompt-executor/prompt-executor-llms/src/jvmTest/kotlin/ai/jetbrains/code/prompt/executor/llms/MultipleSystemMessagesPromptIntegrationTest.kt
```diff
@@ -0,0 +1,78 @@
...
+    private val anthropicToken = System.getenv("ANTHROPIC_TOKEN")
+
+    @Test
+    fun `test OpenAI model with multiple system messages`() = runBlocking {
+        // Create real OpenAI client with token from environment
+        val openAIClient = OpenAILLMClient(openAIToken)
+
+        // Create MultiLLMPromptExecutor with OpenAI client
+        val executor = MultiLLMPromptExecutor(
+            LLMProvider.OpenAI to openAIClient
```
zarechneva-a, line 32: Let's just add LLMProvider.Anthropic to anthropicClient here and do 2 separate calls below, WDYT?

zarechneva-a: This way, we can omit the `test Anthropic model with multiple system messages` test that almost fully repeats this one.

EugeneTheDev: That was the initial idea, but I though it would be nice to have a separate test per provider, to see which one fails in case of problems

zarechneva-a: Okay, we can leave it as is and add parametrization further, so one test will run twice with different executors.

PR #46, opened 2025-05-14
Fix links to the repo after rename

fzhinkin:


settings.gradle.kts
```diff
@@ -1,3 +1,4 @@
+// TODO: don't forget to rename it
```
fzhinkin: Should the root name be `koog-agents` or just `koog`, WDYT?

Ololoshechkin: The website is gonna be koog.ai . So maybe koog-ai ?

Ololoshechkin: And maybe we'll be adding more things not only related to agents (who knows), so it's a safer name. WDYT?

fzhinkin: Nice idea! Renamed it to `koog-ai`.

PR #47, opened 2025-05-14
Fix small bugs (mainly in subgraphWithTask)

23jura23:
Fixes:
- `subgraphWithTask` component
- Adds result of tool calls after their respective call, not to the end of the prompt
- fix `onToolCall`, as it threw deserialization error in case of `it.tool != tool.name`


prompt/prompt-model/src/commonMain/kotlin/ai/jetbrains/code/prompt/dsl/PromptBuilder.kt
```diff
@@ -200,7 +200,11 @@ public class PromptBuilder internal constructor(private val id: String, private
          * @param result The tool result message to add
          */
         public fun result(result: Message.Tool.Result) {
-            messages.add(result)
+            messages
+                .indexOfLast { it is Message.Tool.Call && it.id == result.id }
+                .let { index -> if (index == -1) null else index }
```
EugeneTheDev: Looks a bit better: `.takeIf { it != -1 }`

PR #56, opened 2025-05-14
JBAI-13625. Delete CoroutineScope from an agent class constructor

sdubov:
PR contains the following changes:

- JBAI-13906. Implement the `Closable` interface for AIAgent (needed for keeping resources until agent is closed)
- JBAI-13625. Get rid of `CoroutineScope` in an agent class constructor
- JBAI-13625. Replace interceptor on an agent created with one before an agent started


agents/agents-core/src/commonMain/kotlin/ai/grazie/code/agents/core/utils/Closeable.kt
```diff
@@ -0,0 +1,26 @@
+package ai.grazie.code.agents.core.utils
+
+/**
+ * Represents a resource or entity that can be closed to release any associated resources.
+ */
+public interface Closeable {
```
fzhinkin: There's identical iface in the `ai.grazie.code.agents.local.features.common.message` package. Does it make sense to replace its usages with this one?

sdubov: Absolutely. I thought that we might need a general package named `agents-common` or `common` or `utils` with all general code. I noticed some duplicate code in tests as well. What do you think?

fzhinkin: Sounds good

sdubov: Please see my last commit. Added the `agents-utils` module.

agents/agents-core/src/commonMain/kotlin/ai/grazie/code/agents/core/feature/AIAgentPipeline.kt
```diff
@@ -111,15 +112,14 @@ public class AIAgentPipeline {
...
-        registeredFeatures.values.flatMap { config -> config.messageProcessor.map { provider -> provider.isReady } }
-            .awaitAll()
+    internal suspend fun prepareFeatures() {
+        registeredFeatures.values.forEach { featureConfig ->
+            featureConfig.messageProcessor.map { processor ->
+                withContext(featurePrepareDispatcher) {
+                    launch { processor.initialize() }
+                }
+            }.joinAll()
+        }
```
Rizzen, line 122: launch(featurePrepareDispatcher) instead of withContext - launch

sdubov: You are not in the coroutine context when you need to launch. So you cannot simply replace one with another. I found another problems with this code - `coroutineContext()` will suspend until your job is finished. So, this code does not do any parallel execution. Fixed it here - https://github.com/JetBrains/koog-agents/pull/60.

PR #59, opened 2025-05-15
Get rid of grazie dependecies

devcrocod:
- remove unused dependencies
- replace the grazie `UUID` to Kotlin `Uuid` implementation
- switch from `LoggerFactory` to `KotlinLogging`
- clean up examples


agents/agents-features/agents-features-common/src/commonMain/kotlin/ai/grazie/code/agents/local/features/common/files/FileSystemProvider.kt
```diff
@@ -0,0 +1,80 @@
...
+
+import kotlinx.io.Sink
+import kotlinx.io.Source
+
+/**
+ * High-level, platform-agnostic gateway to the file-system.
+ *
+ * @param Path concrete path type used by an implementation
+ */
+interface FileSystemProvider<Path> {
```
sproshev: is it a copy of Code Engine interfaces? I don't remember there was a plan to remove dependencies on CE, what is the reason for that?

devcrocod: This is a limited version of what’s available in the code engine. What we really need is just the `FileSystem` with extensions

The problem is that dependencies bring in too much with it. For example, `ai.grazie.model:model-text`. I’d prefer to remove such dependencies, as it would help us better control and manage them. Especially since I see little benefit in using them

buildSrc/src/main/kotlin/ai.kotlin.configuration.gradle.kts
```diff
@@ -19,8 +19,8 @@ extensions.getByType<KotlinProjectExtension>().apply {
     }
 }
 
-val kotlinLanguageVersion = KotlinVersion.KOTLIN_1_9
-val kotlinApiVersion = KotlinVersion.KOTLIN_1_9
+val kotlinLanguageVersion = KotlinVersion.KOTLIN_2_2
```
fzhinkin: Do you know why it was previously restricted to 1.9?
Also, I'd recommend keeping it at 2.1 until the 2.2 is actually released :)

agents/agents-features/agents-features-common/src/commonMain/kotlin/ai/grazie/code/agents/local/features/common/files/FileSystemProvider.kt
sproshev: but these file systems and their integrations are supported by the Code Engine, otherwise we do the same work twice -- (interfaces + implementations) * (CE + Agents framework).

Most of (if not all) Grazie dependencies in the CE are implementations, not api, so they are not exported. CE or Agents framework users should specify them manually if they want to use them for some reason.

buildSrc/src/main/kotlin/ai.kotlin.configuration.gradle.kts
devcrocod: Unfortunately, I don’t know why it was restricted.
Nevertheless, it was necessary to bump the version to use `Uuid`

You’re right, I’ll change it to 2.1

agents/agents-features/agents-features-common/src/commonMain/kotlin/ai/grazie/code/agents/local/features/common/files/FileSystemProvider.kt
devcrocod: You’re right, most of it is just implementation, and it’s code duplication

Here's how I see it:
- We have an `api` dependency, and `FileSystemProvider` is part of the public API. This means users depend and need to use on an entity without documentation or open-source code
- It gives us better control over our dependencies and the build
- The code engine code is still closed — no documentation, no license
- Little parts of the code need to be duplicated

In my opinion, we should eventually get rid of this provider and switch to `kotlinx.io`. However, since that would be an API change, I’d prefer to design and align this solution first, not with this PR

sproshev: Does `kotlinx.io` provide implementations for IntelliJ VFS and Fleet FS? CE does. `FSProvider` was implemented to allow providing AI features into any product.

If there are only little parts of the code that need to be duplicated, it means there are no big issues with using them. Otherwise, we are opening a source code and violate one of the main open source principles by re-implementing stuff instead of reusing it.

sproshev: Based on the issue text https://youtrack.jetbrains.com/issue/JBAI-13869, only Grazie platform dependencies should be dropped

devcrocod: To me, the whole project currently looks like a copy of part of code-engine)

I’ve created a separate issue about this because I see using closed-source code brings risks and problems

In this PR, I’ll revert the changes

agents/agents-features/agents-features-common/build.gradle.kts
```diff
@@ -45,7 +42,7 @@ kotlin {
 
         jvmTest {
             dependencies {
-                implementation("ai.jetbrains.code.files:code-files-jvm:1.0.0-beta.55+0.4.45") {
+                implementation(libs.code.engine.files.jvm.get().toString()) {
```
sproshev: can `get().toString()` be dropped?

agents/agents-features/agents-features-memory/build.gradle.kts
```diff
@@ -39,7 +38,7 @@ kotlin {
 
         jvmTest {
             dependencies {
-                implementation("ai.jetbrains.code.files:code-files-jvm:1.0.0-beta.55+0.4.45") {
+                implementation(libs.code.engine.files.jvm.get().toString()) {
```
sproshev, line 42: `get().toString()`

agents/agents-features/agents-features-trace/build.gradle.kts
```diff
@@ -42,7 +41,7 @@ kotlin {
 
         jvmTest {
             dependencies {
-                implementation("ai.jetbrains.code.files:code-files-jvm:1.0.0-beta.55+0.4.45") {
+                implementation(libs.code.engine.files.jvm.get().toString()) {
```
sproshev: `get().toString()`

examples/src/main/kotlin/ai/grazie/code/agents/example/errors/Documents.kt
```diff
@@ -0,0 +1,43 @@
+package ai.grazie.code.agents.example.errors
+
+import java.nio.file.Path
+import kotlin.io.path.readLines
+import kotlin.io.path.writeLines
+import kotlin.io.path.writeText
+
+class DocumentPosition(val line: Int, val column: Int) : Comparable<DocumentPosition> {
```
sproshev: major: since `examples` are not published, it seems there is no reason to limit CE usage here

agents/agents-features/agents-features-common/src/commonMain/kotlin/ai/grazie/code/agents/local/features/common/files/FileSystemProvider.kt
sproshev: > To me, the whole project currently looks like a copy of part of code-engine)

That's right, it was started in the code engine and is opening now via a separate repository

agents/agents-features/agents-features-common/build.gradle.kts
devcrocod: Unfortunately, no
`libs.code.engine.files.jvm` returns a special object, and to call `exclude` later, need to get a string

examples/src/main/kotlin/ai/grazie/code/agents/example/errors/Documents.kt
devcrocod: But they will be publicly available on GitHub, right?

It uses Fleet code there, so I decided to get rid of CE here

agents/agents-features/agents-features-common/build.gradle.kts
sproshev: I did not know that :( Found the issue https://github.com/gradle/gradle/issues/20274

examples/src/main/kotlin/ai/grazie/code/agents/example/errors/Documents.kt
sproshev: no commitment at the moment, can it be an issue? CE provides jars and sources under Apache 2 license, If I understand correctly, it does not require disclosing the source and does not require using only open-sourced dependencies.

Regarding Fleet usage, do we need to care in the examples?

Ololoshechkin: Let's just drop product-specific examples that are too much related to the code.

devcrocod: okay, then I will drop `memory` and `errors` examples
upd: only `errors`

PR #65, opened 2025-05-15
[JBAI-13882] Ollama client done right

Rizzen:


prompt/prompt-executor/prompt-executor-clients/prompt-executor-ollama-client/build.gradle.kts
```diff
@@ -1,54 +1,57 @@
-import ai.grazie.gradle.publish.maven.Publishing.publishToGraziePublicMaven
-
-group = "${rootProject.group}.prompt"
-version = rootProject.version
-
```
sproshev, line 5: why? isn't it going to be published?

Rizzen: Does it prevent it from publish?

sproshev: fixed in #75

sproshev: sorry, I made a comment in the wrong place, there is no longer call of `publishToGraziePublicMaven()`

PR #68, opened 2025-05-15
Reduce usages of the Code Engine (JBAI-13966)

sproshev:


agents/agents-features/agents-features-common/src/commonMain/kotlin/ai/grazie/code/agents/local/features/common/writer/FeatureMessageFileWriter.kt
```diff
@@ -119,16 +104,6 @@ public abstract class FeatureMessageFileWriter<Path>(
...
 
-    private suspend fun resolvePath(fs: FileSystemProvider.ReadWrite<Path>, root: Path): Path {
-        // Root is an existing file
-        if (fs.exists(root) && fs.isFile(root)) {
-            return root
-        }
-
-        val targetFileName = "agent-trace-${Clock.System.now().toEpochMilliseconds()}.log"
-        return fs.fromRelativeString(root, targetFileName)
-    }
```
devcrocod: losing part of logic here. If the user passes a directory, they will get an exception

agents/agents-features/agents-features-trace/src/commonMain/kotlin/ai/grazie/code/agents/local/features/tracing/writer/TraceFeatureMessageFileWriter.kt
```diff
@@ -41,15 +41,15 @@ import ai.grazie.code.files.model.FileSystemProvider
...
- * @param path The path where trace events will be written
+ * @param targetPath The path where feature messages will be written.
+ * @param sinkOpener Returns a [Sink] for writing to the file, this class manages its lifecycle.
  * @param format Optional custom formatter for trace events
  */
 public class TraceFeatureMessageFileWriter<Path>(
-    fs: FileSystemProvider.ReadWrite<Path>,
-    path: Path,
+    targetPath: Path,
+    sinkOpener: (Path) -> Sink,
```
devcrocod: Passing two lambdas in a row is not the best approach.
From the code, I don’t see `targetPath` being used elsewhere.
I suggest simply accepting a sink object in this case

sproshev: I do agree, but at the same time I wanted to make as few changes as possible

agents/agents-features/agents-features-common/src/commonMain/kotlin/ai/grazie/code/agents/local/features/common/writer/FeatureMessageFileWriter.kt
sproshev: yep, we discussed it with @sdubov. To drop FS usage here, we decided to accept a target file as a parameter and specify this update in the docs

PR #70, opened 2025-05-16
JBAI-13989. Extract agents, strategies and tools impls into agents-ext module

sdubov:


agents/agents-ext/src/commonMain/kotlin/ai/grazie/code/agents/ext/tool/AskUser.kt
```diff
@@ -1,4 +1,4 @@
-package ai.grazie.code.agents.core.tools.tools
+package ai.grazie.code.agents.ext.tool
```
EugeneTheDev, line 1: How about we rename these to `StdAskUser`, `StdSayToUser` to better communicate the idea that they use stdin/stdout. And also not to claim more generic `AskUser` `SayToUser` names, because for a more advanced usecases users would probably implement their own tools like this

sdubov: I like the initial naming. It is just a set of tools, it can have any names. Why would we think it is some Std tool?

EugeneTheDev: Because this is a tool with default implemenation, which interacts with std, which I want to make more clear. Since in a lot of cases we might not have an easy access to std (server, mobile device). So I want to state it in the name clearly to avoid confusing users

agents/agents-ext/build.gradle.kts
```diff
@@ -0,0 +1,44 @@
...
+kotlin {
+    sourceSets {
+        commonMain {
+            dependencies {
+                api(project(":agents:agents-core"))
+                api(project(":agents:agents-features:agents-features-common"))
+                api(project(":agents:agents-tools"))
+                api(project(":agents:agents-utils"))
+
+                implementation(project(":prompt:prompt-executor:prompt-executor-clients:prompt-executor-openai-client"))
```
devcrocod: I’m curious why we use OpenAI as the default instead of Anthropic?
Also, if it’s the default, shouldn’t it be api?

sdubov: Good catch. It should have no default implementation. Updated.

PR #78, opened 2025-05-16
Add API to add tool in existing ToolRegistry instance

sdubov:


agents/agents-tools/src/commonMain/kotlin/ai/grazie/code/agents/core/tools/ToolRegistry.kt
```diff
@@ -61,12 +64,19 @@ public class ToolRegistry private constructor(
...
     }
 
-    public infix fun with(toolRegistry: ToolRegistry): ToolRegistry {
+    public operator fun plus(toolRegistry: ToolRegistry): ToolRegistry {
         val mergedTools = (this.tools + toolRegistry.tools).distinctBy { it.name }
         return ToolRegistry(mergedTools)
     }
 
-    public operator fun plus(toolRegistry: ToolRegistry): ToolRegistry = this with toolRegistry
+    public fun add(tool: Tool<*, *>) {
```
EugeneTheDev, line 72: So `ToolRegistry` becomes mutable now. I'm not sure if it's OK, since it might introduce potential concurrency problems. When you pass this instance somewhere, e.g. to the agent, and in the process change it by appending new tools from the outside. So the behavior of the agent might change in the process too. TBH I like immutable data structures in these cases more. Just like `plus` operator does now, just returning a new ToolRegistry

Ololoshechkin: We can consider having a MutableToolRegistry/ToolRegistry (same as MutableList/List and other collections). WDYT?
At least, then it would make it a conscious decision + we can analyze how many people even use the mutable one + if needed, we can make ConcurrentMutableToolRegistryWithFareworks if we see any demand.

sdubov: > So ToolRegistry becomes mutable now. I'm not sure if it's OK, since it might introduce potential concurrency problems. When you pass this instance somewhere, e.g. to the agent, and in the process change it by appending new tools from the outside.

Correct. Well, it depends on what do you want to achieve with immutability of this class. According to our current architecture, you can change tools within an agent execution how you want as you can mutate tools inside a context. For example, you can do this:
```
val node = node<String, Unit>("my node") { input ->
            llm.writeSession {
                tools = // my new set of tools from outside
            }
        }
```

I do not found a reason to have this instance immutable at the moment. Currently, we use this only before we create an agent and pass it as an argument.

From the other side, to support immutability we now need to use `plus` operator and append tools to the list, spawning the `ToolRegistry` instances. Good example for me is `ai.grazie.code.agents.mcp.McpToolRegistryProvider#fromClient` that returns a `ToolRegistry` at the moment (instead of a list of tools collected from an MCP server, which would be more clear for me).

I would consider the `ToolRegistry` as a container of tool descriptions. So, I would expect the API allows me to append tools into it. You can provide immutability on a caller side when this list of tools is used.

@EugeneTheDev, what do you think?

EugeneTheDev: > According to our current architecture, you can change tools within an agent execution how you want as you can mutate tools inside a context

That's true, but:
1. There's at least some kind of a safeguard (rw lock in read/write sessions)
2. I don't like this that much either :)

But that's probably just me nitpicking, and having a general dislike towards using mutable data structures without a strong need for that (Scala and FP corrupted me, I guess). And I also like that immutable data structures protect you from some errors (when you pass the data somewhere, then modify it in some other place, so that the behavior changes = less predictable and straightforward flow of execution, higher chance of some strange errors) and also gives you concurrence safety "for free" (almost).

And spawning more entities (mutable tool registry, non mutable tool registry, concurrent safety tool registry, etc.) would also increase overhead with maintance and make it more confusing for the end user. 

But well, I don't want to hold everyone back, so you can disregard my comments if you think it would be better and simpler for now

sdubov: > And spawning more entities (mutable tool registry, non mutable tool registry, concurrent safety tool registry, etc.) would also increase overhead with maintance and make it more confusing for the end user.

I would +1 to this comment. I agree, and I think it can be fixed by providing an immutable context wherever it is needed. As simpler API we have is better. 

> But well, I don't want to hold everyone back, so you can disregard my comments if you think it would be better and simpler for now

I really like that you touches those topics! Thank you a lot for that. Let me highlight my point again :) currently, the class is immutable, but the code implementation allows to mutate tools while running the agent. Unless I miss something.

sdubov: >At least, then it would make it a conscious decision + we can analyze how many people even use the mutable one + if needed, we can make ConcurrentMutableToolRegistryWithFareworks if we see any demand.

Agree.

> We can consider having a MutableToolRegistry/ToolRegistry (same as MutableList/List and other collections). WDYT?

As I mentioned above. I would not introduce a separate entities here, but rather use an immutable entities wherever it is needed. 

As for now, as a user, I'm just wondering why cannot I append tools to my `ToolRegistry`.

EugeneTheDev: OK, let's make mutable and immutable versions if you'd like (I guess mutable would extend immutable?). Also, to guarantee thread-safety, let's make `add` suspendable and wrap it in `mutex.withLock`?. I was suspecting that this might cause problems even though it is not suspendable (threads still exist), and Claude verified my concerns. BTW, one possible mitigation it suggested is exactly what are we doing now - just returning a new copy (then no `suspend` and mutex needed, which might also be nice)

Rizzen: I heavily double @EugeneTheDev in favor of immutability - it provides simplicity, thread safety and I don't see a lot of overhead here, I don't see any points in favor to have mutable version.

Moreover, it would be just confusing for user to choose between mutable and immutable version when diving deeper and would require to read actual implementation.

PR #79, opened 2025-05-17
Rewrite memory example to customer support agent

Ololoshechkin:


examples/env.template.properties
```diff
@@ -4,11 +4,6 @@
 # Or, if you want to integrate with 1password, you can put reference, e.g. op://foo/bar:
 # FOO=op://foo/bar
 
-IDEFORMER_VERSION=
-GRAZIE_TOKEN=
-# prod or staging
-GRAZIE_ENVIRONMENT=
+OPEN_AI_TOKEN=
```
devcrocod: Please use the standard naming conventions for the tokens: `OPENAI_API_KEY` and `ANTHROPIC_API_KEY` respectively.

PR #84, opened 2025-05-18
JBAI-13675. Rename AIAgentConfig implementation to avoid name duplication

sdubov:


agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/model/agent/AIAgentSystemPrompt.kt
```diff
@@ -1,30 +0,0 @@
-package ai.koog.agents.core.model.agent
```
sdubov, line 1: Please note deletion for this file. I have not found any dependencies. It seems like this is an artifacts from Ide former.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/config/AIAgentConfig.kt
```diff
@@ -1,45 +1,18 @@
...
-            return AIAgentConfig(
-                prompt = prompt(id) {
-                    system(prompt)
-                },
-                model = llm,
-                maxAgentIterations = maxAgentIterations
-            )
-        }
-    }
+public interface AIAgentConfig {
```
tiginamaria: Maybe continue our style and name it `AIAgentConfigBase` and our main impl as `AIAgentConfig`? I see that it's not abstract class, but I just like the way all classes that are related to the api are named as AIAgentBlaBla

sdubov: Updated.

PR #92, opened 2025-05-19
(JBAI-13980 & JBAI-13981) [prompts/clients] [agents/agents-ext] Fix bugs

EugeneTheDev:
1. Turns out Anthropic sometimes can send empty content. Return emtpy assistant message in this case. Also added `finishReason` to `Message.Assistant` class for better clarity
2. Finish incorrect exit condition in `chatAgentStrategy` that led to infinite loops

### DIFF
--- a/prompt/prompt-executor/prompt-executor-clients/prompt-executor-openrouter-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openrouter/OpenRouterLLMClient.kt
+++ b/prompt/prompt-executor/prompt-executor-clients/prompt-executor-openrouter-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openrouter/OpenRouterLLMClient.kt
@@ -330,7 +330,9 @@ public class OpenRouterLLMClient(
             error("Empty choices in OpenRouter response")
         }
 
-        val (choice, message) = response.choices.first().let { it to it.message }
+        val (choice, message) = response.choices
+            .firstOrNull()
+            ?.let { it to it.message } ?: throw IllegalStateException("No choice found in OpenRouter response")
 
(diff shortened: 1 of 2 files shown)


prompt/prompt-executor/prompt-executor-clients/prompt-executor-openrouter-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openrouter/OpenRouterLLMClient.kt
```diff
@@ -330,7 +330,7 @@ public class OpenRouterLLMClient(
             error("Empty choices in OpenRouter response")
         }
 
-        val message = response.choices.first().message
+        val (choice, message) = response.choices.first().let { it to it.message }
```
Rizzen: I would use firstOrNull because nobody guarantees that choices list would be non-empty

EugeneTheDev: I guess API kinda guarantees it here and in other places. And if the choices list would in some case be empty, that should be an error anyway, since we won't be able to proceed. Do you think I should add a better exception here?

EugeneTheDev: Added more explicit exception on empty choices here and in OpenAI (the API is the same in these clients)

PR #116, opened 2025-05-21
Dokka publish manual + do not publish tests

Rizzen:


publish-docs-manual.sh
```diff
@@ -0,0 +1,20 @@
...
+# Build docs
+./gradlew dokkaGenerate
+
+# Clone or update docs repo
+if [ ! -d "$DOCS_REPO" ]; then
+  git clone git@github.com:JetBrains/koog-api-docs.git "$DOCS_REPO"
+fi
+
+cd "$DOCS_REPO"
+cp -r ../koog-agents/build/dokka/html/* .
```
fzhinkin, line 15: Does it make sense to save the CWD before changing it to `DOCS_REPO`? The repo changed its name twice, so it could be virtually anything, not just `kong-agents` :)

```
AGENTS_ROOT=`pwd`
cd "$DOCS_REPO"
cp -r "$AGENTS_ROOT/build/dokka/html/*" .
```

Rizzen: Actually this is just a script for manual deploy dokka doc, I'll remove it when we'll setup GH Action for that

PR #150, opened 2025-05-23
Setup dependencies resolution for Dokka

### LINKED ISSUE #146: API docs don't have cross-module links
The current Dokka configuration does not have cross-module symbol links set up.

For instance, when on the OllamaClient API docs (https://api.koog.ai/prompt/prompt-executor/prompt-executor-clients/prompt-executor-ollama-client/ai.koog.prompt.executor.ollama.client/-ollama-client/index.html), there is no link on the LLMClient or LLMEmbeddingProvider symbols.

fzhinkin:
- Added a missing module to Dokka dependencies
- Included links to external documentation

Closes #146

---

#### Type of the change
- [ ] New feature
- [ ] Bug fix
- [x] Documentation fix

- [ ] ~~Tests for the changes have been added~~
- [ ] ~~All new and existing tests passed~~

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -168,6 +168,7 @@ dependencies {
     dokka(project(":prompt:prompt-cache:prompt-cache-model"))
     dokka(project(":prompt:prompt-cache:prompt-cache-redis"))
     dokka(project(":prompt:prompt-executor:prompt-executor-cached"))
+    dokka(project(":prompt:prompt-executor:prompt-executor-clients"))
     dokka(project(":prompt:prompt-executor:prompt-executor-clients:prompt-executor-anthropic-client"))
     dokka(project(":prompt:prompt-executor:prompt-executor-clients:prompt-executor-google-client"))
     dokka(project(":prompt:prompt-executor:prompt-executor-clients:prompt-executor-openai-client"))
--- a/buildSrc/src/main/kotlin/ai.kotlin.dokka.gradle.kts
+++ b/buildSrc/src/main/kotlin/ai.kotlin.dokka.gradle.kts
@@ -15,5 +15,20 @@ dokka {
             remoteUrl("https://github.com/JetBrains/koog/tree/main")
             remoteLineSuffix = "#L"
         }
+
+        externalDocumentationLinks.register("ktor-client") {
+            url("https://api.ktor.io/ktor-client/")
+            packageListUrl("https://api.ktor.io/package-list")
+        }
+
+        externalDocumentationLinks.register("kotlinx-coroutines") {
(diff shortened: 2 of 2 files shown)


aozherelyeva (approved):

build.gradle.kts
```diff
@@ -168,6 +168,7 @@ dependencies {
     dokka(project(":prompt:prompt-cache:prompt-cache-model"))
     dokka(project(":prompt:prompt-cache:prompt-cache-redis"))
     dokka(project(":prompt:prompt-executor:prompt-executor-cached"))
+    dokka(project(":prompt:prompt-executor:prompt-executor-clients"))
```
whyoleg, line 171: JIC: You can do the aggregation setup semi-automatically if you know the projects that should not be included.
Something like this:
```kotlin
// TBH it's hard to come up with a good name for this property :)
val projectsWithoutDokka = setOf(
  ":examples",
  ":integration-tests",
  ":koog-agents"
)
dependencies {
  subprojects.forEach {
    if(it.path !in projectsWithoutDokka) dokka(it)
  }
}
```

And it should still support  Gradle Configuration Cache and Isolated Projects - at least at the current moment.

buildSrc/src/main/kotlin/ai.kotlin.dokka.gradle.kts
```diff
@@ -15,5 +15,20 @@ dokka {
             remoteUrl("https://github.com/JetBrains/koog/tree/main")
             remoteLineSuffix = "#L"
         }
+
+        externalDocumentationLinks.register("ktor-client") {
+            url("https://api.ktor.io/ktor-client/")
+            packageListUrl("https://api.ktor.io/package-list")
```
whyoleg, line 21: `url("https://api.ktor.io/")` should most likely be enough, as `packageListUrl` will be automatically calculated by Dokka - and ktor has it in the default location (and it will also work for both client/server, as `koog` depends also on `ktor-server`)

Same for coroutines and serialization - just `url` block will be enough

whyoleg (commented):
just passing by :)

buildSrc/src/main/kotlin/ai.kotlin.dokka.gradle.kts
fzhinkin: Ktor stores the package list in the API docs root, but the actual reference documentation for client and server is located inside the `ktor-client` and `ktor-server` sub-paths and those are not referred from the package list, and these sub paths don't contain package lists either. So it does not work well. :(

While `koog` depends on the `ktor-server`, its symbols are not referred directly from the API reference, so I though that it's a problem for future me. :)

PR #174, opened 2025-05-27
Enhance test prompts for Ollama.

aozherelyeva:
Add a one-shot prompt for Ollama tests.

---

#### Type of the change
- [ ] New feature
- [x] Bug fix
- [ ] Documentation fix

### DIFF
--- a/integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/OllamaAgentIntegrationTest.kt
+++ b/integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/OllamaAgentIntegrationTest.kt
@@ -35,88 +35,90 @@ class OllamaAgentIntegrationTest {
         private val model get() = fixture.model
     }
 
-    private fun createTestStrategy() =
-        strategy("test-ollama") {
-            val askCapitalSubgraph by subgraph<String, String>("ask-capital") {
-                val definePrompt by node<Unit, Unit> {
-                    llm.writeSession {
-                        model = OllamaModels.Meta.LLAMA_3_2
-                        rewritePrompt {
-                            prompt("test-ollama") {
-                                system(
-                                    "You are a top-tier geographical assistant. " +
-                                            "ALWAYS communicate to user via tools!!!\n" +
-                                            "ALWAYS use tools you've been provided.\n" +
-                                            "ALWAYS generate valid JSON responses.\n" +
-                                            "ALWAYS call tool correctly, with valid arguments.\n" +
-                                            "NEVER provide tool call in result body.\n" +
-                                            "\n" +
-                                            "Example tool call:\n" +
-                                            "{\n" +
-                                            "id=ollama_tool_call_3743609160,\n" +
-                                            "tool=geography_query_tool,\n" +
-                                            "content={\"query\":\"capital of France\"}\n" +
-                                            "}.".trimIndent()
-                                )
-                            }
+    private fun createTestStrategy() = strategy("test-ollama") {
+        val askCapitalSubgraph by subgraph<String, String>("ask-capital") {
+            val definePrompt by node<Unit, Unit> {
+                llm.writeSession {
+                    model = OllamaModels.Meta.LLAMA_3_2
+                    rewritePrompt {
+                        prompt("test-ollama") {
+                            system(
+                                """
+                                        You are a top-tier geographical assistant. " +
+                                            ALWAYS communicate to user via tools!!!
+                                            ALWAYS use tools you've been provided.
+                                            ALWAYS generate valid JSON responses.
+                                            ALWAYS call tool correctly, with valid arguments.
+                                            NEVER provide tool call in result body.
+                                            
+                                            Example tool call:
+                                            {
+                                                "id"="ollama_tool_call_3743609160",
+                                                "tool"="geography_query_tool",
+                                                "content"={"query":"capital of France"}
+                                            }
+                                            """.trimIndent()
+                            )
                         }
                     }
                 }
-
-                val callLLM by nodeLLMRequest(allowToolCalls = true)
-                val callTool by nodeExecuteTool()
-                val sendToolResult by nodeLLMSendToolResult()
-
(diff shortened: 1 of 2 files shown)


integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/OllamaAgentIntegrationTest.kt
```diff
@@ -44,12 +44,19 @@ class OllamaAgentIntegrationTest {
...
+                                            "ALWAYS communicate to user via tools!!!\n" +
+                                            "ALWAYS use tools you've been provided.\n" +
+                                            "ALWAYS generate valid JSON responses.\n" +
+                                            "ALWAYS call tool correctly, with valid arguments.\n" +
+                                            "NEVER provide tool call in result body.\n" +
+                                            "\n" +
+                                            "Example tool call:\n" +
+                                            "{\n" +
+                                            "id=ollama_tool_call_3743609160,\n" +
+                                            "tool=geography_query_tool,\n" +
```
ptitjes: There are missing quotes, here. Beware that small models (like llama3.1-8B and sometimes llama3.1-70B) have a tendency to forget quotes around the tool names in tool calls. So if the example is wrong too, that might be even worth. :smile:

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/OllamaAgentIntegrationTest.kt
```diff
@@ -76,11 +83,19 @@ class OllamaAgentIntegrationTest {
...
+                                            "ALWAYS communicate to user via tools!!!\n" +
+                                            "ALWAYS use tools you've been provided.\n" +
+                                            "ALWAYS generate valid JSON responses.\n" +
+                                            "ALWAYS call tool correctly, with valid arguments.\n" +
+                                            "NEVER provide tool call in result body.\n" +
+                                            "\n" +
+                                            "Example tool call:\n" +
+                                            "{\n" +
+                                            "id=ollama_tool_call_3743609160,\n" +
+                                            "tool=answer_verification_tool,\n" +
```
ptitjes: The quotes are missing here too.

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/OllamaAgentIntegrationTest.kt
ptitjes: Also the all the property names are not quoted. This is invalid JSON.

zarechneva-a: Thanks! Good point.

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/OllamaAgentIntegrationTest.kt
```diff
@@ -35,89 +35,91 @@ class OllamaAgentIntegrationTest {
...
+                                            ALWAYS use tools you've been provided.
+                                            ALWAYS generate valid JSON responses.
+                                            ALWAYS call tool correctly, with valid arguments.
+                                            NEVER provide tool call in result body.
+                                            
+                                            Example tool call:
+                                            {
+                                                "id"="ollama_tool_call_3743609160",
+                                                "tool"="geography_query_tool",
+                                                "content"={"query":"capital of France"}
```
ptitjes: Sorry again, I just saw you used equal signs instead of colons! :stuck_out_tongue:

Rizzen (approved):

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/OllamaAgentIntegrationTest.kt
```diff
@@ -35,74 +35,91 @@ class OllamaAgentIntegrationTest {
...
+                llm.writeSession {
+                    model = OllamaModels.Meta.LLAMA_3_2
+                    rewritePrompt {
+                        prompt("test-ollama") {
+                            system(
+                                """
+                                        You are a top-tier geographical assistant. " +
+                                            ALWAYS communicate to user via tools!!!
+                                            ALWAYS use tools you've been provided.
+                                            ALWAYS generate valid JSON responses.
```
kpavlov, line 50: Wouldn't be better to incorporate such guardrails into the LLM client itself, so it automatically appends it to the system prompt?
These "typical" guardrails should always be present and it's easy to forget, so the option to add them to the final system prompt  automatically might be useful.

Then the "human written" system prompt will be more focusing on the main problem, rather than on the json syntax and guardrails.

zarechneva-a: @Rizzen could you please take a look? I think adding the instructions to the client itself is a good idea.

ptitjes: Those instructions look very application-specific to me. You might not want "ALWAYS communicate to user via tools!!!" for all applications.

Rizzen: I don't think that such hardcoded annotations would be good idea for library code, because we don't want to introduce unexpected side effects for users and also do want to give them full control on agents behaviour. 

Moreover, such prompts require constant evaluation against all the new models. So here I agree with @ptitjes - these instructions are project-specific and should be introduced by user.

Rizzen: And in addition - in that particular case these instructions was added to solve pretty specific problem - to increase tool calling rate for Ollama, because llama3.2 in that case does not have mechanism to enforce tool calling. So we ended up with such instructions and one-shot prompting. 
We wouldn't need these struggles on 4o, for example.

PR #182, opened 2025-05-28
Extend OpenAIClientSettings with chat completions API path and embeddings API path to make it configurable

matusekma:
Because there can be deployments of OpenAI models outside of the official OpenAI API, e.g. Azure OpenAI, the path of the APIs can vary. This PR makes the chat completions and embeddings api paths configurable.

Issue opened: #183 
Discussion on slack: https://kotlinlang.slack.com/archives/C08SLB97W23/p1748430862443289

---

#### Type of the change
- [x] New feature
- [ ] Bug fix
- [ ] Documentation fix

### DIFF
--- a/prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openai/OpenAILLMClient.kt
+++ b/prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openai/OpenAILLMClient.kt
@@ -34,12 +34,16 @@ import kotlin.uuid.Uuid
 /**
  * Represents the settings for configuring an OpenAI client.
  *
- * @property baseUrl The base URL of the OpenAI API. Defaults to "https://api.openai.com/v1".
+ * @property baseUrl The base URL of the OpenAI API. Defaults to "https://api.openai.com".
  * @property timeoutConfig Configuration for connection timeouts, including request, connect, and socket timeouts.
+ * @property chatCompletionsPath The path of the OpenAI Chat Completions API. Defaults to "v1/chat/completions".
+ * @property embeddingsPath The path of the OpenAI Embeddings API. Defaults to "v1/embeddings".
  */
 public class OpenAIClientSettings(
     public val baseUrl: String = "https://api.openai.com",
-    public val timeoutConfig: ConnectionTimeoutConfig = ConnectionTimeoutConfig()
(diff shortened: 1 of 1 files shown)


prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openai/OpenAILLMClient.kt
```diff
@@ -34,12 +34,16 @@ import kotlin.uuid.Uuid
...
+ * @property baseUrl The base URL of the OpenAI API. Defaults to "https://api.openai.com".
  * @property timeoutConfig Configuration for connection timeouts, including request, connect, and socket timeouts.
+ * @property chatCompletionsPath The path of the OpenAI Chat Completions API. Defaults to "v1/chat/completions".
+ * @property embeddingsPath The path of the OpenAI Embeddings API. Defaults to "v1/embeddings".
  */
 public class OpenAIClientSettings(
     public val baseUrl: String = "https://api.openai.com",
-    public val timeoutConfig: ConnectionTimeoutConfig = ConnectionTimeoutConfig()
+    public val timeoutConfig: ConnectionTimeoutConfig = ConnectionTimeoutConfig(),
+    public val chatCompletionsPath: String = "v1/chat/completions",
```
kpavlov, line 45: Since baseUrl doesn't end with slash '/', shouldn't path always start with a slash: '/v1/chat/completions'? Shouldn't it be a consistent behavior for all connectors?

matusekma: It is a reasonable comment, thanks! I just moved the default constant values from the companion object to the default argument values without modification. Also, I think based on the ktor client documentation, starting the path without a slash is fine: https://ktor.io/docs/client-requests.html#url

Ololoshechkin (approved):
@matusekma Thanks for your PR.
Just checked that we can actually apply this approach to all other clients (Anthropic, OpenRouter, Google, Ollama). It's not a blocker for this MR, though, so we can do it in a separate MR after.

matusekma:
> @matusekma Thanks for your PR. Just checked that we can actually apply this approach to all other clients (Anthropic, OpenRouter, Google, Ollama). It's not a blocker for this MR, though, so we can do it in a separate MR after.

@Ololoshechkin agreed :) I was thinking the same. Should I create an issue for those as well?
Also, could you please approve the workflow run?

PR #184, opened 2025-05-28
JBAI-14351: Add token count and timestamp support to Message.Response, add Tokenizer

Ololoshechkin:

---

#### Type of the change
- [x] New feature
- [ ] Bug fix
- [ ] Documentation fix


EugeneTheDev (changes requested):
Can you please elaborate, why custom tokenizer implementation? AFAIK different models count tokens differently, so we should use these parameters from LLM API response instead if we want to provide accurate information

Ololoshechkin:
@EugeneTheDev tkenizer is relevant and used for OLLama only, because it doesn't return used tokens in response unlike OpenAI/Anthropic/Google. Sure -- different models use different tokenizers, for now I left an interface with just a few simple implementations that :
1. Anyone can extend and implement
2. Those simple tokenizers already provide some rough estimate needed for conditional history compression.
3. We'll add more in the future in separate MRs.

This MR was actually not about tokenizers, but about the tokenCount, and Tokenizer was introduced to cover this for OLLama

ptitjes:
Sorry, I only now understand that having the up-to-now `totalTokenCount` in the response messages would be sufficient enough for a "max-token" summarisation strategy. (That indicator will always be an estimation anyway.)

And indeed using a feature to add a finer-grained token counts using a `Tokenizer` makes total sense!!

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/extension/AIAgentLLMActions.kt
```diff
@@ -11,6 +12,10 @@ public fun AIAgentLLMWriteSession.leaveLastNMessages(n: Int) {
     prompt = prompt.withUpdatedMessages { takeLast(n) }
 }
 
+public fun AIAgentLLMWriteSession.leaveMessagesFromTimestamp(timestamp: Instant) {
```
EugeneTheDev, line 15: nit: shouldn't it be `>=`, because `From` sounds like it is inclusive? Or, alternatively, let's rename it to `After`.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/extension/HistoryCompressionStrategies.kt
```diff
@@ -11,7 +12,9 @@ import ai.koog.prompt.message.Message
  * Example implementations:
  * - [HistoryCompressionStrategy.WholeHistory]
  * - [HistoryCompressionStrategy.FromLastNMessages]
+ * - [HistoryCompressionStrategy.FromTimestamp]
  * - [HistoryCompressionStrategy.Chunked]
+ * - [ai.koog.agents.memory.feature.history.RetrieveFactsFromHistory]
```
EugeneTheDev, line 17: Unresolved reference, let's remove

agents/agents-features/agents-features-tokenizer/src/commonMain/kotlin/ai/koog/agents/features/tokenizer/feature/MessageTokenizer.kt
```diff
@@ -0,0 +1,221 @@
...
+ * This property retrieves the tokenizer from the agent's storage using the `MessageTokenizer.Feature`,
+ * which must be initialized in the pipeline's features. The `PromptTokenizer` allows
+ * for tokenization operations on prompts and messages during the agent's execution.
+ *
+ * It facilitates operations such as calculating token counts for messages and prompts,
+ * which are critical in managing and optimizing interactions with language models.
+ *
+ * Throws an exception if the `MessageTokenizer.Feature` is not available in the context.
+ */
+public val AIAgentContextBase.tokenizer: PromptTokenizer get() = feature(MessageTokenizer.Feature)!!.promptTokenizer
```
EugeneTheDev: This would throw NPE, which is not really descriptive. It's better to throw exception manually with more descriptive message

agents/agents-features/agents-features-tokenizer/src/jvmTest/kotlin/ai/koog/agents/features/tokenizer/feature/MessageTokenizerTest.kt
```diff
@@ -0,0 +1,212 @@
...
+        fun reset() {
+            _totalTokens = 0
+        }
+    }
+
+    /**
+     * A simple implementation of PromptTokenizer for testing.
+     * This is similar to the OnDemandTokenizer in the main code.
+     */
+    class TestPromptTokenizer(private val tokenizer: Tokenizer) {
```
EugeneTheDev: The comment above says that it's an implementation of the `PromptTokenizer`, but it doesn't actually implement it. Looks like it should

agents/agents-features/agents-features-tokenizer/src/jvmTest/kotlin/ai/koog/agents/features/tokenizer/feature/MessageTokenizerTest.kt
```diff
@@ -0,0 +1,212 @@
...
+        fun tokensFor(message: Message): Int = tokenizer.countTokens(message.content)
+
+        /**
+         * Counts total tokens for all messages in a prompt.
+         */
+        fun totalTokensSpent(prompt: Prompt): Int = prompt.messages.sumOf { tokensFor(it) }
+    }
+
+    @Test
+    fun testTokenizerFeature() = runTest {
```
EugeneTheDev: I don't quite see the purpose of this test. It just tests mock implementations, which is not that useful. We don't want to test mocks in our tests, we want to use these mocks to test some actual behavior.

agents/agents-features/agents-features-tokenizer/src/jvmTest/kotlin/ai/koog/agents/features/tokenizer/feature/MessageTokenizerTest.kt
```diff
@@ -0,0 +1,212 @@
...
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+
+/**
+ * Test for the MessageTokenizer feature.
+ *
+ * This test verifies that the MessageTokenizer correctly tracks token usage.
+ */
+class MessageTokenizerTest {
```
EugeneTheDev, line 30: Also, why did you put this test inside `jvmTest`? I don't see it using any platform-specific funcionality, let's move it in `commonMain`?

prompt/prompt-tokenizer/Module.md
```diff
@@ -0,0 +1,54 @@
...
+
+The prompt-tokenizer module offers tools for estimating the number of tokens in text strings, which is essential for working with LLMs that have token limits or require token counting. The module includes:
+
+- A `Tokenizer` interface that defines the contract for token counting
+- Multiple implementations with varying levels of accuracy and performance:
+  - `NoTokenizer`: A minimal implementation that always returns 0 (useful when token counting is not needed)
+  - `RegexBasedTokenizer`: A simple implementation that provides reasonable approximations for most LLMs
+  - `TiktokenEncoder`: An advanced implementation based on OpenAI's tiktoken library, using byte pair encoding (BPE) for more accurate token counting
+
+This module is particularly useful for LLMs like Ollama that don't provide token counts in their responses, requiring client-side estimation.
```
EugeneTheDev: This is not true for Ollama though, as it turned out, right? Ollama should provide token information, as was mentioned in the conversation above? (link: https://github.com/ollama/ollama/blob/main/docs/api.md#response)

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/message/Message.kt
```diff
@@ -83,7 +94,8 @@ public sealed interface Message {
     @Serializable
     public data class Assistant(
         override val content: String,
-        val finishReason: String? = null
+        val finishReason: String? = null,
```
EugeneTheDev: I think you can move `finishReason` to metadata class now, it looks like it fits there

agents/agents-features/agents-features-trace/src/jvmTest/kotlin/ai/koog/agents/features/tracing/writer/TraceFeatureMessageFileWriterTest.kt
```diff
@@ -67,19 +67,24 @@ class TraceFeatureMessageFileWriterTest {
...
                 "${AIAgentFinishedEvent::class.simpleName} (strategy name: $strategyName, result: Done)",
             )
 
             val actualMessages = writer.targetPath.readLines()
 
             assertEquals(expectedMessages.size, actualMessages.size)
-            assertContentEquals(expectedMessages, actualMessages)
+            assertContentEquals(expectedMessages, actualMessages.map {
+                it.replace(
+                    ""","metadata":\{"timestamp":"(\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(?:\.\d+)?Z)"\}""".toRegex(),
```
EugeneTheDev: Here and in other similar tests you are trying to filter metadata out because of always changing timestamps. This doesn't look like a proper way to do it. See, you are using `Clock.System` directly in your classes implementation, which is explicitly discouraged by the docs (https://kotlinlang.org/api/kotlinx-datetime/kotlinx-datetime/kotlinx.datetime/-clock/). What you should probably do here instead is to utilize the common approach for such cases: don't use `Clock.System` directly in your implementation, pass `Clock` instance externally. This is recommended by the docs too. So then in real world `Clock.System` would be used, but in tests you can supply some fixed or determenistic `Clock` implementation. It's very simple interface, it only has one method `now()`, so it should be relatively easy to create your own test `Clock`.

EugeneTheDev (changes requested):
I have no major questions for now, generally looks good. There are some comments however regarding implementation details, please address them

agents/agents-features/agents-features-tokenizer/src/commonMain/kotlin/ai/koog/agents/features/tokenizer/feature/MessageTokenizer.kt
```diff
@@ -0,0 +1,221 @@
...
+ * An interface that provides utilities for tokenizing and calculating token usage in messages and prompts.
+ */
+public interface PromptTokenizer {
+    /**
+     * Calculates the number of tokens required for a given message.
+     *
+     * @param message The message for which the token count should be determined.
+     * @return The number of tokens required to encode the message.
+     */
+    public fun tokensFor(message: Message): Int
```
ptitjes: Can we change the name for `tokenCountFor`? The current name feels like it will return the token list.

agents/agents-features/agents-features-tokenizer/src/commonMain/kotlin/ai/koog/agents/features/tokenizer/feature/MessageTokenizer.kt
```diff
@@ -0,0 +1,221 @@
...
+
+    /**
+     * Calculates the total number of tokens spent in a given prompt.
+     *
+     * @param prompt The prompt for which the total tokens spent need to be calculated.
+     * @return The total number of tokens spent as an integer.
+     */
+    public fun totalTokensSpent(prompt: Prompt): Int
```
ptitjes: I may be nit-picking here too, but I'd rather we name that method as we named the method that takes a single message. Also having the notion of "spent" is confusing here, as this is the result of the tokenizer and not the one of the LLM API.

I think it would be more accurate to both call them `tokenCountFor(...)`.

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/message/Message.kt
```diff
@@ -150,8 +164,60 @@ public sealed interface Message {
...
+ *              Optionally, a custom timestamp can be provided; otherwise, the current system time is used.
+ *
+ * Example:
+ * - Message 1: "Hello" (3 tokens) → tokensCount = 3
+ * - Message 2: "How are you?" (4 tokens) → tokensCount = 3 + 4 = 7
+ * - Message 3: "I am fine, thank you." (6 tokens) → tokensCount = 7 + 6 = 13
+ */
+@Serializable
+public data class ResponseMetadata(
+    public val tokensCount: Int? = null,
```
ptitjes: Maybe `tokenCount` without an "s"?

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/message/Message.kt
```diff
@@ -150,8 +164,60 @@ public sealed interface Message {
...
+    override val timestamp: Instant = Clock.System.now()
```
Rizzen: I would not introduce Clock.System.now() as default value for timestamp, because it's kinda error prone. For me cleaner way here - to require calling side to stamp their requests

prompt/prompt-tokenizer/Module.md
```diff
@@ -0,0 +1,54 @@
...
+
+A module that provides interfaces and implementations for tokenizing text and counting tokens when working with Large Language Models (LLMs).
+
+### Overview
+
+The prompt-tokenizer module offers tools for estimating the number of tokens in text strings, which is essential for working with LLMs that have token limits or require token counting. The module includes:
+
+- A `Tokenizer` interface that defines the contract for token counting
+- Multiple implementations with varying levels of accuracy and performance:
+  - `NoTokenizer`: A minimal implementation that always returns 0 (useful when token counting is not needed)
```
Rizzen, line 11: We have `null` as default value for token count, and NoTokenizer assigning value = 0.
I suggest to align these entities either in the way we would have NoTokenizer assigning default value. WDYT?

prompt/prompt-tokenizer/src/commonMain/kotlin/ai/koog/prompt/tokenizer/tiktoken/TiktokenEncoder.kt
```diff
@@ -0,0 +1,117 @@
...
+ * This implementation is based on the reference Rust implementation from
+ * https://github.com/openai/tiktoken/blob/main/src/lib.rs.
+ *
+ * @constructor Initializes the TiktokenEncoder with the specified vocabulary, regex pattern for
+ * matching tokens, and a token ID for unknown tokens.
+ * @param vocabulary A mapping of ByteArrayKey to token ID that represents the token encoding vocabulary.
+ * @param pattern A regular expression used to match text patterns for initial token matching.
+ * @param unkTokenId The token ID used for unknown tokens not present in the vocabulary.
+ */
+public open class TiktokenEncoder(
```
Rizzen, line 19: Do we really want it open?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/extension/HistoryCompressionStrategies.kt
Ololoshechkin: I added it intentionally. `RetrieveFactsFromHistory` is in another module and requires a dependency (unless you use  the all-in-one dependency.

That's making it not so discoverable. So I think it's better to tell that it exists at least here in this kdoc

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/message/Message.kt
Ololoshechkin: It's not present in the Tool Call -- only in assistant message.
I don't want to make a separate ad-hock metadata for just assistant message

prompt/prompt-tokenizer/Module.md
Ololoshechkin: Hmm... Then I would have to make tokenCount() return Int? and all users would have to always write this !! or ?. everywhere.

> `We have null as default value for token count`

`null` is default value in metadata of LLM response because some LLMs might not return this data.
And tokenizer is an exact implementation of Tokenizer interface that counts symbols/tokens in the text: String (not in a LLM response) -- so why would it be `null`?

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/message/Message.kt
Ololoshechkin: Then, whenever one wants to create any Message even in tests -- they would have to manually create `ResponseMetadata` and pass the timestamp

prompt/prompt-tokenizer/src/commonMain/kotlin/ai/koog/prompt/tokenizer/tiktoken/TiktokenEncoder.kt
Ololoshechkin: Yes, I didn't add implementations for it because that is quite out of scope of this MR. There are multiple versions of TiktokenEncoder that we may add in the future

agents/agents-features/agents-features-tokenizer/src/jvmTest/kotlin/ai/koog/agents/features/tokenizer/feature/MessageTokenizerTest.kt
Ololoshechkin: `runBlocking`

agents/agents-features/agents-features-trace/src/jvmTest/kotlin/ai/koog/agents/features/tracing/writer/TraceFeatureMessageFileWriterTest.kt
Ololoshechkin: @jetbrains-junie Fix this issue and refactor as it was suggested

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/message/Message.kt
ptitjes: I agree with @Rizzen. I think that particularly in tests you want predictable timestamps.

Are those class constructors really meant to be called directly anyway?
Most people will use `PromptBuilder`.

I would propose that we remove the default timestamp values on the Message subclasses.
And that we enhance `PromptBuilder` so that:
- `PromptBuilder` takes an additional `Clock` parameter.
- It would have as default value `Clock.System`.
- It would use its clock when it creates Message subclasses.

In that way, in tests we can call `prompt("my-prompt", clock = testClock) { }` with a `TestClock` clock implementation.

Maybe that's too much?

agents/agents-features/agents-features-trace/src/jvmTest/kotlin/ai/koog/agents/features/tracing/writer/TraceFeatureMessageFileWriterTest.kt
Ploffi: @jetbrains-junie Fix this issue and refactor as it was suggested

ptitjes (commented):
That looks great!

agents/agents-features/agents-features-trace/src/jvmTest/kotlin/ai/koog/agents/features/tracing/writer/TraceFeatureMessageFileWriterTest.kt
Ploffi: @Ololoshechkin I'm afraid you can't run a Junie on a branch `vbr/token-count` which doesn't have a workflow file. Now it's present only in the `develop` branch.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-openrouter-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openrouter/OpenRouterLLMClient.kt
```diff
@@ -348,19 +348,27 @@ public class OpenRouterLLMClient(
             .firstOrNull()
             ?.let { it to it.message } ?: throw IllegalStateException("No choice found in OpenRouter response")
 
+        // Extract token count from the response
+        val tokensCount = response.usage?.totalTokens
```
tiginamaria: Don’t we want to support all token counts? It would be helpful for calculating billing from traces and evaluating agents. Maybe it’s not urgent, but it seems like a good PR to add into
https://platform.openai.com/docs/api-reference/responses/create
```
  "usage": {
    "input_tokens": 36,
    "input_tokens_details": {
      "cached_tokens": 0
    },
    "output_tokens": 87,
    "output_tokens_details": {
      "reasoning_tokens": 0
    },
    "total_tokens": 123
  },
```

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/message/Message.kt
```diff
@@ -83,7 +94,8 @@ public sealed interface Message {
     @Serializable
     public data class Assistant(
         override val content: String,
-        val finishReason: String? = null
+        val finishReason: String? = null,
+        override val metadata: ResponseMetadata = ResponseMetadata()
```
tiginamaria: Also metadata seems like existing field in LLM response so maybe consider other naming to avoid misunderstanding (e.g. `tokensUsage`) https://platform.openai.com/docs/api-reference/chat/create#chat-create-metadata

agents/agents-features/agents-features-tokenizer/src/jvmTest/kotlin/ai/koog/agents/features/tokenizer/feature/MessageTokenizerTest.kt
EugeneTheDev: You should use `runTest`, it's multiplatform and is generally preferred in tests
https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-test/kotlinx.coroutines.test/run-test.html. You can check example usage in [MCP tests](https://github.com/JetBrains/koog/blob/0efd4150d4ecdf4c15fefa6f33b0a3b6d1d11946/agents/agents-mcp/src/jvmTest/kotlin/ai/koog/agents/mcp/McpToolTest.kt#L41)

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/message/Message.kt
EugeneTheDev: But you already have separate metadata for request and response. Why don't you want to extend it further and make a special metadata for assistant messages too? It doesn't seem that bad to me

prompt/prompt-executor/prompt-executor-clients/prompt-executor-openrouter-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openrouter/OpenRouterLLMClient.kt
Ololoshechkin: Thanks, good point! I was actually thinking about it but looks like not all models and providers support such details. For instance, Ollama only returns input tokens for the whole prompt before the response, and response tokens.

But probably this we can include:
- total tokens
- input tokens
- output tokens

And if something is missing, they are all optional anyway

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/message/Message.kt
Ololoshechkin: Hmm. It also contains timestamps. Maybe let's call it "metaInfo" ?

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/message/Message.kt
Rizzen: >Then, whenever one wants to create any Message even in tests -- they would have to manually create ResponseMetadata and pass the timestamp

I don't think it's a problem. That's DTO after all, it shouldn't timestamp itself. Moreover, users might want to control what stamps they want to put on the messages.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/environment/ReceivedToolResult.kt
```diff
@@ -5,21 +5,24 @@ import ai.koog.agents.core.tools.ToolResult
...
+import kotlinx.datetime.Clock
 
 public data class ReceivedToolResult(
     val id: String?,
     val tool: String,
     val content: String,
     val result: ToolResult?
 ) {
-    public fun toMessage(): Message.Tool.Result = Message.Tool.Result(
+    internal fun toMessage(clock: Clock): Message.Tool.Result = Message.Tool.Result(
```
EugeneTheDev: Why did you make it internal? It's a useful function, let's not hide it. You can just add default parameter for `clock`

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/environment/ReceivedToolResult.kt
```diff
@@ -5,21 +5,24 @@ import ai.koog.agents.core.tools.ToolResult
...
         id = id,
         tool = tool,
         content = content,
+        metaInfo = RequestMetaInfo.create(clock)
     )
 }
 
-public fun EnvironmentToolResultToAgentContent.toResult(): ReceivedToolResult {
+internal fun EnvironmentToolResultToAgentContent.toResult(): ReceivedToolResult {
```
EugeneTheDev: Same here, why did you make all these functions internal?

EugeneTheDev (approved):
Can you please clarify, why did you change the visibility modifier (checky comments below)? I would prefer them to stay public

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/environment/ReceivedToolResult.kt
Ololoshechkin: It was used only in one single place:

```
public fun PromptBuilder.ToolMessageBuilder.result(result: ReceivedToolResult) {
    result(result.toMessage(clock))
}
```

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/environment/ReceivedToolResult.kt
Ololoshechkin: Same, was only used inside one and only inside `internal` object `AIAgentEnvironmentUtils`

Issue #178, opened 2025-05-28
LLMDescription not included in the generated schema


ionTea:
It seems to me like the `JsonStructuredData` schema doesn't actually contain the annotated data from `@LLMDescription`

For example this is what the result is when serializing an annotated class: 

```
    @Serializable
    @LLMDescription("A simple data class for testing serialization")
    data class Hero(val name: String, val age: Int, @LLMDescription("The hero's superpower") val superpower: String)

    test("Serializing simple data") {
      val heroStructure =
        JsonStructuredData.createJsonStructure<Hero>(
          schemaFormat = JsonSchemaGenerator.SchemaFormat.Simple,
          schemaType = JsonStructuredData.JsonSchemaType.SIMPLE,
        )

      expectSelfie(serialize.encodeToString(heroStructure.schema))
        .toBe(
          """{
    "type": "ai.koog.prompt.params.LLMParams.Schema.JSON.Simple",
    "name": "Hero",
    "schema": {
        "type": "object",
        "properties": {
            "name": {
                "type": "string"
            },
            "age": {
                "type": "integer"
            },
            "superpower": {
                "type": "string"
            }
        },
        "required": [
            "name",
            "age",
            "superpower"
        ]
    }
}"""
        )
    }
  })
```

I would have expected the schema to be this:

```
{
    "type": "ai.koog.prompt.params.LLMParams.Schema.JSON.Simple",
    "name": "Hero",
    "description": "A simple data class for testing serialization", <----
    "schema": {
        "type": "object",
        "properties": {
            "name": {
                "type": "string"
            },
            "age": {
                "type": "integer"
            },
            "superpower": {
                "type": "string",
                "description": "The hero's superpower" <----
            }
        },
        "required": [
            "name",
            "age",
            "superpower"
        ]
    }
}
```

dpolugic:
I think perhaps `LLMDescription` needs to be annotated with [`kotlinx.serialization.SerialInfo`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-serial-info/) in order to be included in the descriptor. 🤔

ionTea:
@Ololoshechkin adding `@property:LLMDescription` doesn't help either 🤔

PR #188, opened 2025-05-29
Add assertion for schema description in StructuredDataTest.kt

dpolugic:
Schema descriptions added with the `@LLMDescription` annotation are currently not included in the generated JSON Schema. This PR adds an assertion to one of the structured data tests to demonstrate the behavior.

The reason it isn't included is that `kotlinx.serialization` only includes annotations relevant for serialization when generating the descriptor. Adding the `@SerialInfo` annotation to `LLMDescription` ensures that it is included, and should solve this problem. The only potential issue I see with that is that it requires opting into `ExperimentalSerializationApi`, and I'm not sure what the maintainers' stance on that is.

An alternate solution could be to not use the descriptor to extract the annotations, but to use the Kotlin reflection API, as is done for tools. I don't know how viable that is though.

Related to #178 

---

#### Type of the change
- [ ] New feature
- [x] Bug fix
- [ ] Documentation fix

### DIFF
--- a/agents/agents-tools/src/commonMain/kotlin/ai/koog/agents/core/tools/annotations/LLMDescription.kt
+++ b/agents/agents-tools/src/commonMain/kotlin/ai/koog/agents/core/tools/annotations/LLMDescription.kt
@@ -1,11 +1,14 @@
 package ai.koog.agents.core.tools.annotations
 
+import kotlinx.serialization.SerialInfo
+
 /**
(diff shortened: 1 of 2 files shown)


EugeneTheDev:
Thank you for discovering this bug. I used your branch as a foundation to properly implement annotation handling during JSON schema generation in #191. So I'm closing this PR now.

PR #191, opened 2025-05-29
(#178) Fix structured output JSON descriptions

### LINKED ISSUE #178: LLMDescription not included in the generated schema

EugeneTheDev:
Based on #188 and fixes #178 by properly handling `@LLMDescription` during JSON schema generation. Also updated the tests to include the annotation too.


prompt/prompt-structure/src/commonMain/kotlin/ai/koog/prompt/structure/json/JsonStructuredData.kt
```diff
@@ -67,26 +66,29 @@ public class JsonStructuredData<TStruct>(
...
         /**
          * Factory method to create JSON structure with auto-generated JSON schema.
+         *
+         * @param id Unique identifier for the structure.
+         * @param serializer Serializer used for converting the data to and from JSON.
+         * @param json JSON configuration instance used for serialization.
+         * @param schemaFormat Format of the generated schema, can be simple or detailed.
+         * @param maxDepth Maximum recursion depth when generating schema to prevent infinite recursion for circular references.
+         * @param descriptions Optional map of serial class names and property names to descriptions.
+         * If a property/type is already described with [LLMDescription] annotation, value from the map will override this description.
```
Ololoshechkin, line 128: Let's also please add good example in KDoc about how to form the descriptions map.
I mean, `"className.propertyName" to "property description"`

Ololoshechkin (approved):
@EugeneTheDev please also add a test with a lot of different descriptors:
- class descriptor via annotation
- property descriptor via annotation
- property of Serializable type, also with field, each of them annotated
- one field is annotated and also overriden by the map
- one field is NOT annotated but provided through the map
- what else can be an edge case?

prompt/prompt-structure/src/commonMain/kotlin/ai/koog/prompt/structure/json/JsonStructuredData.kt
Ololoshechkin: Also, maybe let's name it `descriptionOverrides` if it's an overriding map?

EugeneTheDev: Regarding testing edge cases, that's what I already modified in tests, to cover these cases you described. I didn't mark each field with annotation though because it looks redundant, and it's also nice to test that description is added only for these fields where it is specified, and is skipped for other fields.

Regarding examples and parameter renaming: sounds good, will do it

Rizzen (approved):

PR #195, opened 2025-05-30
Devcrocod/support media types

### LINKED ISSUE #165: Support audio input and output
Could you please add support for audio generation. This is essential for conversational assistants.

It would be nice if it could support:
- speech to speech flows
- and chained flows: input audio -> input transcription -> model text response -> audio output

### LINKED ISSUE #189: Missing capability to send complex content (text and images) in PromptDSL
Hi Koog Team,

I've noticed that the current version of PromptDSL lacks the ability to send more complex structures in the content field, a feature supported by some language models. Specifically, I'm referring to the capability to send a list of objects, where each object defines its type (e.g., text or image) and its respective content.

An example of such a structure, which I currently cannot implement using PromptDSL, is:

```json
{
  "role" : "user"
  "content": [
    {
      "type": "input_text",
      "text": "what's in this image?"
    },
    {
      "type": "input_image",
      "image_url": "data:image/jpeg;base64,{base64_image}"
    }
  ]
}
```
This functionality would be useful for working with multimodal models.

Are there any plans to add support for this type of complex content object in PromptDSL? I would appreciate any information on this.

Thank you and best regards!

devcrocod:
close #165 
close #189

---

#### Type of the change
- [ ] New feature
- [ ] Bug fix
- [ ] Documentation fix


agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/session/AIAgentLLMWriteSession.kt
```diff
@@ -441,7 +441,7 @@ public class AIAgentLLMWriteSession internal constructor(
         if (definition != null) {
             val prompt = prompt(prompt) {
                 user {
-                    definition.definition(this)
+                    text { definition.definition(this) }
                 }
```
Rizzen: Consider making plain text the default input type, with other media types (e.g., images, audio) explicitly defined. This way, users won't need to wrap text inputs in `text { "some string" }`.

devcrocod: @Rizzen 
Yeap, I agree that need to improve dsl. what do you suggest?
now, it looks like this:

```kotlin
prompt() {
   user("string")
}
```

```kotlin
prompt() {
   user {
	text("string")
	audio(...)
	image(...)
	video(...)
	document(...)
   }
}
```

```kotlin
prompt() {
   user {
	text {
		...
	}
	audio(...)
	image(...)
	video(...)
	document(...)
   }
}
```

prompt/prompt-executor/prompt-executor-clients/prompt-executor-anthropic-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/anthropic/AnthropicLLMClient.kt
```diff
@@ -309,18 +265,70 @@ public open class AnthropicLLMClient(
...
+            is MediaContent.File -> {
+                require(model.capabilities.contains(LLMCapability.Vision.Image)) {
+                    "Model ${model.id} does not support files"
+                }
+                val listOfContent = buildList {
+                    if (content.isNotEmpty()) {
+                        add(AnthropicContent.Text(content))
+                    }
+                    val docSource = when {
+                        media.isUrl() -> DocumentSource.PDFUrl(media.source)
```
Ololoshechkin: What if the passed file is not a PDF? Maybe we should handle this error explicitly before sending to anthropic?

Ololoshechkin:
Also please consider adding some example with media types:
1. Simple example with images and prompt executor (ex: here are my 3 instagram photos that I want to share in one post, please invent a good description under the post)

2. Example with agent (ex: chat-support conversation that includes image attachments and image responses from the agent)

Ololoshechkin (changes requested):
Please, add agent example with media types, and add required edges/nodes to the DSL

prompt/prompt-executor/prompt-executor-clients/prompt-executor-anthropic-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/anthropic/AnthropicLLMClient.kt
devcrocod: There will be an error, but the URL doesn’t always provide information about the type of document. 
Usually, it looks like: `https://..../document.pdf`

But for example, pdfs from arxiv look like this: `https://arxiv.org/pdf/2505.23896`

That's why I removed check document extension for Anthropic, leaving this responsibility to user

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/session/AIAgentLLMWriteSession.kt
Rizzen: I was thinking about option as following:
```
prompt() {
   user {
	+"Hello"
        +"There"
	audio(...)
	image(...)
	video(...)
	document(...)
   }
}
```

Just to simplify simplest cases like following:

```
prompt() {
   system {
	+"Hello"
        +"There"
   }
   user {
	+"General"
        +"Kenobi"
   }
}

File: prompt/prompt-executor/prompt-executor-clients/prompt-executor-anthropic-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/anthropic/AnthropicLLMClient.kt (hunk 2 above)
Comment (Ololoshechkin) [reply prompt/prompt-executor/prompt-executor-clients/prompt-executor-anthropic-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/anthropic/AnthropicLLMClient.kt]: okay, let's then leave a noticeable description in kdoc and also ask our tech writers to document this very explicity

File: agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/session/AIAgentLLMWriteSession.kt (hunk 1 above)
Comment (devcrocod) [reply agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/session/AIAgentLLMWriteSession.kt]: @Rizzen In my opinion, this is not good design

I'm not a big fan of overriding `unaryPlus` for String because such an API is not very obvious.
That's my personal pref, but in this case, such api creates confusion. 

How should message to LLM look if we write it like this:
```kotlin
prompt() {
   user {
        +"General"
        document(...)
        +"Kenobi"
   }
}
```

What do you think about explicitly separating the methods?  Keep old `user` method and create a new one like `userWithMedia`?

Rizzen: >Keep old user method and create a new one like userWithMedia?

 I think we can introduce `userAttachment` builder for media types. @Ololoshechkin WDYT?

prompt/prompt-executor/prompt-executor-clients/prompt-executor-anthropic-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/anthropic/AnthropicLLMClient.kt
devcrocod: btw, it’s not very clear where to add this
The user doesn’t create the prompt directly for the model, and right now, each model supports different types of content:
- Google only supports local PDFs, to work with other files, we need to support the Files API
- OpenAI also supports only a limited number of media content types, and we would need to handle a different REST API and different file upload logic
- Anthropic would also need Files API support

I noted the supported types and models where I could, but I might have missed something.
The problem is that when the user creates a prompt or a node, they won’t see what exactly the Anthropic models (or any other models) support

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/session/AIAgentLLMWriteSession.kt
Ololoshechkin: I'm not the biggest fan of the unary plus, either. It's quite error prone imo. But having `attachments` sounds actually good -- it will allow us to not mix up the text/image/text/audio/text without any order, and will be much more like in the chat UI -- you click "attach" and attach :)

If it's just text -- `user("aaa")` and `text("...")` covers this case
If its something that needs a builder:
```
user {
   text("aaa")
   markdown { ... }
   
   attachments {
      image(...)
      image(...)
      document(...) // Note: maybe if we only support PDFs, let's call it `pdf(...)` or `pdfDocument()` ???
   }
}
```

Though, `voice` doesn't really fit into attachments category to me, as you can just dictate it.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-anthropic-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/anthropic/AnthropicLLMClient.kt
Ololoshechkin: Sorry, I might have missed all the places where you added it. But let's probably add it in the `document` DSL method at least (probably, the most user-facing place)

Ololoshechkin:
> noted that very few models support media output. Right now, from all the models we have, only OpenAI 4o-audio can return audio
Anthropic only returns text.
Google only returns text, and only special models can return audio or images.

Looks like we have to actually invest some time in IDE plugin with a few inspections for all these model differences...
And mention that in docs. I'm thinking of having some badges about models/providers like [jvm/native/js] badges on kotlin libraries website (cc: @innateteniuk  WDYT? )

> I also faced a problem when creating a node for media. For example, if we want to create a strategy that sends a specific image or file, it’s straightforward. But what if we have a lot of images or files?

Tried to check out your branch to take a look how the node is defined, but didn't find it there,
But in general, would something like `nodeLLMRequestMultiple` / `nodeLLMSendMultipleToolResults` work for your case?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/session/AIAgentLLMWriteSession.kt
devcrocod: >  Note: maybe if we only support PDFs, let's call it `pdf(...)` or `pdfDocument()`

Not just PDF,
Anthropic also supports other types by sending text
Google supports many types too

> Though, voice doesn't really fit into attachments category to me, as you can just dictate it.

But audio also like attachment 
Later, we can make a separate `userVoice` when we support the Realtime API

PR #202, opened 2025-06-02
Separate Ollama tests and set temperature to 0

aozherelyeva:
- Add an Ollama test type
- add a dedicated workflow for Ollama tests
- update Ollama temperature in tests
- add Google server error handling in retry logic
- disable tests that fail due to known bugs

---

#### Type of the change
- [ ] New feature
- [ ] Bug fix
- [ ] Documentation fix
- [x] Test enhancement


.github/workflows/ollama-tests.yml
```diff
@@ -0,0 +1,53 @@
...
+
+      # Configure Gradle for optimal use in GitHub Actions, including caching of downloaded dependencies.
+      # See: https://github.com/gradle/actions/blob/main/setup-gradle/README.md
+      - name: Setup Gradle
+        uses: gradle/actions/setup-gradle@af1da67850ed9a4cedd57bfd976089dd991e2582 # v4.0.0
+
+      - name: JvmOllamaTest with Gradle Wrapper
+        env:
+          OLLAMA_IMAGE_URL: ${{ vars.OLLAMA_IMAGE_URL }}
+        run: ./gradlew jvmOllamaTest --no-parallel --continue
```
sproshev, line 45: why have you decided to run them separately?

kpavlov:
> If you want reproducible outputs, you can also pass a seed to Ollama. Don't know if that helps...

tip: and if the model does not support the seed, the temperature can be used as seed...

.github/workflows/ollama-tests.yml
zarechneva-a: It seems that the `--no-parallel ` approach indeed helped, but I don't want to run all integration tests without parallelism. So I figured out having a separate test config for the Ollama tests (and maybe other local providers if we support such ones in the future?) is a good solution.

sproshev: Ok, let's try, if it saves time, then why not

zarechneva-a: Thank you! It should. I believe, as our test number grows, this distinction might become crucial for the tests' performance.

KarpovSS (approved):

PR #203, opened 2025-06-02
Change OPEN_AI_TOKEN to OPENAI_API_KEY

kpgalligan:
The examples use `ai.koog.agents.example.ApiKeyService` to pull model keys from environment variables. When open ai isn't found, you get an exception with 'OPENAI_API_KEY env is not set'. It took some extra time to debug because 'OPENAI_API_KEY' isn't actually the name of the value that `ApiKeyService` is looking for, but 'OPEN_AI_TOKEN'.

Since other docs use 'OPENAI_API_KEY', I figured that was the convention to use. If somebody has the examples currently running with 'OPEN_AI_TOKEN', the change will break that, but it seems more consistent to use 'OPENAI_API_KEY'.
---

#### Type of the change
- [ ] New feature
- [x] Bug fix
- [ ] Documentation fix

- [ ] Tests for the changes have been added (doesn't really seem like a thing to "test")

##### Additional steps for pull requests adding a new feature (not a new feature so N/A)

### DIFF
--- a/examples/src/main/kotlin/ai/koog/agents/example/ApiKeyService.kt
+++ b/examples/src/main/kotlin/ai/koog/agents/example/ApiKeyService.kt
@@ -2,7 +2,7 @@ package ai.koog.agents.example
 
 internal object ApiKeyService {
     val openAIApiKey: String
-        get() = System.getenv("OPEN_AI_TOKEN") ?: throw IllegalArgumentException("OPENAI_API_KEY env is not set")
+        get() = System.getenv("OPENAI_API_KEY") ?: throw IllegalArgumentException("OPENAI_API_KEY env is not set")
 
     val anthropicApiKey: String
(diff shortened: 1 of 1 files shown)


examples/src/main/kotlin/ai/koog/agents/example/ApiKeyService.kt
```diff
@@ -2,7 +2,7 @@ package ai.koog.agents.example
 
 internal object ApiKeyService {
     val openAIApiKey: String
-        get() = System.getenv("OPEN_AI_TOKEN") ?: throw IllegalArgumentException("OPENAI_API_KEY env is not set")
+        get() = System.getenv("OPENAI_API_KEY") ?: throw IllegalArgumentException("OPENAI_API_KEY env is not set")
```
kpavlov, line 5: It would be nice to add a unit test with https://github.com/webcompere/system-stubs

PR #205, opened 2025-06-02
Add Linear Strategy API with History Compression Support 

Ololoshechkin:

---

#### Type of the change
- [x] New feature
- [ ] Bug fix
- [ ] Documentation fix
- [x] API enhancement

### DIFF
--- a/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/builder/AIAgentSubgraphBuilder.kt
+++ b/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/builder/AIAgentSubgraphBuilder.kt
@@ -2,7 +2,11 @@ package ai.koog.agents.core.dsl.builder
 
 import ai.koog.agents.core.agent.context.AIAgentContextBase
 import ai.koog.agents.core.agent.entity.*
+import ai.koog.agents.core.dsl.extension.HistoryCompressionStrategy
+import ai.koog.agents.core.dsl.extension.HistoryCompressionStrategy.NoCompression
+import ai.koog.agents.core.dsl.extension.nodeLLMCompressHistory
 import ai.koog.agents.core.tools.Tool
+import ai.koog.agents.core.utils.Some
 import kotlin.reflect.KProperty
 
 // TODO: rename *BuilderBase to *Builder and use specific prefixes (or suffixes) for subclasses
@@ -54,13 +58,76 @@ public abstract class AIAgentSubgraphBuilderBase<Input, Output> {
     /**
      * Connects the sequence of nodes with edges between them.
      * @param nextNode Node to connect to
-     * @return The next node
+     * @return The [LinearStrategyIntermediate] object that can be then passed to [applyLinearStrategy]
      */
-    public infix fun <IncomingOutput, OutgoingInput, OutgoingOutput> AIAgentNodeBase<IncomingOutput, OutgoingInput>.then(
-        nextNode: AIAgentNodeBase<OutgoingInput, OutgoingOutput>
-    ): AIAgentNodeBase<OutgoingInput, OutgoingOutput> {
+    public infix fun <IncomingInput, IncomingOutput, OutgoingOutput> AIAgentNodeBase<IncomingInput, IncomingOutput>.then(
+        nextNode: AIAgentNodeBase<IncomingOutput, OutgoingOutput>
+    ): LinearStrategyIntermediate<IncomingInput, OutgoingOutput> {
         edge(this forwardTo nextNode)
-        return nextNode
+        return LinearStrategyIntermediate(firstNode = this, lastNode = nextNode, allNodes = listOf(this, nextNode))
+    }
+
+    /**
+     * Connects the last node of the current linear strategy intermediate to a new node in the sequence
+     * and returns a new linear strategy intermediate representing the updated sequence.
+     *
+     * @param nextNode The next node to which the last node of the current intermediate will be connected.
+     * @return A new instance of [LinearStrategyIntermediate] with the updated sequence of nodes,
+     *  starting from the original first node and ending at the provided next node.
+     */
+    public infix fun <IncomingInput, IncomingOutput, OutgoingOutput> LinearStrategyIntermediate<IncomingInput, IncomingOutput>.then(
+        nextNode: AIAgentNodeBase<IncomingOutput, OutgoingOutput>
+    ): LinearStrategyIntermediate<IncomingInput, OutgoingOutput> {
+        return LinearStrategyIntermediate(
+            firstNode = this.firstNode,
+            lastNode = nextNode,
+            allNodes = this.allNodes + nextNode
+        )
+    }
+
+    /**
+     * Applies a defined linear strategy by connecting a sequence of nodes in a predefined order
+     * and optionally applying a history compression strategy between each pair of nodes.
+     *
+     * @param steps The linear strategy consisting of interconnected nodes, represented by an instance of [LinearStrategyIntermediate].
+     *              It defines the sequence of nodes to be connected in the agent's graph.
+     * @param historyCompressionStrategy A strategy for compressing history between each step in the linear strategy. Defaults to [HistoryCompressionStrategy.NoCompression].
+     */
+    @Suppress("UNCHECKED_CAST")
+    public fun applyLinearStrategy(
+        steps: LinearStrategyIntermediate<Input, Output>,
+        historyCompressionStrategy: HistoryCompressionStrategy = NoCompression
+    ) {
+        check(!steps.allNodes.contains(nodeStart)) {
+            "`nodeStart` can't be one of the steps of the linear strategy. It will be already included automatically"
+        }
+        check(!steps.allNodes.contains(nodeFinish)) {
+            "`nodeFinish` can't be one of the steps of the linear strategy. It will be already included automatically"
+        }
+
+        fun addEdge(from: AIAgentNodeBase<Any?, Any?>, to: AIAgentNodeBase<Any?, Any?>) {
+            from.addEdge(AIAgentEdge(to, { _, output -> Some(output) }))
+        }
+
+        var currentNode: AIAgentNodeBase<Any?, Any?> = nodeStart as AIAgentNodeBase<Any?, Any?>
+
+        steps.allNodes.forEachIndexed { index, node ->
+            // We only compress history if it's needed AND if it's not between the nodeStart and the first node
+            //   (doesn't make sense to compress in this case as nodeStart doesn't perform any additional actions)
+            if (historyCompressionStrategy !is NoCompression && index > 0) {
+                val compressHistory by nodeLLMCompressHistory<Any?>(
+                    name = "compressHistory_before_${node.name}",
+                    strategy = historyCompressionStrategy
+                )
+                currentNode.addEdge(AIAgentEdge(compressHistory, { _, output -> Some(output) }))
+                currentNode = compressHistory
+            }
+
+            currentNode.addEdge(AIAgentEdge(node as AIAgentNodeBase<Any?, Any?>, { _, output -> Some(output) }))
+
+            currentNode = node
+        }
+        addEdge(currentNode, nodeFinish as AIAgentNodeBase<Any?, Any?>)
     }
 
     /**
--- a/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/builder/AIAgentSubgraphBuilder.kt
+++ b/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/builder/AIAgentSubgraphBuilder.kt
@@ -56,7 +56,9 @@ public abstract class AIAgentSubgraphBuilderBase<Input, Output> {
      * @param nextNode Node to connect to
      * @return The next node
      */
-    public infix fun <IncomingOutput, OutgoingInput, OutgoingOutput> AIAgentNodeBase<IncomingOutput, OutgoingInput>.then(nextNode: AIAgentNodeBase<OutgoingInput, OutgoingOutput>): AIAgentNodeBase<OutgoingInput, OutgoingOutput> {
+    public infix fun <IncomingOutput, OutgoingInput, OutgoingOutput> AIAgentNodeBase<IncomingOutput, OutgoingInput>.then(
+        nextNode: AIAgentNodeBase<OutgoingInput, OutgoingOutput>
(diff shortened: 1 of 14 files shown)


integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/KotlinAIAgentWithMultipleLLMIntegrationTest.kt
```diff
@@ -473,7 +473,7 @@ class KotlinAIAgentWithMultipleLLMIntegrationTest {
 
             val compressHistoryNode by nodeLLMCompressHistory<Unit>("compress_history")
 
-            nodeStart then anthropicSubgraph then compressHistoryNode then openaiSubgraph then nodeFinish
+            applyLinearStrategy(anthropicSubgraph then compressHistoryNode then openaiSubgraph)
```
Ololoshechkin: So, basically, one side of this, is that users don't have to write `nodeStart` and `nodeFinish` all the time

agents/agents-core/src/commonTest/kotlin/ai/koog/agents/core/dsl/extension/LinearStrategyTest.kt
```diff
@@ -0,0 +1,152 @@
...
+
+            val node3 by node<String, String>("node3") { input ->
+                "Node 3 processed: $input"
+            }
+
+            // Apply the linear strategy with history compression
+            applyLinearStrategy(
+                steps = node1 then node2 then node3,
+                historyCompressionStrategy = HistoryCompressionStrategy.CompressWholeHistory
+            )
```
Ololoshechkin: Another side -- we can now add history compression between steps

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/extension/AIAgentNodes.kt
```diff
@@ -35,16 +35,35 @@ public fun <T> AIAgentSubgraphBuilderBase<*, *>.nodeDoNothing(name: String? = nu
  * @param name Optional node name, defaults to delegate's property name.
  * @param body Lambda to modify the prompt using PromptBuilder.
  */
-public fun AIAgentSubgraphBuilderBase<*, *>.nodeUpdatePrompt(
+public fun <T> AIAgentSubgraphBuilderBase<*, *>.nodeUpdatePrompt(
```
Ololoshechkin, line 38: Also made `nodeUpdatePrompt` typed so that when you need to do smth. then update prompt, than do smth. else -- you won't loose the `T` value.

Previously I had to write something like:
```
 val createStructure by subgraph(tools = listOf(CreateStructureTool)) {
      val nodeInitStage by node<String, String> { input ->
          llm.writeSession {
              updatePrompt { system(createStructureSystemPrompt) }
          }
          input
      }
      val nodeSendInput by nodeLLMRequest()
      ...

      edge(nodeStart forwardTo nodeInitStage)
      edge(nodeInitStage forwardTo nodeSendInput)
```

Now, I can do it like this:
```
 val createStructure by subgraph(tools = listOf(CreateStructureTool)) {
      val nodeInitStage by nodeUpdatePrompt<String> { system(createStructureSystemPrompt) }
      val nodeSendInput by nodeLLMRequest()
      ...

      edge(nodeStart forwardTo nodeInitStage)
      edge(nodeInitStage forwardTo nodeSendInput)
```

agents/agents-core/src/commonTest/kotlin/ai/koog/agents/core/dsl/extension/LinearStrategyTest.kt
Ololoshechkin: Previously, one had to write :

```
val clearHistory by node<String, String> {
            llm.writeSession {
                clearHistory()
            }
            it
}


nodeStart then createStructure then clearHistory then createDependencies then nodeFinish
```

Now:
1. `val clearHistory by nodeLLMCompressHistory(strategy = HistoryCompressionStrategy.ClearHistory)` or even `val clearHistory by nodeClearHistory()`
2. whole code is just `applyLinearStrategy(steps = createStructure then createDependencies, historyCompressionStrategy = HistoryCompressionStrategy.ClearHistory)`

agents/agents-core/src/commonTest/kotlin/ai/koog/agents/core/dsl/extension/AIAgentNodesHistoryCompressionTest.kt
```diff
@@ -200,4 +200,112 @@ class AIAgentNodesHistoryCompressionTest {
...
+        agent.run("")
+
+        // After compression, we should have one result
+        assertEquals(1, results.size)
+        assertEquals("Done", results.first())
+
+        // Verify that no TLDR messages were created
+        assertEquals(0, testExecutor.tldrCount, "NoCompression strategy should not create any TLDR messages")
+
+        // Verify that no TLDR messages are in the final history
```
EugeneTheDev, line 253: In this test and the one below: seems like this check for tldr is redundant, since we already check `tldrCount` from `testExecutor`. But I would also check the resulting prompt, that it indeed contains messages we would expect. The main goal of history strategies is to do something with the prompt, so it would be appropriate to check it

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/builder/AIAgentSubgraphBuilder.kt
```diff
@@ -54,11 +58,76 @@ public abstract class AIAgentSubgraphBuilderBase<Input, Output> {
...
+    /**
+     * Applies a defined linear strategy by connecting a sequence of nodes in a predefined order
+     * and optionally applying a history compression strategy between each pair of nodes.
+     *
+     * @param steps The linear strategy consisting of interconnected nodes, represented by an instance of [LinearStrategyIntermediate].
+     *              It defines the sequence of nodes to be connected in the agent's graph.
+     * @param historyCompressionBetweenSteps A strategy for compressing history between each step in the linear strategy. Defaults to [HistoryCompressionStrategy.NoCompression].
+     */
+    @Suppress("UNCHECKED_CAST")
+    public fun applyLinearStrategy(
```
EugeneTheDev: I have some questions/thoughts:
1.  Am I correct that the main idea is to make it simpler to specify history compression strategy between such nodes?
2. The idea itself isn't bad. It now looks more consistent with the general way to define edges: you have to wrap `forwardTo` into `edge` function, but with `then` before this PR they were just kinda hanging around (there were some pros and cons). However, the name looks somewhat confusing to me. We are, yet again, overusing the word "strategy". We already have agent strategy (graph), history compression strategy, and now this. Let's make it more clear that it's some linear edge and call it `edgesLinear` or something like that (and the class too), to better reflect it's purpose, that it's just a bunch of edges connected in a line (and also to avoid introducing yet another strategy)

EugeneTheDev (commented):
Generally looks good, I have a couple of questions

agents/agents-core/src/commonTest/kotlin/ai/koog/agents/core/dsl/extension/AIAgentNodesHistoryCompressionTest.kt
Ololoshechkin: > would be appropriate to check it

It is already checked in another test. And we use mock executors anyway

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/builder/AIAgentSubgraphBuilder.kt
Ololoshechkin: > 1. Am I correct that the main idea is to make it simpler to specify history compression strategy between such nodes?

1. Yes
2. Also not to write `nodeStart then` and `then nodeFinish` every time (it's error prone and easy to forget about them)
3. Also to make more unified with `edge(...)`. Why edges are inside `edge()` and `then` is just hanging randomly inside the strategy context?

Ololoshechkin: > 2.

About the name, good point. I was initially thinking about `applyStrategy` because we apply it to the current strategy (I mean to the current `strategy()  {...` if we are inside it.

Maybe `applyLinearSteps` then?

Ololoshechkin: Or maybe just `linearPipeline` ?

EugeneTheDev: Wait, I guess I missed this part, my bad: this function also adds start and finish nodes automatically? And what if I want to write a linear segment, then some custom edges with branching or smth, then another linear segment. It won't be possible with this API, right?

About pipeline, we already have an agent pipeline, so again, name clashes and also not too clear that it's about edges

Ololoshechkin: I believe we shouldn't create another way to write simple nodes?
We want to cover the most popular use case by making it simpler. For the current proposed API there are at least 3 use cases that I know in JB.

If someone needs to connect 3 nods linearly and that's not the whole graph - then why not just write
`edge(a forwardTo b)`
`edge(b forwardTo c)`

Why should we provide 2 almost identical DSLs with just small differences?

The current `linearSteps(a then b then c, ..params)` solves specific problem:
you have high-level subgraphs/nodes, and you want to apply them one by one easily, and also sometimes you may want to have some compressions/other changes in that pipeline. That's quite popular for the whole graph, and can be optimized. Also you don't have to switch from your high-level components to low-level `nodeStart`, `nodeFinish`.

If someone doesn't fit into this use case, that's very likely some non-linear (in general) strategy. Hence, they would use more "advanced" (i.e. requiring more detailed attention and understanding) API with ordinary nodes and edges.

ptitjes: A couple of notes:

- I am not much of a fan of the agent's graph being called a strategy. This leads to the `AIAgentStrategy` builder which IMO should not exist. `AIAgentSubgraphBuilderBase` should simply be called `AgentGraphBuilder` and an `Agent<String, String>` (rather than `AIAgent`...) should simply take a `AgentGraph<String, String>` and not an `AIAgentStrategy`.

- I was hopping that when Context Parameters are in the language, we could get rid of `edge`. And have the `AgentGraphBuilder` be a context parameter of the Graph builder function.

```kotlin
fun graph(
    name: String,
    toolSelectionStategy: ToolSelectionStrategy,
    init: context(AgentGraphBuilder) () -> Unit,
)

context(AgentGraphBuilder)
infix fun <O, I> AgentNode<*, O> forwardTo(otherNode: AgentNode<I, *>): AgentEdgeBuilder<O, O, I>

context(AgentGraphBuilder)
infix fun <IO, OI, OO> AgentNode<IO, OI> then(otherNode: AgentNode<OI, OO>): AgentNode<OI, OO>
```

So that we can do:
```kotlin
val graph = graph("qa") {
    val llmRequest by llmRequest()

    start forwadTo llmRequest
    llmRequest forwardTo finish

    // Or here,
    // start then llmRequest then finish 
}

val agent = Agent(
    graph = graph,
    // ...
)
```

- Finally having a `linearSteps` builder just to be able to add a node in-between each edge doesn't seem a great idea to me. I would propose that an agent graph has by default the ability to introduce intermediate nodes to each edge. And that in-fine from the user's point of view you only have to use a subgraph to instrument the edges.

For instance, in the following graph definition, a historyCompression node would be inserted between each of the nodes of the "qa" subgraph. Also the `maxAgentInteractions` would be temporarily set to 50 during the execution of the "qa" subgraph.
```kotlin

val graph = graph("qa-loop") {
    val userInput by userInput()

    val qa = subgraph(
        name = "qa",
        maxAgentInteractions = 50,
        historyCompressionStategy = NoCompression,
    ) {
        val llmRequest by llmRequest()
        val executeTools by executeToolCalls()

        start then llmRequest
        llmRequest forwardTo executeTools onToolCalls { true }
        llmRequest forwardTo finish onToolCalls { false }
    }

    start then userInput
    userInput forwardTo qa onUserMessage { it == "Quit" }
    userInput forwardTo finish onUserMessage { it != "Quit" }
    qa then start
}
```

EugeneTheDev: +1 for just `graph`, I also suggested something like this before

Ololoshechkin: WDYT about `graphStrategy` ?

`graph` sounds too generic to me and doesn't reflect the connection to AI Agents.

EugeneTheDev: I thought it would be nice to remove `strategy` and `subgraph` completely from api, and leave only `graph`. Because under the hood it's actually the same implementation (`AIAgentSubgraph`). So we can have only graph entity, and it's usage (is it a strategy or a subgraph of another strategy) would be determined by where we pass it (in the agent/embedding it in other strategy). We also touched this topic in our discussions before briefly. But maybe that's too radical, although I believe it might make things a bit more straightforward and reduce the number of entities in user-facing API

tiginamaria: I propose to add `sequence` inside the `graph` instead of defining it as a strategy, because one of the common case is also to build linear `subgraph` which probably should have common interface. Moreover, as for parallel execution, I'm thinking about adding `parallel` so they will be named in the same manner. Also I'd rather keep naming as simple as it could be and having no camel case in dsl, as it looks just nice like this.

Ololoshechkin: I like the sequence/parallel proposal.

But still think that "graph" looks like an implementation detail for the problem of defining agent strategies

EugeneTheDev:
Since this PR become stale and out of sync, I'm closing it. If we still want to, we can revisit and reimplement this idea

PR #219, opened 2025-06-04
Move Tool.Args from Tool

tiginamaria:
- [x] Move `Tool.Args` from Tool to separate `ToolArgs` class
- [ ] Remove explicit args serialized from `Tool` class 

The numerous attempts to fix the second bullet like showed that this refactoring can not be done due to the `TArgs` type erasure in `Tool` -> we are not able to create object from string (received from LLM). So the only way to have the **serializer** in singleton instance of the `Tool` class in `ToolRegistry`

### DIFF
--- a/agents/agents-tools/src/commonMain/kotlin/ai/koog/agents/core/tools/Tool.kt
+++ b/agents/agents-tools/src/commonMain/kotlin/ai/koog/agents/core/tools/Tool.kt
@@ -3,7 +3,6 @@ package ai.koog.agents.core.tools
 import ai.koog.agents.core.tools.annotations.InternalAgentToolsApi
 import ai.koog.agents.core.tools.serialization.ToolJson
 import kotlinx.serialization.KSerializer
-import kotlinx.serialization.Serializable
 import kotlinx.serialization.json.JsonObject
 import kotlinx.serialization.json.jsonObject
 
@@ -32,7 +31,7 @@ public interface DirectToolCallsEnabler
  * Represents a tool that, when executed, makes changes to the environment.
  */
 @Suppress("UNCHECKED_CAST", "unused")
-public abstract class Tool<TArgs : Tool.Args, TResult : ToolResult> {
+public abstract class Tool<TArgs : ToolArgs, TResult : ToolResult> {
     /**
      * Serializer responsible for encoding and decoding the arguments required for the tool execution.
      * This abstract property is used to define the specific [KSerializer] corresponding to the type of arguments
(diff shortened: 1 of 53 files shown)


Ololoshechkin (approved):
@tiginamaria Please also tell @innateteniuk about this so that she adjusts the documentation

agents/agents-tools/src/commonMain/kotlin/ai/koog/agents/core/tools/Tool.kt
```diff
@@ -32,7 +31,7 @@ public interface DirectToolCallsEnabler
  * Represents a tool that, when executed, makes changes to the environment.
  */
 @Suppress("UNCHECKED_CAST", "unused")
-public abstract class Tool<TArgs : Tool.Args, TResult : ToolResult> {
+public abstract class Tool<TArgs : ToolArgs, TResult : ToolResult> {
```
tiginamaria, line 36: Please take a look at how `Tool` class was refactored. I've tried to remove methods used only for testing and God method `executeAndSerialize` as it was not useful anymore

agents/agents-tools/src/commonMain/kotlin/ai/koog/agents/core/tools/Tool.kt
```diff
@@ -28,11 +27,12 @@ import kotlinx.serialization.json.jsonObject
...
 public interface DirectToolCallsEnabler
 
+
 /**
  * Represents a tool that, when executed, makes changes to the environment.
  */
 @Suppress("UNCHECKED_CAST", "unused")
-public abstract class Tool<TArgs : Tool.Args, TResult : ToolResult> {
+public abstract class Tool<TArgs : ToolArgs, TResult : ToolResult> {
     /**
```
Ololoshechkin, line 37: Let's also keep the Tool.Args here with a deprecation and replaceWith warning?

Ololoshechkin:
```kotlin
@Deprecated("Use ToolArgs instead", ReplaceWith("ToolArgs", "ai.koog.agents.core.tools.ToolArgs"))
public interface Args : ToolArgs
```

Ololoshechkin: And we'll remove it after the next-next release

agents/agents-tools/src/commonMain/kotlin/ai/koog/agents/core/tools/ToolArgs.kt
```diff
@@ -0,0 +1,15 @@
+package ai.koog.agents.core.tools
+
+import kotlinx.serialization.Serializable
+
+/**
+ * Represents the arguments for a tool operation.
+ */
```
Ololoshechkin, line 9: Let's also state that it should be Serializable ?

Ololoshechkin (approved):
@tiginamaria please check my comments before merging

PR #220, opened 2025-06-04
Add parallel nodes invocation

tiginamaria:
Add parallel nodes execution implementation
https://github.com/JetBrains/koog/issues/255

### DIFF
--- a/agents/agents-test/src/commonMain/kotlin/ai/koog/agents/testing/tools/DummyAIAgentContext.kt
+++ b/agents/agents-test/src/commonMain/kotlin/ai/koog/agents/testing/tools/DummyAIAgentContext.kt
@@ -15,6 +15,13 @@ import org.jetbrains.annotations.TestOnly
 import kotlin.uuid.ExperimentalUuidApi
 import kotlin.uuid.Uuid
 
+/**
+ * A mock implementation of the [AIAgentContextBase] interface, used for testing purposes.
+ *
+ * @constructor Creates a new instance of `DummyAIAgentContext` using a predefined
+ * `AIAgentContextMockBuilder`.
+ * @param builder A builder object used to initialize the mock properties of the context.
+ */
 @TestOnly
 @OptIn(ExperimentalUuidApi::class)
 public class DummyAIAgentContext(


agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/builder/AIAgentSubgraphBuilder.kt
```diff
@@ -56,11 +60,29 @@ public abstract class AIAgentSubgraphBuilderBase<Input, Output> {
...
+     * Executes multiple nodes in parallel and merges their contexts.
+     *
+     * @param nodes List of nodes to execute in parallel
+     * @param reduce Function to merge the contexts after parallel execution
+     * @param name Optional name for the parallel node
+     * @return A node that executes the given nodes in parallel
+     */
+    public fun <Input, Output> parallel(
+        vararg nodes: AIAgentNodeBase<Input, Output>,
+        reduce: suspend AIAgentContextBase.(List<Triple<String, AIAgentContextBase, Output>>) -> Pair<AIAgentContextBase, Output>,
```
tiginamaria: Should I make reduce separate node?

tiginamaria: Should I use Triple here or create separate data class like `ParallelNodeResult`?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/context/AIAgentContext.kt
```diff
@@ -31,17 +31,57 @@ import kotlin.uuid.Uuid
...
+    llm: AIAgentLLMContext,
+    stateManager: AIAgentStateManager,
+    storage: AIAgentStorage,
+    sessionUuid: Uuid,
+    strategyId: String,
     @OptIn(InternalAgentsApi::class)
-    override val pipeline: AIAgentPipeline,
+    pipeline: AIAgentPipeline,
 ) : AIAgentContextBase {
+    private var _environment: AIAgentEnvironment = environment
```
tiginamaria: To make attributes private set, but not sure this is the right way to go. Looks not so ugly as we discussed to make `AIAgentContextDelegate` with `AIAgentContext` context inside

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/context/AIAgentContext.kt
```diff
@@ -31,17 +31,57 @@ import kotlin.uuid.Uuid
...
+    override val sessionUuid: Uuid
+        get() = _sessionUuid
+
+    override val strategyId: String
+        get() = _strategyId
+
+    @OptIn(InternalAgentsApi::class)
+    override val pipeline: AIAgentPipeline
+        get() = _pipeline
+
```
Rizzen: These are basically the same, but a bit decompiled :D
I dont really understand why we want to do that, because it's just conversion from val fields to properties with baking field without any logic on get, so it's the same but written other way.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/builder/AIAgentSubgraphBuilder.kt
Rizzen: I would prefer to see data class instead of triple here, yes

Rizzen: yeah, it would be useful and consistent

agents/agents-tools/src/commonMain/kotlin/ai/koog/agents/core/tools/ToolDescriptors.kt
```diff
@@ -75,7 +75,33 @@ public sealed class ToolParameterType(public val name: kotlin.String) {
...
+            if (this === other) return true
+            if (other == null || this::class != other::class) return false
+
+            other as Enum
+
+            if (!entries.contentEquals(other.entries)) return false
+
+            return true
+        }
+    }
```
Rizzen: What the purpose on overriding default data class equality comparers?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/builder/AIAgentSubgraphBuilder.kt
```diff
@@ -137,3 +159,35 @@ public open class AIAgentSubgraphDelegate<Input, Output> internal constructor(
...
+                async {
+                    val nodeContext = initialContext.copy()
+                    val result = node.execute(nodeContext, input)
+                    Triple(node.name, nodeContext, result)
+                }
+            }
+
+            val results = mapResults.awaitAll()
+
+            reduce(results)
```
Rizzen: I think it would be good to provide ability to schedule such parallel execution on user provided Dispatcher, with default = Dispatchers.Default

Rizzen (changes requested):

Rizzen:
@tiginamaria could you also rebase on latest develop? You have QD inspections because of little bit outdated branch

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/context/AIAgentContext.kt
Rizzen: Ah, I see private VAR here, but why do we need that?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/context/AIAgentContext.kt
```diff
@@ -119,4 +159,22 @@ internal class AIAgentContext(
...
+        _agentInput = context.agentInput
+        _config = context.config
+        _llm = context.llm
+        _stateManager = context.stateManager
+        _storage = context.storage
+        _sessionUuid = context.sessionUuid
+        _strategyId = context.strategyId
+        _pipeline = @OptIn(InternalAgentsApi::class) this.pipeline
+    }
+}
```
Rizzen: I suggest make copy on write, because making mutable context in concurrent environment is good way to provide good amount of leg shooting in future

agents/agents-tools/src/commonMain/kotlin/ai/koog/agents/core/tools/ToolDescriptors.kt
tiginamaria: This is something that should not be there...

tiginamaria: This is something that should not be there...

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/context/AIAgentContext.kt
tiginamaria: I'm unable to do copy on write as I we have no control on context instance from outside the graph (only from nodes, but they are executed “on” context, so I can not substitute "this")

agents/agents-core/src/commonTest/kotlin/ai/koog/agents/core/dsl/extension/ParallelNodesTest.kt
```diff
@@ -0,0 +1,107 @@
...
+                "Result from node3"
+            }
+
+            // Create a parallel node that executes all three nodes
+            val parallelNode by parallel<Unit, String>(
+                node1, node2, node3,
+                name = "parallelNode",
+            )
+
+            val reduceNode by reduce<Unit, String>(name = "reduceNode") { results ->
```
Ololoshechkin: parallel + reduce -- is it a good pair of names?

Also, is there a way to not call reduce right after parallel?

Something like parallel(....).transform(...).transform(...).transform(...).reduce(...)

Something similar to:

myList.map { // parallel
   it.toInt()
}.map { // transform
   it + 5
}.map{ // transform 
   println(it)
   it-1
}.max() // reduce

tiginamaria: Map is taken so maybe fork + reduce?

tiginamaria: And yes, if you make a node from list of parallel result to the list of parallel results you can you, but I can make such a node with hidden types lite `transformParallel<Input, Output>` otherwise you need to declare it like this `node<List<ParallelNodeResult<Input, Output>>, List<ParallelNodeResult<Input, Output>>>`

tiginamaria: Or fork + merge?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/builder/AIAgentSubgraphBuilder.kt
```diff
@@ -88,11 +91,39 @@ public abstract class AIAgentSubgraphBuilderBase<Input, Output> {
...
+     * @param execute Function to merge the contexts after parallel execution
+     * @param name Optional node name
+     */
+    public fun <Input, Output> merge(
+        name: String? = null,
+        execute: suspend AIAgentContextBase.(List<ForkedNodeResult<Input, Output>>) -> Pair<AIAgentContextBase, Output>,
+    ): AIAgentNodeDelegateBase<List<ForkedNodeResult<Input, Output>>, Output> {
+        return AIAgentNodeDelegate(name, MergeNodeBuilder(execute))
+    }
+
```
ptitjes: One drawback I see in this design is the impossibility to do different things and have different output types in the parallel nodes. There are many use cases where you'll want to have parallel processing with different outputs (e.g. advanced rag, etc.) Having a unique node that does both the forking and the merge would allow to have different outputs and still keep type-safety.

Will you guys support these use-cases? Something like:

```kotlin
fun <Input, Output1, Output2, Output3, Output> parallel(
    name: String? = null,
    dispatcher: CoroutineDispatcher = Dispatchers.Default,
    node1: AIAgentNodeBase<Input, Output1>,
    node2: AIAgentNodeBase<Input, Output2>,
    node3: AIAgentNodeBase<Input, Output3>,
    merge: suspend AIAgentContextBase.(
        ForkedNodeResult<Input, Output1>,
        ForkedNodeResult<Input, Output2>,
        ForkedNodeResult<Input, Output3>,
    ) -> Pair<AIAgentContextBase, Output>,
): AIAgentNodeDelegateBase<Input, Output>
```

This requires having multiple overloads though, one for each number of parallel nodes.

tiginamaria: Fair point, thank you! My assumption was that in the most cases you need to do something similar in parallel (call llms, process something) so the output will have the same type. Also, most of the outputs of the nodes are strings right now (so the output type will automatically match). Moreover, in any case, you can store the node result in storage, return Unit and then get one in the merge from context, this is not so fancy, but works. 

So my proposal is leave it as it is rn, but if we will face the usage you proposed, we should consider adding something like you proposed for sure. 
WDYT?

agents/agents-test/src/commonMain/kotlin/ai/koog/agents/testing/tools/DummyAIAgentContext.kt
```diff
@@ -0,0 +1,447 @@
...
+            llm?.let { this.llm = it }
+            stateManager?.let { this.stateManager = it }
+            storage?.let { this.storage = it }
+            sessionUuid?.let { this.sessionUuid = it }
+            strategyId?.let { this.strategyId = it }
+        }
+    )
+
+    override suspend fun fork(): AIAgentContextBase {
+        TODO("Not yet implemented")
```
tiginamaria, line 118: Should I implement fork/replace for the `DummyAIAgentContext`? And why we actually need it (I don't see any functionality that blocks from using just `AIAgentContext`, except all nullable values, or I'm missing something)?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/builder/AIAgentSubgraphBuilder.kt
ptitjes: Sure, this can be added later.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/builder/AIAgentSubgraphBuilder.kt
```diff
@@ -218,3 +249,63 @@ public open class AIAgentSubgraphDelegate<Input, Output> internal constructor(
...
+    private val nodes: List<AIAgentNodeBase<Input, Output>>,
+    private val dispatcher: CoroutineDispatcher
+) : AIAgentNodeBuilder<Input, List<ForkResult<Input, Output>>>(
+    execute = { input ->
+        val initialContext: AIAgentContextBase = this
+        val mapResults = supervisorScope {
+            nodes.map { node ->
+                async(dispatcher) {
+                    val nodeContext = (initialContext as? ai.koog.agents.core.agent.context.AIAgentContext)?.fork()
+                        ?: initialContext.fork()
```
ptitjes: It looks to me like you don't need this first case anymore, now that `fork` is defined on the interface. Or maybe I am missing something.

```suggestion
                    val nodeContext = initialContext.fork()
```

ptitjes (commented):
I was not sure about the cast in the `fork` node. But apart from that LGTM.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/builder/AIAgentSubgraphBuilder.kt
tiginamaria: For sure, thank you! Don't know how come I had this change in my diff

agents/agents-core/src/commonTest/kotlin/ai/koog/agents/core/dsl/extension/ParallelNodesTest.kt
Ololoshechkin: > Map is taken

Sorry, I meant `transform` :)

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/builder/AIAgentSubgraphBuilder.kt
```diff
@@ -88,11 +91,39 @@ public abstract class AIAgentSubgraphBuilderBase<Input, Output> {
...
         return nextNode
     }
 
+    /**
+     * Creates a node that executes multiple nodes in parallel.
+     * @param nodes List of nodes to execute in parallel
+     * @param dispatcher Coroutine dispatcher to use for parallel execution
+     * @param name Optional node name
+     */
+    public fun <Input, Output> fork(
```
Ololoshechkin: Maybe let's call it `parallel` ? Just seems a bit more searchable and common word. WDYT?

examples/src/main/kotlin/ai/koog/agents/example/parallelexecution/BestJokeAgent.kt
```diff
@@ -0,0 +1,146 @@
...
+                    }
+                }
+
+
+                val response = requestLLMStructured(
+                    structure = JsonStructuredData.createJsonStructure<JokeWinner>(
+                        schemaFormat = JsonSchemaGenerator.SchemaFormat.JsonSchema,
+                        schemaType = JsonStructuredData.JsonSchemaType.FULL
+                    )
+                )
```
Ololoshechkin: nit: this can be simplified to just `val response = requestLLMStructured(JsonStructuredData.createJsonStructure<JokeWinner>())`

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/builder/AIAgentSubgraphBuilder.kt
Ololoshechkin: Also, I still can't find a way to write more than a single step in the parallel pipeline.
"fork" takes Input and gives you a  List<ForkResult<Input, Output>>
The only node that takes `List<ForkResult<Input, Output>>` and can do something with it is "merge".

so the only option is to write "fork" and then immediately "merge".

Let's add some way to transform `List<ForkResult<Input, Output>>` to `List<ForkResult<Input, AnotherOutput>>`, for example with a transformation lambda.

Then we can write something like (I'll also change `fork` to `parallel` in my example:

```
val openAIJoke by node { ... }
val anthropicJoke by node { ... }
val googleJoke by node { ... }


val generateJokes by parallel(openAIJoke, anthropicJoke, googleJoke)
val criticiseJokes by transform(generateJokes) { 
    llm.writeSession {
       ...
    }
 }
 val selectJoke by merge(...) { ... }
```

Ololoshechkin: Just made an experiment and realized that to support this we would need to have `AsyncForkResult` and `ForkResult` independently (don't take my naming and specific API as the silver bullet, it's just a rough proposal):
```
public data class ResultWithContext<Output>(val context: AIAgentContextBase, val output: Output)

public data class AsyncParallelResult<Input, Output>(
    val nodeName: String,
    val input: Input,
    val asyncResult: Deferred<ResultWithContext<Output>>
) {
    public suspend fun await(): ForkResult<Input, Output> = ParallelResult(nodeName, input, asyncResult.await())
}

public data class ParallelResult<Input, Output>(
    val nodeName: String,
    val input: Input,
    val result: ResultWithContext<Output>
)
```

And then you don't have to `await()` in the `parallel` (formerly `fork`) function -- just leave all async results as Deferred. You should only "await"/"join" them in the final `merge` function:


- `parallel` : 
```
public fun <Input, Output> parallel(
        vararg nodes: AIAgentNodeBase<Input, Output>,
        dispatcher: CoroutineDispatcher = Dispatchers.Default,
        name: String? = null,
    ): AIAgentNodeDelegateBase<Input, List<AsyncParallelResult<Input, Output>>> {
        return AIAgentNodeDelegate(name, AIAgentForkNodeBuilder(nodes.asList(), dispatcher))
}

public class AIAgentForkNodeBuilder<Input, Output> internal constructor(
    private val nodes: List<AIAgentNodeBase<Input, Output>>,
    private val dispatcher: CoroutineDispatcher
) : AIAgentNodeBuilder<Input, List<AsyncParallelResult<Input, Output>>>(
    execute = { input ->
        val initialContext: AIAgentContextBase = this
        val mapResults = supervisorScope {
            nodes.map { node ->
                val asyncResult = async(dispatcher) {
                    val nodeContext = initialContext.fork()
                    val result = node.execute(nodeContext, input)

                    ResultWithContext(nodeContext, result)
                }

                AsyncParallelResult(node.name, input, asyncResult)
            }
        }

        mapResults
    }
)
```


- `transform` :


```

public fun <Input, OldOutput, NewOutput> transform(
        name: String? = null,
        transform: suspend AIAgentContextBase.(OldOutput) -> NewOutput,
    ): AIAgentNodeDelegateBase<List<AsyncParallelResult<Input, OldOutput>>, List<AsyncParallelResult<Input, NewOutput>>> {
        return AIAgentNodeDelegate(name, AIAgentParallelTransformNodeBuilder(transform))
    }
    
    
public class AIAgentParallelTransformNodeBuilder<Input, OldOutput, NewOutput> internal constructor(
    transform: suspend AIAgentContextBase.(OldOutput) -> NewOutput,
) : AIAgentNodeBuilder<List<AsyncParallelResult<Input, OldOutput>>, List<AsyncParallelResult<Input, NewOutput>>>(
    execute = { input ->
        supervisorScope {
            input.map {
                AsyncParallelResult(it.nodeName, it.input, async {
                    val result = it.asyncResult.await()
                    with(result.context) {
                        ResultWithContext(this@with, transform(result.output))
                    }
                })
            }
        }
    }
)
```



- `merge`:

```
public fun <Input, ParallelOutput, MergedOutput> merge(
        name: String? = null,
        execute: suspend AIAgentContextBase.(List<ParallelResult<Input, ParallelOutput>>) -> Pair<AIAgentContextBase, MergedOutput>,
    ): AIAgentNodeDelegateBase<List<AsyncParallelResult<Input, ParallelOutput>>, MergedOutput> {
        return AIAgentNodeDelegate(name, AIAgentMergeNodeBuilder(execute))
    }
    
    
    
public class AIAgentMergeNodeBuilder<Input, ParallelOutput, MergedOutput> internal constructor(
    execute: suspend AIAgentContextBase.(List<ParallelResult<Input, ParallelOutput>>) -> Pair<AIAgentContextBase, MergedOutput>,
) : AIAgentNodeBuilder<List<AsyncParallelResult<Input, ParallelOutput>>, MergedOutput>(
    execute = { input ->
        val (context, output) = execute(input.map { it.await() })
        this.replace(context)

        output
    }
)
```



And then you can write something like:

```
val nodeGenerateJokes by parallel(nodeOpenAI, nodeAnthropicSonnet, nodeAnthropicOpus,)

val transformJokes by transform<String, String, String> { joke -> "This is joke:\n$joke" }

val nodeSelectBestJoke by merge<String, String, String>() { results -> .... }
```

Ololoshechkin: Another thing that I added was -- merge now can produce different output type other than initial one. In the initial API it was only possible to "select" the result or concatenate/add primitives (String, numbers) as it was String -> String.

Ololoshechkin: Another issue that would be great to address is - having Pair<Context, Output> in the API and having to explicitly define it. For sure, as a generic case it makes sense, but let's add at least some defaults that would allow one not to think about context explicitly at least in the very simple use-cases. The ones I can think about are:

- selection (as in your example about jokes)
- folding (like in `fold(list, {a, b -> a+b}`)

Ololoshechkin: And also once we have `val p by parallel`, `val t by transform` and `val m by merge` it should be possible to also define parallel computations with a DSL that would create explicit nodes under the hood.

Instead of 

```
val nodeGenerateJokes by parallel(nodeOpenAI, nodeAnthropicSonnet, nodeAnthropicOpus)
val transformJokes by transform<String, String, String> { joke -> "This is joke:\n$joke" }
val nodeSelectBestJoke by merge<String, String, String>() { results -> .... }
```

and instead of having to define explicit <Input, OldOutput, NewOutput> types all the time, it can be made all implicit with a DSL:

```
val selectBestJoke by parallel( // create "parallel" parallel node(s) under the hood
    nodeOpenAI, nodeAnthropicSonnet, nodeAnthropicOpus,
) transform { joke ->  // create "transform-1" parallel node(s) under the hood
    "This is joke:\n$joke"
} transform { wrappedJoke -> // create another "transform-2" parallel node(s) under the hood
    wrappedJoke.removePrefix("This is joke:\n")
} select { results -> // create "merge" node under the hood
   ...
}
```


which without my comments would look very short and clear:


```
val selectBestJoke by parallel(nodeOpenAI, nodeAnthropicSonnet, nodeAnthropicOpus) transform { joke -> 
    "This is joke:\n$joke"
} transform { wrappedJoke ->
    wrappedJoke.removePrefix("This is joke:\n")
} select { results -> 
   ...
}
```

The `select` extension might have an API like this:

```

/**
* User has to select an index of the best output.
* Context merge strategy is -- selection of one. 
*/
fun ParallelChainBuilder<Input, Output>.select(selectIndex: suspend AIAgentContextBase.(List<Output>) -> Int) ...
```



And we can also provide other optons like `fold` for example

```
/**
* User has to define how to combine all outputs into one by applying some operation
* Context merge strategy can be customized, but by default it's merging into one
*/
fun <Output> fold(
    mergeContexts: (suspend AIAgentContextBase.(AIAgentContextBase, AIAgentContextBase) -> AIAgentContextBase) = { ctx1, ctx2 ->
        copy(
            llm = llm.copy(
                prompt = prompt.copy(
                    messages = ctx1.llm.prompt.messages + ctx2.llm.prompt.messages
                )
            )
        )
    },
    operation: suspend AIAgentContextBase.(Output, Output) -> Int
) ...
```

for example, joke concatenation would look like:
```
val allJokes by parallel(googleJoke, anthropicJoke, openAIJoke, ollamaJoke, deepSeekJoke) fold { joke1, joke2 -> joke1 + joke2 }
```

And of course, we would leave a generic `merge` that would require context selection.

tiginamaria: Thank you for your comment and suggestions! I've added the `transform` node so now it's possible to apply transformations to parallel. About the rest dsl you proposed, let's introduce it next PR as it deserves separate discussion and not so trivial design

Ololoshechkin (approved):
@tiginamaria I pushed a small change , please take a look :)

Ololoshechkin:
@tiginamaria please add more tests for the `AIAgentParallelNodesMergeContext` methods (let's cover them)

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/context/AIAgentContext.kt
```diff
@@ -119,4 +167,30 @@ internal class AIAgentContext(
...
+     * @param context The context to replace the current context with.]]
+     */
+    override suspend fun replace(context: AIAgentContextBase) {
+        mutableAIAgentContext.replace(
+            context.llm,
+            context.stateManager,
+            context.storage
+        )
+    }
+}
```
Rizzen, line 196: Maybe fork+join for consistency? :)

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/entity/AIAgentState.kt
```diff
@@ -44,4 +50,10 @@ public class AIAgentStateManager internal constructor(
 
         result
     }
+
+    public suspend fun copy(): AIAgentStateManager {
```
Rizzen: Are we sure in public here? Internal maybe?

Rizzen (approved):

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/context/AIAgentContext.kt
tiginamaria: It's definitely `fork` (or deepCopy) and `replace` (supstitute all the mutable fields with given) by implementation, don't think it should be `join`

ptitjes: `clone`/`replace`? 😊

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/context/AIAgentContext.kt
```diff
@@ -119,4 +167,30 @@ internal class AIAgentContext(
...
+    )
+
+    /**
+     * Replaces the current context with the provided context.
+     * This method is used to update the current context with values from another context,
+     * particularly useful in scenarios like parallel node execution where contexts need to be merged.
+     *
+     * @param context The context to replace the current context with.]]
+     */
+    override suspend fun replace(context: AIAgentContextBase) {
```
EugeneTheDev, line 189: There's a potential problem with this method that might confuse someone (it confused me initially, at least). From the method signature and according to KDocs it looks like the whole context will be replaced, but in fact it's not the case and only three properties are actually replaced.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/context/AIAgentContext.kt
```diff
@@ -30,18 +31,65 @@ import kotlin.uuid.Uuid
...
+                stateManager?.let { this.stateManager = stateManager }
+                storage?.let { this.storage = storage }
+            }
+        }
+    }
+
+    private val mutableAIAgentContext = MutableAIAgentContext(llm, stateManager, storage)
+
+    override val llm: AIAgentLLMContext
+        get() = mutableAIAgentContext.llm
```
EugeneTheDev, line 85: Also, for all these properties there's no RW lock. Meaning, when the replace operation is in progress, it's still possible to read from these fields, potentially getting inconsistent data (although the chances are probably very low)

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/context/AIAgentContext.kt
```diff
@@ -30,18 +31,65 @@ import kotlin.uuid.Uuid
...
+        var stateManager: AIAgentStateManager,
+        var storage: AIAgentStorage,
+    ) {
+        private val rwLock = RWLock()
+
+        /**
+         * Creates a copy of the current [MutableAIAgentContext].
+         * @return A new instance of [MutableAIAgentContext] with copies of all mutable properties.
+         */
+        suspend fun copy(): MutableAIAgentContext {
```
EugeneTheDev, line 61: This method is not used

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/context/AIAgentLLMContext.kt
```diff
@@ -36,6 +36,26 @@ public data class AIAgentLLMContext(
...
+     * Creates a deep copy of this LLM context.
+     *
+     * @return A new instance of [AIAgentLLMContext] with deep copies of mutable properties.
+     */
+    public suspend fun copy(): AIAgentLLMContext {
+        return rwLock.withReadLock {
+            AIAgentLLMContext(
+                tools.toList(),
+                toolRegistry,
+                prompt.copy(),
```
EugeneTheDev, line 49: There's no need to unnecessary copy prompt and tools, these are already fully immutable

examples/src/main/kotlin/ai/koog/agents/example/parallelexecution/BestJokeAgent.kt
```diff
@@ -0,0 +1,158 @@
...
+            "My favorite joke: $joke"
+        }
+
+        val nodeSelectBestJoke by merge<String, String> {
+            selectByIndex { jokes ->
+                // Another LLM (ex: GPT4o) would find the funniest joke:
+                llm.writeSession {
+                    model = OpenAIModels.Chat.GPT4o
+                    updatePrompt {
+                        prompt("best-joke-selector") {
```
EugeneTheDev, line 90: This part wouldn't work. `updatePrompt` accepts `PromptBuilder` and in its lambda returns `Unit`. It is meant to add more messages to existing prompt. In this case a new prompt is created and returned, but since the lambda is `Unit`, this operation essentially has no effect and the prompt remains the same.

PR #229, opened 2025-06-05
[prompt] Fix LLM clients after #195, make LLM request construction again more explicit in LLM clients

EugeneTheDev:
Fix #195 to make Anthropic client work properly again and to also make request construction in LLM clients again more explicit


Ololoshechkin (approved):
@EugeneTheDev please also create GH issue , mention in MR and change the MR name so that it explains the problem and the solution concisely

PR #252, opened 2025-06-10
Added support for nullable tool parameters

denyska:
Added support for nullable tool parameters. A lot of useful MCP servers are developed with popular Python FastMCP library. For ex., this [graphiti_mcp_server.py#L675](https://github.com/getzep/graphiti/blob/e839c774aa50cfc4abd5757a340d56396302afe0/mcp_server/graphiti_mcp_server.py#L675). FastMCP treats nullable params as `anyOf`, i.e. `my_param: str | None = None` is getting translated into 
```
            "my_param": {
              "anyOf": [{
                          "type": "string"
            		     },
                       {
                          "type": "null"
                       }],
              "title": "Nullable string parameter"
            }
```

kotlinx.serialization.json.JsonBuilder#getExplicitNulls is not disabled, hence null Args are not getting serialized by `ai.koog.agents.mcp.McpTool.ArgsSerializer`
---

#### Type of the change
- [ ] New feature
- [x] Bug fix
- [ ] Documentation fix

### DIFF
--- a/agents/agents-mcp/src/jvmMain/kotlin/ai/koog/agents/mcp/McpToolDefinitionParser.kt
+++ b/agents/agents-mcp/src/jvmMain/kotlin/ai/koog/agents/mcp/McpToolDefinitionParser.kt
@@ -45,7 +45,33 @@ public object DefaultMcpToolDescriptorParser : McpToolDescriptorParser {
     private fun parseParameterType(element: JsonObject): ToolParameterType {
         // Extract the type string from the JSON object
         val typeStr = element["type"]?.jsonPrimitive?.content
-            ?: throw IllegalArgumentException("Parameter type must have type property")
+
+        if (typeStr == null) {
+            // special case for nullable types
+            //"nullableParam": {
+            //  "anyOf": [{
+            //              "type": "string"
+            //		     },
+            //           {
+            //              "type": "null"
+            //           }],
+            //  "title": "Nullable string parameter"
+            //}
+            val anyOf = element["anyOf"]?.jsonArray
+            if (anyOf != null && anyOf.size == 2) {
+                val types = anyOf.map { it.jsonObject["type"]?.jsonPrimitive?.content }
+                if (types.contains("null")) {
+                    val nonNullType = anyOf.first {
+                        it.jsonObject["type"]?.jsonPrimitive?.content != "null"
+                    }.jsonObject
+                    return parseParameterType(nonNullType)
+                }
+            }
+            val title =
+                element["title"]?.jsonPrimitive?.content ?: element["description"]?.jsonPrimitive?.content.orEmpty()
+            throw IllegalArgumentException("Parameter $title must have type property")
+        }
+
 
         // Convert the type string to a ToolParameterType
         return when (typeStr.lowercase()) {
--- a/agents/agents-mcp/src/jvmTest/kotlin/ai/koog/agents/mcp/McpToolTest.kt
+++ b/agents/agents-mcp/src/jvmTest/kotlin/ai/koog/agents/mcp/McpToolTest.kt
@@ -61,6 +61,13 @@ class McpToolTest {
                         type = ToolParameterType.String,
                         description = "A name to greet",
                     )
+                ),
+                optionalParameters = listOf(
+                    ToolParameterDescriptor(
+                        name = "title",
+                        type = ToolParameterType.String,
+                        description = "Title to use in the greeting",
+                    )
                 )
             )
         )
@@ -81,5 +88,18 @@ class McpToolTest {
 
         val content = result.promptMessageContents.first() as TextContent
         assertEquals("Hello, Test!", content.text)
+
+        val argsWithTitle = McpTool.Args(buildJsonObject {
+            put("name", "Test")
+            put("title", "Mr.")
+        })
+        val (resultWithTitle, _) = withContext(Dispatchers.Default.limitedParallelism(1)) {
+            withTimeout(1.minutes) {
+                greetingTool.executeAndSerialize(argsWithTitle, TestToolEnabler)
+            }
+        }
+
+        val contentWithTitle = resultWithTitle.promptMessageContents.first() as TextContent
+        assertEquals("Hello, Mr. Test!", contentWithTitle.text)
     }
 }
--- a/agents/agents-mcp/src/jvmTest/kotlin/ai/koog/agents/mcp/DefaultMcpToolDescriptorParserTest.kt
+++ b/agents/agents-mcp/src/jvmTest/kotlin/ai/koog/agents/mcp/DefaultMcpToolDescriptorParserTest.kt
@@ -93,17 +93,61 @@ class DefaultMcpToolDescriptorParserTest {
                     put("type", "string")
                     put("description", "String parameter")
                 }
+                putJsonObject("nullableStringParam") {
+                    putJsonArray("anyOf") {
+                        addJsonObject {
+                            put("type", "null")
+                        }
+                        addJsonObject {
+                            put("type", "string")
+                        }
+                    }
+                    put("description", "Nullable string parameter")
+                }
                 putJsonObject("integerParam") {
                     put("type", "integer")
                     put("description", "Integer parameter")
+                }
+                putJsonObject("nullableIntegerParam") {
+                    putJsonArray("anyOf") {
+                        addJsonObject {
+                            put("type", "null")
+                        }
+                        addJsonObject {
+                            put("type", "integer")
+                        }
+                    }
+                    put("description", "Nullable integer parameter")
                 }
                 putJsonObject("numberParam") {
                     put("type", "number")
                     put("description", "Number parameter")
                 }
+                putJsonObject("nullableNumberParam") {
+                    putJsonArray("anyOf") {
+                        addJsonObject {
+                            put("type", "null")
+                        }
+                        addJsonObject {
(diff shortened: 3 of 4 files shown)


agents/agents-mcp/src/jvmMain/kotlin/ai/koog/agents/mcp/McpToolDefinitionParser.kt
```diff
@@ -17,6 +17,9 @@ public interface McpToolDescriptorParser {
  * Default implementation of [McpToolDescriptorParser].
  */
 public object DefaultMcpToolDescriptorParser : McpToolDescriptorParser {
+    // Maximum depth of recursive parsing
+    private const val MAX_DEPTH = 10
```
tiginamaria: Can you add this value to the constructor, in case someone will want to have > 10 recursion limit?

agents/agents-mcp/src/jvmMain/kotlin/ai/koog/agents/mcp/McpToolDefinitionParser.kt
```diff
@@ -42,10 +45,39 @@ public object DefaultMcpToolDescriptorParser : McpToolDescriptorParser {
...
+                }
+            }
+            val title =
+                element["title"]?.jsonPrimitive?.content ?: element["description"]?.jsonPrimitive?.content.orEmpty()
+            throw IllegalArgumentException("Parameter $title must have type property")
+        }
+
 
         // Convert the type string to a ToolParameterType
         return when (typeStr.lowercase()) {
```
tiginamaria, line 96: Could you please add recursion limit to all recursion `parseParameterType` calls? 
see line 98

agents/agents-mcp/src/jvmTest/kotlin/ai/koog/agents/mcp/McpToolTest.kt
```diff
@@ -81,5 +88,18 @@ class McpToolTest {
 
         val content = result.promptMessageContents.first() as TextContent
         assertEquals("Hello, Test!", content.text)
+
+        val argsWithTitle = McpTool.Args(buildJsonObject {
+            put("name", "Test")
+            put("title", "Mr.")
+        })
+        val (resultWithTitle, _) = withContext(Dispatchers.Default.limitedParallelism(1)) {
+            withTimeout(1.minutes) {
```
tiginamaria, line 97: Why we need timeout here? Will it make test fluky if it >1m

agents/agents-mcp/src/jvmMain/kotlin/ai/koog/agents/mcp/McpToolDefinitionParser.kt
```diff
@@ -42,10 +45,39 @@ public object DefaultMcpToolDescriptorParser : McpToolDescriptorParser {
...
+             *       { "type": "string" },
+             *       { "type": "null" }
+             *     ],
+             *     "title": "Nullable string parameter"
+             *   }
+             * }
+             */
+            val anyOf = element["anyOf"]?.jsonArray
+            if (depth  < MAX_DEPTH && anyOf != null && anyOf.size == 2) {
+                val types = anyOf.map { it.jsonObject["type"]?.jsonPrimitive?.content }
```
tiginamaria, line 81: Shouldn't we also add this parameter to the `optionalParameters` in `ToolDescriptor` then?

tiginamaria (changes requested):
Thank you for your contribution! It's really interesting to know that FastMCP selected such a way to interpret the arguments, because the most common way to add "nullable" parameters is to include them in the optional field (or simply omit them from the required field).
I'll appreciate it if you can share any simple mcp server (without any special credentials preferably) so I can test this changes on

agents/agents-mcp/src/jvmMain/kotlin/ai/koog/agents/mcp/McpToolDefinitionParser.kt
denyska: This is object (i.e. singleton), adding constructor is not trivial. One option is to change this to class, or change depth to something larger like 30 that is going to work for most of the use cases, yet don't cause stack overflow.

agents/agents-mcp/src/jvmTest/kotlin/ai/koog/agents/mcp/McpToolTest.kt
denyska: This code mirrors the implementation above. If we proceed with refactoring, both instances should be updated concurrently to maintain consistency.

`TestMcpServer` is embedded and lightweight, so it should execute quickly under normal conditions. However, performance may degrade if the host machine is under heavy load or experiencing resource constraints.

agents/agents-mcp/src/jvmMain/kotlin/ai/koog/agents/mcp/McpToolDefinitionParser.kt
denyska: done

agents/agents-mcp/src/jvmMain/kotlin/ai/koog/agents/mcp/McpToolDefinitionParser.kt
denyska: on line 43, required vs. optional parameters are driven by `inputSchema.required` json attribute. 

Looking at the Tools schema produced by simple Python FastMCP library, None (aka nullable) params are not included in the required section. See sample Python FastMCP server and Koog client here https://github.com/denyska/koog_python_mcp

Sample response from Tools endpoint.

```
{
  "tools": [
    {
      "name": "hello",
      "description": "Says hello",
      "inputSchema": {
        "type": "object",
        "properties": {
          "name": {
            "anyOf": [
              {
                "type": "string"
              },
              {
                "type": "null"
              }
            ],
            "default": null,
            "title": "Name"
          }
        },
        "title": "helloArguments"
      }
    }
  ]
}
``` 

image

denyska:
I added Typescript sample MCP server with nullable / optional arguments and schema is even more complex.  See TypeScript MCP server  here https://github.com/denyska/koog_python_mcp/tree/main/typescript_mcp

```json
{
  "tools": [
    {
      "name": "hello",
      "inputSchema": {
        "type": "object",
        "properties": {
          "name": {
            "type": [
              "string",
              "null"
            ]
          }
        },
        "additionalProperties": false,
        "$schema": "http://json-schema.org/draft-07/schema#"
      }
    }
  ]
}
```

agents/agents-mcp/src/jvmMain/kotlin/ai/koog/agents/mcp/McpToolDefinitionParser.kt
```diff
@@ -48,10 +51,46 @@ public object DefaultMcpToolDescriptorParser : McpToolDescriptorParser {
         )
     }
 
-    private fun parseParameterType(element: JsonObject): ToolParameterType {
+    private fun parseParameterType(element: JsonObject, depth: Int = 0 ): ToolParameterType {
+        if (depth >= MAX_DEPTH) {
+            throw IllegalArgumentException(
+                "Maximum recursion depth ($MAX_DEPTH) exceeded. " +
```
harrytmthy, line 57: Hi @denyska, just a small wording suggestion on the exception message:

```kotlin
if (depth >= MAX_DEPTH) {
    throw IllegalArgumentException(
        "Maximum recursion depth ($MAX_DEPTH) exceeded..."
    )
}
```

Since the condition is `>=`, the message **"exceeded"** might be slightly misleading. It technically throws even when `depth == MAX_DEPTH`. You might want to adjust the message for clarity, such as:

```kotlin
"Maximum recursion depth ($MAX_DEPTH) reached..."
```

or, if you want to keep "exceeded", update the condition to `> MAX_DEPTH` instead.

Just a minor thing, but can help avoid confusion.

denyska: Great suggesting, addressed

agents/agents-mcp/src/jvmTest/kotlin/ai/koog/agents/mcp/McpToolTest.kt
tiginamaria: @aozherelyeva can you help me here, please! What is the best way to design test here?

tiginamaria (approved):
Hi! Sorry for waiting, I was unavailable for the last week.
Thank you a lot for example, it's all crystal clear for me now and I know how to reproduce it! 
About your changes in MCP, all is great! 
I have a little doubt about the test, but I've asked tester to take a look, so after her approve we can merge it!

agents/agents-mcp/src/jvmTest/kotlin/ai/koog/agents/mcp/McpToolTest.kt
zarechneva-a: I think the current approach works well. The `TestMcpServer` is actually lightweight, and such test usually runs in a couple of seconds. I believe this approach is also good to indicate the issue in case the test becomes too slow.

denyska:
Thank you, @tiginamaria and @aozherelyeva. Is there anything else that needs to be addressed before this can be merged? I noticed that the Qodana check failed in previous runs, but after reviewing the logs, I wasn't able to identify the root cause.

Issue #253, opened 2025-06-10
Support `data:` URLs for media attachments


ptitjes:
In some of my use-cases, I receive the images' data already in base64 format inside a `data:` URL.
Currently, `MediaContent` only supports local paths. It would need a small refactoring to support `data:` URLs (with mime-type embedded in the data URL).

EugeneTheDev:
Fixed in #264

PR #264, opened 2025-06-11
[prompt] Improve attachments (former media content) support in prompt messages

### LINKED ISSUE #269: Improve attachments (media content) support in prompt messages
We need to make attachments more versatile, unified and extensible to support more data types for more attachment types. E.g. certain users might have some constraints regarding fetching attachment content, so we need to provide the option to pass just a binary content for all attachment types. Also, current code that handles attachments can be refactored after API changes to make it more readable and extensible

EugeneTheDev:
This PR improves on the changes made in #195 and improves existing attachments API, making it more clean, straightforward and versatile. Support providing binary data directly for all attachment types to support more cases, e.g. mobile devices or server that need to retrieve attachment content their own custom way. Also refactor the code, making it more structured and extensible.
Fixes #269, #253


prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/dsl/Prompt.kt
```diff
@@ -17,6 +17,7 @@ import kotlin.time.Duration
  * @property id The unique identifier for the prompt.
  * @property params The language model pa rameters associated with the prompt. Defaults to [LLMParams].
  */
+// FIXME move it from dsl package up to the module root package?
```
EugeneTheDev, line 20: This is not related directly to this PR, but I've noticed it while preparing it and want to raise a question. Currently, `Prompt` class is located in the `dsl` package, which seems kinda strange, since it's not about DSL, but rather a core structure of this whole module, a prompt model. So I think it's better to move it to the root module package or at least together with other models to `model`

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/message/Message.kt
```diff
@@ -78,14 +78,14 @@ public sealed interface Message {
...
-     * @property mediaContent Optional media content attached to the message.
+     * @property attachment Optional media content attached to the message.
      * @property role The role of the message, which is fixed as [Role.User] for this implementation.
      */
     @Serializable
     public data class User(
         override val content: String,
         override val metaInfo: RequestMetaInfo,
-        val mediaContent: List<MediaContent> = emptyList(),
+        val attachment: List<Attachment> = emptyList(),
```
ptitjes: Shouldn't this be called `attachments`?

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/message/Attachment.kt
```diff
@@ -0,0 +1,135 @@
...
+     * URL of the content (e.g. image or a document)
+     */
+    @Serializable
+    public data class URL(val url: String): AttachmentContent
+
+    /**
+     * Binary content represented as byte array.
+     */
+    @Serializable
+    public data class Binary(val data: ByteArray): AttachmentContent {
```
ptitjes: Could we have an additional `Base64` class in the sealed interface hierarchy? That would allow to not have to decode base64 and then re-encode to base64, in the use cases where the data is already in base 64? C.f. https://github.com/JetBrains/koog/issues/253. As an example, when Koog is used behind a REST API, then the data would already be in base 64 format.

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/message/Message.kt
EugeneTheDev: Good catch, thanks

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/message/Attachment.kt
EugeneTheDev: Sounds good, thank you, will support it

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/dsl/AttachmentBuilder.kt
```diff
@@ -1,121 +1,210 @@
...
-     * - **Ollama** — does not support audio
-     * - **OpenAI** — formats: `wav`, `mp3`
-     * - **OpenRouter** — formats: `wav`, `mp3`
+     * @param path Path to local image file
+     * @throws IllegalArgumentException if the path is not valid, the file does not exist, or is not a regular file.
+     */
+    public fun imageFromPath(path: Path) {
+        val fileData = path.fileData()
+        image(Attachment.Image(content = AttachmentContent.Binary.Bytes(path.readByteArray()), format = fileData.extension, fileName = fileData.name))
+    }
```
Ololoshechkin: These 2 methods already are differentiated by the parameter (`url: String` vs `path: Path`), no need to name them explicitly differently. Let's just provide 2 overloads of `image`.
Also having just `image` / `audio` / `document` provides better consistency and looks much simpler for the user imo.

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/dsl/AttachmentBuilder.kt
```diff
@@ -1,121 +1,210 @@
...
+     * Adds [Attachment.Audio] with [AttachmentContent.Binary.Bytes] content from the provided local file path.
      *
-     * @param data The audio data as a byte array
-     * @param format The audio file format (e.g., "mp3", "wav", "ogg")
+     * @param path Path to local audio file
+     * @throws IllegalArgumentException if the path is not valid, the file does not exist, or is not a regular file.
+     */
+    public fun audioFromPath(path: Path) {
+        val fileData = path.fileData()
+        audio(Attachment.Audio(content = AttachmentContent.Binary.Bytes(path.readByteArray()), format = fileData.extension, fileName = fileData.name))
```
Ololoshechkin: Same here, let's just provide 2 overloads of one `audio` method.

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/dsl/AttachmentBuilder.kt
```diff
@@ -1,121 +1,210 @@
...
-     * - Ollama — does not support documents yet
-     * - OpenAI — supports only local PDF files
-     * - OpenRouter — supports only local PDF files
+     * @param path Path to local video file
+     * @throws IllegalArgumentException if the path is not valid, the file does not exist, or is not a regular file.
+     */
+    public fun videoFromPath(path: Path) {
+        val fileData = path.fileData()
+        video(Attachment.Video(content = AttachmentContent.Binary.Bytes(path.readByteArray()), format = fileData.extension, fileName = fileData.name))
+    }
```
Ololoshechkin: Same here

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/dsl/AttachmentBuilder.kt
```diff
@@ -1,121 +1,210 @@
...
+     * @param url File URL
+     * @param mimeType MIME type of the file (e.g., "application/pdf", "text/plain")
+     * @throws IllegalArgumentException if the URL is not valid or no file in the URL was found.
+     */
+    public fun fileFromUrl(url: String, mimeType: String) {
+        val fileData = url.urlFileData()
+        file(Attachment.File(content = AttachmentContent.URL(url), format = fileData.extension, mimeType = mimeType, fileName = fileData.name))
+    }
+
+    /**
```
Ololoshechkin: And here

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/dsl/AttachmentBuilder.kt
```diff
@@ -1,121 +1,210 @@
...
+     * @param path Path to local file
+     * @param mimeType MIME type of the file (e.g., "application/pdf", "text/plain")
+     * @throws IllegalArgumentException if the path is not valid, the file does not exist, or is not a regular file.
      */
-    public fun document(source: String) {
-        mediaContents.add(MediaContent.File(source))
+    public fun binaryFileFromPath(path: Path, mimeType: String) {
+        val fileData = path.fileData()
+        file(Attachment.File(content = AttachmentContent.Binary.Bytes(path.readByteArray()), format = fileData.extension, mimeType = mimeType, fileName = fileData.name))
     }
```
Ololoshechkin: can be just `binaryFile`

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/dsl/AttachmentBuilder.kt
```diff
@@ -1,121 +1,210 @@
...
-     * Constructs and returns the accumulated list of media content items.
+     * Adds [Attachment.File] with [AttachmentContent.PlainText] content from the provided local file path.
      *
-     * This method finalizes the building process and returns all the media content
-     * items that were added through the various builder methods.
+     * @param path Path to local file
+     * @param mimeType MIME type of the file (e.g., "application/pdf", "text/plain")
+     * @throws IllegalArgumentException if the path is not valid, the file does not exist, or is not a regular file.
+     */
+    public fun textFileFromPath(path: Path, mimeType: String) {
```
Ololoshechkin: Can be just `textFile(...)`

PR #272, opened 2025-06-12
Add embedding-based file storage with local files and in-memory implementations

Ololoshechkin:

---

#### Type of the change
- [ ] New feature
- [ ] Bug fix
- [ ] Documentation fix


rag/rag-base/src/commonMain/kotlin/ai/koog/rag/base/RankedDocumentStorage.kt
```diff
@@ -0,0 +1,60 @@
...
+        query: String,
+        count: Int = Int.MAX_VALUE,
+        similarityThreshold: Double = 0.0
+    ): Iterable<Document> = rankDocuments(query)
+        .filter { it.similarity >= similarityThreshold }
+        .toList()
+        .sortedByDescending { it.similarity }
+        .map { it.document }
+        .toList()
+}
```
ptitjes: @Ololoshechkin I played with it locally and it seems to me that `mostRelevantDocuments(...)` should be "reversed". According to the docs of `Embedder.diff()`, for two documents, the closer to zero their cosine similarity is, the more similar they are.

Also, you were not taking `count` into account.

Here is what I had to write to make the whole thing work correctly:
```kotlin
suspend fun mostRelevantDocuments(
    query: String,
    count: Int = Int.MAX_VALUE,
    similarityThreshold: Double = 1.0,
): Iterable<Document> = rankDocuments(query)
    .filter { it.similarity <= similarityThreshold }
    .toList()
    .sortedBy { it.similarity }
    .take(count)
    .map { it.document }
    .toList()
```

(cf. https://github.com/ptitjes/konvo/blob/experiments/rag/konvo-core/src/commonMain/kotlin/io/github/ptitjes/konvo/core/ai/koog/Rag.kt#L20-L30)

rag/vector-storage/src/commonMain/kotlin/ai/koog/rag/vector/InMemoryVectorStorage.kt
```diff
@@ -0,0 +1,93 @@
...
+ * @param embedder The embedder used for generating vectorized representations of text.
+ * @param documentReader The document provider facilitating access to document contents.
+ */
+public class InMemoryTextDocumentEmbeddingStorage<Document, Path>(
+    embedder: Embedder,
+    documentReader: DocumentProvider<Path, Document>
+) : EmbeddingBasedDocumentStorage<Document>(
+    embedder = TextDocumentEmbedder(documentReader, embedder),
+    storage = InMemoryVectorStorage()
+)
```
ptitjes: Maybe, reuse the above class definition?

```suggestion
) : InMemoryDocumentEmbeddingStorage<Document>(
    embedder = TextDocumentEmbedder(documentReader, embedder),
)

```

rag/rag-base/src/commonMain/kotlin/ai/koog/rag/base/RankedDocumentStorage.kt
Ololoshechkin: Thanks a lot for noticing this! Indeed, it has to be reversed.
Actually, it should stay as it is `similarity >= similarityThreshold` here, but reversed in `EmbeddingBasedDocumentStorage.rankDocuments`:
```
similarity = 1.0 - embedder.diff(queryVector, documentVector)
```
(initially similarity was equal to `diff` which was wrong 😆)

build.gradle.kts
```diff
@@ -168,8 +168,8 @@ dependencies {
     dokka(project(":agents:agents-tools"))
     dokka(project(":agents:agents-utils"))
     dokka(project(":agents:agents-ext"))
-    dokka(project(":embeddings:embeddings-base"))
-    dokka(project(":embeddings:embeddings-llm"))
+    dokka(project(":rag:embeddings:embeddings-base"))
```
kpavlov: nit: embedding models are valuable not only in the context or RAG. Example usecase could be calculating cosine similarity of two given text fragments

rag/rag-base/Module.md
```diff
@@ -0,0 +1,83 @@
...
+A foundational module that provides core interfaces for document storage and retrieval in Retrieval-Augmented Generation (RAG) systems.
+
+### Overview
+
+The rag-base module defines the fundamental abstractions for working with document storage in RAG applications. It includes:
+
+- The `DocumentStorage` interface that defines core operations for storing, reading, and deleting documents
+- The `DocumentStorageWithPayload` interface that extends document storage with support for associated metadata or payload
+- The `RankedDocumentStorage` interface that extends document storage with ranking capabilities based on query relevance
+- The `TextDocumentReader` interface for transforming documents into text representation
```
kpavlov, line 12: QQ: Is PDF considered a TextDocument? Should it be a separation between the _Reader_ (read document from file system, url, s3 etc) and _Transformer_ (transform text, markdown, pdf, docx to plain text)?

rag/rag-base/src/commonMain/kotlin/ai/koog/rag/base/DocumentStorageWithPayload.kt
```diff
@@ -0,0 +1,114 @@
...
+     */
+    public suspend fun readWithPayload(documentId: String): DocumentWithPayload<Document, Payload>?
+
+
+    /**
+     * Iterates over the documents
+     *
+     * @return An iterable collection of documents associated with the specified document ID.
+     */
+    public suspend fun allDocuments(): Flow<Document>
```
kpavlov: Here and in other places: Function returning a Flow should **not** be suspending

rag/rag-base/src/commonMain/kotlin/ai/koog/rag/base/RankedDocumentStorage.kt
```diff
@@ -0,0 +1,60 @@
...
+
+    /**
+     * Retrieves the most relevant documents matching the provided query, ranked by their similarity scores
+     * in descending order. Only documents with a similarity score greater than or equal to the specified
+     * similarity threshold are included, and the result set is limited to the specified count.
+     *
+     * @param query The search query used to find relevant documents.
+     * @param count The maximum number of documents to return. Defaults to Int.MAX_VALUE if not specified.
+     * @param similarityThreshold The minimum similarity score a document must have to be considered relevant.
+     *        Defaults to 0.0 if not specified.
```
kpavlov, line 52: Does "not specified" make any sense? Should it be mandatory without the default value? Or have some sensible default value, e.g. 0.9

rag/rag-base/src/commonMain/kotlin/ai/koog/rag/base/files/DocumentProvider.kt
```diff
@@ -0,0 +1,122 @@
...
+import kotlinx.serialization.Serializable
+
+/**
+ * Interface representing a provider for managing documents and their contents.
+ * It allows retrieval of documents, manipulation of text, and extraction of content fragments.
+ *
+ * @param Path The type representing the path or identifier for locating documents.
+ * @param Document The type representing the document being operated on.
+ */
+public interface DocumentProvider<Path, Document> {
```
kpavlov, line 12: How would it work for S3 document provider? What would be the Path?
Maybe, let's get rid of the Path in the interface?
Or this is a FilesystemDocumentProvider

rag/rag-base/src/commonMain/kotlin/ai/koog/rag/base/files/DocumentUtils.kt
```diff
@@ -0,0 +1,214 @@
...
+}
+
+/**
+ * Determines whether the character is a line break.
+ *
+ * A line break is identified if the character is either `\n` (LF) or `\r` (CR).
+ *
+ * @return `true` if the character is a line break, `false` otherwise.
+ */
+internal fun Char.isLineBreak(): Boolean {
```
kpavlov, line 131: It's a very generic function. Maybe it should go to some utils module

kpavlov:
Thank you, @Ololoshechkin 🙏🏻

Would it be possible to make smaller PRs to make reviews easier and reduce the chance of missing something important? E.g. introduce new interfaces separately to the implementations. 3k+ lines do not fit into my context window

rag/rag-base/src/commonMain/kotlin/ai/koog/rag/base/RankedDocumentStorage.kt
ptitjes: @Ololoshechkin yeah, `1 - diff` works too! :smile: but don't forget to take `count` results.

build.gradle.kts
Ololoshechkin: Yes, true, thought about that. Will move it back to a separate external module, thanks

rag/rag-base/Module.md
Ololoshechkin: I think we don't have to separate. Reader takes a Document and produces a String text. So it's also a transformer. I don't want to introduce extra abstractions here unless we have a request/ise case where it's necessary

rag/rag-base/src/commonMain/kotlin/ai/koog/rag/base/RankedDocumentStorage.kt
Ololoshechkin: Without that value(or with 0.0) if the user just specifies number of documents (say, 10) it will return the top-10 most relevant documents for the context.
If the user specifies also the similarity threshold, say 0.7 -- it will return no more than 10 documents that are the most relevant and have relevance at least 70% (there might be less then 10 documents that match this condition for example)

rag/rag-base/src/commonMain/kotlin/ai/koog/rag/base/files/DocumentUtils.kt
Ololoshechkin: It's internal and it's in DocumentUtils.

This particular part was moved from our internal SDK where we have all the system-agnostic and product-agnostic file-providers, document readers and other things that allow to reuse same feature logic across different environments.

So (at least currently) it has no usages in Koog but it was tested and is adopted in products of JB.
Not sure if it needs to be public API in Koog for now, so let's keep it as internal and see if it's required elsewhere.

We can then move it somewhere and make it public API but for the scope of this PR let's keep it here

cc: @sproshev WDYT?

rag/rag-base/src/commonMain/kotlin/ai/koog/rag/base/files/DocumentProvider.kt
Ololoshechkin: You can have a data class containing Bucket + String object name as a Path. Or just a full s3 key.

cc: @sproshev do we have any examples with S3 ?

Ololoshechkin: For example, a GitHub path to a file is a combination of organizationId, repositoryID, and relative path.

And in order to "read" the file (or to "write" it) through a GitHub application you have to get the right app token for the installation of the app for the specific organisationId, then go to repositories rest API and read a file by it's path. Or to "write" a file you would do a pull request.

For sure it can be just a String like "orgId/repoId/file-path" that you would parse and split every time to obtain the correct installation token, but it looks more natural and less error-prone if you work with a custom path object for the specific environment

sproshev: I don't have any examples with S3, but yes, it was designed to be used with any custom type as a path

rag/rag-base/src/commonMain/kotlin/ai/koog/rag/base/files/DocumentUtils.kt
sproshev: I agree to have a full copy here but export only the necessary parts

PR #275, opened 2025-06-13
[agent] Refactor subgraph with task. Simplify message handling to make it more straightforward and easier to use.

EugeneTheDev:
* Refactor the implementation of `subgraphWithTask` to make the code more readable 
* Don't make a task message `system`, it's just a `user` message now
* Remove built-in complicated message history handling logic, make the whole process more straightforward to avoid non-obvious side-effects 
* It's mandatory now to add your `finishTool` to the global agent tool registry, since it is actually executed now as a regular tool
* Also, turns it `onRequestContains` in `MockLLMBuilder` for mocking tool calls was missing, which was required to adjust tests in this PR. So I added it as well

### DIFF
--- a/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/entity/AIAgentStorage.kt
+++ b/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/entity/AIAgentStorage.kt
@@ -52,6 +52,19 @@ public class AIAgentStorage internal constructor() {
         storage[key] as T?
     }
 
+
+    /**
+     * Retrieves the non-null value associated with the given key from the storage.
+     * If the key does not exist in the storage, a [NoSuchElementException] is thrown.
+     *
+     * @param key The key of type [AIAgentStorageKey] used to identify the value in the storage.
+     * @return The value associated with the key, of type [T].
+     * @throws NoSuchElementException if the key does not exist in the storage.
+     */
+    public suspend fun <T : Any> getValue(key: AIAgentStorageKey<T>): T {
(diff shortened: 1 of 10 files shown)


Ololoshechkin (changes requested):
@EugeneTheDev could you please check if we can make `defineTask` produce not just a string , but actually build a prompt?

Ololoshechkin:
Also, maybe cleaning older system messages can be added by a flag. WDYT?

ptitjes:
For any LLM prompt, whatever the model, there can only be one system message. So I'd say, if the caller adds a system message, it should replace the previous one.

In Langchain4j, their ChatMemory actually has a special treatment for the system message: it is always replacing the previous one if a new one is added to the chat memory and it is always the first message of the chat memory.

EugeneTheDev:
@Ololoshechkin I decided to leave `defineTask` as a simple string deliberately after some consideration. Prompt builder seems like a more advanced usage here. This is a more simplified, high-level API, so we want the user to just provide us with a single "prompt" - task description. And then we will do our logic under the hood. If a user wants to have more control over the prompt, they can use a custom node and manipulate the prompt directly using the lower level API we provide, since doing it would require a deeper understanding of how the agent works anyway. For a better structure users can use TextContentBuilder here. That's why I don't like the idea of allowing the users to pass multiple messages here, especially system ones.

And what @ptitjes said about system message too, system messages are special, so we shouldn't be adding a lot of them in the prompt just like that (although it is possible to put several system messages in the prompt, at least OpenAI and Anthropic return successful responses in such cases)

Ololoshechkin (approved):

PR #221, opened 2025-06-04
Create a type safe DSL for edges

karloti:
### The 'edge' function takes a lambda as an argument.
```
edge {
    nodeStart forwardTo nodeCallLLM
}
```
---

#### Type of the change
- [x] New feature
- [ ] Bug fix
- [ ] Documentation fix

#### Checklist for all pull requests
- [x] The pull request has a description of the proposed change
- [x] I read the [Contributing Guidelines](https://github.com/JetBrains/koog/blob/main/CONTRIBUTING.md) before opening the pull request
- [x] The pull request uses **`develop`** as the base branch
- [x] Tests for the changes have been added
- [x] All new and existing tests passed

##### Additional steps for pull requests adding a new feature
- [ ] An issue describing the proposed change exists
- [ ] The pull request includes a link to the issue
- [ ] The change was discussed and approved in the issue
- [ ] Docs have been added / updated

Issue #217, opened 2025-06-04
[API design] Graph DSL improments


ptitjes:
[Disclaimer: This is an opinionated proposal that only reflects my views.]

#### Rational

Currently writing a graph is a bit cumbersome for various reasons:
- there is no DSL marker annotation, so that makes discovery of the builders complicated
- all the node builder functions have a `node` prefix, which is a certainly due to not having a DSL marker annotation
- you have to wrap edges in `edge`, sometimes...

#### Proposal

Update the style
  - Add a `GraphDsl` DSL marker annotation
  - Remove the `node` prefix from all node builders (except `node` itself obviously...)

Update the implementation
  1. Make node builder functions immediately register the node to graph it is created in
  2. Have `forwardTo` immediately register an edge to the graph the left node is in
  3. Have `AgentEdgeBuilderIntermediate` conditionnal/transform methods just enrich the already registered edge

Having 1. will make possible to detect orphan nodes.
Having 2. and 3. will make it possible to remove the `edge` function.

In fine, you would do:

```kotlin

val graph = graph("qa") {
    val llmRequest by llmRequest

    start forwardTo llmRequest
    llmRequest forwardTo finish
}

```

Or in this instance, as there are no conditionnals:

```kotlin

val graph = graph("qa") {
    val llmRequest by llmRequest

    start then llmRequest then finish
}

```

Ololoshechkin:
Hi! Thanks for your great suggestions!

> Remove the node prefix from all node builders (except node itself obviously...)

Actually, I was thinking about this multiple times myself, too. Extra word "node" is annoying sometimes.
The reason why all nodes start with `node` word is basically the IDE completion: you start typing "nod...." and you'll see all the available options in the completion popup, so even if you don't know/remember what is available, you can scroll through them.

>  Have forwardTo immediately register an edge to the graph the left node is in

The thing with `edge` is that it only accepts `EdgeIntermediate<Input, Output, Output>`. That means that if you have some not-yet-registered edge with a type `EdgeIntermediate<Input, SomethingElse, Output>` it will not allow it and require you to add a `transformed {...}` with a `(SomethingElse) -> Output` lambda that would convert it to `EdgeIntermediate<Input, Output, Output>` so that the input of the `to` node will match it's expected input type. And as you can stack these `transformed` + `onCondition` + `transformed` up on each other multiple times we don't know when the expression will finish.

> Have AgentEdgeBuilderIntermediate conditionnal/transform methods just enrich the already registered edge

[...]

Issue #216, opened 2025-06-04
[API design] Agent/Strategy/Graph API enhancement


ptitjes:
[Disclaimer: This is an opinionated proposal that only reflects my views.]

#### Rational

One of the first things I noticed while starting to use Koog is that is that you can't put a strategy in a strategy. So if you are building above Koog, you can't ask your users to build strategies and wrap them in your own.

The second thing I noticed is that `AgentStrategy` have fixed `String` to `String` type parameters. This means you can only do `String` to `String` agents. This is IMO a problematic restriction. Why would you limit yourself to string to string when even the LLMs can produce structured data.

#### Proposal

- Remove `AgentStrategy` and rename `AgentSubgraph` to `AgentGraph`
- Rename `AgentSubgraphBuilder` to `AgentGraphBuilder`
- Make Agent take an `agentGraph` argument instead of a `strategy` argument
- Rename the `strategy { }` builder function to `graph`
- Make `Agent` have type parameters for `Input` and `Output`

Ololoshechkin:
Hi! Thanks for raising this discussion!

> The second thing I noticed is that AgentStrategy have fixed String to String type parameters. This means you can only do String to String agents. This is IMO a problematic restriction. Why would you limit yourself to string to string when even the LLMs can produce structured data.

Actually, we were considering adding a support for <T1, T2> agents while keeping the <String, String> option as a default. It was also already requested internally a few times, so we will definitely add this support soon.

> Rename the strategy { } builder function to graph
> Remove AgentStrategy and rename AgentSubgraph to AgentGraph

That's a good suggestion, but at the same time I would like to have it discussed carefully and gathering more opinions from the users. It makes total sense from the perspective of an advanced user, but at the same time I feel it might look less intuitive to a newcomer ("I am writing AI agents, when did we go to the concept of graphs?"). 

But for sure, we'll consider these renames and maybe will change those in the future once we get more reviews from the wide audience.

ptitjes:
> Actually, we were considering adding a support for <T1, T2> agents while keeping the <String, String> option as a default. It was also already requested internally a few times, so we will definitely add this support soon.

Good. I can contribute on that if you will.

> That's a good suggestion, but at the same time I would like to have it discussed carefully and gathering more opinions from the users. It makes total sense from the perspective of an advanced user, but at the same time I feel it might look less intuitive to a newcomer ("I am writing AI agents, when did we go to the concept of graphs?").

[...]

PR #280, opened 2025-06-14
Improve edge creation DSL [Demo]

ptitjes:
This PR presents an alternative DSL design for the graph edge creation. It solves part of #217.

## Explanation of the current situation

Currently, edge creation requires a call to `AIAgentSubgraphBuilderBase<*,*>.edge(...)`.
IMO this leads to overly verbose code, and forces having public APIs to do back and forth calls between the graph and the node recipient of the edge.

#221 tried a first alternative DSL design to fix these issues, but:
- it was forced to mutably replace edges when further `onCondition`/`transformed` calls were made to modify those,
- there were type-safety issues

## Proposed change

To achieve the same expressibility while not having edge-replacing or type-safety issues, the trick is to swap the order of `onCondition`/`transformed` calls w.r.t. the `forwardTo` call.

That is, we move from
```kotlin
edge(source forwardTo target transformed { ... } onCondition { ... })
```
to
```kotlin
source transformed { ... } onCondition { ... } forwardTo target
```

Not only does it fix all the above problems, but I believe it does also read better, as the transforms and conditions to walk through an edge comes before the target of the edge, and the edge can be naturally read from left-to-right.

So we introduce an alternate `AIAgentEdgeBuilder` which `AIAgentNode` inherits from.
This `AIAgentEdgeBuilder<I, O>` can create wrapping `AIAgentEdgeBuilder<I, O'>`s via calls to `transformed` and `onCondition`, or create the final edge via a call to `forwardTo`.

Note that `forwardTo` also returns the target node, which does also render `then` obsolete. And you can write similar chains as with `then`, but with the added possibility to transform/condition edges:
```kotlin
node1 forwardTo node2 onCondition { ... } forwardTo node3 forwardTo node4
```

> [WARNING] I did not fix the Kdocs for now.

### Far less type-parameters and cleaner API

- `AIAgentEdgeBuilderIntermediate`, which has three type-parameters, can be removed.
- `AIAgentEdgeBuilder` doesn't extend `BuilderBase`, which does remove some API bloat
- `edge` and `then` can be removed
- I believe it does also improve type-inference, as the right operand types are now fully depend on the left operand types and additional parameters

### Unfinished edges error at build time

With the upcoming "Unused return value checker", we will have the opportunity to catch unfinished edges at build time.

We would annotate `transformed` and `onCondition` with `@MustUseReturnValue`. It is not clear to me yet, whether this `@MustUseReturnValue` will be inherited by methods wrapping `transformed` and `onCondition` (i.e. `onToolCall`, etc.), but we might have to annotate all the transform/condition methods. This should be discussed with the Kotlin team and bring our use-case, but I don't how far they are yet in their design.

The `forwardTo` method in turn, would be annotated with `@IgnorableReturnValue`, in order to allow the end of `forwardTo` chains.

### `addEdge` could be made internal

Both `AIAgentNode<>.addEdge` and `AIAgentEdge` could potentially be made internal and become an implementation detail. This would simplify the API a lot and avoid exposing the internal edge management.

### Using a `context(GraphBuilder)` marker in the future

When context parameters are a thing, we will be able to mark `forwardTo` and others with this `context(GraphBuilder` parameter, in addition to using `@GraphDsl`, in order to restrict their use when inside a graph builder.

## Conclusion

I really believe this is the way to go, as it does greatly simplify the API and improve readability, while still preserving the same expressibility.

In this PR, I only modified `singleRunStrategy()` as to demonstrate the concept. It is fully functional and the tests in `SimpleAgentMockedTest` pass as they were before. That means the DSL changes do not impact any other features.

I placed the new DSL stuff in a `dsl2` because there cannot be any possible migration from the current situation to the proposed change if we want to keep the `forwardTo` name (which IMO, and until someone convinces me otherwise, is the right name). I tried playing with `@Deprecated` to make the transition smooth, but this is not easy as we are swapping the `forwardTo` and `transformed`/`onCondition` calls.

However, I think such a big improvement to the API might call for a breaking change in `0.3`.

Also, while making a breaking change, we should take the opportunity to improve the whole graph DSL further (cf. #216 and #217).

@Ololoshechkin , @tiginamaria , @EugeneTheDev , @Rizzen please guys tell me what you think about this.

---

#### Type of the change
- [ ] New feature
- [ ] Bug fix
- [ ] Documentation fix

### DIFF
--- a/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
+++ b/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
@@ -2,8 +2,19 @@ package ai.koog.agents.core.agent
 
 import ai.koog.agents.core.agent.config.AIAgentConfig
 import ai.koog.agents.core.agent.config.AIAgentConfigBase
-import ai.koog.agents.core.agent.entity.*
 import ai.koog.agents.core.agent.context.AIAgentContext
+import ai.koog.agents.core.agent.context.AIAgentLLMContext
+import ai.koog.agents.core.agent.entity.AIAgentStateManager
+import ai.koog.agents.core.agent.entity.AIAgentStorage
+import ai.koog.agents.core.agent.entity.AIAgentStrategy
+import ai.koog.agents.core.dsl.builder.strategy
+import ai.koog.agents.core.dsl.extension.nodeDoNothing
+import ai.koog.agents.core.dsl.extension.nodeExecuteTool
+import ai.koog.agents.core.dsl.extension.nodeLLMRequest
+import ai.koog.agents.core.dsl.extension.nodeLLMSendToolResult
+import ai.koog.agents.core.dsl2.builder.forwardTo
+import ai.koog.agents.core.dsl2.extension.onAssistantMessage
+import ai.koog.agents.core.dsl2.extension.onToolCall
 import ai.koog.agents.core.environment.AIAgentEnvironment
 import ai.koog.agents.core.environment.AIAgentEnvironmentUtils.mapToToolResult
 import ai.koog.agents.core.environment.ReceivedToolResult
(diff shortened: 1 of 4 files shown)


Rizzen:
@ptitjes Hello! Sorry for such a long response time. For me the idea looks great, but I'm not sure about removing old APIs part. I would prefer to preserve basic edge API.

aozherelyeva:
I think we can close this one as outdated. Please rebase to the `develop` and open a new PR in case the change is still actual 🙂

Issue #162, opened 2025-05-24
AWS Bedrock Model Support


blakeromano:
Koog should support AWS Bedrock Models to support Agentic AI use cases.

whatdoidoherereally:
Anyone at JetBrains want to chime in and let us know if this is a roadmap thing? Otherwise I’m game to take a stab at mapping the Converse API to Koog.

SomeDatabaseDude:
@EugeneTheDev @Ololoshechkin @tiginamaria I have a local implementation of the bedrock client here. It's probably a bit _rough_, but would you guys be up to giving feedback on it and helping me refine at all?

kaeawc:
@SomeDatabaseDude I'm eagerly awaiting this and in talking on the kotlin lang slack Jetbrains is happy to accept contributions. If you can get your version up I'm happy to help work on yours, otherwise I'm going to start working on my own contribution because I need this working asap

PR #285, opened 2025-06-16
AWS Bedrock support

kaeawc:
Wrote this to fulfill https://github.com/JetBrains/koog/issues/162

Adds model support for the current slate of available Bedrock models with supporting unit tests and MCP tool calling. I'm happy to add more documentation than just the kdoc, let me know what's appropriate here.

---

#### Type of the change
- [x] New feature
- [ ] Bug fix
- [ ] Documentation fix


examples/src/main/kotlin/ai/koog/agents/example/mcp/GoogleMapsMcpClient.kt
```diff
@@ -58,6 +60,21 @@ fun main() {
...
             )
         }
+
+        // Example of using the Bedrock executor
+        val awsAccessKey = System.getenv("AWS_ACCESS_KEY_ID") ?: error("AWS_ACCESS_KEY_ID not set")
+        val awsSecretKey = System.getenv("AWS_SECRET_ACCESS_KEY") ?: error("AWS_SECRET_ACCESS_KEY not set")
+
+        val bedrockAgent = AIAgent(
+            executor = simpleBedrockExecutor(awsAccessKey, awsSecretKey),
+            systemPrompt = "You are a helpful assistant. Answer user questions concisely using AWS Bedrock.",
```
zarechneva-a: Nit: this may be extensive information to include in the system prompt. As `simpleBedrockExecutor` is an instance of `SingleLLMPromptExecutor`, the agent won't be able to use any other executor anyways.

examples/src/main/kotlin/ai/koog/agents/example/mcp/GoogleMapsMcpClient.kt
```diff
@@ -58,6 +60,21 @@ fun main() {
                         "You can only call tools. Get it by calling maps_geocode and maps_elevation tools."
             )
         }
+
+        // Example of using the Bedrock executor
```
sdubov: @kaeawc, thank you a bunch for updating examples with using Bedrock models. Sorry, it might be I miss something, but I would propose to extract the example that shows how to use Bedrock model into a separate file to not loose it here in mpc example.

You can add a separate package: "ai.koog.agents.example.client" and add a file "BedrockAgent.kt" with a simple example (similar to `BasicAgentWithTools.kt` ), but using the Bedrock model.

Please let me know if I miss something with using Bedrock models.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/bedrock/BedrockLLMClient.jvm.kt
```diff
@@ -0,0 +1,645 @@
...
+import aws.smithy.kotlin.runtime.net.url.Url
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.FlowPreview
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.channelFlow
+import kotlinx.coroutines.flow.map
+import kotlinx.datetime.Clock
+import kotlinx.serialization.json.Json
+import kotlinx.serialization.json.JsonArray
+import kotlinx.serialization.json.JsonElement
```
sdubov: Could you please clear unused imports from the file header?

prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/build.gradle.kts
```diff
@@ -0,0 +1,66 @@
...
+
+plugins {
+    id("ai.kotlin.multiplatform")
+    alias(libs.plugins.kotlin.serialization)
+}
+
+kotlin {
+    targets.withType(org.jetbrains.kotlin.gradle.targets.jvm.KotlinJvmTarget::class.java).configureEach {
+        compilations.all {
+            kotlinOptions.jvmTarget = "1.8"
```
sdubov: Could you please clarify why do you need to set a JVM target version explicitly here? Currently, it is defined inside the `ai.kotlin.configuration` gradle plugin inside `buildSrc` which come in the config with the `ai.kotlin.multiplatform` gradle plugin. Is it any specific reason to re-define a version for this module?

prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/bedrock/BedrockLLMClient.jvm.kt
```diff
@@ -0,0 +1,645 @@
...
+internal actual class StaticCredentialsProvider actual constructor(
+    private val awsAccessKeyId: String,
+    private val awsSecretAccessKey: String
+) : CredentialsProvider {
+    override suspend fun resolve(attributes: Attributes): Credentials {
+        return Credentials(accessKeyId = awsAccessKeyId, secretAccessKey = awsSecretAccessKey)
+    }
+}
+
+public actual fun createBedrockLLMClient(
```
sdubov: @kaeawc, could you please add KDoc to the public function?

prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/bedrock/BedrockLLMClient.jvm.kt
```diff
@@ -0,0 +1,645 @@
...
+            this.modelId = model.id
+            this.contentType = "application/json"
+            this.accept = "*/*"
+            this.body = requestBody.toString().encodeToByteArray()
+        }
+
+        logger.debug { "Bedrock InvokeModel Request (JVM): ModelID: ${model.id}, Body: $requestBody" }
+
+        try {
+            val response = bedrockClient.invokeModel(invokeRequest)
```
sdubov: Could you please wrap network operations execution using an IO dispatcher? Usually, for other client we use:
```
withContext(Dispatchers.SuitableForIO) {
   ...
}
```
from the `agent-utils` module (please make sure you have a dependency on this module in build.gradle.kts).

prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/bedrock/BedrockLLMClient.jvm.kt
```diff
@@ -0,0 +1,645 @@
...
+            val responseBodyString = response.body?.decodeToString()
```
sdubov: It seems that the `body` parameter from a response is a non-nullable type. It seams that you can drop the `?` from here. It also would need to adjust the check for `isNullOrBlank()` below. Please correct me if I miss anything.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/bedrock/BedrockLLMClient.jvm.kt
```diff
@@ -0,0 +1,645 @@
...
+            this.accept = "*/*"
+            this.body = requestBody.toString().encodeToByteArray()
+        }
+        logger.debug { "Bedrock InvokeModelWithResponseStream Request (JVM): ModelID: ${model.id}, Body: $requestBody" }
+
+        return channelFlow {
+            try {
+                bedrockClient.invokeModelWithResponseStream(streamRequest) { response: InvokeModelWithResponseStreamResponse ->
+                    response.body?.collect { event: ResponseStream ->
+                        val chunkBytes = event.asChunk()?.bytes
```
sdubov: It seems that `asChun()` function return a non-nullable result. You can safely drop the `?` and update the check below.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/bedrock/BedrockLLMClient.jvm.kt
```diff
@@ -0,0 +1,645 @@
...
+
+        val streamRequest = InvokeModelWithResponseStreamRequest {
+            this.modelId = model.id
+            this.contentType = "application/json"
+            this.accept = "*/*"
+            this.body = requestBody.toString().encodeToByteArray()
+        }
+        logger.debug { "Bedrock InvokeModelWithResponseStream Request (JVM): ModelID: ${model.id}, Body: $requestBody" }
+
+        return channelFlow {
```
sdubov: May I ask you to use a IO dispatcher for this operation as well? We utilize `Dispatchers.SuitableForIO` extension from `agents-utils` module for such operations to make sure all IO operations are running in a proper dispatcher. For example:
```
return channelFlow { 
  ...
}.flowOn(Dispatchers.SuitableForIO)
```

prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/bedrock/BedrockLLMClient.jvm.kt
```diff
@@ -0,0 +1,645 @@
...
+        val messages = mutableListOf<JsonObject>()
+        val systemMessages = mutableListOf<String>()
+
+        prompt.messages.forEach { msg ->
+            when (msg) {
+                is Message.System -> systemMessages.add(msg.content)
+                is Message.User -> {
+                    // Check for image content if present
+                    if (msg.mediaContent.isNotEmpty()) {
+                        require(model.capabilities.contains(LLMCapability.Vision.Image)) {
```
sdubov: @kaeawc could you please clarify this check for me? As far as I understand, the model can return some media type, that is not an image, but we explicitly check for Image capabilities here. Please correct me if I miss anything.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/bedrock/BedrockLLMClient.jvm.kt
```diff
@@ -0,0 +1,645 @@
...
+                                add(buildJsonObject {
+                                    put("type", "text")
+                                    put("text", msg.content)
+                                })
+                            }
+
+                            msg.mediaContent.forEach { media ->
+                                when (media) {
+                                    is MediaContent.Image -> {
+                                        val imageData = if (media.isUrl()) {
```
sdubov: It seems that `imageData` variable is unused after creating. Could you please clarify if it is used anywhere below?

prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jsMain/kotlin/ai/koog/prompt/executor/clients/bedrock/BedrockLLMClient.js.kt
```diff
@@ -0,0 +1,42 @@
...
+import kotlinx.datetime.Clock
+
+internal actual class StaticCredentialsProvider actual constructor(
+    @Suppress("UNUSED_PARAMETER") awsAccessKeyId: String,
+    @Suppress("UNUSED_PARAMETER") awsSecretAccessKey: String
+) {
+    // JS specific: No actual credentials needed as it will be unsupported.
+}
+
+public actual fun createBedrockLLMClient(
```
sdubov: Could you please add KDoc to the public function?

prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jsMain/kotlin/ai/koog/prompt/executor/clients/bedrock/BedrockLLMClient.js.kt
```diff
@@ -0,0 +1,42 @@
...
+        model: LLModel,
+        tools: List<ToolDescriptor>
+    ): List<Message.Response> {
+        console.warn(unsupportedMessage)
+        throw UnsupportedOperationException(unsupportedMessage)
+    }
+
+    override fun executeStreaming(prompt: Prompt, model: LLModel): Flow<String> {
+        console.warn(unsupportedMessage)
+        return flowOf()
```
sdubov: Could you please clarify the difference in the behavior between the `execute` (throw) and the `executeStreaming` (return empty flow) methods? Would it be more convenient to throw in both cases?

prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jvmTest/kotlin/ai/koog/prompt/executor/clients/bedrock/BedrockLLMClientTest.kt
```diff
@@ -0,0 +1,511 @@
...
+import kotlinx.serialization.json.jsonArray
+import kotlinx.serialization.json.jsonPrimitive
+import kotlin.test.Test
+import kotlin.test.assertNotNull
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+import kotlin.test.assertContains
+import kotlin.test.assertFails
+
+class BedrockLLMClientTest {
```
sdubov, line 31: Nice! Thank you for adding all those tests :+1:

kaeawc:
@sdubov thanks for the very thorough and thoughtful review! I've addressed all feedback and made sure tests are still passing. I rebased from latest `develop` to resolve merge conflicts. Let me know if there's anything else you'd like me to address.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jsMain/kotlin/ai/koog/prompt/executor/clients/bedrock/BedrockLLMClient.js.kt
```diff
@@ -0,0 +1,52 @@
...
+    ): List<Message.Response> {
+        console.warn(unsupportedMessage)
+        throw UnsupportedOperationException(unsupportedMessage)
+    }
+
+    override fun executeStreaming(prompt: Prompt, model: LLModel): Flow<String> {
+        console.warn(unsupportedMessage)
+        throw UnsupportedOperationException(unsupportedMessage)
+    }
+}
```
Ololoshechkin: Let's please make the whole module JVM-only in gradle without a JS option at all, as we don't support it anyway right now.

Ololoshechkin (changes requested):

prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jsMain/kotlin/ai/koog/prompt/executor/clients/bedrock/BedrockLLMClient.js.kt
kaeawc: Sure thing!

kaeawc: @Ololoshechkin let me know if that change works - it isn't how the other existing modules are written so I was confused.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/bedrock/BedrockLLMClient.kt
```diff
@@ -0,0 +1,666 @@
...
+ */
+public class BedrockClientSettings(
+    public val region: String = "us-east-1",
+    public val timeoutConfig: ConnectionTimeoutConfig = ConnectionTimeoutConfig(),
+    public val endpointUrl: String? = null,
+    public val maxRetries: Int = 3,
+    public val enableLogging: Boolean = false
+)
+
+internal class StaticCredentialsProvider(
```
EugeneTheDev: AWS SDK already provides `aws.sdk.kotlin.runtime.auth.credentials.StaticCredentialsProvider`, why do we need a custom class here?

prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/bedrock/BedrockLLMClient.kt
```diff
@@ -0,0 +1,666 @@
...
+ * @return A configured [LLMClient] instance for Bedrock
+ */
+public fun createBedrockLLMClient(
+    awsAccessKeyId: String,
+    awsSecretAccessKey: String,
+    settings: BedrockClientSettings,
+    clock: Clock
+): LLMClient = BedrockLLMClientImpl(awsAccessKeyId, awsSecretAccessKey, settings, clock)
+
+private class BedrockLLMClientImpl(
```
EugeneTheDev: Why did you make the class private and called it `..Impl`? This looks a bit odd and not in line with existing LLM clients. I suggest removing `createBedrockLLMClient` function and simply using constructor here, since the function just delegates to constructor anyway. And lets remove `Impl` from the class name.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/bedrock/BedrockLLMClient.kt
```diff
@@ -0,0 +1,666 @@
...
+    awsAccessKeyId: String,
+    awsSecretAccessKey: String,
+    private val settings: BedrockClientSettings,
+    private val clock: Clock
+) : LLMClient {
+
+    private val logger = KotlinLogging.logger {}
+
+    private val bedrockClient = BedrockRuntimeClient {
```
EugeneTheDev: Let's make `bedrockClient: BedrockRuntimeClient` a primary constructor parameter instead of current `settings: BedrockClientSettings`. The reason is, `BedrockRuntimeClient` contains multiple other properties as well, so for advanced use-cases users might want to configure them too or to just pass existing client here. So in the primary constructor we provide them with this option, and then make secondary constructor with `settings: BedrockClientSettings` which configures `bedrockClient` like you do here, to cover simpler use-cases.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/bedrock/BedrockLLMClient.kt
```diff
@@ -0,0 +1,666 @@
...
+            this.accept = "*/*"
+            this.body = requestBody.toString().encodeToByteArray()
+        }
+
+        logger.debug { "Bedrock InvokeModel Request (JVM): ModelID: ${model.id}, Body: $requestBody" }
+
+        try {
+            return withContext(Dispatchers.SuitableForIO) {
+                val response = bedrockClient.invokeModel(invokeRequest)
+                val responseBodyString = response.body?.decodeToString()
```
EugeneTheDev: nit: unnecessary safe call here

prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/bedrock/BedrockLLMClient.kt
```diff
@@ -0,0 +1,666 @@
...
+                    model.id.startsWith("amazon.titan") -> parseTitanResponse(responseBodyString)
+                    model.id.startsWith("ai21.j2") -> parseJurassicResponse(responseBodyString)
+                    model.id.startsWith("cohere.command") -> parseCohereResponse(responseBodyString)
+                    model.id.startsWith("meta.llama") -> parseLlamaResponse(responseBodyString)
+                    else -> throw IllegalArgumentException("Unsupported Bedrock model: ${model.id}")
+                }
+            }
+        } catch (e: Exception) {
+            logger.error(e) { "Error executing Bedrock prompt for model ${model.id} (JVM)" }
+            return listOf(Message.Assistant("Error: ${e.message}", metaInfo = ResponseMetaInfo.create(clock)))
```
EugeneTheDev: Returning successful `Message` in case of an error looks wrong to me, we should properly throw an exception to indicate that something is wrong, just like in other clients. Otherwise it would be tricky to handle errors

prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/bedrock/BedrockLLMClient.kt
```diff
@@ -0,0 +1,666 @@
...
+            "Model ${model.id} does not support chat completions"
+        }
+
+        // Check tool support
+        if (tools.isNotEmpty() && !model.capabilities.contains(LLMCapability.Tools)) {
+            throw IllegalArgumentException("Model ${model.id} does not support tools")
+        }
+
+        val requestBody = when {
+            model.id.startsWith("anthropic.claude") -> createAnthropicRequest(prompt, model, tools)
```
EugeneTheDev: I see these strings with model id patterns are repeated several times in the client, which would make it harder to maintain. Is it possible to extract them into `const`, or maybe extract the whole matching process into a separate function to reuse it? Or maybe even better: if different models in AWS Bedrock require different message format, what if we create several sub-providers for the `Bedrock` LLM provider? So that `Bedrock` provider would be an interface, and then providers like `BedrockAnthropic`, `BedrockMeta` etc. This way, we would be able to specify concrete provider when declaring the model and then match it here in a nice static manner.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/bedrock/BedrockLLMClient.kt
```diff
@@ -0,0 +1,666 @@
...
+                }
+            } catch (e: Exception) {
+                logger.warn(e) { "Failed to parse Bedrock stream chunk (JVM): $chunkJsonString" }
+                ""
+            }
+        }
+    }
+
+    // Anthropic Claude specific methods
+    private fun createAnthropicRequest(prompt: Prompt, model: LLModel, tools: List<ToolDescriptor>): JsonObject {
```
EugeneTheDev: For all `create` and `parse` methods in the client: let's use proper classes instead of JSON objects when parsing requests/responses, just like in other existing LLM clients. Using JSON objects directly makes it harder to understand the format and maintain.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/bedrock/BedrockLLMClient.kt
```diff
@@ -0,0 +1,666 @@
...
+                                put("content", json.parseToJsonElement(msg.content))
+                            })
+                        })
+                    })
+                }
+            }
+        }
+
+        return buildJsonObject {
+            put("anthropic_version", "bedrock-2023-05-31")
```
EugeneTheDev: Hardcoded version, shouldn't it be moved to constructor?

EugeneTheDev (changes requested):

kaeawc:
@EugeneTheDev thanks for the thoughtful review! I've done most of the changes requested but have a question:

The last part have left to finish is using data classes instead of json objects. I was wondering if we could reuse the data classes that are already in koog for providers like Anthropic since those objects should be exactly the same. That might help with maintenance overall. Alternatively, if y'all don't want those data classes to be a public APl on those Gradle modules I could refactor them into separate serialization modules? But don't know if it's a concern at all.

Let me know what you think, happy to do what works best for koog.

kaeawc:
@EugeneTheDev @Ololoshechkin let me know if there's anything else you'd like me to change or update. The serialization data class change was a pretty big refactor, I went through the AWS model lists and available documentation for the current models. I also scoped down the number of providers supported out of the gate so it would be easier to get this out.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/bedrock/BedrockLLMClient.kt
```diff
@@ -0,0 +1,292 @@
...
+        }
+
+        // Check tool support
+        if (tools.isNotEmpty() && !model.capabilities.contains(LLMCapability.Tools)) {
+            throw IllegalArgumentException("Model ${model.id} does not support tools")
+        }
+
+        val requestBody = when (modelFamily) {
+            is BedrockModelFamilies.AI21Jamba ->
+                if (model.id.startsWith("ai21.jamba")) {
```
EugeneTheDev: Looks like these if's here and below are redundant, since this check is already performed in `getBedrockModelFamily`

prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/bedrock/BedrockLLMClient.kt
```diff
@@ -0,0 +1,292 @@
...
+            "Model ${model.id} does not support chat completions"
+        }
+
+        val requestBody = when (modelFamily) {
+            is BedrockModelFamilies.AI21Jamba -> json.encodeToString(
+                JambaRequest.serializer(),
+                BedrockAI21JambaSerialization.createJambaRequest(prompt, model, emptyList())
+            )
+            is BedrockModelFamilies.AmazonNova -> {
+                if (model.id.startsWith("amazon.nova")) {
```
EugeneTheDev: Similar, redundant if

prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/bedrock/BedrockLLMClient.kt
```diff
@@ -0,0 +1,292 @@
...
+                            // Amazon Titan
+                            throw IllegalArgumentException("Model ${model.id} not supported in koog")
+                        }
+                    }
+                    is BedrockModelFamilies.AnthropicClaude -> BedrockAnthropicClaudeSerialization.parseAnthropicStreamChunk(chunkJsonString)
+                    is BedrockModelFamilies.Meta -> BedrockMetaLlamaSerialization.parseLlamaStreamChunk(chunkJsonString)
+                }
+            } catch (e: Exception) {
+                logger.warn(e) { "Failed to parse Bedrock stream chunk: $chunkJsonString" }
+                ""
```
EugeneTheDev: Let's rethrow an exception explicitly here instead of sliently swallowing it

prompt/prompt-executor/prompt-executor-llms-all/src/commonMain/kotlin/ai/koog/prompt/executor/llms/all/SimplePromptExecutors.kt
```diff
@@ -44,4 +47,18 @@ public fun simpleGoogleAIExecutor(apiKey: String): SingleLLMPromptExecutor = Sin
...
+ * Creates an instance of `SingleLLMPromptExecutor` with a `BedrockLLMClient`.
+ *
+ * @param awsAccessKeyId Your AWS Access Key ID.
+ * @param awsSecretAccessKey Your AWS Secret Access Key.
+ * @param settings Custom client settings for region and timeouts.
+ */
+public fun simpleBedrockExecutor(
+    awsAccessKeyId: String,
+    awsSecretAccessKey: String,
+    settings: BedrockClientSettings = BedrockClientSettings()
```
EugeneTheDev: This doesn't compile, Bedrock client has to be added to the module dependencies

prompt/prompt-executor/prompt-executor-clients/src/commonMain/kotlin/ai/koog/prompt/executor/clients/InternalAPI.kt
```diff
@@ -0,0 +1,17 @@
...
+    AnnotationTarget.CLASS,
+    AnnotationTarget.TYPEALIAS,
+    AnnotationTarget.FUNCTION,
+    AnnotationTarget.PROPERTY,
+    AnnotationTarget.FIELD,
+    AnnotationTarget.CONSTRUCTOR,
+    AnnotationTarget.PROPERTY_SETTER,
+    AnnotationTarget.PROPERTY_SETTER
+)
+public annotation class InternalAPI
```
EugeneTheDev: Let's rename it to `InternalLLMClientApi` to better highlight the scope and purpose. We already have `InternalAgentToolsApi` annotation, so let's continue with this scoped annotations approach for now. We can introduce some single global `InternalKoogAPI` annotation later.

Note for maintainers: we will need to update custom inspection `MissingKDocForPublicAPI.inspection.kts` so that it doesn't require KDocs on public symbols marked with this annotation.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/build.gradle.kts
```diff
@@ -0,0 +1,60 @@
...
+                api(libs.ktor.serialization.kotlinx.json)
+                implementation(libs.oshai.kotlin.logging)
+            }
+        }
+
+        jvmMain {
+            dependencies {
+                api(libs.ktor.client.cio)
+                implementation(project(":prompt:prompt-executor:prompt-executor-clients:prompt-executor-anthropic-client"))
+                implementation(project(":prompt:prompt-executor:prompt-executor-clients:prompt-executor-google-client"))
```
EugeneTheDev: Do we need dependencies on OpenAI and Google clients here? Looks like only Anthropic DTOs are reused for now, please consider removing these dependencies

examples/src/main/kotlin/ai/koog/agents/example/client/BedrockAgent.kt
```diff
@@ -0,0 +1,48 @@
...
+import ai.koog.agents.core.tools.reflect.asTools
+import ai.koog.agents.example.ApiKeyService
+import ai.koog.agents.example.simpleapi.Switch
+import ai.koog.agents.example.simpleapi.SwitchTools
+import ai.koog.prompt.executor.clients.bedrock.BedrockClientSettings
+import ai.koog.prompt.executor.clients.bedrock.BedrockModels
+import ai.koog.prompt.executor.llms.all.simpleBedrockExecutor
+import kotlinx.coroutines.runBlocking
+
+fun main() = runBlocking {
```
EugeneTheDev: nit: Please specify return type (`Unit`) explicitly, otherwise there's a warning about potential non-Unit return type in main (since `runBlocking` can return some value).

EugeneTheDev (approved):
Generally looks good, thank you for this PR. I left a few minor comments which should be easy to fix, mostly style issues and couple small fixes, please check these. And also, please mark all public symbols in the Bedrock client module as `internal`, except for these that are actually meant to be public (client, config, models list).

Ololoshechkin (approved):

prompt/prompt-executor/prompt-executor-llms-all/src/commonMain/kotlin/ai/koog/prompt/executor/llms/all/SimplePromptExecutors.kt
kaeawc: It is - the issue is that Bedrock is in `jvmMain` and this code is in `commonMain`. Going to think about it a bit more before changing

PR #320, opened 2025-06-24
Retry component

antoniibelyshev:
Add retry node, which runs a node several times until a success condition is satisfied

Ticket: [JBAI-12616](https://youtrack.jetbrains.com/issue/JBAI-12616/Retry-component)

---

#### Type of the change
- [x] New feature
- [ ] Bug fix
- [ ] Documentation fix


agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/context/AIAgentContextBase.kt
```diff
@@ -180,4 +181,40 @@ public interface AIAgentContextBase {
...
+ * with the provided new context.
+ */
+internal suspend fun AIAgentContextBase.update(newContext: AIAgentContextBase) {
+    val newPrompt = newContext.llm.readSession { prompt }
+    llm.writeSession { rewritePrompt { newPrompt } }
+
+    stateManager.withStateLock { state ->
+        state.iterations = newContext.stateManager.withStateLock { newState -> newState.iterations }
+    }
+}
```
ptitjes: It looks to me like @tiginamaria is already adding something similar in https://github.com/JetBrains/koog/pull/220.

tiginamaria: It's already in develop, but has slightly different naming (`fork` and `replace`), feel free to reuse / modify / rename them

antoniibelyshev: I've updated my code to use `fork` and `replace` methods. @tiginamaria could you please take a look?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/extension/AIAgentNodes.kt
```diff
@@ -330,3 +331,91 @@ public inline fun <reified ToolArg : ToolArgs, reified TResult : ToolResult> AIA
...
+    node: AIAgentNodeBase<Input, Output>,
+    maxRetries: Int = 3,
+    successCondition: (Output) -> Boolean,
+): AIAgentNodeDelegateBase<Input, RetryResult<Output>> {
+    require(maxRetries > 0) { "maxRetries must be greater than 0" }
+    return node("retry_${node.name}") execute@{ input ->
+        (1..maxRetries).mapNotNull {
+            try {
+                val newContext = fork()
+                val output = node.execute(newContext, input)
```
tiginamaria: Looks a bit unsafe to call `node.execute` manually as it breaks the paradigm that node can not call node, but maybe @EugeneTheDev has another opinion on that. The workaround here can be implementing it as a `subgraph` (which make more sense to me) where define it as a normal workflow with edges and condition of going to finish or retry

tiginamaria (changes requested):
Thank you, looks nice! I just need another opinion here on how it should look like (as a `subgraph` or as a `node` calling another node). @EugeneTheDev @Ololoshechkin can you also comment here, as maybe it's already made decision

EugeneTheDev (changes requested):
I agree with @tiginamaria here. I would suggest implementing this component as a subgraph (similar to `subgraphWithTask`), since it looks like a higher-level component than just a node. 

Here's my proposed implementaiton, feel free to take it and check if it works and adjust some parts/names if needed. And let's put this component in the `agents-ext` module in the `agent/SubgraphWithRetry.kt` file. This looks like an appropriate place to start building a library of higher-level components. 

Note: I had to specify upper bound for `Input` and `Output`, that's because currently `AIAgentStorage` has `Any` as an upper bound for its elements. I would suggest making it `Any?` in storage implementation, then we won't need these bound here.

```kotlin
import ai.koog.agents.core.agent.entity.ToolSelectionStrategy
import ai.koog.agents.core.agent.entity.createStorageKey
import ai.koog.agents.core.dsl.builder.AIAgentSubgraphBuilderBase
import ai.koog.agents.core.dsl.builder.AIAgentSubgraphDelegateBase
import ai.koog.agents.core.dsl.builder.forwardTo

/**
 * Represents the result of [subgraphWithRetry].
 *
 * @param result The result of the subgraph operation.
 * @param success A boolean indicating whether the action was successful.
 * @param retryCount The number of retries attempted.
 */
public data class RetrySubgraphResult<T>(
    val result: T,
    val success: Boolean,
    val retryCount: Int,
)

/**
 * Creates a subgraph with retry mechanism, allowing a specified action subgraph to be retried multiple
 * times until a given condition is met or the maximum number of retries is reached.
 *
 * @param toolSelectionStrategy The strategy used to select a tool for executing the action.
 * @param condition A suspend function that evaluates whether the output meets the desired condition.
 * @param maxRetries The maximum number of allowed retries. Must be greater than 0.
 * @param name The optional name of the subgraph.
 * @param defineAction A lambda defining the action subgraph to perform within the retry subgraph.
 */
public fun <Input : Any, Output : Any> AIAgentSubgraphBuilderBase<*, *>.subgraphWithRetry(
    toolSelectionStrategy: ToolSelectionStrategy,
    condition: suspend (Output) -> Boolean,
    maxRetries: Int,
    name: String? = null,
    defineAction: AIAgentSubgraphBuilderBase<Input, Output>.() -> Unit,
): AIAgentSubgraphDelegateBase<Input, RetrySubgraphResult<Output>> {
    require(maxRetries > 0) { "maxRetries must be greater than 0" }

    return subgraph(name = name) {
        val retriesKey = createStorageKey<Int>("retires")
        val initialInputKey = createStorageKey<Any>("initial_input")

        val actionSubgraph by subgraph(
            name = "${name}_action",
            toolSelectionStrategy = toolSelectionStrategy,
            define = defineAction
        )

        val beforeAction by node<Input, Input> { input ->
            val retries = storage.get(retriesKey) ?: 0

            // Store initial input on the first run
            if (retries == 0) {
                storage.set(initialInputKey, input)
            }
            // Increment retries
            storage.set(retriesKey, retries + 1)

            input
        }

        val decide by node<Output, RetrySubgraphResult<Output>> { output ->
            val retries = storage.getValue(retriesKey)
            val success = condition(output)

            RetrySubgraphResult(
                result = output,
                success = success,
                retryCount = retries
            )
        }

        nodeStart then beforeAction then actionSubgraph then decide

        // Repeat the action with initial input when condition is not met and the number of retries does not exceed max retries.
        edge(
            decide forwardTo beforeAction
                onCondition { result -> !result.success && result.retryCount <= maxRetries }
                transformed {
                    @Suppress("UNCHECKED_CAST")
                    storage.getValue(initialInputKey) as Input
                }
        )

        // Otherwise return the last iteration result.
        edge(
            decide forwardTo nodeFinish
                onCondition { result -> result.success || result.retryCount > maxRetries }
        )
    }
}
```

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/SubgraphWithRetry.kt
```diff
@@ -0,0 +1,163 @@
...
+    condition: suspend (Output) -> Boolean,
+    maxRetries: Int,
+    toolSelectionStrategy: ToolSelectionStrategy = ToolSelectionStrategy.ALL,
+    name: String? = null,
+    defineAction: AIAgentSubgraphBuilderBase<Input, Output>.() -> Unit,
+): AIAgentSubgraphDelegate<Input, RetrySubgraphResult<Output>> {
+    require(maxRetries > 0) { "maxRetries must be greater than 0" }
+
+    return subgraph(name = name) {
+        val retriesKey = createStorageKey<Int>("${name}_retires")
```
tiginamaria, line 43: Not sure if we should use storage here, as it can probably rewrite user's key. Maybe we can reserve some name for us and come up with naming convention for them (e.g., \_\_name\_\_)

tiginamaria: As a workaround here, we can operate with the input wrapped into the class which also contains the retries info (something like `RetrySubgraphResult`)

tiginamaria: And also add initial input and context there

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/SubgraphWithRetry.kt
```diff
@@ -0,0 +1,163 @@
...
+            input
+        }
+
+        val actionSubgraph by subgraph(
+            name = "${name}_retryableAction",
+            toolSelectionStrategy = toolSelectionStrategy,
+            define = defineAction
+        )
+
+        val decide by node<Output, RetrySubgraphResult<Output>> { output ->
```
tiginamaria, line 72: Do we want to keep it as a node or move this logic to edge (condition + transform)?

tiginamaria (changes requested):
Thank you for changes, looks much better, but one more time need second though of Andrey on some high-level design questions

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/SubgraphWithRetry.kt
EugeneTheDev: > Not sure if we should use storage here, as it can probably rewrite user's key

We won't rewirte anything, because the key in the storage is not a string with a name (it's here only for convenience), but the object. So, every time you call `createStorageKey`, a unique object (key) is created, even if the names are the same. So it's safe to do it like this, we won't overwrite any existing data.

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/SubgraphWithRetry.kt
EugeneTheDev: Initially I thought the same, but then you get duplicated transformation code, that's why it looks a bit cleaner to put it in a single place - in a node.

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/SubgraphWithRetry.kt
```diff
@@ -0,0 +1,163 @@
...
+ *     maxRetries = 2,
+ * ) {
+ *     val nodeCallLLM by nodeLLMRequest("sendInput")
+ *     nodeStart then nodeCallLLM then nodeFinish
+ * }
+ * val nodeExecuteTool by nodeExecuteTool("nodeExecuteTool")
+ * edge(subgraphRetryCallLLM forwardTo nodeExecuteTool onToolCall { true })
+ * ```
+ */
+public fun <Input : Any, Output> AIAgentSubgraphBuilderBase<*, *>.subgraphWithRetryPlain(
```
EugeneTheDev: It's better to call it `...Strict` then, this would be more descriptive

EugeneTheDev (approved):

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/SubgraphWithRetry.kt
```diff
@@ -0,0 +1,163 @@
...
+            storage.remove(initialContextKey)
+            result
+        }
+
+        nodeStart then beforeAction then actionSubgraph then decide
+
+        // Repeat the action with initial input when condition is not met and the number of retries does not exceed max retries.
+        edge(
+            decide forwardTo beforeAction
+                    onCondition { result -> !result.success && result.retryCount < maxRetries }
```
EugeneTheDev, line 95: Oh, actually it should be `retryCount <= maxRetries`, since we allow up to and including `maxRetries`. And the same for condition below, strict `>`

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/SubgraphWithRetry.kt
antoniibelyshev: Strict is a flag: this subgraph could represent either a strict retry or not. Plain refers to the Output: it is not wrapped into the `RetryResult`. I can try to think of a better name, but I do not think `...Strict` describes this subgraph

EugeneTheDev: Oh, sorry, missed that part. Then yes, maybe something like `...Simple` would sound better

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/SubgraphWithRetry.kt
antoniibelyshev: That probably depends on what do you call retry: specifically, the first time you do the action is it your zeroth or first retry? I think it is more convenient if `maxRetries` represents the number of times you perform the action (i.e. the first action is the first retry), then the current implementation works as expected. Do you think I should change the name of `maxRetries` then?

EugeneTheDev: I think the number of retries is the number of subgraph executions. So your first execution should be your first retry. It seems more intutive to me.

EugeneTheDev: So in your current implementation you are actually allowing `maxRetries - 1` retries, which to me seems unintuitive

EugeneTheDev: Oh, sorry, actually I realized just now after rereading the code that I was wrong, so your current implementation is indeed correct. Sorry

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/SubgraphWithRetry.kt
antoniibelyshev: I think neither of the two describes what happens, but yours is _simpler_, so I'll go with it

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/SubgraphWithRetry.kt
antoniibelyshev: I've faced a similar dilemma a few times. Should we discuss introduction of something like `preprocessed` / `postprocessed` functions analogous to `transformed` but applied to a single node rather than an edge? Probably, that's a topic for the technical meet

tiginamaria (approved):
Looks good, thanks!

PR #326, opened 2025-06-24
[agent] Add custom input/output types support for AIAgent

EugeneTheDev:
Currently, agent can take only string as an input and produce string as an output. This PR adds generic parameters to agent and agent strategy to allow specifying custom types. 

I also updated structured data example to demonstrate custom input/output in `examples/src/main/kotlin/ai/koog/agents/example/structureddata/StructuredDataExample.kt`.

Also, I made a couple of related API simplifications that became more evident once I introduced arbitrary input/output types. I made them as a separate commits so I can revert these changes easier if someone is not happy with them. But I do believe they are actually benefical, so let's discuss them if anyone has any questions:
1. Removed `sendTerminate` from `AIAgentEnvironment`. 
    * The major problem, it doesn't actually terminate the agent. The agent continues to run, so this method is confusing and can potentially be misused.
    * The idea of this method is not really clear, because it's not consistent with how we handle agent flow currently. We do not report that the agent has been started and its input, so we should either add that, or, like I did here, remove `sendTermination`.
    * And if we do want to support it for some reason, then API gets a bit scarier, because we now need to add input/output serializers to the `AIAgent` constructor parameters (to be able to serialize and send these to the environment). 
2. Removed `runAndGetResult` from `AIAgent` and made `run` always return the output. This one seems quite logical, because the previous name was kinda long. And since agent always returns something, we can just make the primary `run` method to always return the result, and you can just ignore the output if you're not interested in it.

And I also want to propose upgrading to the latest Kotlin version, `2.2.0`. With the current version I sometimes get what I think is a memory leak in the Kotlin compiler (`OutOfMemory Java heap space` during compilation). I suppose this is very likely due to the autmatic resoultion of generic parameters. Basically, in certain cases Kotlin is supposed to resolve generic types in `strategy` builder function automagically. But instead it fails to do so, and not with a compilation error, which would be clear and easy to fix, but with this `OutOfMemory`. This is not very intutivie and makes devex worse, because now you are forced to find and fix all such cases by hand, with no actual feedback from the compiler. But I tested it on the latest `2.2.0` and it worked, so maybe we can upgrade to that. Or maybe I'm doing something wrong here and it should be possible to workaroud this issue without upgrading Kotlin.


Ololoshechkin (approved):

tiginamaria (approved):
Looks just perfect for me, no concerns and comments. 
The only thing that can appear in soon is that if (when) we will have A2A or remote agents we will need this termination with result call, but it will be the problem of future =)

Issue #295, opened 2025-06-17
[Suggestion] Ability to send num_ctx parameter to Ollama to change Context Window size


krissrex:
By default, ollama will use the `num_ctx` set in a modelfile parameters, or fall back to a low value between 1k and 8k. I think the default depends on how ollama is used (cli vs api).

In a chat, I can change the context window with `/set parameter num_ctx 131072` for much higher memory usage and full context of `llama3.2`.
In the API, the `options` object can take a `num_ctx` https://ollama.readthedocs.io/en/api/#request_7 .

For some tasks, we want a much higher context window.

### Workaround

The currently available option is to create a new model with `num_ctx`, like installing a new Modelfile or running `/set parameter num_ctx 20000` followed by `/save llama3.2-20k_ctx` .
Or set the global default when starting ollama, with environment variable `OLLAMA_CONTEXT_LENGTH=20000`.

ollama logs:
```text
llama_context: constructing llama_context
llama_context: n_seq_max     = 1
llama_context: n_ctx         = 131072
llama_context: n_ctx_per_seq = 131072
llama_context: n_batch       = 512
llama_context: n_ubatch      = 512
llama_context: causal_attn   = 1
llama_context: flash_attn    = 1
llama_context: freq_base     = 500000.0
llama_context: freq_scale    = 1
```

---

## Details and notes

Enable debug logging
```sh
OLLAMA_DEBUG=1 ollama serve
```

Ollama will log during model loading, pay attention to `runner.num_ctx=8192`:

```text
time=2025-06-17T15:46:27.259+02:00 level=DEBUG source=sched.go:495 msg="finished setting up" runner.name=registry.ollama.ai/library/llama3.2:latest runner.inference=metal runner.devices=1 runner.size="3.3 GiB" runner.vram="3.3 GiB" runner.parallel=2 runner.pid=36317 runner.model=/Users/kristian/.ollama/models/blobs/sha256-dde5aa3fc5ffc17176b5e8bdc82f587b24b2678c6c66101bf7da77af9f7ccdff runner.num_ctx=8192
```

```text
llama_context: constructing llama_context
llama_context: n_seq_max     = 2
llama_context: n_ctx         = 8192
llama_context: n_ctx_per_seq = 4096
llama_context: n_batch       = 1024
llama_context: n_ubatch      = 512
llama_context: causal_attn   = 1
llama_context: flash_attn    = 1
llama_context: freq_base     = 500000.0
llama_context: freq_scale    = 1
[...]

PR #335, opened 2025-06-25
Feature/ollama request customization

ptitjes:
This PR add the ability to customize Ollama chat requests as discussed in #295.

@aozherelyeva I had to make two modification to the `OllamaTestFixture` and I hope it's OK:
- Added a check to pull the llama3.2 model in the container, because by default the image that you suggested in TESTING.md doesn't come with the llama3.2 model. Now it works without touching anything (provided you have docker obviously).
- Exposed the `baseUrl` to allow some tests to initialize a custom `OllamaClient` if required.

---

#### Type of the change
- [x] New feature
- [ ] Bug fix
- [ ] Documentation fix


ptitjes:
And I had forgotten to add Kdocs. That's now fixed.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-ollama-client/src/commonMain/kotlin/ai/koog/prompt/executor/ollama/client/OllamaClient.kt
```diff
@@ -42,7 +76,8 @@ public class OllamaClient(
     private val baseUrl: String = "http://localhost:11434",
     baseClient: HttpClient = HttpClient(engineFactoryProvider()),
     timeoutConfig: ConnectionTimeoutConfig = ConnectionTimeoutConfig(),
-    private val clock: Clock = Clock.System
+    private val requestBuilderAction: (OllamaRequestBuilder.(prompt: Prompt, model: LLModel) -> Unit)? = null,
+    private val clock: Clock = Clock.System,
```
Ololoshechkin: Why not `requestBuilderAction: OllamaRequestBuilder.(prompt: Prompt, model: LLModel) -> Unit = {}` ?

also why do you need the `prompt` here as a parameter?

Ololoshechkin: Also if you make this lambda the last parameter -- you would be able to write 
```
val client = OllamaClient(...) {
   seed = 0
   numCtx = 100
}
```

Also, could you please clarify once again -- why do you need to configure it with a lambda and not just pass a data object with this parameter, ex:

```
val client = OllamaClient(..., requestOptions = OllamaRequestOptions(seed = 0, numCtx = 100))
```

Ololoshechkin (changes requested):

prompt/prompt-executor/prompt-executor-clients/prompt-executor-ollama-client/src/commonMain/kotlin/ai/koog/prompt/executor/ollama/client/OllamaClient.kt
ptitjes: > Why not `requestBuilderAction: OllamaRequestBuilder.(prompt: Prompt, model: LLModel) -> Unit = {}` ?

You're right making it nullable doesn't bring any real optimization.

> Also if you make this lambda the last parameter

I was actually trying to avoid putting it as last parameter, because those are sensible parameters and I wanted for the user to consciously decide to implement the lambda, and not implement a trailing  lambda because IDEA suggests it.

> also why do you need the `prompt` here as a parameter?
> could you please clarify once again -- why do you need to configure it with a lambda and not just pass a data object with this parameter

So, Ollama needs to allocate some VRAM to handle your context. `num_ctx` tells Ollama that you want it to be able to take `num_ctx` tokens in account when generating the next token. If you do not set `num_ctx`, it will use the constant value configured for the server (`OLLAMA_CONTEXT_LENGTH`, which is 4096 by default). **Any token that doesn't fit in this context window, will be arbitrarily truncated**!

When working with local models, you want the user to be able to finely tune this because you might be in a constrained environment. Also, you cannot ask the server admin to put a big value for the `OLLAMA_CONTEXT_LENGTH`, because a big context is often not needed by most of the requests.

The size of the context window needs to be decided at each request, if there is a big context in the prompt. If the user can estimate (either by looking at `lastTokenUsage` or using a tokenizer) that the prompt contains than 4096 tokens, then they need to allocate more context with `num_ctx`.

Proprietary LLMs behind APIs do the same. At some point, they have to decide how much VRAM they allocate for your request (or rather, on what node with preallocated models they will forward your request). You just don't see it.

Unfortunately, with Ollama, it is the responsibility of the caller to define the allocated context size (in the limits of the actual max context length of the model, obviously).

When handling long-running conversations or maybe doing RAG with lots of augmented data, a typical implementation would look something like, for instance:
```kotlin
  requestBuilderAction = { prompt, model ->
    val tokenCount = prompt.lastTokenUsage
    val maxContextSize = model.contextSize // <-- custom extension property here
    numCtx = minOf(maxOf(2048, ((tokenCount + 500) / 1024 + 1) * 1024), maxContextSize)
  }
```

BTW, it seems nothing is done still on the AI Assistant side: https://youtrack.jetbrains.com/issue/LLM-13677/Lift-context-size-restrictions-for-local-Ollama-model-or-make-it-configurable

ptitjes: Ideally, the `numCtx` would be an LLMParam in the prompt. But there is currently no way to have provider-specific LLMParams.

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/OllamaClientIntegrationTest.kt
```diff
@@ -37,6 +41,30 @@ class OllamaClientIntegrationTest {
...
+
+        val prompt = Prompt.build("test-with-options") {
+            system("You are a helpful assistant.")
+            user("What is the capital of France?")
+        }
+
+        val response = executor.execute(prompt = prompt, model = model)
+
+        assertTrue(response.content.isNotEmpty(), "Response should not be empty")
+        assertEquals("The capital of France is Paris.", response.content)
```
zarechneva-a: I'd suggest just checking for some keywords ("capital", "Paris") rather than checking for the whole line correspondence. I'm afraid that this can make test flaky. 

Although I understand the idea of checking the predictable output with `seed=0`, maybe we could run the `execute` method twice and make sure that the outputs are equal without comparing them to an exact string?

ptitjes: > Although I understand the idea of checking the predictable output with `seed=0`, maybe we could run the `execute` method twice and make sure that the outputs are equal without comparing them to an exact string?

Yeah, that would lower the risks of flaky tests. I will do that.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-ollama-client/src/commonMain/kotlin/ai/koog/prompt/executor/ollama/client/OllamaClient.kt
krissrex: > The size of the context window needs to be decided at each request, if there is a big context in the prompt.

I think estimating the size based on the prompt may be tricky. You can't take into account the future output of the llm itself. It would be sufficient for me to just set the num_ctx to the max possible. But perhaps there are scenarios where this dynamic `num_ctx` is useful.

I assume this `requestBuilderAction` also runs before tool call results? So the llm can upsize in case of a big tool result.

devcrocod: I agree with @Ololoshechkin that using a lambda as the last parameter is better.
I understand your motivation to make it more explicit, but in my opinion, code readability is more important, and this change makes it worse.

Also, if possible, it’s better to use only one receive parameter inside the lambda. it’s not yet clear to me why both `prompt` and `model` are needed there

ptitjes: > It would be sufficient for me to just set the num_ctx to the max possible.

I think that would impair a lot of users, because not everybody has great hardware...

For example, I run my agents on my tiny AMD 5700 with 8GB of VRAM. I usually use `hf.co/mradermacher/ToolACE-2-Llama-3.1-8B-i1-GGUF:Q4_K_M` for tool related stuff. It is a llama3.1 fine-tune, and thus supports a 128K (131072) context length. If I set `OLLAMA_CONTEXT_LENGTH=131072`, this mean 16384.00 MiB for the KV cache in addition to the 4685.30 MiB of the model. I get all of the model and cache loaded into RAM and using CPU. When I try to keep the context relatively small, I get at some parts of the model and cache loaded in the VRAM and using the GPU, and the generation is very much faster... So, automatically setting `num_ctx` to the max context length of the model would make Koog unusable for me.

Also, newer models like llama4 scout support up to 10M context length. Automatically setting `num_ctx` to the max would make any request (even with the smallest context) unmanageable even on the beefiest hardware. (In q8_0 KV cache type, Ollama stores 16K context per GiB, which means 640 GiB of (V)RAM needed!!! :eyes: )

> I assume this `requestBuilderAction` also runs before tool call results? So the llm can upsize in case of a big tool result.

It runs at each LLM request, so also before the request after adding the tool call results to the prompt.

ptitjes: > I agree with @Ololoshechkin that using a lambda as the last parameter is better.
> I understand your motivation to make it more explicit, but in my opinion, code readability is more important, and this change makes it worse.

Sure, I really don't mind putting it as last parameter. I will do that then.

> Also, if possible, it’s better to use only one receive parameter inside the lambda. it’s not yet clear to me why both `prompt` and `model` are needed there

- The prompt is needed to estimate how much context length is needed, without having Ollama truncate the input.
- The model is needed to know how much context length is supported.

I don't how we could do without both those parameters. Please, give me ideas...

krissrex: The context calculation has prompt and model, but does this prompt also include the tools formulation? I'm thinking if we have a list of tools with names and descriptions, this fills the context as well. Either from Koog if it builds the description, or inside ollama if it expands the model template.

ptitjes: > The context calculation has prompt and model, but does this prompt also include the tools formulation? I'm thinking if we have a list of tools with names and descriptions, this fills the context as well. Either from Koog if it builds the description, or inside ollama if it expands the model template.

No, the prompt don't contain the tool formulation. Indeed, Ollama is expanding that itself in the model template. So, in any case, we can only do a token count estimation.

But isn't that better than doing nothing?

ptitjes:
Rebased on develop and applied requested changes:
- made `requestBuilderAction` the last (non-nullable) constructor parameter
- checked that the response is the same when seed=0

prompt/prompt-executor/prompt-executor-clients/prompt-executor-ollama-client/src/commonMain/kotlin/ai/koog/prompt/executor/ollama/client/OllamaClient.kt
krissrex: Yes, it should be sufficient for my case of using a static context value. Anyone with specific needs could heuristically add some more overhead for tools and templates. LGTM 👍

aozherelyeva (approved):

ptitjes:
I'll have to rebase because the llama3.2 pull of https://github.com/JetBrains/koog/pull/335/commits/1012015883cd4bf9adb3f76c4c0efacf84eaf385 was merged in #371.

ptitjes:
Rebased on develop and removed the commit which @aozherelyeva also implemented in develop.

integration-tests/src/jvmMain/kotlin/ai/koog/integration/tests/OllamaTestFixture.kt
```diff
@@ -18,6 +18,7 @@ class OllamaTestFixture {
 
     private lateinit var ollamaContainer: GenericContainer<*>
 
+    lateinit var baseUrl: String
```
devcrocod: Why is `baseUrl` `latenit` here?

prompt/prompt-executor/prompt-executor-clients/prompt-executor-ollama-client/src/commonMain/kotlin/ai/koog/prompt/executor/ollama/client/OllamaClient.kt
```diff
@@ -42,7 +78,8 @@ public class OllamaClient(
     private val baseUrl: String = "http://localhost:11434",
     baseClient: HttpClient = HttpClient(engineFactoryProvider()),
     timeoutConfig: ConnectionTimeoutConfig = ConnectionTimeoutConfig(),
-    private val clock: Clock = Clock.System
+    private val clock: Clock = Clock.System,
+    private val requestBuilderAction: OllamaRequestBuilder.(prompt: Prompt, model: LLModel) -> Unit = { _, _ -> },
```
devcrocod: As I mentioned earlier, it’s better to use two params in context receivers only when necessary, since it complicates code completion. In my opinion, we can get rid of `model` here. Besides, it introduces a potential source of bugs, as there’s no guarantee that the model here will be the same as the one used in execute.

could apply a not-so-pretty workaround: for example, introduce a flag in RequestBuilder to indicate that a large contextSize should be used. Alternatively, could set a maximum and truncate based on the current model’s limit — but that would be implicit and unclear to the user

prompt/prompt-executor/prompt-executor-clients/prompt-executor-ollama-client/src/commonMain/kotlin/ai/koog/prompt/executor/ollama/client/dto/OllamaModels.kt
```diff
@@ -71,6 +71,9 @@ internal data class OllamaChatRequestDTO(
     @Serializable
     internal data class Options(
         val temperature: Double? = null,
+        val seed: Int? = null,
+        @SerialName("num_ctx") val numCtx: Int? = null,
```
devcrocod: Doesn’t the Ollama json config assume the use of `snake_case` flag, so that we don’t have to use `@SerialName`?

devcrocod (commented):
I’m worried that Ollama’s implementation is drifting further and further from the others.

Also, I think it would be better to add additional functions or constructors to creating so that I can store the request parameters in a variable and pass them that way. It’s more convenient when writing custom wrapper functions on top of koog

integration-tests/src/jvmMain/kotlin/ai/koog/integration/tests/OllamaTestFixture.kt
ptitjes: Because it is initialised in `setUp` (so after construction) from the ollama test container's host and port, just like `client` and `executor`.

Would you prefer that I change its type to `String?`?

prompt/prompt-executor/prompt-executor-clients/prompt-executor-ollama-client/src/commonMain/kotlin/ai/koog/prompt/executor/ollama/client/OllamaClient.kt
ptitjes: > As I mentioned earlier, it’s better to use two params in context receivers only when necessary, since it complicates code completion. In my opinion, we can get rid of `model` here.

I understand and agree with that. But I think it is necessary here. Otherwise, how will the framework user be able to handle the context size correctly if they cannot know what model they are setting `num_ctx` for?

For instance, llama3.1 8B has a max context length of 128K tokens whereas granite3.2-vision 2B has a max context length of 16K tokens. Those values are really different and as Ollama is often used in constraint environments, it is indispensable to know what model we are tweaking `num_ctx` for. 

> Besides, it introduces a potential source of bugs, as there’s no guarantee that the model here will be the same as the one used in execute.

I don't understand what you mean here. This `requestBuilderAction` lambda is exactly called inside `execute` passing the very prompt and model provided to the `execute()` call. How come is there "no guarantee that the model here will be the same as the one used in execute"?

> could apply a not-so-pretty workaround: for example, introduce a flag in RequestBuilder to indicate that a large contextSize should be used. Alternatively, could set a maximum and truncate based on the current model’s limit — but that would be implicit and unclear to the user

I can set `model` as an immutable val field of `OllamaRequestBuilder`, but that would be even way worse here, because then `OllamaRequestBuilder` would both have mutable fields to build the request and an immutable field to get information about the model... This would make it confusing and hardly discoverable for the user, which is bad design IMO.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-ollama-client/src/commonMain/kotlin/ai/koog/prompt/executor/ollama/client/dto/OllamaModels.kt
ptitjes: Well, there are `@SerialName(...)` annotations everywhere in the Ollama DTO definitions.

To be honest, I didn't even know about kotlinx-serialization `namingStrategy` option. So thanks for making me discover that.

However, it seems that using global naming strategies may not be the best idea. To quote `JsonNamingStrategy`'s KDocs:
> ### Controversy about using global naming strategies
>
> Global naming strategies have one key trait that makes them a debatable and controversial topic: They are very implicit. It means that by looking only at the definition of the class, it is impossible to say which names it will have in the serialized form. As a consequence, naming strategies are not friendly to refactorings. Programmer renaming myId to userId may forget to rename my_id, and vice versa. Generally, any tools one can imagine work poorly with global naming strategies: Find Usages/Rename in IDE, full-text search by grep, etc. For them, the original name and the transformed are two different things; changing one without the other may introduce bugs in many unexpected ways. The lack of a single place of definition, the inability to use automated tools, and more error-prone code lead to greater maintenance efforts for code with global naming strategies. However, there are cases where usage of naming strategies is inevitable, such as interop with an existing API or migrating a large codebase. Therefore, one should carefully weigh the pros and cons before considering adding global naming strategies to an application.

integration-tests/src/jvmMain/kotlin/ai/koog/integration/tests/OllamaTestFixture.kt
devcrocod: Oh, I see
I thought the URL was already set at the time the class was initialized

prompt/prompt-executor/prompt-executor-clients/prompt-executor-ollama-client/src/commonMain/kotlin/ai/koog/prompt/executor/ollama/client/OllamaClient.kt
devcrocod: > I don't understand what you mean here. This requestBuilderAction lambda is exactly called inside execute passing the very prompt and model provided to the execute() call. How come is there "no guarantee that the model here will be the same as the one used in execute"?

Yea, i missed that
you're absolutely right

> I can set model as an immutable val field of OllamaRequestBuilder, but that would be even way worse here, because then OllamaRequestBuilder would both have mutable fields to build the request and an immutable field to get information about the model... This would make it confusing and hardly discoverable for the user, which is bad design IMO.

true

prompt/prompt-executor/prompt-executor-clients/prompt-executor-ollama-client/src/commonMain/kotlin/ai/koog/prompt/executor/ollama/client/dto/OllamaModels.kt
devcrocod: Ideally, all of this should be caught by tests, so I don’t see a big issue with it
Besides, we already use `namingStrategy` in other clients
So far, the only real advantage is a slight reduction in code

I don’t think it’s critical, since we can remove it or add it to the other clients at any time if needed

ptitjes:
Let's put this on hold for now and take some time discussing a better solution. I can live with configuring bigger context sizes directly on Ollama for some time. I just hope we can agree on an acceptable design in the next Koog releases, because this is at the expense of memory and performance.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-ollama-client/src/commonMain/kotlin/ai/koog/prompt/executor/ollama/client/OllamaClient.kt
Ololoshechkin: Hi @ptitjes ! I actually looked through this conversation again, and it looks like adding an example to the `examples` module that would showcase the Ollama with fine-tuned dynamic context lenght based on the prompt and model -- will be ideal. Plus, if you could put some textual explanations there in this example, that would benefit everyone's understanding and enrich the knowledgebase around Koog. WDYT? :)

Ololoshechkin: Also would be great to reuse the `Tokenizer` here  for prompt estimation.

Ololoshechkin: Other than that, might be a great idea to make the main kdoc above the `OllamaClient` much longer and explain all these details, and even provide a few code samples how to use this new configuration lambda (feel free to check the kdoc of `AgentMemory.Feature` or `EventHandler.Feature` for instance -- we already have quite long explanations in Kdocs :)). Also it's nice for LLMs to learn on these things once they read them :)

prompt/prompt-executor/prompt-executor-clients/prompt-executor-ollama-client/src/commonMain/kotlin/ai/koog/prompt/executor/ollama/client/OllamaClient.kt
Ololoshechkin: @ptitjes I believe we can just move all not-so-much-popular params to `OllamaRequestBuilder` receiver. Then, `model` will be always available (those who know they know :) ), while `prompt` will be the only param of the lambda. We usually follow such approach in the DSLs in general.

WDYT?

Ololoshechkin: > I can set model as an immutable val field of OllamaRequestBuilder, but that would be even way worse here, because then OllamaRequestBuilder would both have mutable fields to build the request and an immutable field to get information about the model... This would make it confusing and hardly discoverable for the user, which is bad design IMO.

> true

Hmm... actually I don't think it's a problem tbh. I would assume that the most popular case will be using one single Ollama model for the whole agent. Hence, the average user won't be accessing this `model` param at all.

Also, on a separate note -- we are trying to also provide some common solutions in Koog (likeout-of-the-box history compression strategies, or `selectByIndex` / `fold` / `selectBy` / ... DSL in parallel-merge , etc.) while keeping the ability to write the "raw" code. Maybe we can consider making some extensions on the `OllamaRequestBuilder` receiver so that users won't have to write this:

```kotlin
OllamaClient { prompt, model ->
    val tokenCount = prompt.lastTokenUsage
    val maxContextSize = model.contextSize // <-- custom extension property here
    numCtx = minOf(maxOf(2048, ((tokenCount + 500) / 1024 + 1) * 1024), maxContextSize)
  }
  ```
  
  But instead, they would write something like this:
  ```kotlin
OllamaClient {
    truncateRequestTokensBy(500)
    
    // and same for other things.
}
```

Also, it would be amazing to just have some default strategy of context truncations (i.e. default implementation of this lambda) so that one wouldn't have to write it at all and still be fine with performance and almost never overfit (but also we should provide a bunch of non-default but out-of-the-box strategies to select from, and only the last shelter is -- to write it manually, that is also possible).

Btw, while writing this, I just thought -- maybe it would make sense to actually introduce some `ContextWindowStrategy` interface that would have some default nested objects-implementation to select from. 

And then, one would just write something like:

- default usage: `val client = OllamaClient()`
- equivalent default usage: `val client = OllamaClient(contextWindowStrategy = ContextWindowStrategy.Default)`
- some custom usage: `val client = OllamaClient(contextWindowStrategy = ContextWindowStrategy.TruncateRequestBy(500)`
- some other custom usage: `val client = OllamaClient(contextWindowStrategy = ContextWindowStrategy.Fixed(windowSize = 1024)`
- some other custom usage: `val client = OllamaClient(contextWindowStrategy = ContextWindowStrategy.ModelBased(ratio = 0.75)` // 0.75 of the model max size, whatever model is
- and if someone really want -- they would implement their own:
```
class MyVeryCustomContextWindowStrategy: ContextWindowStrategy {

    override val seed = 100

    override fun estimateRequestWindow(prompt: Prompt, model: Model): Int {
        val tokenCount = prompt.lastTokenUsage
        val maxContextSize = model.contextSize // <-- custom extension property here

        return minOf(maxOf(2048, ((tokenCount + 500) / 1024 + 1) * 1024), maxContextSize)
    }
    
    override fun estimateResponseWindow(prompt: Prompt, model: Model): Int {
        ...
    }
}
```

Very much similar to the `nodeLLMCompressHistory()` with optional `strategy` and a bunch of default ones to select from.


WDYT?

Ololoshechkin:
> I have faith that in the future small LLMs with great performances will become more and more common. That is good for people and good for the climate. IMO Koog has to support these use-cases.

Totally agree with you here, @ptitjes !

prompt/prompt-executor/prompt-executor-clients/prompt-executor-ollama-client/src/commonMain/kotlin/ai/koog/prompt/executor/ollama/client/OllamaClient.kt
ptitjes: > adding an example to the examples module that would showcase the Ollama with fine-tuned dynamic context lenght based on the prompt and model -- will be ideal. Plus, if you could put some textual explanations there in this example, that would benefit everyone's understanding and enrich the knowledgebase around Koog. WDYT? :)

> make the main kdoc above the OllamaClient much longer and explain all these details

Yeah, definitely.

ptitjes:
> I don't think we can make it for 0.3.0 already, unfortunately :(
>
> But in fact -- adding ollama as a first thing and as experimental in 0.3.* is fine -- we don't have to be waiting for 0.4.0.

Yes, I prefer we get an almost right design, rather than rush it.

> Experimental annotation -- no, we currently don't have it. So might be a great time to add it.

I definitely think that we should have such an annotation and start being more conservative (maybe not from 0.3 or 0.4, but soon) on the new APIs we introduce. Otherwise that might bite us later.

ptitjes:
@Ololoshechkin I really like your `ContextWindowStrategy` interface proposal, coming with some built-in implementations and a sensible default.

I will rework this PR in this way. I need a few days to think about the appropriate built-in implementations to cover the main use-cases.

Some initial thoughts though:

1. I wood remove `seed` there, as it is completely unrelated. But at the same time, I think it's only useful for tests so I wouldn't mind having that as a simple constructor parameter of `OllamaClient`. If you are OK with that, I will do a separate PR with only this and the corresponding integration test.

2. This `ContextWindowStrategy` would be something specific to Ollama, as I believe none of the other clients need anything similar. As far as I can see, all the proprietary models behind API handle this by themselves. So that wouldn't make much sense for those, right?

3. I am convinced we should add new attributes to `LLModel`, namely `contextLength : Long` and `embeddingLength : Long`. That would be really helpful to inform the context summarization strategies (and also this `ContextWindowStrategy` as a result). WDYT?

prompt/prompt-executor/prompt-executor-clients/prompt-executor-ollama-client/src/commonMain/kotlin/ai/koog/prompt/executor/ollama/client/OllamaClient.kt
ptitjes: @Ololoshechkin I started implementing something like this, but I am really wondering if the `maxResponseTokenCount` shouldn't simply be in `LLMParams` as all the proprietary-models behind APIs also have some form of max generated tokens.

If we decide to make this one an `LLMParams`, then I would only compute the max context window length in the strategy.

Does this make sense? WDYT?

ptitjes:
So, finally coming back to this. I rewrote everything following @Ololoshechkin's advice.

I defined a `ContextWindowStrategy` with a single method that is responsible to compute the context length value to send to Ollama as `num_ctx`. This value computed value can be null in case the user wishes to let the Ollama server deal with that. (This is fully documented in the KDocs of `ContextWindowStrategy`.)

I implemented three basic strategies:
- `None`, that simply returns `null` to let the Ollama server deal with this.
- `Fixed`, that returns a constant value. If this value is greater than what the current model supports, then we fallback to the maximum context length supported by the model (and log a warning).
- `FitPrompt`, that computes a context length value so that the current prompt fits in the context window. This strategy uses either the given prompt tokenizer or the last reported token usage in the prompt. The strategy also ensured that computed context length is a multiple of the given `granularity` (in order to avoid too many context length changes, and thus model reloads by Ollama) and is coerced between a given minimum context length and the max supported context length of the model.

I had to move `PromptTokenizer` and its implementation to the `prompt-tokenizer` module. (I don't know why it was lost near the Tokenizer agent feature.)

I added unit tests that check that the correct `num_ctx` value is sent to Ollama, by using a MockHttpClient. I considered that using integration tests would be really difficult (given we would have to produce very long prompts) and so not worth the cost.

aozherelyeva (approved):
The part regarding the `OllamaTestFixture` is more than okay, thanks for the contribution!

prompt/prompt-executor/prompt-executor-clients/prompt-executor-ollama-client/src/commonMain/kotlin/ai/koog/prompt/executor/ollama/client/ContextWindowStrategy.kt
```diff
@@ -0,0 +1,112 @@
...
+         * @param granularity The granularity to use for computing the context window length. Defaults to 2048.
+         * @param minimumContextLength The minimum context window length,
+         *   if the prompt length is less than it or cannot be computed yet.
+         *   If not null, [minimumContextLength] must be a multiple of the [granularity].
+         *   If null, we let Ollama decide the context window length.
+         */
+        public data class FitPrompt(
+            val promptTokenizer: PromptTokenizer? = null,
+            val granularity: Long = 2048,
+            val minimumContextLength: Long? = null,
```
Ololoshechkin, line 82: If `minimumContextLength` should be a multiple of `granularity` , maybe should we consider changing this to:
`contextChunkSize` and `chunksPerContext` (or similar).

(`granularity = contextChunkSize` , `minimumContextLength = contextChunkSize * chunksPerContext`)

WDYT?

ptitjes: Yeah, I like it. Avoiding a potential runtime error is always better.

I think I will go for `chunkSize` and `minimumChunkCount`. And I'll maybe also introduce an optional upper bound, with `maximumChunkCount`.

PR #336, opened 2025-06-25
[examples] Fix demo app to work with newer version of Koog

EugeneTheDev:
Fix demo android app to work properly with newer version of Koog (`0.2.1`)


Ololoshechkin (approved):
@EugeneTheDev let's add the compilation of examples in CI checks (maybe separately)

PR #344, opened 2025-06-26
Simple single run strategy working with multiple message

Rizzen:
Modified singleRunStrategy in order to support parallel tools calls.
Also introduced node that calls tools and sends results to back to LLM.
To test it I modified MockLLMExecutor a little bit to support mocking multiresponses.

#### Type of the change
- [ ] New feature
- [x] Bug fix
- [ ] Documentation fix

### DIFF
--- a/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
+++ b/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
@@ -7,22 +7,9 @@ import ai.koog.agents.core.agent.context.AIAgentLLMContext
 import ai.koog.agents.core.agent.entity.AIAgentStateManager
 import ai.koog.agents.core.agent.entity.AIAgentStorage
 import ai.koog.agents.core.agent.entity.AIAgentStrategy
-import ai.koog.agents.core.dsl.builder.forwardTo
-import ai.koog.agents.core.dsl.builder.strategy
-import ai.koog.agents.core.dsl.extension.nodeExecuteMultipleTools
-import ai.koog.agents.core.dsl.extension.nodeExecuteTool
-import ai.koog.agents.core.dsl.extension.nodeLLMRequest
-import ai.koog.agents.core.dsl.extension.nodeLLMRequestMultiple
-import ai.koog.agents.core.dsl.extension.nodeLLMSendMultipleToolResults
-import ai.koog.agents.core.dsl.extension.nodeLLMSendToolResult
-import ai.koog.agents.core.dsl.extension.onAssistantMessage
-import ai.koog.agents.core.dsl.extension.onMultipleAssistantMessages
-import ai.koog.agents.core.dsl.extension.onMultipleToolCalls
-import ai.koog.agents.core.dsl.extension.onToolCall
 import ai.koog.agents.core.environment.AIAgentEnvironment
 import ai.koog.agents.core.environment.AIAgentEnvironmentUtils.mapToToolResult
 import ai.koog.agents.core.environment.ReceivedToolResult
-import ai.koog.agents.core.environment.TerminationTool
 import ai.koog.agents.core.exception.AgentEngineException
 import ai.koog.agents.core.feature.AIAgentFeature
 import ai.koog.agents.core.feature.AIAgentPipeline
@@ -40,14 +27,12 @@ import ai.koog.prompt.llm.LLModel
 import ai.koog.prompt.message.Message
 import ai.koog.prompt.params.LLMParams
 import io.github.oshai.kotlinlogging.KotlinLogging
-import kotlinx.coroutines.CompletableDeferred
 import kotlinx.coroutines.async
 import kotlinx.coroutines.awaitAll
 import kotlinx.coroutines.supervisorScope
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
 import kotlinx.datetime.Clock
-import kotlinx.serialization.json.Json
 import kotlin.uuid.ExperimentalUuidApi
 import kotlin.uuid.Uuid
 
--- a/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
+++ b/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
@@ -368,8 +368,8 @@ public fun AIAgent(
     installFeatures = installFeatures
 )
 public fun singleRunStrategy(): AIAgentStrategy = strategy("single_run") {
-    val nodeCallLLMMultiple by nodeLLMRequestMultiple("sendInput")
-    val executeToolAndSendResult by nodeLLMExecuteMultipleToolsAndSendResults("executeToolAndSendResult")
+    val nodeCallLLMMultiple by nodeLLMRequestMultiple()
+    val executeToolAndSendResult by nodeLLMExecuteMultipleToolsAndSendResults()
 
     edge(nodeStart forwardTo nodeCallLLMMultiple)
     edge(nodeCallLLMMultiple forwardTo executeToolAndSendResult onMultipleToolCalls { true })
--- a/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
+++ b/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
@@ -7,6 +7,18 @@ import ai.koog.agents.core.agent.context.AIAgentLLMContext
 import ai.koog.agents.core.agent.entity.AIAgentStateManager
 import ai.koog.agents.core.agent.entity.AIAgentStorage
 import ai.koog.agents.core.agent.entity.AIAgentStrategy
+import ai.koog.agents.core.dsl.builder.forwardTo
+import ai.koog.agents.core.dsl.builder.strategy
+import ai.koog.agents.core.dsl.extension.nodeExecuteMultipleTools
+import ai.koog.agents.core.dsl.extension.nodeExecuteTool
+import ai.koog.agents.core.dsl.extension.nodeLLMRequest
+import ai.koog.agents.core.dsl.extension.nodeLLMRequestMultiple
+import ai.koog.agents.core.dsl.extension.nodeLLMSendMultipleToolResults
+import ai.koog.agents.core.dsl.extension.nodeLLMSendToolResult
+import ai.koog.agents.core.dsl.extension.onAssistantMessage
+import ai.koog.agents.core.dsl.extension.onMultipleAssistantMessages
+import ai.koog.agents.core.dsl.extension.onMultipleToolCalls
+import ai.koog.agents.core.dsl.extension.onToolCall
 import ai.koog.agents.core.environment.AIAgentEnvironment
 import ai.koog.agents.core.environment.AIAgentEnvironmentUtils.mapToToolResult
 import ai.koog.agents.core.environment.ReceivedToolResult
@@ -367,67 +379,3 @@ public fun AIAgent(
     toolRegistry = toolRegistry,
     installFeatures = installFeatures
 )
-
-/**
- * Creates an AI agent strategy based on the specified single-run mode.
- *
- * This method determines the type of strategy to be used for executing single-run tasks,
- * depending on the provided mode. It supports sequential, parallel, and single execution modes.
- *
- * @param runMode The mode in which the single-run strategy should operate. Defaults to SingleRunMode.SINGLE.
- *                - SingleRunMode.SINGLE: Executes without allowing multiple simultaneous tool calls.
- *                - SingleRunMode.SEQUENTIAL: Executes simultaneous tool calls sequentially.
- *                - SingleRunMode.PARALLEL: Executes multiple tool calls in parallel.
- * @return An instance of AIAgentStrategy configured according to the specified single-run mode.
- */
-public fun singleRunStrategy(runMode: ToolCalls = ToolCalls.SINGLE): AIAgentStrategy =
-    when (runMode) {
-        ToolCalls.SEQUENTIAL -> singleRunWithParallelAbility(false)
-        ToolCalls.PARALLEL   -> singleRunWithParallelAbility(true)
-        ToolCalls.SINGLE     -> singleRunModeStrategy()
-    }
-
-private fun singleRunWithParallelAbility(parallelTools: Boolean) = strategy("single_run_sequential") {
-    val nodeCallLLM by nodeLLMRequestMultiple()
-    val nodeExecuteTool by nodeExecuteMultipleTools(parallelTools = parallelTools)
-    val nodeSendToolResult by nodeLLMSendMultipleToolResults()
-
-    edge(nodeStart forwardTo nodeCallLLM)
-    edge(nodeCallLLM forwardTo nodeExecuteTool onMultipleToolCalls { true })
-    edge(nodeCallLLM forwardTo nodeFinish
-                                        onMultipleAssistantMessages { true }
-                                        transformed { it.joinToString("\n") { message -> message.content } })
-
-    edge(nodeExecuteTool forwardTo nodeSendToolResult)
-
-    edge(nodeSendToolResult forwardTo nodeFinish
-                                                onMultipleAssistantMessages { true }
-                                                transformed { it.joinToString("\n") { message -> message.content } })
-
-    edge(nodeSendToolResult forwardTo nodeExecuteTool onMultipleToolCalls { true })
-}
-
-private fun singleRunModeStrategy() = strategy("single_run") {
-    val nodeCallLLM by nodeLLMRequest()
-    val nodeExecuteTool by nodeExecuteTool()
-    val nodeSendToolResult by nodeLLMSendToolResult()
-
-    edge(nodeStart forwardTo nodeCallLLM)
-    edge(nodeCallLLM forwardTo nodeExecuteTool onToolCall { true })
-    edge(nodeCallLLM forwardTo nodeFinish onAssistantMessage { true })
-    edge(nodeExecuteTool forwardTo nodeSendToolResult)
-    edge(nodeSendToolResult forwardTo nodeFinish onAssistantMessage { true })
-    edge(nodeSendToolResult forwardTo nodeExecuteTool onToolCall { true })
-}
-
-/**
- * Enum representing the modes in which a single-run strategy for an AI agent can be executed.
- *
- * These modes define how tasks or operations are processed during the agent's run:
- * - SEQUENTIAL: Multiple tool calls allowed but will be executed sequentially.
- * - PARALLEL: Tool calls executed in parallel.
- * - SINGLE: Multiple tool calls are not allowed.
- */
-public enum class ToolCalls {
-    SEQUENTIAL, PARALLEL, SINGLE
-}
--- a/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
+++ b/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
@@ -367,3 +367,14 @@ public fun AIAgent(
     toolRegistry = toolRegistry,
     installFeatures = installFeatures
 )
+public fun singleRunStrategy(): AIAgentStrategy = strategy("single_run") {
+    val nodeCallLLMMultiple by nodeLLMRequestMultiple("sendInput")
+    val executeToolAndSendResult by nodeLLMExecuteMultipleToolsAndSendResults("executeToolAndSendResult")
+
+    edge(nodeStart forwardTo nodeCallLLMMultiple)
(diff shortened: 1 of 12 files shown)


agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
```diff
@@ -441,14 +441,13 @@ public open class AIAgent(
...
-    val nodeExecuteTool by nodeExecuteTool("nodeExecuteTool")
-    val nodeSendToolResult by nodeLLMSendToolResult("nodeSendToolResult")
-
-    edge(nodeStart forwardTo nodeCallLLM)
-    edge(nodeCallLLM forwardTo nodeExecuteTool onToolCall { true })
-    edge(nodeCallLLM forwardTo nodeFinish onAssistantMessage { true })
-    edge(nodeExecuteTool forwardTo nodeSendToolResult)
-    edge(nodeSendToolResult forwardTo nodeFinish onAssistantMessage { true })
-    edge(nodeSendToolResult forwardTo nodeExecuteTool onToolCall { true })
+    val nodeCallLLMMultiple by nodeLLMRequestMultiple("sendInput")
```
Ololoshechkin: Nit: Let's remove string names? 😃 
They will be deduced from variable names

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/extension/AIAgentNodes.kt
```diff
@@ -251,6 +251,50 @@ public fun AIAgentSubgraphBuilderBase<*, *>.nodeLLMSendToolResult(
...
+                    result(toolExecResult)
+                }
+            }
+
+            requestLLM()
+        }
+    }
+
+/**
+ *
```
harrytmthy: Nit: a couple of small doc-related things:
- L255: consider adding a `@param name` to the KDoc for consistency with the other node helpers.
- L274: looks like an empty KDoc block. Perhaps we want to put something here?

Minor cleanup, but helps keep things polished 🙂

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/extension/AIAgentNodes.kt
```diff
@@ -251,6 +251,50 @@ public fun AIAgentSubgraphBuilderBase<*, *>.nodeLLMSendToolResult(
...
+ */
```
Ololoshechkin: Add proper KDoc please

agents/agents-test/src/commonMain/kotlin/ai/koog/agents/testing/tools/MockLLMBuilder.kt
```diff
@@ -151,12 +151,51 @@ public class MockLLMBuilder(private val clock: Clock, private val tokenizer: Tok
...
+    }
+
+    /**
+     * Adds a partial pattern match for an LLM answer that triggers a tool call.
+     *
+     * @param pattern The exact input string to match
+     * @param tool The tool to be called when the input matches
+     * @param args The arguments to pass to the tool
+     */
+    public fun <Args : ToolArgs> addLLMAnswerPartialPattern(pattern: String, toolCalls: List<Pair<Tool<Args, *>, Args>>) {
```
Ololoshechkin, line 180: Is it tested?
Please add test for these methods.

Also -- please consider covering callable tools for jvm, and test it also.

Ololoshechkin (approved):

agents/agents-test/src/commonMain/kotlin/ai/koog/agents/testing/tools/MockLLMBuilder.kt
Rizzen: >Is it tested?

I don't see particular tests for same mockLLMBuilder methods

agents/agents-test/src/commonMain/kotlin/ai/koog/agents/testing/tools/MockLLMBuilder.kt
```diff
@@ -151,12 +151,51 @@ public class MockLLMBuilder(private val clock: Clock, private val tokenizer: Tok
...
-            )
+            ))
+        }
+    }
+
+    /**
+     * Adds a partial pattern match for an LLM answer that triggers a tool call.
+     *
+     * @param pattern The exact input string to match
+     * @param tool The tool to be called when the input matches
```
zarechneva-a: Could you update the doc here, please?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
```diff
@@ -441,14 +441,13 @@ public open class AIAgent(
...
-    val nodeSendToolResult by nodeLLMSendToolResult("nodeSendToolResult")
-
-    edge(nodeStart forwardTo nodeCallLLM)
-    edge(nodeCallLLM forwardTo nodeExecuteTool onToolCall { true })
-    edge(nodeCallLLM forwardTo nodeFinish onAssistantMessage { true })
-    edge(nodeExecuteTool forwardTo nodeSendToolResult)
-    edge(nodeSendToolResult forwardTo nodeFinish onAssistantMessage { true })
-    edge(nodeSendToolResult forwardTo nodeExecuteTool onToolCall { true })
+    val nodeCallLLMMultiple by nodeLLMRequestMultiple()
+    val executeToolAndSendResult by nodeLLMExecuteMultipleToolsAndSendResults()
```
tiginamaria: Name looks super long, maybe move `multiple` to the node parameter?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
```diff
@@ -441,14 +441,13 @@ public open class AIAgent(
...
+
+    edge(nodeStart forwardTo nodeCallLLMMultiple)
+    edge(nodeCallLLMMultiple forwardTo executeToolAndSendResult onMultipleToolCalls { true })
+    edge((nodeCallLLMMultiple forwardTo nodeFinish) transformed { it.first() } onAssistantMessage { true })
```
tiginamaria: Do we need exra () now or it's just for readability?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
tiginamaria: I see now, the output type is the problem. I personally against for having God nodes that do several things inside, so for me, it's just clearer to have two nodes here. Or we can define such a node only for here and do not add it to public API

examples/src/main/kotlin/ai/koog/agents/example/calculator/OllamaCalculatorExample.kt
```diff
@@ -40,7 +42,7 @@ fun main() = runBlocking {
 
     val agent = AIAgent(
         promptExecutor = executor,
-        strategy = CalculatorStrategy.strategy,
+        strategy = singleRunStrategy(),
```
tiginamaria: Do not forget to roll back it

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
```diff
@@ -441,14 +441,13 @@ public open class AIAgent(
...
-    val nodeExecuteTool by nodeExecuteTool("nodeExecuteTool")
-    val nodeSendToolResult by nodeLLMSendToolResult("nodeSendToolResult")
-
-    edge(nodeStart forwardTo nodeCallLLM)
-    edge(nodeCallLLM forwardTo nodeExecuteTool onToolCall { true })
-    edge(nodeCallLLM forwardTo nodeFinish onAssistantMessage { true })
-    edge(nodeExecuteTool forwardTo nodeSendToolResult)
-    edge(nodeSendToolResult forwardTo nodeFinish onAssistantMessage { true })
-    edge(nodeSendToolResult forwardTo nodeExecuteTool onToolCall { true })
+    val nodeCallLLMMultiple by nodeLLMRequestMultiple()
```
tiginamaria: It's not obvious change for users which currently using `singleRunStrategy` as the behavior of their agent can change dramatically with parallel tool calls. Maybe we should add separate strategy for multiple or provide parameter to decide it inside here?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
Rizzen: My idea was that nodeSendResultToLLM was redundant here, because here we're always want to send result to llm and get something back, ant it's pretty common case. Wouldn't the users want to have simpler graph in that case?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
Rizzen: We do need it, because `(nodeCallLLMMultiple forwardTo nodeFinish)` is the first argument for `transformed`

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
Rizzen: Well, that's actually good question, because initial problem was the fact that singleRunStrategy doesn't work with parallel tool calls which Ollama could produce and that was confusing for users.

Rizzen:
@tiginamaria raised really good question, because I agree - that change is kinda dangerous. Maybe we want to introduce another simpleStrategy?
@Ololoshechkin WDYT?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
tiginamaria: I totally agree! Let's leave it as it is now and wait until big nodes API refine PR from Andrey

ptitjes:
If I may suggest two things.

### Remove prompt modification from llm request node

We are assisting to a combinatorial explosion of the number of nodes, only because the basic LLM request node does add the incoming and outgoing messages to the prompt.

Rather than creating lots of new nodes, why not extract the prompt modifications from the LLM request node, and create two simple `addToPrompt` (which adds to the prompt and forwards the messages) and `dumpToPrompt` (which add to the prompts and forwards `Unit`)

Similar to what I did [there](https://github.com/ptitjes/koog/blob/stash/alternative-graph-dsl-edge-modifiers/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt#L466-L482).

Said `addToPrompt` and `dumpToPrompt` nodes are defined [here](https://github.com/ptitjes/koog/blob/stash/alternative-graph-dsl-edge-modifiers/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl2/extension/AIAgentNodes.kt#L63-L163) along with the new simpler `llmRequest` node variants.

### Pass `parallelTools` to `singleRunStategy`

We could pass a `parallelTools: Boolean` to the `singleRunStrategy` that would be passed to the `executeToolCalls` node.

Also, in this way, this drastically the number of necessary nodes. All the nodes handle list inputs and/or list outputs, and this is a lot less confusing for the user to not have the tens of nodes with slightly different names.

See the only needed plural [executeToolCalls](https://github.com/ptitjes/koog/blob/stash/alternative-graph-dsl-edge-modifiers/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl2/extension/AIAgentNodes.kt#L227-L237).

Notice that the whole file with [all the nodes](https://github.com/ptitjes/koog/blob/stash/alternative-graph-dsl-edge-modifiers/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl2/extension/AIAgentNodes.kt) is way shorter. And it allows to write stategies like ReAct without having to introduce custom nodes.

Using all this, there is no difference with the actual behavior of the current `singleRunStrategy` and it naturally handles multiple message (and is almost ready to naturally handle thinking messages).

agents/agents-test/src/commonMain/kotlin/ai/koog/agents/testing/tools/MockLLMBuilder.kt
Ololoshechkin: Then add them please :)

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
Ololoshechkin: Actually, is it really parallel?
If we don't call them in parallel but still make actual tool executions sequential -- it should be fine (users wouldn't notice unless they review logs carefully). But I agree that we should definitely provide an optional customisation flag to the strategy to make single tool calls

Rizzen: >Actually, is it really parallel?

It depends on agent's logic :) Basically, situation is - LLM returned a few tool calls messages in response to one request. Currently we're only taking the first one, skipping all the remaining. 

So we need to fix that behaviour for sure, because it's confusing. Whether these call being really parallel or sequential - we could make that customizable via singleRunStrategy(allowParallel: Boolean) as @ptitjes suggested

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/extension/AIAgentEdges.kt
```diff
@@ -199,6 +200,24 @@ public infix fun <IncomingOutput, IntermediateOutput, OutgoingInput>
...
 }
 
+
+/**
+ * Creates an edge that filters assistant messages based on a custom condition and extracts their content.
+ *
+ * @param block A function that evaluates whether to accept an assistant message
+ */
+public infix fun <IncomingOutput, IntermediateOutput, OutgoingInput>
+        AIAgentEdgeBuilderIntermediate<IncomingOutput, IntermediateOutput, OutgoingInput>.onMultipleAssistantMessages(
```
Ololoshechkin: Also `AIAgentEdgeBuilderIntermediate<IncomingOutput, List<Message.Response>, OutgoingInput>` ?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
```diff
@@ -430,25 +430,65 @@ public open class AIAgent(
...
+/**
+ * Enum representing the modes in which a single-run strategy for an AI agent can be executed.
+ *
+ * These modes define how tasks or operations are processed during the agent's run:
+ * - SEQUENTIAL: Multiple tool calls allowed but will be executed sequentially.
+ * - PARALLEL: Tool calls executed in parallel.
+ * - SINGLE: Multiple tool calls are not allowed.
+ */
+public enum class SingleRunMode {
+    SEQUENTIAL, PARALLEL, SINGLE
```
Ololoshechkin: Let's maybe be more verbose about tools semantics here?
`SingleRunMode` -> `ToolCalls`?

`SEQUENTIAL` -> `SEQUENTIAL_TOOLS_CALLS`
`PARALLEL ` -> `PARALLEL_TOOL_CALLS`
`SINGLE` -> `INDIVIDUAL_TOOL_CALLS`

or smth similar

WDYT ?

Ololoshechkin (approved):

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
```diff
@@ -204,7 +204,7 @@ public open class AIAgent(
         )
 
         strategy.execute(context = agentContext, input = agentInput)
-
+        print(agentContext.llm)
```
zarechneva-a: Was it for debugging or intentional? If intentional, may it be worth it adding a "[DEBUG_LOG]" line or something like that to elaborate the line content?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
```diff
@@ -430,25 +430,65 @@ public open class AIAgent(
...
- * 4. Send the tool result back to the LLM.
- * 5. Repeat until LLM indicates no further tool calls are needed or the agent finishes.
+ * Creates an AI agent strategy based on the specified single-run mode.
+ *
+ * This method determines the type of strategy to be used for executing single-run tasks,
+ * depending on the provided mode. It supports sequential, parallel, and single execution modes.
+ *
+ * @param runMode The mode in which the single-run strategy should operate. Defaults to SingleRunMode.SINGLE.
+ *                - SingleRunMode.SINGLE: Executes without allowing multiple simultaneous tool calls.
+ *                - SingleRunMode.SEQUENTIAL: Executes simultaneous tool calls sequentially.
```
zarechneva-a: nit: `multiple` instead of `simultaneous` to reduce potential confusion?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
```diff
@@ -430,25 +430,65 @@ public open class AIAgent(
...
-    val nodeExecuteTool by nodeExecuteTool("nodeExecuteTool")
-    val nodeSendToolResult by nodeLLMSendToolResult("nodeSendToolResult")
+public fun singleRunStrategy(runMode: SingleRunMode = SingleRunMode.SINGLE): AIAgentStrategy =
+    when (runMode) {
+        SingleRunMode.SEQUENTIAL -> singleRunWithParallelAbility(false)
+        SingleRunMode.PARALLEL   -> singleRunWithParallelAbility(true)
+        SingleRunMode.SINGLE     -> singleRunModeStrategy()
+    }
+
+private fun singleRunWithParallelAbility(parallelTools: Boolean) = strategy("single_run_sequential") {
```
zarechneva-a: Let's rename "single_run_sequential"? In case it's parallel, this can lead to confusion.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
Rizzen: I do like it, but in that case naming like `SEQUENTIAL_TOOLS_CALLS` is not necessary, as we'll use it as `ToolCalls.SEQUENTIAL`. There's no need to duplicate "tool calls" naming.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/extension/AIAgentEdges.kt
```diff
@@ -199,6 +200,24 @@ public infix fun <IncomingOutput, IntermediateOutput, OutgoingInput>
...
 }
 
+
+/**
+ * Creates an edge that filters assistant messages based on a custom condition and extracts their content.
+ *
+ * @param block A function that evaluates whether to accept an assistant message
+ */
+public infix fun <IncomingOutput, OutgoingInput>
+        AIAgentEdgeBuilderIntermediate<IncomingOutput, List<Message.Response>, OutgoingInput>.onMultipleAssistantMessages(
```
tiginamaria, line 210: Not sure any model can even do it, but just in case let's keep it

tiginamaria (approved):
Thank you for changes! Looks good to me now

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
Rizzen: Oh, good catch, thank you!

Rizzen: But looks like it's gone after rebase

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
Rizzen: For me "multiple" does not sounds implying simultaneous, and I wanted to highlight exactly the case with model returning few tool calls in one trip to LLM. I'm not sure in current wording though.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/extension/AIAgentEdges.kt
Rizzen: Well, basically it depends on node, `PromptExecutor.execute` always return the list and most of the times it's the list of one element, but it's still qualifies as multiple.
`requestLLM(): Message.Response` just makes `requestMultiple().first()`

aozherelyeva (approved):

PR #359, opened 2025-06-27
Cover SafeTool with tests

skarpovdev:
This PR adds tests for SafeTool.

---

#### Type of the change
- [ ] New feature
- [ ] Bug fix
- [ ] Documentation fix

### DIFF
--- a/agents/agents-core/src/jvmTest/kotlin/ai/koog/agents/core/environment/SafeToolTest.kt
+++ b/agents/agents-core/src/jvmTest/kotlin/ai/koog/agents/core/environment/SafeToolTest.kt
@@ -94,10 +94,6 @@ class SafeToolTest {
         override suspend fun reportProblem(exception: Throwable) {
             throw exception
         }
-
-        override suspend fun sendTermination(result: String?) {
-            // No-op for testing
-        }
     }
 
 
@@ -220,10 +216,6 @@ class SafeToolTest {
             override suspend fun reportProblem(exception: Throwable) {
                 throw exception
             }
-
-            override suspend fun sendTermination(result: String?) {
-                // No-op for testing
-            }
         }
 
         val safeTool = SafeToolFromCallable(::testFunction, directCallEnvironment, testClock)
@@ -319,10 +311,6 @@ class SafeToolTest {
 
             override suspend fun reportProblem(exception: Throwable) {
                 throw exception
-            }
-
-            override suspend fun sendTermination(result: String?) {
-                // No-op for testing
             }
         }
 


agents/agents-core/src/jvmTest/kotlin/ai/koog/agents/core/environment/SafeToolTest.kt
```diff
@@ -0,0 +1,130 @@
...
+
+class SafeToolTest {
+
+    private fun testFunction(param1: String, param2: Int): String {
+        return "Result: $param1 - $param2"
+    }
+
+    private class MockEnvironment(
+        private val shouldSucceed: Boolean = true,
+        private val resultContent: String = "Success content",
```
zarechneva-a, line 68: Let's put strings "Success content", "Test result", "Error: Test error" into variables/constants?

agents/agents-core/src/jvmTest/kotlin/ai/koog/agents/core/environment/SafeToolTest.kt
```diff
@@ -0,0 +1,130 @@
...
+        val result = safeTool.executeRaw("test", 123)
+
+        assertEquals("Raw result content", result)
+    }
+
+    @Test
+    fun testResultSuccessHelpers() = runTest {
+        val success = SafeToolFromCallable.Result.Success("Test result", "Success content")
+
+        assertEquals("Test result", success.asSuccessful().result)
```
zarechneva-a: Is it worth adding
```
assertTrue(success.isSuccessful())
assertFalse(success.isFailure())
```
here as well? (based on the checks in the test below)

agents/agents-core/src/jvmTest/kotlin/ai/koog/agents/core/environment/SafeToolTest.kt
```diff
@@ -0,0 +1,130 @@
...
+        assertEquals("Error message", failure.content)
+    }
+
+    @Test
+    fun testInvalidArgumentCount() = runTest {
+        val mockEnvironment = MockEnvironment(shouldSucceed = true)
+        val safeTool = SafeToolFromCallable(::testFunction, mockEnvironment, testClock)
+
+        assertThrows<IllegalStateException> {
+            safeTool.execute("test")
```
zarechneva-a: I'd add parametrisation here to check a couple of other cases as well:
- zero arguments
- more arguments than the tool expects (3 in this case)

agents/agents-core/src/jvmTest/kotlin/ai/koog/agents/core/environment/SafeToolTest.kt
```diff
@@ -0,0 +1,130 @@
...
+        }
+    }
+
+    @Test
+    fun testWithNullArgument() = runTest {
+        val mockEnvironment = MockEnvironment(shouldSucceed = true)
+        val safeTool = SafeToolFromCallable(::testFunction, mockEnvironment, testClock)
+
+        val result = safeTool.execute("test", null)
```
zarechneva-a, line 224: Is it OK that we don't throw any error in this case, although the arguments are not nullable in the `testFunction`? Is it because we are using a `SafeToolFromCallable` wrap?

agents/agents-core/src/jvmTest/kotlin/ai/koog/agents/core/environment/SafeToolTest.kt
```diff
@@ -0,0 +1,130 @@
...
+
+        assertTrue(result.isSuccessful())
+        assertEquals("Test result", result.asSuccessful().result)
+    }
+}
```
zarechneva-a, line 333: I'd suggest adding some more tests.

1. Check that `safeTool.toolFunction` corresponds the initial function;
2. Check that the `safeTool.toolFunction`'s parameters are correct;
3. Check cases when `testFunction` have nullable arguments, arguments with default values, complex arguments structures (e.g. nested lists, enums, etc. – the cases that can potentially break serialization);
4. (optional, can be created in scope of another PR) Create some more functions and check the fields that are provided by the `KFunction` interface: `isExternal`, `isInline`, `isOperator`, `isInfix`, `isSuspend`, etc. – maybe also by parametrisation?.

WDYT?

agents/agents-core/src/jvmTest/kotlin/ai/koog/agents/core/environment/SafeToolTest.kt
skarpovdev: I'll keep "Success content" because it's a default argument already, and extracting it into a constant and then assigning would decrease readability. Will do for the other two 👍🏻

agents/agents-core/src/jvmTest/kotlin/ai/koog/agents/core/environment/SafeToolTest.kt
skarpovdev: Yes, since here we’ve mocked the response and we bypass the actual function call.
I added another test which does call the function which throws exception, and then we check that it was successfully written to `content`.

agents/agents-core/src/jvmTest/kotlin/ai/koog/agents/core/environment/SafeToolTest.kt
skarpovdev: 1. Sure, I'll do it in each test right after `safeTool` assignments.
2. ✅
3. ✅
4. I think this one is not necessary. Those are the part of a standard library, and they are not directly used by the code we are testing.

agents/agents-core/src/jvmTest/kotlin/ai/koog/agents/core/environment/SafeToolTest.kt
skarpovdev: I added these checks, but decided to use a very simple kind of parametrization to not increase complexity and reduce readability.

aozherelyeva (approved):
LGTM!

PR #369, opened 2025-06-30
Support OpenTelemetry formatting for Koog agent events

sdubov:

Add support for Open Telemetry events formatting for Koof agent events.

Include the following changes:
- Move pipeline handler parameters for all agent events under a context data class;
- Add missing events to AIAgentPipeline: `onBeforeAgentClose`, `onStartLLMStreaming`, `onBeforeExecuteMultipleChoices`, `onAfterExecuteMultipleChoices`
- Extract common test code into agent-test module;
- Add OpenTelemetry feature that produces spans from a pipeline event;
- Rename sessionUuid to runId and update type from Uuid to String.

#### Type of the change
- [x] New feature
- [ ] Bug fix
- [ ] Documentation fix


gradle/libs.versions.toml
```diff
@@ -47,6 +49,14 @@ mcp = { module = "io.modelcontextprotocol:kotlin-sdk", version.ref = "mcp" }
 slf4j-simple = { module = "org.slf4j:slf4j-simple", version.ref = "slf4j" }
 jetsign-gradle-plugin = { module = "com.jetbrains:jet-sign", version.ref = "jetsign" }
 testcontainers = { module = "org.testcontainers:testcontainers", version.ref = "testcontainers" }
+opentelemetry-api = { module = "io.opentelemetry:opentelemetry-api", version.ref = "opentelemetry" }
```
EugeneTheDev: Let's use [OpenTelemetry BOM](https://mvnrepository.com/artifact/io.opentelemetry/opentelemetry-bom) here instead of specifying versions for all packages manually. Using BOM should offer more reliability when it comes to managing compatible dependencies

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
```diff
@@ -121,63 +121,77 @@ public open class AIAgent<Input, Output>(
...
+            }
 
-        return result
+            result
+        }
     }
 
     override suspend fun executeTools(toolCalls: List<Message.Tool.Call>): List<ReceivedToolResult> {
-        logger.info { formatLog("Executing tools: [${toolCalls.joinToString(", ") { it.tool }}]") }
+        val agentRunInfo = currentCoroutineContext().getAgentRunInfoElement() ?: throw IllegalStateException("Agent run info not found")
```
EugeneTheDev: Looks like we always expect some value here, how about we move this null check into the `getAgentRunInfoElement` and make it return non-nullable value instead?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/handler/AgentEventHandlerContext.kt
```diff
@@ -0,0 +1,96 @@
...
+
+/**
+ * Represents the context for creating and managing an AI agent within a specific strategy.
+ *
+ * @param FeatureT The type of the feature associated with the context.
+ * @property strategy The AI agent strategy that defines the workflow and execution logic for the AI agent.
+ * @property agent The AI agent being managed or operated upon in the context.
+ * @property feature An additional feature or configuration associated with the context.
+ */
+public class AgentTransformEnvironmentContext<FeatureT>(
```
EugeneTheDev: nit: consistent generic parameter naming - `TFeature`, like in the classes below

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/handler/AgentEventHandlerContext.kt
```diff
@@ -0,0 +1,96 @@
...
+    public val strategy: AIAgentStrategy<*, *>,
+    public val agent: AIAgent<*, *>,
+    public val feature: FeatureT
+) : AgentEventHandlerContext {
+    /**
+     * Executes a given block of code with the `AIAgentStrategy` instance of this context.
+     *
+     * @param block A suspending lambda function that receives the `AIAgentStrategy` instance.
+     */
+    public suspend fun readStrategy(block: suspend (AIAgentStrategy<*, *>) -> Unit) {
```
EugeneTheDev: It's not qutie clear to me, why this method is needed? Why is simply accessing `strategy` property and passing it wherever you need is not enough? Looks this method can be removed

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/handler/AgentEventHandlerContext.kt
```diff
@@ -0,0 +1,96 @@
...
+    public val runId: String,
+    public val strategy: AIAgentStrategy<*, *>,
+    public val feature: TFeature,
+) : AgentEventHandlerContext {
+    /**
+     * Reads the current AI agent strategy and executes the provided block of logic with it as a parameter.
+     *
+     * @param block A suspendable block of code that receives the current [AIAgentStrategy] as its parameter.
+     */
+    public suspend fun readStrategy(block: suspend (AIAgentStrategy<*, *>) -> Unit) {
```
EugeneTheDev: Same as above

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/handler/AgentEventHandlerContext.kt
```diff
@@ -0,0 +1,96 @@
...
+ * This interface extends the foundational event handling context, `EventHandlerContext`,
+ * and is specialized for scenarios involving agents and their associated workflows or features.
+ *
+ * The `AgentEventHandlerContext` enables implementation of event-driven systems within
+ * the AI Agent framework by offering hooks for custom event handling logic tailored to agent operations.
+ */
+public interface AgentEventHandlerContext : EventHandlerContext
+
+/**
+ * Represents the context for creating and managing an AI agent within a specific strategy.
```
EugeneTheDev: This KDoc and the class name do not make it clear enough, what's exactly the purpose of this class. So maybe KDoc or class name can be adjusted.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/handler/StrategyEventHandlerContext.kt
```diff
@@ -0,0 +1,62 @@
...
+    public val feature: FeatureT
+) : StrategyEventHandlerContext {
+    /**
+     * Provides read-only access to the current AI agent strategy within the execution context.
+     *
+     * @param block A suspending lambda function to process the current strategy. The strategy is
+     *              provided as an instance of [AIAgentStrategy] and allows reading its configuration
+     *              or properties without modifying the state.
+     */
+    public suspend fun readStrategy(block: suspend (AIAgentStrategy<*, *>) -> Unit) {
```
EugeneTheDev: Same, can probably be removed

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/handler/StrategyEventHandlerContext.kt
```diff
@@ -0,0 +1,62 @@
...
+    public val result: Any?
+) : StrategyEventHandlerContext {
+    /**
+     * Provides read-only access to the current AI agent strategy within the execution context.
+     *
+     * @param block A suspending lambda function to process the current strategy. The strategy is
+     *              provided as an instance of [AIAgentStrategy] and allows reading its configuration
+     *              or properties without modifying the state.
+     */
+    public suspend fun readStrategy(block: suspend (AIAgentStrategy<*, *>) -> Unit) {
```
EugeneTheDev: Same

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetry.kt
```diff
@@ -0,0 +1,292 @@
...
+
+        override fun createInitialConfig(): OpenTelemetryConfig {
+            return OpenTelemetryConfig()
+        }
+
+        override fun install(
+            config: OpenTelemetryConfig,
+            pipeline: AIAgentPipeline
+        ) {
+            val interceptContext = InterceptContext(this, OpenTelemetry())
```
EugeneTheDev, line 53: So the actual feature implementation is just empty? Maybe adding actual methods on the `OpenTelemetry` class itself and then moving the logic from interceptors to these methods and making interceptors invoke these methods would be a bit cleaner.

agents/agents-features/agents-features-opentelemetry/README.md
```diff
@@ -0,0 +1,61 @@
...
+The server will keep running until you press ENTER in the console.
+
+### Integrating with Agents
+
+To collect telemetry data from your agents, add the OpenTelemetry feature to your agent pipeline:
+
+```kotlin
+import ai.koog.agents.features.opentelemetry.feature.OpenTelemetry
+
+val agent = AIAgent.builder()
```
EugeneTheDev: This code snippet looks wrong, we don't have such API

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/messageFormat.kt
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * The string is structured as: `role: <role>, message: <content>`.
+ *
+ * This property is useful for logging or debugging purposes where a concise yet descriptive
+ * summary of the message content and its associated role is required.
+ */
+public val Message.Response.traceString: String
+    get() {
+        return "role: ${role}, message: $content"
+    }
```
sproshev: why are these properties better than toString?

EugeneTheDev: These are custom-purpose ones, I guess. We want to keep regular .toString more generic

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
Rizzen: We also need more clear error message here.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/context/AIAgentContextBase.kt
```diff
@@ -92,7 +89,7 @@ public interface AIAgentContextBase {
      * employed during the execution pipeline of an AI agent. It can be used
      * for logging, debugging, and switching between different strategies dynamically.
      */
-    public val strategyId: String
+    public val strategyName: String
```
Rizzen, line 88: >the unique identifier for the strategy being used in the current AI agent context

I believe that "unique identifier" should be called id, not name. Btw why we're changing that?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/PromptExecutorProxy.kt
```diff
@@ -31,28 +28,40 @@ public class PromptExecutorProxy(
...
 
         return responses
     }
 
     override suspend fun executeStreaming(prompt: Prompt, model: LLModel): Flow<String> {
-        return executor.executeStreaming(prompt, model)
+        logger.debug { "Executing LLM streaming call (prompt: $prompt)" }
+        val stream = executor.executeStreaming(prompt, model)
+
+        return stream
```
Rizzen, line 45: Inline variable?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/PromptExecutorProxy.kt
```diff
@@ -31,28 +28,40 @@ public class PromptExecutorProxy(
...
-        logger.debug { "Finished LLM call with response: $response" }
+        responses.forEachIndexed { index, response ->
+            messageBuilder.appendLine("- Response #${index}")
+            response.forEach { message ->
+                messageBuilder.appendLine("  -- [${message.role}] ${message.content}")
+            }
+        }
 
-        return response
+        logger.debug { "Finished LLM call with responses: $messageBuilder" }
```
Rizzen: As we're making that only for debug level logs, I suggest moving all the message builder logic to  `logger.debug` supplier in order to avoid excessive string memory traffic in cases where log level excludes debug.

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/span/SpanStorage.kt
```diff
@@ -0,0 +1,82 @@
...
+                id = NodeExecuteSpan.createId(agentId, runId, nodeName)
+                if (toolName != null) {
+                    id = ToolCallSpan.createId(agentId, runId, nodeName, toolName)
+                    if (promptId != null) {
+                        id = LLMCallSpan.createId(agentId, runId, nodeName, promptId)
+                    }
+                }
+            }
+        }
+
```
Rizzen: val id = when {
        runId == null -> AgentSpan.createId(agentId)
        nodeName == null -> AgentRunSpan.createId(agentId, runId)
        toolName == null -> NodeExecuteSpan.createId(agentId, runId, nodeName)
        promptId == null -> ToolCallSpan.createId(agentId, runId, nodeName, toolName)
        else -> LLMCallSpan.createId(agentId, runId, nodeName, promptId)
    }

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/messageFormat.kt
```diff
@@ -0,0 +1,37 @@
...
+import ai.koog.prompt.message.Message
+
+/**
+ * Constructs a string representation of the `Prompt` object, detailing its unique identifier,
+ * messages, and temperature parameter.
+ *
+ * This property is primarily intended for debugging or logging purposes, offering a concise overview of
+ * the `Prompt` object's state.
+ */
+public val Prompt.traceString: String
```
ptitjes: Maybe this can be made internal?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/messageFormat.kt
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Provides a formatted string representation of a `Response` message that includes its role and content.
+ *
+ * The string is structured as: `role: <role>, message: <content>`.
+ *
+ * This property is useful for logging or debugging purposes where a concise yet descriptive
+ * summary of the message content and its associated role is required.
+ */
+public val Message.Response.traceString: String
```
ptitjes: Maybe this can be made internal?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/model/featureEventTypes.kt
```diff
@@ -305,16 +336,29 @@ public data class ToolCallFailureEvent(
...
+/**
+ * A property that combines the provider ID and the model ID of an `LLModel` instance into a single string.
+ *
+ * It constructs a formatted identifier in the form of `providerId:modelId`, where:
+ * - `providerId` is the unique identifier of the `LLMProvider` associated with the model.
+ * - `modelId` is the unique identifier for the specific model instance.
+ *
+ * This property is typically used to uniquely identify an LLM instance for logging, tracing, or serialization purposes.
+ */
+public val LLModel.eventString: String
```
ptitjes: Maybe this can be made internal?

agents/agents-features/agents-features-opentelemetry/README.md
```diff
@@ -0,0 +1,264 @@
...
+
+### Basic configuration
+
+Here is an example of installing the OpenTelemetry feature with a basic set of configuration items:
+
+```kotlin
+install(OpenTelemetry) {
+    serviceName = "my-agent-service"  // Name of your service
+    serviceVersion = "1.0.0"          // Version of your service
+    addOtlpExporter()                 // Add OTLP exporter with default settings
```
ptitjes: Maybe simply call it `addExporter()`, as we are already in the `OpenTelemetry` feature DSL? Or `addDefaultExporter`?

gradle/libs.versions.toml
sdubov: @EugeneTheDev , BOM pull too many dependencies. I've updated telemetry dependencies in our module and deleted unused ones. Currently, there are only three:
- `io.opentelemetry:opentelemetry-sdk`
- `io.opentelemetry:opentelemetry-exporter-otlp`
- `io.opentelemetry:opentelemetry-exporter-logging` 

that uses the same otel version. Please see latest updates. I found it might be better for a users to add necessary dependencies when needed. Please let me know it you have other comments on this.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
sdubov: Good point. Moved checked inside a method and renamed it as well. Also, fixed the error message as per @Rizzen 's request.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/handler/AgentEventHandlerContext.kt
sdubov: Fixed all places.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/handler/AgentEventHandlerContext.kt
sdubov: Dropped the method. Indeed, it is not needed.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/handler/AgentEventHandlerContext.kt
sdubov: Done.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/handler/AgentEventHandlerContext.kt
sdubov: Updated KDoc for several mentioned methods.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/handler/StrategyEventHandlerContext.kt
sdubov: Fixed.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/handler/StrategyEventHandlerContext.kt
sdubov: Fixed.

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetry.kt
sdubov: Please see the new structure. The main preparation logic is located in the Config. We need to do it here because SDK and tracer configuration can be adjusted. The rest of logic inside interceptors are interceptors-related - create a particular span on a particular interceptor, add attributes, etc. Please let me know if I miss anything from your point.

agents/agents-features/agents-features-opentelemetry/README.md
sdubov: Updated README.md file according to the initial draft from @markomitos .

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/messageFormat.kt
sdubov: Indeed, extensions can add some feature-specific implementation, e.g. I would like to have a tracing-specific log message. Overriding `toString` here sounds too aggressive from my perspective.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/context/AIAgentContextBase.kt
sdubov: Is not intended to be unique :) If you look on assignment, we use a context parameter named `strategyId` and set a `strategy.name` to it. Since we have no strategy ids, I found it would be correct to rename the parameter.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/PromptExecutorProxy.kt
sdubov: I would like to keep it as is if possible. I had here the pipeline interceptor call for handing the stream start event. Let's consider this a placeholder if you don't mind :)

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/PromptExecutorProxy.kt
sdubov: Good catch. Fixed.

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/span/SpanStorage.kt
sdubov: I've dropped this method since I do not use it anywere. Thank you for the nice implementation. I like it better.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/messageFormat.kt
sdubov: It is definitely not a public API. However, I use it in multiple modules. For now, I've just copy duplicate extensions when needed. I would consider the better modules structure later. @ptitjes, thank you for catching this!

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/model/featureEventTypes.kt
sdubov: Done.

agents/agents-features/agents-features-opentelemetry/README.md
sdubov: Updated README.md file according to the initial draft.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/messageFormat.kt
sdubov: Done.

gradle/libs.versions.toml
EugeneTheDev: BOMs itself do not pull any dependencies. They only **define** how dependencies should be resolved once added explicitly to your build: which versions to pull, how to resolve some conflicts between dependencies, if any, etc. Take a look at Kotlin BOM e.g. So BOMs provide convenient and more reliable way to ensure that you would get dependencies that would be compatible with each other and there will be no conflicts, to put it simply. The way it's done is you add BOM as a regular dependency, specify version for this BOM, and then add actual dependencies, wihtout specifying their versions - these would be handled by BOM. In our case it's not critical, and it won't even change anything significantly, probably. But I just prefer to use BOMs when I deal with complex frameworks with multple dependencies, since, again, it gives you a bit more stability and convenience. If you are against BOM in this particular case, this is fine, these were just my 5 cents. We can proceed without it.

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/attribute/Attribute.kt
```diff
@@ -0,0 +1,11 @@
+package ai.koog.agents.features.opentelemetry.attribute
+
+internal interface Attribute {
+    val key: String
+    val value: Any
+
+    val verbose: Boolean
+        get() = false
+
+    fun String.concatKey(other: String) = this.plus(".${other}")
```
EugeneTheDev, line 10: nit: Putting extension functions inside some namespaces (interfaces/objects/classes) is usually not very convenient and idiomatic way (while we don't have proper context parameters) - it's harder to use them. Just a suggestion (feel free to ignore it if you don't agree with it) - let's make it top level in this file?

EugeneTheDev (approved):

ptitjes (commented):
Thanks for applying my small suggestions. LGTM.

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/attribute/CommonAttributes.kt
```diff
@@ -0,0 +1,42 @@
...
+            override val key: String = super.key.concatKey("address")
+            override val value: String = address
+        }
+
+        // server.port
+        data class Port(private val port: Int) : Server {
+            override val key: String = super.key.concatKey("port")
+            override val value: Int = port
+        }
+    }
```
Rizzen, line 41: I believe that these comments over the classes are excessive - we either use KDoc (which is IDE-friendly) or we leave them undocumented if we don't want to provide any useful information

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/attribute/SpanAttributes.kt
```diff
@@ -0,0 +1,277 @@
...
+
+        // gen_ai.agent.name
+        data class Name(private val name: String) : Agent {
+            override val key: String = super.key.concatKey("name")
+            override val value: String = name
+        }
+    }
+
+    // gen_ai.conversation
+    sealed interface Conversation : GenAIAttribute {
```
Rizzen, line 89: Same as above

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/span/GenAIAgentSpan.kt
```diff
@@ -0,0 +1,49 @@
...
+    abstract val attributes: List<Attribute>
+
+
+    fun setContext(context: Context) {
+        _context = context
+    }
+
+    fun setSpan (span: Span) {
+        _span = span
+    }
```
Rizzen: These could be the setters of `context` and `span` properties

sproshev (approved):

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/span/GenAIAgentSpan.kt
sdubov: Fixed.

PR #383, opened 2025-07-02
Skarpov/agents module test coverage

skarpovdev:
1. I’ve fixed a small bug in `describeToolCall` and `describeToolResult` where we only included the id when `[message.id](http://message.id/)` was null.
2. There is another problem - `describeToolCall` fails with `JsonDecodingException` in case content is null or contains an invalid JSON. I added `parseJsonContentSafely` which should handle both cases. But should we just use `JsonPrimitive(message.content)` like in `describeToolResult`?
3. minor: should we change `id: String = "code-engine-agents”` to `id: String = "koog-agents”`?
4. Modified `withSystemPrompt` to require positive `maxAgentIterations`.
---

#### Type of the change
- [ ] New feature
- [ ] Bug fix
- [ ] Documentation fix


agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/config/AIAgentConfig.kt
```diff
@@ -48,13 +48,16 @@ public class AIAgentConfig(
             llm: LLModel = OpenAIModels.Chat.GPT4o,
             id: String = "code-engine-agents",
             maxAgentIterations: Int = 3,
-        ): AIAgentConfigBase =
-            AIAgentConfig(
+        ): AIAgentConfigBase {
+            require(maxAgentIterations > 0) { "maxAgentIterations must be greater than 0" }
```
EugeneTheDev: It is still possible to create config directly without this check. Let's put it to `init` instead

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/config/ToolCallDescriber.kt
```diff
@@ -59,9 +57,9 @@ public interface ToolCallDescriber {
             return Assistant(
                 content = Json.encodeToString(
                     buildJsonObject {
-                        message.id ?: put("tool_call_id", JsonPrimitive(message.id))
+                        message.id?.let { put("tool_call_id", JsonPrimitive(it)) }
                         put("tool_name", JsonPrimitive(message.tool))
-                        put("tool_args", message.contentJson)
+                        put("tool_args", parseJsonContentSafely(message.content))
```
EugeneTheDev: I have my doubts regarding this method. Why do we need it? Why is it a problem? There should be no invalid tool call content if it was generated by LLM, and in case it was constructed manually or something went wrong I guess we want to fail instead of just silently putting malformed content as string here.

EugeneTheDev (changes requested):

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/config/ToolCallDescriber.kt
skarpovdev: I agree that it's unlikely that an invalid content may be passed, but I thought that why not add this extra safety in case it happens. But in case the content is completely malformed I agree that it makes sense to just fail. I'll revert the change. Thank you!

skarpovdev: @EugeneTheDev what do you think about https://github.com/JetBrains/koog/blob/b9fd3c711e12ca8a9a0b06a9b0dfed451fdb7478/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/config/ToolCallDescriber.kt#L78-L84?
In case of a malformed content, is it ok that we still return it or should we use the same logic as for tool call and fail?

EugeneTheDev: Tool call result is not necessary JSON, so we don't know what is malformed and what is not. So it is fine like this, yes

agents/agents-core/src/commonTest/kotlin/ai/koog/agents/core/agent/config/AIAgentConfigTest.kt
```diff
@@ -0,0 +1,115 @@
...
+
+class AIAgentConfigTest {
+    private companion object {
+        private val testModel = OpenAIModels.Chat.GPT4o
+        private const val MAX_ITERATIONS = 5
+    }
+
+    @Test
+    fun testConstructorWithAllParameters() {
+        val testPrompt = prompt("test-id") {
```
zarechneva-a: Let's put it as a private val in a companion object, too?

agents/agents-core/src/commonTest/kotlin/ai/koog/agents/core/agent/config/AIAgentConfigTest.kt
```diff
@@ -0,0 +1,115 @@
...
+
+        assertEquals(testPrompt, config.prompt)
+        assertEquals(testModel, config.model)
+        assertEquals(MAX_ITERATIONS, config.maxAgentIterations)
+        assertTrue(config.missingToolsConversionStrategy is MissingToolsConversionStrategy.Missing)
+    }
+
+    @Test
+    fun testWithSystemPromptAllParameters() {
+        val testPromptContent = "Test system prompt"
```
zarechneva-a, line 53: Let's put this one as a val as well, please

aozherelyeva:
> minor: should we change id: String = "code-engine-agents” to id: String = "koog-agents”?

Sounds valid! Would you please do it?

agents/agents-core/src/commonTest/kotlin/ai/koog/agents/core/agent/config/AIAgentConfigTest.kt
```diff
@@ -0,0 +1,115 @@
...
+
+    @Test
+    fun testWithSystemPromptDefaultParameters() {
+        val testPromptContent = "Test system prompt"
+
+        val config = AIAgentConfig.withSystemPrompt(
+            prompt = testPromptContent
+        )
+
+        assertEquals(OpenAIModels.Chat.GPT4o, config.model)
```
zarechneva-a: `testModel`?

agents/agents-core/src/commonTest/kotlin/ai/koog/agents/core/agent/config/AIAgentConfigTest.kt
```diff
@@ -0,0 +1,115 @@
...
+        assertEquals(3, config.maxAgentIterations)
```
zarechneva-a, line 77: `MAX_ITERATIONS`?

agents/agents-core/src/commonTest/kotlin/ai/koog/agents/core/agent/config/AIAgentConfigTest.kt
```diff
@@ -0,0 +1,115 @@
...
+            prompt = testPrompt,
+            model = testModel,
+            maxAgentIterations = MAX_ITERATIONS,
+            missingToolsConversionStrategy = testStrategy,
+        )
+
+        assertEquals(testPrompt, config.prompt)
+        assertEquals(testModel, config.model)
+        assertEquals(MAX_ITERATIONS, config.maxAgentIterations)
+        assertEquals(testStrategy, config.missingToolsConversionStrategy)
```
zarechneva-a, line 35: Let's put `config.missingToolsConversionStrategy` into variable as well? Like `defaultStrategy`

agents/agents-core/src/commonTest/kotlin/ai/koog/agents/core/agent/config/MissingToolsConversionStrategyTest.kt
```diff
@@ -0,0 +1,257 @@
...
+import kotlin.test.assertTrue
+
+class MissingToolsConversionStrategyTest {
+    private companion object {
+        private val testClock = object : Clock {
+            override fun now() = fromEpochMilliseconds(123)
+        }
+
+        private val testToolDescriptor = ToolDescriptor(
+            name = "test-tool",
```
zarechneva-a, line 24: I'd rather put these strings (`test-call-id`, `test-tool`, `Test tool description`, `"""{"param": "value"}"""` etc.) and the other ones used in tests and assertions to variables, but this is more a matter of taste.

agents/agents-core/src/commonTest/kotlin/ai/koog/agents/core/agent/config/MissingToolsConversionStrategyTest.kt
```diff
@@ -0,0 +1,257 @@
...
+
+        private val regularMessage = Message.User(
+            content = "Regular message content",
+            metaInfo = RequestMetaInfo.create(testClock),
+        )
+    }
+
+    @Test
+    fun testConvertMessageWithToolCall() {
+        val strategy = MissingToolsConversionStrategy.All(ToolCallDescriber.JSON)
```
zarechneva-a: Let's put the strategy into companion object?

aozherelyeva (approved):

agents/agents-core/src/commonTest/kotlin/ai/koog/agents/core/agent/config/AIAgentConfigTest.kt
skarpovdev: `MAX_ITERATIONS` is test-specific while in this test we check the default values of `withSystemPrompt` and its `maxAgentIterations` is 3

agents/agents-core/src/commonTest/kotlin/ai/koog/agents/core/agent/config/AIAgentConfigTest.kt
skarpovdev: I gave it a second thought and it seems `config.missingToolsConversionStrategy is MissingToolsConversionStrategy.Missing` checks don't make much sense. should something change, the tests will fail in other places

agents/agents-core/src/commonTest/kotlin/ai/koog/agents/core/agent/config/MissingToolsConversionStrategyTest.kt
skarpovdev: I tried that and I don't like that the companion object takes as much space as the tests 😃
I believe hardcoded values are fine if it's not in the test method themselves.

agents/agents-core/src/commonTest/kotlin/ai/koog/agents/core/agent/config/MissingToolsConversionStrategyTest.kt
skarpovdev: Huh, I thought I did that, thanks!

agents/agents-core/src/commonTest/kotlin/ai/koog/agents/core/agent/config/MissingToolsConversionStrategyTest.kt
zarechneva-a: I see! Frankly speaking, as this is more a unit test, I don't think we'd be tweaking it too much unless the functionality completely changes, so no problem.

PR #387, opened 2025-07-03
Add standard TimeTool

kpavlov:
# Add standard TimeTool

- Implement TimeTool + unit test
- Update QuickStart guide and [`Tool` KDoc](https://github.com/JetBrains/koog/pull/387/files#diff-88141b61c9a662860a93872a719104fa58a5338b666c7530c58b60c60f5a3578R59)

#### Type of the change
- [x] New feature
- [ ] Bug fix
- [x] Documentation fix


agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/tool/TimeTool.kt
```diff
@@ -0,0 +1,93 @@
...
+ * specified via argument, e.g. `"UTC"`, `"Europe/Paris"`, `"+01:00"`.
+ *
+ * It provides an option to truncate the time precision to seconds.
+ *
+ * @param clock The clock used to retrieve the current time. Defaults to the system clock.
+ * @param timezone The default timezone used when no timezone is specified in the arguments.
+ * Defaults to the system's default timezone.
+ * @param truncateToSeconds A flag to determine whether the time should be truncated to seconds. Defaults to `true`.
+ */
+public class TimeTool(
```
kpavlov, line 29: Should this class somehow make use of the [TimeProvider](https://github.com/JetBrains/koog/blob/develop/agents/agents-features/agents-features-memory/src/commonMain/kotlin/ai/koog/agents/memory/model/TimeProvider.kt) 🤔

gradle/libs.versions.toml
```diff
@@ -50,6 +51,7 @@ mcp = { module = "io.modelcontextprotocol:kotlin-sdk", version.ref = "mcp" }
 slf4j-simple = { module = "org.slf4j:slf4j-simple", version.ref = "slf4j" }
 jetsign-gradle-plugin = { module = "com.jetbrains:jet-sign", version.ref = "jetsign" }
 testcontainers = { module = "org.testcontainers:testcontainers", version.ref = "testcontainers" }
+kotest-assertions-core = { module = "io.kotest:kotest-assertions-core", version.ref = "kotest" }
```
sdubov: Would it be possible to use assertions from the default `kotlin.test` lib? I would keep consistency in the test framework and keep it between modules. Currently, we mostly use `kotlin.test` assertions. Please correct me if anybody has different vision here. cc @aozherelyeva.

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/tool/TimeTool.kt
```diff
@@ -0,0 +1,93 @@
...
+
+        val localDateTime = now.toLocalDateTime(timezone)
+        logger.debug { "Returning current time: \"$localDateTime\" at timeZone: \"$timezone\"" }
+        return "Current time: $localDateTime (Timezone: ${timezone.id})"
+    }
+
+    override val argsSerializer: KSerializer<Args> = Args.serializer()
+
+    override val descriptor: ToolDescriptor = ToolDescriptor(
+        name = "__time__",
```
sdubov: Could you please clarify why do you use a special formatting for a tool name with "__" as a prefix and postfix?

agents/agents-ext/src/commonTest/kotlin/ai/koog/agents/ext/tool/TimeToolTest.kt
```diff
@@ -0,0 +1,103 @@
...
+        truncateToSeconds: Boolean
+    ) {
+
+        val expectedTime = if (truncateToSeconds) {
+            Instant.fromEpochSeconds(currentTime.epochSeconds)
+        } else {
+            currentTime
+        }.toLocalDateTime(timezone)
+
+        println(toolResponse)
```
sdubov: Maybe it worth adding it to the assertion custom message, or relay on the standard test framework error message? If there is some special case of this `println`, I would suggest to define a logger and write this message to a logger instead of stdout:

agents/agents-ext/src/commonTest/kotlin/ai/koog/agents/ext/tool/TimeToolTest.kt
```diff
@@ -0,0 +1,103 @@
...
+
+        val result = subject.doExecute(TimeTool.Args())
+
+        verifyTimeAtTimezone(result, defaultTimezone, false)
+    }
+
+
+    @Test
+    fun testTimeToolWithSpecificTimezone() = runTest {
+        for (timezone in listOf("UTC", "Europe/Paris", "+01:00")) {
```
sdubov: It seems for me we should split this to different tests to verify and get separate errors about each failure. Wdyt?

sdubov (changes requested):

gradle/libs.versions.toml
zarechneva-a: I agree :) We do use JUnit assertions in the integration tests, I believe `kotlin.test` and JUnit is a good and sufficient combo.

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/tool/TimeTool.kt
kpavlov: Other tools like [AskUser](https://github.com/kpavlov/koog/blob/develop/agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/tool/AskUser.kt#L23) in this package are having "__" as a prefix and suffix. Added it here for consistency.

agents/agents-ext/src/commonTest/kotlin/ai/koog/agents/ext/tool/TimeToolTest.kt
kpavlov: Agree

agents/agents-ext/src/commonTest/kotlin/ai/koog/agents/ext/tool/TimeToolTest.kt
kpavlov: I agree. I'd also prefer to convert it to `@ParametrizedTest`. @sdubov, do you know something suitable for multiplatform tests? I can't use JUnit's `@ParametrizedTest` in common module. Otherwise, it might not make much sense splitting it into different test methods: good error message should be enough to diagnose.

gradle/libs.versions.toml
kpavlov: Thanks for the feedback, makes sense to aim for consistency.

That said, I proposed [Kotest assertions](https://kotest.io/docs/assertions/assertions.html) because they offer a more fluent, expressive API compared to kotlin.test and JUnit. They improve readability, reduce boilerplate, and help avoid common pitfalls like parameter ambiguity in Junit/KotlinTest `assertEquals(expected, actual)`—which is especially error-prone and hard to catch.

In Java, AssertJ is often used solve this. Kotest brings similar clarity and safety to Kotlin, together with infix functions, which let you write assertions almost in pure English.

Happy to align with the team direction, but I believe Kotest adds real value, especially for larger test suites.
Would you be open to review Kotest APIs and consider it for wider adoption in the project?

zarechneva-a: Sure! I think we can at least compare it with the existing approach and estimate whether it'd be challenging to switch to them. I didn't have time to check it out yet, but maybe you know if it supports the backwards compatibility with the KotlinTest assertions?

kpavlov: @aozherelyeva, it's just an assertion library, and it can be used together with KotlinTest assertions. There is no need for rewriting code or migrations

kpavlov: @aozherelyeva , let me [break one test](https://github.com/JetBrains/koog/pull/387/commits/8245a4c3fb8c2c9d9e25ba6a2faa266379184c7b) so [you can check](https://github.com/JetBrains/koog/actions/runs/16199910435/job/45736004950?pr=387#step:6:1087). I also added [soft asserts](https://kotest.io/docs/assertions/soft-assertions.html) for demonstration:

```
TimeToolTest[jvm] > testCreateTimeToolWithoutParams()[jvm] FAILED
    io.kotest.assertions.MultiAssertionError: The following 2 assertions for "Current time: 2025-07-10T15:56:20 (Timezone: Etc/UTC)" failed:
    1) 💔😉Response should contain current time
    "Current time: 2025-07-10T15:56:20 (Timezone: Etc/UTC)" should include substring "Current ⏰ time: "
       at ai.koog.agents.ext.tool.TimeToolTest$testCreateTimeToolWithoutParams$1.invokeSuspend(TimeToolTest.kt:56)
    2) 💔😉Response should contain system timezone
    "Current time: 2025-07-10T15:56:20 (Timezone: Etc/UTC)" should include substring " (🌍Timezone: Etc/UTC)"
       at ai.koog.agents.ext.tool.TimeToolTest$testCreateTimeToolWithoutParams$1.invokeSuspend(TimeToolTest.kt:59)
```

agents/agents-ext/build.gradle.kts
```diff
@@ -26,6 +26,7 @@ kotlin {
                 implementation(kotlin("test"))
                 implementation(libs.kotlinx.coroutines.test)
                 implementation(project(":agents:agents-test"))
+                implementation(libs.kotest.assertions.core)
```
kpavlov: It makes sense to add this to `commonTest` dependencies in all modules

gradle/libs.versions.toml
```diff
@@ -92,4 +94,4 @@ kotlin-serialization = { id = "org.jetbrains.kotlin.plugin.serialization", versi
 shadow = { id = "com.github.johnrengelman.shadow", version.ref = "shadow" }
 kotlinx-kover = { id = "org.jetbrains.kotlinx.kover", version.ref = "kover" }
 kotlin-spring = { id = "org.jetbrains.kotlin.plugin.spring", version.ref = "kotlin" }
-spring-management = { id = "io.spring.dependency-management", version.ref = "spring-management" }
\ No newline at end of file
+spring-management = { id = "io.spring.dependency-management", version.ref = "spring-management" }
```
kpavlov, line 101: Should this be deleted?
```suggestion
```

gradle/libs.versions.toml
zarechneva-a: Well, this output indeed looks pretty good. 

@sdubov we had a discussion with @kpavlov, and he suggested trying to use `kotest-assertions` in other modules as well – as the output indeed looks more readable with them. Let's give it a try? If it works for us, I'll start updating the code accordingly. And if it doesn't, updating assertions here would be pretty easy. WDYT?

sdubov: @kpavlov, thank you for your detailed examples. I got your comments about readability. 

I have several notes from my side related to this topic:

1. Any new dependency that we bring to the SDK should solves some particular problem. As far as I know, we did not notice any significant issues with current test approach. @aozherelyeva , @Ololoshechkin , please correct me if you have any issues with `kotlin.test` assertions on your side;
2. Adding new dependency introduce extra load on release process: you need to care about versioning, need to perform security checks, etc.;
3. From my perspective, it is better to have consistency across all tests (using one assertion framework) rather than having a single test class with inconsistent, but more readable assertions. It might be quite noticeable when you start using different libraries and on each new test class have no clear understanding what framework should I use to perform assert logic;
4. If we decide to go with a new assertion library, I would prefer doing it as a separate PR, with (not necessarily, but I would vote for) updating existing test code base for consistency.

zarechneva-a: > please correct me if you have any issues with kotlin.test assertions on your side

Absolutely no problems :) The stack traces are less polished but still are readable and comprehensible.

zarechneva-a: > If we decide to go with a new assertion library, I would prefer doing it as a separate PR, with (not necessarily, but I would vote for) updating existing test code for consistency.

I indeed support this approach!

agents/agents-ext/src/commonTest/kotlin/ai/koog/agents/ext/tool/TimeToolTest.kt
sdubov: @kpavlov , `verifyTimeAtTimezone` does not have any error handling logic (try/catch). The test execution will be terminated by the first error you get. I think the easiest way that I would go is to write three tests instead of one with clear names.

Probably, you can update the helper function logic a bit in the following way to make test easier:
```
@Test
fun test timezone - UTCT`() = verifyTimeAtTimezone("UTC")

@Test
fun test timezone - EuropeParis`() = verifyTimeAtTimezone("Europe/Paris")

@Test
fun `test timezone - +01:00`() = verifyTimeAtTimezone("+01:00")

private fun verifyTimeAtTimezone(timeZone: String) {
    // Expected
    val expectedTimeZone = TimeZone.of(timeZone)
    val expectedTime = currentTime.toLocalDateTime(expectedTimeZone)
    val expectedToolResponse = "Current time: $expectedTime (Timezone: $expectedTimeZone)"

    // Actual
    val actualToolResponse = subject.doExecute(TimeTool.Args(timeZoneToolArg))

    actualToolResponse shouldBe expectedToolResponse
}
```

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/tool/TimeTool.kt
sdubov: @kpavlov , this is just an old naming convention. Please note there is another tool in the same package with name `say_to_user`. We do not need to follow it in Kotlin code. For tool names it would be better to use "snake case" naming without prefixes.

sdubov (changes requested):

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/tool/TimeTool.kt
kpavlov: Changed name, updated KDoc in Tool interface

Ololoshechkin (approved):

PR #393, opened 2025-07-04
Tweak prompt for test & remove windows heavy tests

aozherelyeva:
1. Tweak the prompt to handle long responses from OpenAI `audio` LLMs.
2. Increase test timeouts for audio models (from time to time they took longer than 2 minutes).
3. Remove windows runs from heavy tests (agreed with @skarpovdev that this makes sense for integration checks).

---

#### Type of the change
- [ ] New feature
- [ ] Bug fix
- [ ] Documentation fix
- [x] Tests fix

### DIFF
--- a/.github/workflows/heavy-tests.yml
+++ b/.github/workflows/heavy-tests.yml
@@ -22,7 +22,7 @@ jobs:
 
     strategy:
       matrix:
-        os: [ ubuntu-latest, windows-latest ]
+        os: [ ubuntu-latest ]
 
     steps:
       - name: Configure Git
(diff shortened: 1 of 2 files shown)


.github/workflows/heavy-tests.yml
```diff
@@ -22,7 +22,7 @@ jobs:
 
     strategy:
       matrix:
-        os: [ ubuntu-latest, windows-latest ]
```
sproshev, line 25: why? I read the description, but it does not name the actual reasons

zarechneva-a: Sure!

For me, there are three main reasons:

1. The 2-environment setup is inherited from the code-engine unit tests. The integration tests are running with agents and executors, which use the components that are already checked by the unit tests. The main goal of them is to check the behaviour of the agent/executor with a real LLM profile, which is (in my understanding) similar for both platforms.

2. Windows tests are slow, pricey to support (in terms of time of debugging and fixing them) and didn't yet find any OS-specific problems (only the same ones that reproduce on Ubuntu as well).

3. The most important: the integration tests check more complex, high-level flows that are usually not platform-specific. For sure, the usual unit tests (jvmTest) remain running on both ubuntu and windows.

sproshev: thank you!

PR #395, opened 2025-07-04
Support content moderation requests and moderating models (for Ollama, OpenAI, and Bedrock)

Ololoshechkin:
Fixes: https://github.com/JetBrains/koog/issues/396

---

#### Type of the change
- [x] New feature
- [ ] Bug fix
- [ ] Documentation fix


agents/agents-core/src/commonTest/kotlin/ai/koog/agents/core/dsl/extension/AIAgentNodesHistoryCompressionTest.kt
```diff
@@ -68,6 +69,31 @@ class AIAgentNodesHistoryCompressionTest {
...
+
+                if (latestTokenUsage() > 100500) {
+                    compressHistory()
+                }
+
+                callTools(tools)
+            }
+
+            "Should not be reached"
+        }
```
ptitjes: That's really interesting. I was thinking recently that one drawback of the graph DSL was that you have to know in advance the types of the nodes you need. And when write/edit a strategy, you find yourself moving back and forth between your nodes and your edges to adjust types.

This is mainly happening when you use passthrough nodes to try to avoid executing twice the same conditional logic in your edges. This is also happening when you want to do type-safe parallelism.

Would the Koog project be interested in having a second DSL to allow "code" strategies? We could have a `codeStrategy() { }` factory method to handle this case.

I'd be interested in experimenting on this.

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/MultipleLLMPromptExecutorIntegrationTest.kt
```diff
@@ -1042,4 +1043,60 @@ class MultipleLLMPromptExecutorIntegrationTest {
             )
         }
     }
+
+    @Test
+    fun integration_testOpenAISingleMessageModeration() = runTest {
+        val prompt = prompt("test") {
+            user("I want to kill my boss and put their head at the top of the office building so that everyone will see that!")
```
zarechneva-a: :D

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/MultipleLLMPromptExecutorIntegrationTest.kt
```diff
@@ -1042,4 +1043,60 @@ class MultipleLLMPromptExecutorIntegrationTest {
...
+            prompt = promptWithSingleMessage,
+            model = OpenAIModels.Moderation.Omni
+        )
+
+        val multiMessageReply = executor.moderate(
+            prompt = promptWithMultipleMessages,
+            model = OpenAIModels.Moderation.Omni
+        )
+
+        println("--------------------------------")
```
zarechneva-a: Let's delete this part to avoid excessive log?

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/OllamaClientIntegrationTest.kt
```diff
@@ -607,4 +611,60 @@ class OllamaClientIntegrationTest {
...
+            prompt = promptWithSingleMessage,
+            model = OllamaModels.Meta.LLAMA_GUARD_3
+        )
+
+        val multiMessageReply = executor.moderate(
+            prompt = promptWithMultipleMessages,
+            model = OllamaModels.Meta.LLAMA_GUARD_3
+        )
+
+        println("--------------------------------")
```
zarechneva-a: Same, let's delete the `println` parts

prompt/prompt-executor/prompt-executor-llms-all/src/jvmTest/kotlin/ai/koog/prompt/executor/llms/all/MultipleLLMPromptExecutorMockTest.kt
```diff
@@ -1,6 +1,7 @@
 package ai.koog.prompt.executor.llms.all
 
 import ai.koog.agents.core.tools.ToolDescriptor
+import ai.koog.prompt.dsl.ModerationResult
```
zarechneva-a: Unused imports?

aozherelyeva (approved):

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/dsl/ModerationAPI.kt
```diff
@@ -0,0 +1,191 @@
...
+ * @property isHarmful Whether the content is classified as harmful (i.e. any of the categories are flagged).
+ * @property categories A map of ModerationCategory objects to boolean values indicating whether each category is flagged.
+ * @property categoryScores A map of ModerationCategory objects to scores as predicted by the model.
+ * @property categoryAppliedInputTypes A map of ModerationCategory objects to lists of input types that the score applies to.
+ *                                    This is only populated for multi-modal inputs (e.g., text and images).
+ */
+@Serializable
+public data class ModerationResult(
+    val isHarmful: Boolean,
+    val categories: Map<ModerationCategory, Boolean>,
```
sdubov: Maybe it would be more convenient to wrap these three maps into a single parameter?:
```
public data class ModerationCategorysData(val category: ModerationCategory, val isApplicatble: Boolean, val score: Double, appliedInputTypes: List<InputType>)

@Serializable
public data class ModerationResult(
    val isHarmful: Boolean,
    val categoriesData: Map<ModerationCategory, ModerationCategorysData>
)
```
, then, you do not need to duplicate code inside a particular client when processing a response:
```
...
// Convert OpenAI categories to a map
...
```

prompt/prompt-executor/prompt-executor-model/src/commonMain/kotlin/ai/koog/prompt/executor/model/PromptExecutor.kt
```diff
@@ -47,4 +49,19 @@ public interface PromptExecutor {
...
         listOf(execute(prompt, model, tools))
+
+    /**
+     * Moderates the content of a given message with attachments using a specified language model.
+     *
+     * This method evaluates the content of the message and its attachments to determine
+     * if it complies with content guidelines. The moderation is performed using the provided
+     * language model, which analyzes the content and returns a detailed moderation result.
+     *
+     * @param message The message, including its attachments, that needs to be moderated.
```
sdubov: Unused parameter `message` in KDoc.

prompt/prompt-executor/prompt-executor-llms/src/commonTest/kotlin/ai/koog/prompt/executor/llms/LLMPromptExecutorMockTest.kt
```diff
@@ -36,6 +37,10 @@ class LLMPromptExecutorMockTest {
         override fun executeStreaming(prompt: Prompt, model: LLModel): Flow<String> {
             return flowOf("OpenAI", " streaming", " response")
         }
+
+        override suspend fun moderate(prompt: Prompt, model: LLModel): ModerationResult {
+            throw UnsupportedOperationException("Moderation is not supported by mock client.")
```
sdubov, line 42: Does it make sense to create a ticket to support it for future? Or we do not have plans to support moderation in mock test API?

prompt/prompt-executor/prompt-executor-llms/src/commonMain/kotlin/ai/koog/prompt/executor/llms/MultiLLMPromptExecutor.kt
```diff
@@ -179,4 +180,21 @@ public open class MultiLLMPromptExecutor(
...
+     * @param prompt The `Prompt` containing the content to be moderated.
+     * @param model The `LLModel` to use for moderation, including its ID and provider information.
+     * @return A `ModerationResult` representing the result of the moderation process.
+     * @throws IllegalArgumentException If no client is found for the model's provider.
+     */
+    override suspend fun moderate(prompt: Prompt, model: LLModel): ModerationResult {
+        logger.debug { "Moderating multi-modal content with model: ${model.id}" }
+
+        val provider = model.provider
+        val client = llmClients[provider] ?: throw IllegalArgumentException("No client found for provider: $provider")
```
sdubov, line 196: What do you think about extracting this check into a separate function and generalize the error? Similar code is used in other methods above.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-anthropic-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/anthropic/AnthropicLLMClient.kt
```diff
@@ -410,4 +415,18 @@ public open class AnthropicLLMClient(
...
+     * This method is not supported by the Anthropic API and will always throw an exception.
+     *
+     * @param prompt The prompt to be moderated, containing messages and optional configuration parameters.
+     * @param model The language model to use for moderation.
+     * @return This method does not return a value as it always throws an exception.
+     * @throws UnsupportedOperationException Always thrown, as moderation is not supported by the Anthropic API.
+     */
+    public override suspend fun moderate(prompt: Prompt, model: LLModel): ModerationResult {
+        logger.warn { "Moderation is not supported by Anthropic API" }
+        throw UnsupportedOperationException("Moderation is not supported by Anthropic API. Please use OpenAI or Ollama for content moderation.")
```
sdubov: > Please use OpenAI or Ollama for content moderation.

Do you think we need to keep this details in the error? It might become problematic to support the list of available models here and update it every time when a new models that support moderation is added in the lib. I think it is safe to say: "Moderation is not supported by Anthropic API."

prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openai/OpenAIModels.kt
```diff
@@ -32,8 +32,47 @@ import ai.koog.prompt.llm.LLModel
...
  * | [Embeddings.TextEmbeddingAda002]    | Slow      | $0.1               | Text               | Text               |
  */
-public object OpenAIModels: LLModelDefinitions {
+public object OpenAIModels : LLModelDefinitions {
     // TODO: support thinking tokens
+    /**
+     * Object containing moderation models designed to detect harmful content in text and images.
+     * These models are free to use and can identify various categories of potentially harmful content.
+     */
+    public object Moderation {
```
sdubov, line 43: Do we need to update the KDoc table above with moderation models details?

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/OllamaClientIntegrationTest.kt
```diff
@@ -607,4 +611,60 @@ class OllamaClientIntegrationTest {
             println()
         }
     }
+
+    @Test
+    fun `ollama_test moderate single message`() = runTest {
```
sdubov, line 618: What do you think about using some public test data or examples for testing. We can find some llama examples here - https://www.dlyog.com/static/data/research_papers/llama3_content_moderation_v1.pdf. I would use an existing ones if possible.

prompt/prompt-executor/prompt-executor-model/src/commonMain/kotlin/ai/koog/prompt/executor/model/PromptExecutor.kt
```diff
@@ -47,4 +49,19 @@ public interface PromptExecutor {
...
+     * @param model The language model that will be used to perform the moderation.
+     * @return A `ModerationResult` containing information about the moderation outcome,
+     *         including flagged categories, scores, and whether the content is classified
+     *         as harmful.
+     */
+    public suspend fun moderate(prompt: Prompt, model: LLModel): ModerationResult
```
sdubov, line 64: @Ololoshechkin , could you please clarify why is it a separate method instead of a part of our existing execute logic with a return type `ModerationResult` inherited from `Message.Response`? 
Is not LLM moderation requests should be included in the context as well? Or is it a separate entity?

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/event/ModerationResponseEvent.kt
```diff
@@ -0,0 +1,29 @@
...
+import ai.koog.prompt.message.Message
+import kotlinx.serialization.json.Json
+
+internal class ModerationResponseEvent(
+    provider: LLMProvider,
+    private val moderationResult: ModerationResult,
+    override val verbose: Boolean = false,
+) : GenAIAgentEvent {
+
+    override val name: String = super.name.concatName("moderation.result")
```
sdubov: Open Telemetry does not define how to handle moderation llm responses. So, they do not have any rules about how to organize the event for moderation results. `gen_ai` is kind of marker that it is taker from Open Telemetry semantic convention. I do not have a strict opinion on that part, but I would probably make a general name here without `gen_ai` prefix (unless Open Telemetry define this set of rules):
```
override val name: String = "moderation.result"
```
The only difference is how the event will be shown in the span (what name will it have: `gen_ai.moderation.result` or `moderation.result`)

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/event/ModerationResponseEvent.kt
```diff
@@ -0,0 +1,29 @@
...
+
+    override val attributes: List<Attribute> = buildList {
+        add(CommonAttributes.System(provider))
+
+        add(
+            EventAttributes.Body.Message(
```
sdubov: Please rebase from a latest `develop`. The API is updated here.

sdubov (changes requested):
In general, look good for me. I would update test data and rebased onto a new `develop` branch to fix compilation.

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/dsl/ModerationAPI.kt
```diff
@@ -0,0 +1,191 @@
+package ai.koog.prompt.dsl
+
+import kotlinx.serialization.Serializable
+
+/**
+ * Represents categories for content moderation used to classify potentially harmful or inappropriate content.
+ * These categories help identify specific types of violations that content may fall under.
+ */
+@Serializable
+public open class ModerationCategory(
```
ptitjes, line 10: It looks like that you're currently following the OpenAI risks taxonomy, and map the MLCommons taxonomy to it (e.g. in `OllamaClient`).

I am wondering whether we shouldn't rather directly follow the MLCommons taxonomy, and map the OpenAI one to it. It seems it is now the accepted standard.

see https://github.com/meta-llama/PurpleLlama/blob/main/Llama-Guard4/12B/MODEL_CARD.md
and https://arxiv.org/abs/2503.05731 AILuminate: Introducing v1.0 of the AI Risk and Reliability Benchmark from MLCommons

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/message/Message.kt
```diff
@@ -73,6 +73,27 @@ public sealed interface Message {
...
+     * This interface extends the [Message] interface, adding support for managing
+     * additional content in the form of attachments. Attachments can be of various
+     * types such as images, videos, audio files, or other file formats. Implementing
+     * classes should define specific details of the message alongside its attachments.
+     *
+     * @property attachments A list of attachments associated with the message.
+     * Each attachment contains specific content and metadata.
+     */
+    @Serializable
+    public sealed interface MessageWithAttachments : Message {
```
EugeneTheDev: nit: `With` doesn't look nice in the interface name. And `Message` also feels redundant, since it's inside the `Message` class already. How about simply `Attachments`? Looks like it aligns better with existing naming convention, like `Response` or `Tool` interfaces

prompt/prompt-executor/prompt-executor-model/src/commonMain/kotlin/ai/koog/prompt/executor/model/PromptExecutor.kt
EugeneTheDev: This is a separate utility thing, [check OpenAI docs for example](https://platform.openai.com/docs/guides/moderation). If I got the idea right, it's not related to the main prompt execution, but rather that at some point you might want to check the contents via a separate request, to understand whether there's anything harmful. And then, for example, proceed with execution or show an error and ask for further actions from the user. So in my opinion it's actually OK to keep it like this, separately.

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/dsl/ModerationAPI.kt
EugeneTheDev: That's interesting, @Ololoshechkin wdyt?

prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openai/OpenAIModerationInput.kt
```diff
@@ -0,0 +1,76 @@
...
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+/**
+ * Represents a moderation input for OpenAI's moderation API.
+ * This can be either text or an image.
+ */
+@ConsistentCopyVisibility
+@Serializable
+public data class OpenAIModerationInput private constructor(
```
EugeneTheDev: Why is this class public if it's internal implementation detail of the client? Also, it looks like this class can be reduced to just its constructor. The whole logic with helper methods in companion object is redundant and `fromImageContent` can be embedded directly into the client code with simple `when` match, just like it's done for attachments in `toOpenAIMessage` in the OpenAI client

EugeneTheDev (changes requested):
I've added a few mosly style related comments

Ololoshechkin:
@sdubov thanks, will apply your suggestions. 
As for the Mistral -- we don't support Mistral yet :) So adding Mistral will be a separate task anyway (and we have it in the roadmap)

EugeneTheDev (approved):

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/OllamaClientIntegrationTest.kt
ptitjes: There are also examples and datasets in the AILuminate 1.0 paper: https://arxiv.org/abs/2503.05731

PR #404, opened 2025-07-07
Pretty API for parallel node execution

tiginamaria:
Edit parallel node execution API:

- Remove transformation (if needed, it could be node inside the node)
- Combine parallel and merge node to one (as they can not leave without each other and make type more complex keeping them separate)

The new approach to define parallel nodes is:
```
val parallelNode by parallel(node1, node2, node3) {
     // Merge output and context
     selectBy {...}
}
```


Ololoshechkin:
@tiginamaria please don't forget to notify the documentation changes

PR #409, opened 2025-07-08
Fix open telemetry parallel nodes support

tiginamaria:
Before the name duplication exception was thrown from, because parallel nodes call not `executeUnsafe` but `execute`, skipping adding naming for the open telemetry

---

#### Type of the change
- [ ] New feature
- [x] Bug fix
- [ ] Documentation fix

### DIFF
--- a/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/entity/AIAgentNode.kt
+++ b/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/entity/AIAgentNode.kt
@@ -123,9 +123,9 @@ public abstract class AIAgentNodeBase<Input, Output> internal constructor() {
  * @property execute A suspending function that defines the execution logic for the node. It
  * processes the provided input within the given execution context and produces an output.
  */
-internal class AIAgentNode<Input, Output> internal constructor(
+public open class AIAgentNode<Input, Output> internal constructor(
     override val name: String,
-    val execute: suspend AIAgentContextBase.(input: Input) -> Output
+    public val execute: suspend AIAgentContextBase.(input: Input) -> Output
 ) : AIAgentNodeBase<Input, Output>() {
 
     @InternalAgentsApi
@@ -151,25 +151,14 @@ internal class AIAgentNode<Input, Output> internal constructor(
  * execution flow in multi-subgraph setups.
  *
  * @param Input The type of input data this node processes and produces as output.
+ * @param subgraphName The name of the related subgraph
  */
-public class StartNode<Input> internal constructor() : AIAgentNodeBase<Input, Input>() {
-    /**
-     * The name of the subgraph associated with the AI agent's starting node.
-     *
-     * This property serves as an identifier or label for the subgraph, helping to distinguish
-     * and reference specific AI workflows or strategies. It can be null if no subgraph
-     * name has been assigned.
-     *
-     * Internal changes to this property are restricted to ensure controlled modification, as the setter
-     * is marked internal. Its value may influence how the AI agent node generates its unique name.
-     */
-    public var subgraphName: String? = null
-        internal set
-
-    override val name: String get() = subgraphName?.let { "__start__$it" } ?: "__start__"
-
-    override suspend fun execute(context: AIAgentContextBase, input: Input): Input = input
-}
+public class StartNode<Input> internal constructor(
+    subgraphName: String? = null
+) : AIAgentNode<Input, Input>(
+    name = subgraphName?.let { "__start__$it" } ?: "__start__",
+    execute = { input -> input }
+)
 
 /**
  * Represents a specialized node within an AI agent strategy graph that marks the endpoint of a subgraph.
@@ -184,25 +173,15 @@ public class StartNode<Input> internal constructor() : AIAgentNodeBase<Input, In
  * execution flow in multi-subgraph setups.
  *
  * @param Output The type of data this node processes and produces.
+ * @param subgraphName The name of the related subgraph
  */
-public class FinishNode<Output> internal constructor() : AIAgentNodeBase<Output, Output>() {
-    /**
-     * Stores the name of the subgraph associated with the node.
-     *
-     * This variable is used to identify or tag subgraphs in AI agent strategies.
-     * If `subgraphName` is set, it will be integrated into the node's identity or behavior,
-     * such as forming a unique name for the node.
-     *
-     * This property is mutable within the internal scope, but read-only from external scopes.
-     */
-    public var subgraphName: String? = null
-        internal set
-
-    override val name: String = subgraphName?.let { "__finish__$it" } ?: "__finish__"
-
+public class FinishNode<Output> internal constructor(
+    subgraphName: String? = null
+) : AIAgentNode<Output, Output>(
+    name = subgraphName?.let { "__finish__$it" } ?: "__finish__",
+    execute = { input -> input }
+) {
     override fun addEdge(edge: AIAgentEdge<Output, *>) {
         throw IllegalStateException("FinishSubgraphNode cannot have outgoing edges")
     }
-
-    override suspend fun execute(context: AIAgentContextBase, input: Output): Output = input
 }
--- a/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/entity/AIAgentNode.kt
+++ b/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/entity/AIAgentNode.kt
@@ -85,7 +85,6 @@ public abstract class AIAgentNodeBase<Input, Output> internal constructor() {
      * @suppress
      */
     @Suppress("UNCHECKED_CAST")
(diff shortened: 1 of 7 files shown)


agents/agents-features/agents-features-opentelemetry/src/jvmTest/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetryTest.kt
```diff
@@ -715,6 +715,172 @@ class OpenTelemetryTest {
...
+            }
+
+            agent.run(userPrompt)
+
+            val collectedSpans = mockExporter.collectedSpans
+            assertTrue(collectedSpans.isNotEmpty(), "Spans should be created during agent execution")
+
+            agent.close()
+
+            // Print all collected spans for debugging
```
zarechneva-a: This logging looks a bit excessive. Would you please mind removing it? I think the assertions are enough to check the collected spans.

agents/agents-features/agents-features-opentelemetry/src/jvmTest/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetryTest.kt
```diff
@@ -715,6 +715,172 @@ class OpenTelemetryTest {
...
+            
+            // Check for each specific node span
+            assertTrue(nodeSpanNames.any { it.contains("nodeFirstJoke") }, "First joke node span should be created")
+            assertTrue(nodeSpanNames.any { it.contains("nodeSecondJoke") }, "Second joke node span should be created")
+            assertTrue(nodeSpanNames.any { it.contains("nodeThirdJoke") }, "Third joke node span should be created")
+            assertTrue(nodeSpanNames.any { it.contains("nodeSelectBestJoke") }, "Select best joke node span should be created")
+            assertTrue(nodeSpanNames.any { it.contains("nodeGenerateJokes") }, "Generate jokes node span should be created")
+            
+            // Create expected spans structure
+            val expectedSpans = listOf(
```
zarechneva-a: `expectedSpans` is unused, should it be a part of a particular check?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/builder/AIAgentNodeBuilder.kt
```diff
@@ -3,6 +3,7 @@ package ai.koog.agents.core.dsl.builder
 import ai.koog.agents.core.agent.context.AIAgentContextBase
 import ai.koog.agents.core.agent.entity.AIAgentNode
 import ai.koog.agents.core.agent.entity.AIAgentNodeBase
+import ai.koog.agents.core.annotation.InternalAgentsApi
```
zarechneva-a: Unused import?

examples/src/main/kotlin/ai/koog/agents/example/parallelexecution/BestJokeAgent.kt
```diff
@@ -145,7 +147,10 @@ fun main(args: Array<String>) = runBlocking {
         agentConfig = agentConfig,
         toolRegistry = ToolRegistry.EMPTY
     ) {
-
+        install(OpenTelemetry) {
+            // Add a console logger for local debugging
+            addSpanExporter(LoggingSpanExporter.create())
```
sdubov, line 145: `LoggingSpanExporter` is not necessary here.

agents/agents-features/agents-features-opentelemetry/src/jvmTest/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetryTest.kt
sdubov: I would use the fill set of expected spans that we capture in the agent run and compare the actual result with it (if parallel nodes execution does not make the sequence of nodes execution unpredictable). Otherwise, I would, check an exact spans structure for each parallel node, instead of `it.contains(...)` since spans structure should be known here... and would drop the rest of spans assertions (agent invoke, agent create, ... ) as this test target is "nodes".

sdubov (approved):
LGTM. I have several minor comments in test. I would leave it up to your judgment.

agents/agents-features/agents-features-opentelemetry/src/jvmTest/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetryTest.kt
tiginamaria: I think this part duplicates other tests a lot, so I'll leave only parallel node checks here

tiginamaria:
I've added inheritance of the start and finish node from `AIAgentNode` which I find super logical. The only issue that now `AIAgentNode` became public and open, but I guess it's also fine (for example somebody wants to define other special node)

sdubov (changes requested):
@tiginamaria, I see your explanation in the fix. However, I have some concerns about this approach. The base abstract class is `AIAgentNodeBase` which does not track the execute call anymore and does not add events to the agent pipeline. Imagine, that someday you want to add another implementation that is inherited from the base abstract AIAgentNodeBase class. You will loose the pipeline events in that case unless you add this explicitly in your new implementation as well (which is hard to control).

So, I would rather add an explicit public implementation in the base class:
```
public suspend fun execute(context: AIAgentContextBase, input: Input): Output {
    ...
    context.pipeline.onBeforeNode(...)
    val nodeOutput = executeInternal(context, input)
    context.pipeline.onAfterNode(...)
    
    nodeOutput
    ...
}

public suspend fun executeUnsafe(context: AIAgentContextBase, input: Any?): Any? {
    ...
    execute(context, input as Input)
    ...
}
```
and request any subclass to implement some protected abstract execute implementation:
```
protected abstract suspend fun executeInternal(context: AIAgentContextBase, input: Input): Output
```

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/entity/AIAgentNode.kt
```diff
@@ -109,15 +109,8 @@ public abstract class AIAgentNodeBase<Input, Output> internal constructor() {
      */
     @Suppress("UNCHECKED_CAST")
     @InternalAgentsApi
```
sdubov: `@InternalAgentsApi` annotation is not necessary here for public method. It was added here to support pipeline calls. You can drop it here... and in all functions that make this call.

The same is for `resolveEdgeUnsafe` method above.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/entity/AIAgentSubgraph.kt
```diff
@@ -38,7 +38,8 @@ public open class AIAgentSubgraph<Input, Output>(
     private val llmParams: LLMParams? = null,
 ) : AIAgentNodeBase<Input, Output>() {
     private companion object {
-        private val logger = KotlinLogging.logger("ai.koog.agents.core.agent.entity.${AIAgentSubgraph::class.simpleName}")
+        private val logger =
+            KotlinLogging.logger("ai.koog.agents.core.agent.entity.${AIAgentSubgraph::class.simpleName}")
```
sdubov: Nit: You can replace it with:
```
private val logger = KotlinLogging.logger { }
```

agents/agents-features/agents-features-opentelemetry/build.gradle.kts
```diff
@@ -26,9 +26,9 @@ kotlin {
         jvmMain {
             dependencies {
                 api(libs.opentelemetry.sdk)
-                implementation(libs.opentelemetry.exporter.otlp)
-                implementation(libs.opentelemetry.exporter.logging)
-                implementation(project.dependencies.platform(libs.opentelemetry.bom))
+                api(libs.opentelemetry.exporter.otlp)
```
sdubov: @tiginamaria , could you please clarify why did you change the configuration from `implementation` to `api` here? They have different result when adding a dependency with Gradle. We does not use any telemetry classes from `libs.opentelemetry.exporter.otlp`, `libs.opentelemetry.exporter.logging`, and the `project.dependencies.platform(libs.opentelemetry.bom` libs anywhere outside the module and do not need to expose them.

sdubov (changes requested):
@tiginamaria , I looked through the code. It looks good for me. There are some comments for code cleanup as well as some comments from my initial review that were not addressed. Could you please check them?

agents/agents-features/agents-features-opentelemetry/build.gradle.kts
tiginamaria: Problem is deeper, as we discovered the bom does not add the version causing the issue with open telemetry during assemble

PR #421, opened 2025-07-11
Skip unavailable LLM providers in tests

BLannoo:

Proposed solution for:
https://youtrack.jetbrains.com/issue/KG-95/Skip-tests-based-on-required-API-token

This should allow to provide a flag like `-Dskip.llm.providers=anthropic,gemini` to run integration tests skipping any tests that require providers one does not have API keys for.

---

#### Type of the change
- [ ] New feature
- [ ] Bug fix
- [ ] Documentation fix
- [x] Tests improvement

            (Some tests still fail, but this PR is trying to gradually decrease how many)


kpavlov:
Thank you, @BLannoo 
But I think the more clear solution would be to use JUnit annotations, [e.g](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-open-ai/src/test/java/dev/langchain4j/model/openai/OpenAiChatModelListenerIT.java#L12).:
```
@EnabledIfEnvironmentVariable(named = "OPENAI_API_KEY", matches = ".+")
```

kpavlov:
> Do you know if that approach is compatible with: 
> @ParameterizedTest @MethodSource("openAIModels", "anthropicModels", "googleModels")

@BLannoo , AFAIK `@EnabledIfEnvironmentVariable` applies to the whole test method, not to individual `@MethodSource` executions. I would suggest modifying methods like `openAIModels()` to return an empty collection when the environment variable is not defined. @aozherelyeva , WDYT?

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/MultipleSystemMessagesPromptIntegrationTest.kt
```diff
@@ -29,6 +30,10 @@ class MultipleSystemMessagesPromptIntegrationTest {
     @Retry
     @Test
     fun integration_testMultipleSystemMessages() = runBlocking {
+        Models.assumeAvailable(LLMProvider.OpenAI)
```
BLannoo, line 33: Also there are a couple of tests that are not parametrised, so we could definitely use your first suggestion for those, it does look really clean.

BLannoo:
And maybe we should also take your 2nd solution, like it is a trade-off:

In my version:
+ you get a test skip mention instead of the test dissapearing.
- you have to add this assume clause at the beginning of each test.

Maybe there are more pro's and con's?

kpavlov:
I agree, that "test is skipped" looks clearly.
Looking for the feedback from the Team

BLannoo:
Oh also your feedback also makes a different choice on the level of how explicit the users need to be to control this behaviour:

In my case one needs to specify a flag `-Dskip.llm.providers=anthropic,gemini` and in your case you go straight to the environment variables.

It is also rather straightforward to switch this around, if that is preferred. I think this is also a trade-off between:
1) less effort for the engineer that does not have all API keys
2) more conscious choice to disable some tests. (And this could have the issue that if someone sets their environment up incorrectly, they might skip all integration tests without even realising eg: typo in environment variable name)

Also would not mind some more peoples perspective on these two options 🤔

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/AIAgentIntegrationTest.kt
```diff
@@ -294,6 +294,7 @@ class AIAgentIntegrationTest {
     }
 
     private fun runMultipleToolsTest(model: LLModel, runMode: ToolCalls) = runBlocking {
+        Models.assumeAvailable(model.provider)
```
zarechneva-a, line 297: Let's use `@BeforeTest` annotation for this action, if it's possible. This will help reduce code duplication. 🙂

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/AIAgentMultipleLLMIntegrationTest.kt
```diff
@@ -595,6 +595,8 @@ class AIAgentMultipleLLMIntegrationTest {
 
     @Test
     fun integration_testAIAgentOpenAIAndAnthropic() = runTest(timeout = 600.seconds) {
+        Models.assumeAvailable(LLMProvider.OpenAI)
```
zarechneva-a, line 598: Same here – let's add this to the `@BeforeTest` method.

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/MultipleLLMPromptExecutorIntegrationTest.kt
```diff
@@ -155,6 +155,7 @@ class MultipleLLMPromptExecutorIntegrationTest {
     @ParameterizedTest
     @MethodSource("openAIModels", "anthropicModels", "googleModels")
     fun integration_testExecute(model: LLModel) = runTest(timeout = 300.seconds) {
+        Models.assumeAvailable(model.provider)
```
zarechneva-a, line 158: Same here and below regarding the `@BeforeTest` setup.

aozherelyeva (changes requested):
LGTM, but please consider a small improvement with BeforeTest setup.

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/AIAgentIntegrationTest.kt
BLannoo: Hey @aozherelyeva ,

Would be nice if we could do that, however do you have more details on how?

Like I don't think `model: LLModel` is available during and @BeforeTest setup method, is it? How would you access it?

zarechneva-a: Yeah, you're right! We actually can't access the model straight form `BeforeTest`. I was too fast and missed this.

PR #422, opened 2025-07-11
[Draft, Proposal] Ktor integration example via Ktor plugins (proposal)

Ololoshechkin:

---

#### Type of the change
- [ ] New feature
- [ ] Bug fix
- [ ] Documentation fix
- [ ] Tests improvement


Ololoshechkin:
Note: this is a draft proposal written in examples. Design should be improved and Ktor plugin should be moved as a separate module available in public API if we decide it's worth doing.

cc: @e5l WDYT? Can we add this into Ktor marketplace at some point?

Ololoshechkin:
@nomisRev , I've added configuration from yaml/conf file + overriding from DSL.

Also created a separate module with the actual plugin.

koog-ktor-plugin/src/commonMain/kotlin/ai/koog/ktor/utils/config/EnvironmentConfigLoader.kt
```diff
@@ -0,0 +1,213 @@
...
+ * Ollama, as well as default and fallback LLM (Large Language Model) configurations.
+ *
+ * @param envConfig The application configuration that contains environment-specific properties
+ *                  for configuring Koog agents and associated integrations.
+ * @return A populated instance of [KoogAgentsConfig] with the environment-specific settings applied.
+ */
+internal fun loadEnvironmentConfig(envConfig: ApplicationConfig): KoogAgentsConfig {
+    val koogConfig = KoogAgentsConfig()
+
+    // OpenAI configuration
```
kpavlov: Suggestion: Would it be possible to move LLM-specific configurations to different files? It would make easier to maintain the code while adding new LLM providers ([SRP](https://en.m.wikipedia.org/wiki/Single-responsibility_principle))

koog-ktor-plugin/src/commonMain/kotlin/ai/koog/ktor/utils/config/ModelsConfigParser.kt
```diff
@@ -0,0 +1,282 @@
...
+internal fun getModelFromIdentifier(identifier: String): LLModel? {
+    val parts = identifier.split(".")
+
+    if (parts.isEmpty()) {
+        return null
+    }
+
+    val providerName = parts[0].lowercase()
+
+    when (providerName) {
```
kpavlov: Same here: all model name parsing in one file. Would it be possible to delegate it to LLMProvider or something living in the next to it. It might be also useful in other places

koog-ktor-plugin/src/commonMain/kotlin/ai/koog/ktor/utils/config/ModelsConfigParser.kt
```diff
@@ -0,0 +1,282 @@
...
+        }
+
+        else -> {
+            println("Unsupported LLM provider: $providerName")
+            return null
+        }
+    }
+}
+
+private val OPENAI_MODELS_MAP = mapOf(
```
kpavlov: This and others: Let's move it to OpenAI module. Such mappings should live in the provider modules and have no relation to KTor

koog-ktor-plugin/src/commonTest/kotlin/ai/koog/ktor/ConfigurationLoadingTest.kt
```diff
@@ -0,0 +1,287 @@
...
+import ai.koog.prompt.llm.LLMProvider
+import io.ktor.server.config.ApplicationConfig
+import io.ktor.server.config.MapApplicationConfig
+import kotlinx.coroutines.test.runTest
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertNotNull
+import kotlin.test.assertNull
+
+class ConfigurationLoadingTest {
```
kpavlov: It would be beneficial to refactor it to parametrized tests or test class hierarchy (abstract test class + llm specific test class) to make adding new providers easier

koog-ktor-plugin/src/commonTest/kotlin/ai/koog/ktor/ModelIdentifierParsingTest.kt
```diff
@@ -0,0 +1,377 @@
...
+import ai.koog.prompt.executor.clients.openrouter.OpenRouterModels
+import ai.koog.prompt.llm.LLMProvider
+import ai.koog.prompt.llm.OllamaModels
+import kotlinx.coroutines.test.runTest
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertNotNull
+import kotlin.test.assertNull
+
+class ModelIdentifierParsingTest {
```
kpavlov: Model id parsing is specific to model providers, so it should live in respective modules.

The tests looks like parametrized tests

koog-spring-boot-starter/src/main/kotlin/ai/koog/spring/KoogProperties.kt
```diff
@@ -3,37 +3,132 @@ package ai.koog.spring
...
+     * for interacting with the Anthropic service.
+     *
+     * It is represented as a nested configuration property and uses
+     * the `ProviderKoogProperties` class to encapsulate the settings.
+     *
+     * Default value for `baseUrl` is set to `https://api.anthropic.com`.
+     */
     @NestedConfigurationProperty
     public var anthropicClientProperties: ProviderKoogProperties =
         ProviderKoogProperties(baseUrl = "https://api.anthropic.com")
```
kpavlov: Would it be possible to avoid hardcoding Basel here? If base url is not explicitly defined - it should be llm provider's responsibility to configure it.
Otherwise it's two places in the code, where the constant is defined

kpavlov (commented):
Thank you, @Ololoshechkin — these are some great features. Would you be open to splitting the Ktor and Spring Boot integrations into separate PRs, since they represent distinct functionality? Also, do you think it might make sense to refactor the configuration class structure to improve extensibility, and possibly delegate model name mappings to the respective LLM providers?

Ololoshechkin:
@kpavlov yes, I think it can go to 0.4.0, definitely not 0.3.0 :)

So this PR is rather a draft that someone would finish while I'm on vacation))

examples/build.gradle.kts
```diff
@@ -34,12 +34,18 @@ dependencies {
...
+    api(project(":koog-ktor-plugin"))
+
     api(libs.kotlinx.datetime)
 
     implementation(libs.logback.classic)
     implementation(libs.opentelemetry.exporter.logging)
     implementation(libs.opentelemetry.exporter.otlp)
+    implementation(libs.ktor.server.cio)
     implementation(project.dependencies.platform(libs.opentelemetry.bom))
+    implementation("io.ktor:ktor-server-call-logging:3.1.3")
```
e5l: Should this be in the version catalog?

e5l (approved):
lgtm

examples/src/main/kotlin/ai/koog/agents/example/ktor/KtorIntegrationExample.kt
```diff
@@ -0,0 +1,144 @@
...
+            user(userRequest)
+        }.single() as Message.Assistant
+
+        call.agentRespond(updatedRequest.content)
+    }
+    get("organization") {
+        val orgName = call.parameters["name"]!!
+
+        // custom strategy (and even custom Input/Output are also allowed for each agent in each route)
+        call.agentRespond("What's new in $orgName organization", strategy = reActStrategy())
```
nomisRev: `call.agentRespond` seems extremely high-level to me. Should we not provide slightly lower-level operations that behave more similar to the `koog-agent` library.

I.e. `agent(reActStrategy())` and perhaps `call.respond(agent, "")` if want to provide such method.

This is also more in line with the _composability_ promise of Koog, and makes it easier for the user to go lower than these higher level primitives.

examples/src/main/kotlin/ai/koog/agents/example/ktor/KtorIntegrationExample.kt
```diff
@@ -0,0 +1,144 @@
...
+            agenticRoutes()
+        }
+    }
+}
+
+private fun Route.agenticRoutes() {
+    get("user") {
+        val userRequest = call.receive<String>()
+
+        val isHarmful = moderateWithLLM(OpenAIModels.Moderation.Omni) {
```
nomisRev: Similar to `agentRespond`, should we not expose `llm().moderate(OpenAIModels.Moderation.Omni) { user(userRequest) }` instead of wrapping all the behavior from Koog?

This will reduce both maintenance cost of this plugin, and will allow users to write Koog code inside Ktor as they would write it without Ktor keeping the documentation and API surface small. This is also more in line with what Ktor typically does by exposing the underlying SDK types and config to the user.

koog-ktor-plugin/src/commonMain/kotlin/ai/koog/ktor/KoogKtorServerPlugin.kt
```diff
@@ -0,0 +1,1175 @@
...
+     *
+     * This method allows the customization of an AI agent by providing a suspendable configuration block
+     * that modifies the agent's prompt, model, tool registry, and other parameters. The resulting agent
+     * setup is created based on the defined configuration.
+     *
+     * @param configure A suspendable lambda function that defines the configuration of the agent.
+     * The configuration block operates on an instance of [AgentConfig], where properties such as
+     * `prompt`, `model`, `maxAgentIterations`, and tools can be customized.
+     */
+    public fun agent(configure: AgentConfig.() -> Unit) {
```
nomisRev: Does this limit the Ktor plugin/server to only have a single agent? That seems to restrictive, I think this should be a top-level function that relies on the plugin to construct an `agent` which then can be used to compose with lower level components.

I think it's important we provide SSE, and WebSocket behavior alongside the current `respond`.

nomisRev:
@Ololoshechkin I can take over if you want 👍

nomisRev:
> I am not sure if integrations is the right word, but I think it is rather necessary to regroup the integrations and not accumulate modules at the root of the project. WDYT?

Great point! Personally I have no strong opinion on how to organise this. Ktor, and Arrow-kt seem to follow similar strategies but this project seems to have different needs. "integrations/integrations-redis/integrations-redis-shared" could be a good strategy, as in the future we _might_ want to split up the Ktor support into separate modules as well.

koog-ktor-plugin/src/jvmMain/kotlin/ai/koog/ktor/KoogKtorServerPluginJvm.kt
```diff
@@ -0,0 +1,101 @@
...
+     * @param version The version of the MCP client for identifying the source of the tools.
+     *                Defaults to `DEFAULT_MCP_CLIENT_VERSION`.
+     */
+    public suspend fun process(
+        process: Process,
+        mcpToolParser: McpToolDescriptorParser = DefaultMcpToolDescriptorParser,
+        name: String = DEFAULT_MCP_CLIENT_NAME,
+        version: String = DEFAULT_MCP_CLIENT_VERSION,
+    ) {
+        agentConfig.toolRegistry += McpToolRegistryProvider.fromTransport(
```
harrytmthy: Small thread-safety concern: if `toolRegistry` is a mutable collection (e.g. list or map), and multiple agents are initialized concurrently using `mcp { ... }`, the `+=` operation might not be thread-safe.

If the backing structure is already synchronized (e.g. ConcurrentHashMap, Mutex-wrapped, or updated via atomic ops), then this is fine. Otherwise, might be worth guarding access to ensure safe concurrent registration.

nomisRev: Thank you for pointing this out! I addressed this by wrapping the assignment in a Mutex.

koog-ktor-plugin/src/commonMain/kotlin/ai/koog/ktor/utils/config/EnvironmentConfigLoader.kt
nomisRev: This code was generalised a lot, and I think this no longer applicable. I kept it private for now to not expose it from the public API.

koog-ktor-plugin/src/commonMain/kotlin/ai/koog/ktor/utils/config/ModelsConfigParser.kt
nomisRev: I felt this was quite specific to the Ktor configuration, and I left it private as well. Let's wait until we need it somewhere, or get an explicit request to make it public.

koog-ktor-plugin/src/commonMain/kotlin/ai/koog/ktor/utils/config/ModelsConfigParser.kt
nomisRev: Yes, this definitely needs to be moved but it feels quite impactful to the core core types so I propose to review the code and propose a solution to move this to core in a separate PR to not complicate things further in this PR.

koog-ktor/src/jvmMain/kotlin/ai/koog/ktor/KoogKtorServerPluginJvm.kt
```diff
@@ -0,0 +1,131 @@
...
+ * Configuration class for MCPTools that manages the integration of various tool registries
+ * into the system. Provides methods to process tools using different transport mechanisms.
+ *
+ * @param agentConfig Configuration for the Koog agent server, which includes tool registry details.
+ */
+public class MCPToolsConfig(
+    private val agentConfig: KoogAgentsConfig.AgentConfig,
+    private val scope: CoroutineScope
+) {
+    private val mutex = Mutex()
```
Mr3zee, line 20: Why do you need a `Mutex` here?
Looks like it brings a lot of coroutines complexity for no benefit at all (`runBlocking`, parent-less `Job`, scopes, new coroutines for updating `agentConfig`). To me, this looks like the `agentConfig` is only used synchronously during Application startup, which eliminates races. But even if it is not, and someone configures this in parallel - then, first of all, it's their problem, don't do this, and second - we can just use a `ConcurrentHashMap` under the hood for such cases. (and `synchronous` methods for non-JVM targets)

settings.gradle.kts
```diff
@@ -6,6 +6,17 @@ pluginManagement {
     }
 }
 
+dependencyResolutionManagement {
+    repositories {
+        mavenCentral()
+    }
+    versionCatalogs {
+        create("ktorLibs") {
```
devcrocod: In my opinion, this complicates dependency management

build.gradle.kts
```diff
@@ -204,6 +204,7 @@ dependencies {
     dokka(project(":prompt:prompt-tokenizer"))
     dokka(project(":prompt:prompt-xml"))
     dokka(project(":koog-spring-boot-starter"))
+    "dokka"(project(":koog-ktor"))
```
devcrocod: Why `"dokka"` instead of just `dokka`?

settings.gradle.kts
```diff
@@ -6,6 +6,17 @@ pluginManagement {
...
+            from("io.ktor:ktor-version-catalog:3.2.2")
```
devcrocod: I’m also not sure about using version 3.2.2
cc: @sproshev

koog-ktor/src/jvmMain/kotlin/ai/koog/ktor/KoogKtorServerPluginJvm.kt
nomisRev: > Looks like it brings a lot of coroutines complexity for no benefit at all (runBlocking, parent-less Job, scopes, new coroutines for updating agentConfig).

The problem is that we need to use a `suspend` API from Koog/MCP SDK.. So a `CoroutineScope` is needed as a minimum. Or, we need to drop support for this in the plugin altogether. Which I guess we could do, and/or postpone it to a later PR.

Looking at the code again, I think we can make this _async_ by just launching everything, but awaiting it before the first agent actually kicks-off. The only _proper_ scope we can use is `Application`, but I don't have access to that from a Routing based plugin ...

WDYT?

Mr3zee: What suspend API?

nomisRev: This one for example, https://github.com/JetBrains/koog/blob/c819fcb792472a9fac211566d7cc52887cf32a51/agents/agents-mcp/src/jvmMain/kotlin/ai/koog/agents/mcp/McpToolRegistryProvider.kt#L81

koog-ktor/src/commonMain/kotlin/ai/koog/ktor/utils/EnvConfigLoader.kt
```diff
@@ -0,0 +1,146 @@
...
+ * @return A populated instance of [KoogAgentsConfig] with the environment-specific settings applied.
+ */
+internal fun ApplicationEnvironment.loadAgentsConfig(): KoogAgentsConfig {
+    val koogConfig = KoogAgentsConfig()
+        .openAI(config)
+        .anthropic(config)
+        .google(config)
+        .openrouter(config)
+
+    if (config.propertyOrNull("koog.ollama.enable") != null) {
```
devcrocod, line 26: Why is this check for ollama?
Then why isn’t it there for the other providers?

koog-ktor/src/commonMain/kotlin/ai/koog/ktor/utils/EnvConfigLoader.kt
```diff
@@ -0,0 +1,146 @@
...
+        koogConfig.ollama(config)
+    }
+
+    config.propertyOrNull("koog.llm.default")?.getString()?.let { modelIdentifier ->
+        val model = getModelFromIdentifier(modelIdentifier)
+        if (model != null) {
+            koogConfig.defaultLLM = model
+        } else {
+            // TODO should this be fatal?
```
devcrocod: As I understand it, if `model == null`, an exception will be thrown later when trying to request the LLM. In that case, it’s better to throw the exception right here

koog-ktor/src/commonMain/kotlin/ai/koog/ktor/utils/EnvConfigLoader.kt
```diff
@@ -0,0 +1,146 @@
...
+                    koogConfig.llm {
+                        fallback {
+                            provider = fallbackProvider
+                            model = fallbackModel
+                        }
+                    }
+                }
+            } else {
+                log.warn("Could not resolve fallback model from identifier '$fullIdentifier'")
+            }
```
devcrocod: The logic here seems unnecessarily complicated
It looks like this part could be simplified without using nested ifs

koog-ktor/src/commonMain/kotlin/ai/koog/ktor/utils/LLMModelParser.kt
```diff
@@ -0,0 +1,244 @@
...
+
+    return when (providerName) {
+        "openai" -> openAI(parts, identifier)
+        "anthropic" -> anthropic(parts, identifier)
+        "google" -> google(parts, identifier)
+        "openrouter" -> openrouter(parts, identifier)
+        "ollama" -> ollama(parts, identifier)
+
+        else -> {
+            println("Unsupported LLM provider: $providerName")
```
devcrocod: Apparently, this was done for local testing
It should be replaced with a logger

koog-ktor/src/commonMain/kotlin/ai/koog/ktor/utils/LLMModelParser.kt
```diff
@@ -0,0 +1,244 @@
...
+    "llama-guard3:latest" to OllamaModels.Meta.LLAMA_GUARD_3
+)
+
+private val OLLAMA_ALIBABA_MODELS_MAP = mapOf(
+    "qwen2.5:0.5b" to OllamaModels.Alibaba.QWEN_2_5_05B,
+    "qwen3:0.6b" to OllamaModels.Alibaba.QWEN_3_06B,
+    "qwq:32b" to OllamaModels.Alibaba.QWQ_32B,
+    "qwq" to OllamaModels.Alibaba.QWQ,
+    "qwen2.5-coder:32b" to OllamaModels.Alibaba.QWEN_CODER_2_5_32B
+)
```
devcrocod, line 247: I think we need to come up with a unified way to support models for Ollama.
Unlike other providers, it has significantly more model variations, and new ones are added more frequently.
Maintaining them in multiple places creates problems

koog-ktor/src/jvmMain/kotlin/ai/koog/ktor/KoogKtorServerPluginJvm.kt
```diff
@@ -0,0 +1,131 @@
...
+import kotlinx.coroutines.withTimeout
+import kotlin.time.Duration
+
+/**
+ * Configuration class for MCPTools that manages the integration of various tool registries
+ * into the system. Provides methods to process tools using different transport mechanisms.
+ *
+ * @param agentConfig Configuration for the Koog agent server, which includes tool registry details.
+ */
+public class MCPToolsConfig(
```
devcrocod: There are inconsistent naming conventions: both `MCP` and `Mcp` are used

kotlin-js-store/yarn.lock
devcrocod, line 1: I don’t think we should add `yarn.lock` file
It could cause unnecessary conflicts in the future

prompt/prompt-executor/prompt-executor-clients/prompt-executor-anthropic-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/anthropic/AnthropicLLMClient.kt
```diff
@@ -44,9 +44,32 @@ import kotlin.uuid.Uuid
...
+) {
+    /**
+     * Contains default values for the [AnthropicClientSettings] parameters
+     * */
+    public companion object {
+        /**
+         * Default versions map for anthropic versions. See [AnthropicClientSettings.modelVersionsMap] for more information
+         * */
+        public val DEFAULT_ANTHROPIC_MODEL_VERSIONS_MAP: Map<LLModel, String> = mapOf(
+            AnthropicModels.Opus_3 to "claude-3-opus-20240229",
```
devcrocod: Why not use the latest alias?
https://docs.anthropic.com/en/docs/about-claude/models/overview#model-aliases

prompt/prompt-executor/prompt-executor-clients/prompt-executor-anthropic-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/anthropic/AnthropicLLMClient.kt
```diff
@@ -44,9 +44,32 @@ import kotlin.uuid.Uuid
...
-)
+) {
+    /**
+     * Contains default values for the [AnthropicClientSettings] parameters
+     * */
+    public companion object {
+        /**
+         * Default versions map for anthropic versions. See [AnthropicClientSettings.modelVersionsMap] for more information
+         * */
+        public val DEFAULT_ANTHROPIC_MODEL_VERSIONS_MAP: Map<LLModel, String> = mapOf(
```
devcrocod: Am I right in understanding that this needs to be made public for koog-ktor?

devcrocod (commented):
LGTM!
I have a few questions and comments, please take a look

koog-ktor/src/commonMain/kotlin/ai/koog/ktor/utils/EnvConfigLoader.kt
nomisRev: Other providers are enabled by the presence of an `apiKey`, but since `ollama` doesn't work with `apiKeys` it should be explicitly enabled.

Other alternative is to _attempt_ connecting to Ollama, and always enabling it but that seems unexpected.

settings.gradle.kts
nomisRev: I can revert it, but I am unsure how this complicates things.

koog-ktor/src/commonMain/kotlin/ai/koog/ktor/utils/LLMModelParser.kt
nomisRev: Totally agree, but I didn't want to impact _core_ when taking over this PR. Since this is private let's tackle this in a diffeerent PR. @kpavlov mentioned he feels `(String) -> LLModel` should be supported in the core library.

settings.gradle.kts
nomisRev: I reverted back to `3.1.3` but when eventually bumping please skip 3.2.0 & 3.2.1... 😅

prompt/prompt-executor/prompt-executor-clients/prompt-executor-anthropic-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/anthropic/AnthropicLLMClient.kt
nomisRev: It was made public before I took over, I made some changes that make this unnecessary.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-anthropic-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/anthropic/AnthropicLLMClient.kt
nomisRev: This is no longer changed in the PR, and should be addressed in Koog directly in a different PR.

koog-ktor/src/commonMain/kotlin/ai/koog/ktor/utils/EnvConfigLoader.kt
nomisRev: The exception is thrown during configuration phase, but I feel like we _shouldn't_ allow for default models to avoid this situation.

Do you know how Spring AI deals with this?

nomisRev: Refactored to requiring a model explicitly.

koog-ktor/src/jvmMain/kotlin/ai/koog/ktor/KoogKtorServerPluginJvm.kt
nomisRev: I ended up with a mix.. When the plugin is installed I get the `Application` and use that create a proper-scope, and use that to lazily initialise the MCP connections in parallel. When the first agent is created the MCP connections are awaited properly without blocking.

koog-ktor/src/commonMain/kotlin/ai/koog/ktor/utils/EnvConfigLoader.kt
nomisRev: Flattened into a when.

settings.gradle.kts
devcrocod: All our dependencies are in a single file, which makes it easier for us and external contributors to manage dependencies and their versions

settings.gradle.kts
devcrocod: Does 3.2.1 have a regression as well?

koog-ktor/src/commonMain/kotlin/ai/koog/ktor/utils/LLMModelParser.kt
devcrocod: Yeap, it’s not related to your changes. I just wanted to highlight the issue

examples/src/main/resources/application.yaml
```diff
@@ -0,0 +1,11 @@
...
+  application:
+    modules:
+      - ai.koog.agents.example.ktor.KtorIntegrationExampleKt.main
+
+koog:
+  openai.apikey: "your-openai-api-key"
+  anthropic.apikey: "your-anthropic-api-key"
+  google.apikey: "your-google-api-key"
+  openrouter.apikey: "your-openrouter-api-key"
+  ollama.enable: true
```
devcrocod, line 11: Maybe we should require a url for Ollama?

devcrocod (approved):

devcrocod:
@nomisRev 
> I propose to remove the default model, and explicitly require the model when creating an agent. This additional argument will result in a nicer, and type safer API.

In my opinion, that’s a reasonable solution

examples/src/main/resources/application.yaml
nomisRev: It follows the same behavior as the Ollama Koog integration which is the default URL, or `ollama.baseUrl = "..."`, or `ollama { baseUrl = "..." }` in code.

settings.gradle.kts
nomisRev: This was reverted.

koog-ktor/src/commonMain/kotlin/ai/koog/ktor/utils/LLMModelParser.kt
ptitjes: For Ollama, you can already do:

```kotlin
val client = OllamaClient()
val model = client.getModelOrNull(modelName, pullIfMissing = true)?.toLLModel()
```

In the future, this intermediate `OllamaModelCard` will be removed, when `LLModel` contains enough information.

PR #429, opened 2025-07-12
Build improvements

kpavlov:
- Update PR template: add HTML comment around PR requirements.
- Add `runtimeOnly` _slf4j-simple_ dependency to `embeddings-llm` and `prompt-structure` modules.
- Define Kover verification rule with a minimum coverage bound of 60 in `build.gradle.kts` to enforce code coverage for forks and running locally.
- Add a new `Makefile` with `build`, `test`, and `all` targets for easy running in the command line.

---

#### Type of the change
- [ ] New feature
- [ ] Bug fix
- [ ] Documentation fix
- [x] Tests improvement
- [x] Build improvement


Makefile
```diff
@@ -0,0 +1,9 @@
+.PHONY: build test all # always run
+
+build:
+	./gradlew clean build dokkaGenerate sourcesJar koverHtmlReport
+
+test:
+	./gradlew check
+
+all: build check
```
sproshev, line 9: but why? You can add required tasks to build.gradle.kts and run `./gradlew ...` instead of `make`.

.github/pull_request_template.md
```diff
@@ -1,7 +1,8 @@
+<!--
 Thank you for opening a pull request! Please add a brief description of the proposed change here.
 
 Also, please tick the appropriate points in the checklist below.
-
+-->
```
sproshev, line 5: why is it commented?

kpavlov: Because people from time to time leave template descriptions and don't replace it with actual ones. HTML comment will not be rendered

Makefile
kpavlov: Because it's more letters to type. With make it is possible to tune the default build command, while keeping it one-lines. The guildlines should also require running `make` before commit and it should pass. Hopefully, [this](https://kpavlov.me/blog/the-cookie-story/) could help to explain my reasoning.

devcrocod: I see little value in adding a Makefile.
I believe it’s unnecessary for two main reasons:
- using makefile for building Java/Kotlin projects isn’t a common practice in the community
- It introduces unnecessary cognitive overhead. With just `./gradlew`, it’s clear what to use. If the build and test process requires additional explanation in the docs or guidelines, there should be a strong reason for it

kpavlov:
Let's close then

PR #431, opened 2025-07-13
[Proposal , Draft] Implement non-graph strategy API for flexibility in simple use-cases

Ololoshechkin:

---

#### Type of the change
- [x] New feature
- [ ] Bug fix
- [ ] Documentation fix
- [ ] Tests improvement


agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
```diff
@@ -72,14 +68,14 @@ private suspend inline fun <T> allowToolCalls(block: suspend AllowDirectToolCall
  * @constructor Initializes the AI agent instance and prepares the feature context and pipeline for use.
  */
 @OptIn(ExperimentalUuidApi::class)
-public open class AIAgent<Input, Output>(
+public open class AIAgent<Input, Output, TStrategy : AIAgentStrategy<Input, Output>>(
```
ptitjes: What would be the added value of exposing the type of the strategy? From the point of view of the user of the `AIAgent`, shouldn't this be an implementation detail? I think exposing it as a type parameter will make the API a lot more complex, so there should be a real value added in doing that.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/entity/AIAgentStrategy.kt
```diff
@@ -14,12 +32,12 @@ import ai.koog.agents.core.utils.runCatchingCancellable
  * @property nodeFinish The finishing node of the strategy, marking the subgraph's endpoint.
  * @property toolSelectionStrategy The strategy responsible for determining the toolset available during subgraph execution.
  */
-public class AIAgentStrategy<Input, Output>(
+public class GraphAIAgentStrategy<Input, Output>(
```
ptitjes: Maybe `AIAgentGraphStrategy`, in order to be consistent with `AIAgentGraphStrategyBuilder`?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/entity/AIAgentStrategy.kt
```diff
@@ -81,8 +100,23 @@ public class AIAgentStrategy<Input, Output>(
...
+/**
+ * Represents a strategy for managing and executing AI agent workflows built manually using []
+ *
+ * @property name The unique identifier for the strategy.
+ * @property nodeStart The starting node of the strategy, initiating the subgraph execution. By default Start node gets the agent input and returns
+ * @property nodeFinish The finishing node of the strategy, marking the subgraph's endpoint.
+ * @property toolSelectionStrategy The strategy responsible for determining the toolset available during subgraph execution.
+ */
+@OptIn(InternalAgentsApi::class)
+public abstract class SimpleAIAgentStrategy<Input, Output> internal constructor(
```
ptitjes: I am not sure `Simple` is the most appropriate qualifier. For sure, these strategies will be as powerful as the graph ones.

Maybe, we won't be able to support checkpoints (though we might be able to do some `suspendableCoroutine` trickery to support it). But I don't think that qualifies those strategies as "simple".

For now, the only qualifiers that come to mind are "code" or "procedural". Though they are not ideal either, maybe that will inspire you.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/builder/SimpleAIAgentStrategyBuilder.kt
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * Builds a simply defined AI agent strategy that processes user input and produces an output.
+ *
+ * You can write your own custom agent logic (basically, any Kotlin code) inside the [simpleStrategy] builder using
+ * the provided SimpleAgen
+ *
+ * @property name The unique identifier for this agent.
+ * @param init Lambda that defines stages and nodes of this agent
+ */
+public fun <Input, Output> simpleStrategy(
```
ptitjes, line 28: I am afraid that putting this new DSL in the same package as the graph DSL will overload the namespace during code completion. That may be error-prone and difficult to understand when the user mis-uses one DSL inside the other.

What would you think about making it live in its own package? E.g. `ai.koog.agents.core.code.dsl.builder`, and then gradually move the current graph DSL to `ai.koog.agents.core.graph.dsl.builder` in a deprecation cycle.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/extension/SimpleStrategyDSL.kt
```diff
@@ -0,0 +1,404 @@
...
+    }
+}
+
+/**
+ * Executes the provided action if the given response is of type [Message.Assistant].
+ *
+ * @param response The response message to evaluate, which may or may not be of type [Message.Assistant].
+ * @param action A lambda function to execute if the response is an instance of [Message.Assistant].
+ */
+public inline fun SimpleAIAgentStrategyContext.onAssistantMessage(
```
ptitjes, line 48: I am not sure that having `onXXX` combinators in this DSL is the right move. I think we should embrace the "code-style" here and allow full use of Kotlin control flow language constructs.

For instance, `onAssistantMessage` would prevent easy looping. In the example you gave initially in the moderation PR, you were forced to use a `while (true)` because of this.

Why not expose conditional operators like `Message.isAssistantMessage()` and similar?

ptitjes (commented):
I love that this DSL style will get into Koog! I know this is a draft, but I added some initial thoughts.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/extension/SimpleStrategyDSL.kt
ptitjes: For example:
```kotlin
        simpleStrategy("calculator") { input ->
            dumpToPrompt(input)

            var responses = requestLLM()
            dumpToPrompt(responses)

            while (responses.containsToolCalls()) {
                if (responses.latestTokenUsage() > 100500) compressHistory()

                val results = executeTools(responses.filterToolCalls())
                dumpToPrompt(results)
                
                responses = requestLLM()
                dumpToPrompt(responses)
            }

            responses.singleOrNull().asAssistantMessage { it.content }
        }
```

Notes on the example above:
- I adopted my favorite style that is to not let LLM requests modify the prompt implicitly, as I still believe this make for more composable operators. You know my point of view... :smile: You can ignore that!
- I considered that the max iteration checking should be done transparently in the LLM request operators. Those checks would throw a short-circuiting exception that would be caught in the strategy `execute` method. We could reuse the `AIAgentMaxNumberOfIterationsReachedException` btw.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
Ololoshechkin: Because this TStrategy actually goes to pipeline and other properties.
And pipeline determines what type of features can be installed to such agent.

Then, I also added a `TStrategy` param to `Feature` and now you can naturally determine what type of Agent each feature is applicable to.

For example, GraphTesting and Persistency is GRAPH-ONLY features for obvious reasons (they use nodes and edges heavily inside, and these features get access to `strategy: GraphAIAgentStrategy` in all interceptors).

At the same time, some other features, like `OpenTelemetry` are more generic (just they won't receive "node started" / "node finished" events... actually, just thought, maybe we should introduce inheritance here for AIAgentPipeline, too...).


Also, the plan is to add more types of strategies (not only "graph" and "code")

Ololoshechkin: But actually, that's just a rough draft right now.. Probably we can go with more inheritance here opposed to generics. Like, we should introduce `GraphBasedAIAgent<I,O>: AIAgent<I,O>`, `SimpleAIAgent: AIAgent<I, O>` and in the future -- others.

And to allow creating all agents with just `AIAgent` keyword with just different strategies in parameter -- we can make companion object `invoke()` overrides.

So yeah, I just added generics to make it compile and for the sake of experiment :)

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/entity/AIAgentStrategy.kt
Ololoshechkin: I was thinking of Code / CodeBased / NonGraph and others... but didn't come up with any good name for it :(

Ololoshechkin: Also Mannual , HandMade and Artisanal :)))))

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/builder/SimpleAIAgentStrategyBuilder.kt
Ololoshechkin: Yes, I'm actually doing that right now :)

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/extension/SimpleStrategyDSL.kt
Ololoshechkin: > I considered that the max iteration checking should be done transparently in the LLM request operators

I agree, that must be done automatically, not manually.

Ololoshechkin: And in general -- I also agree with the proposed approach without `onAssistantMessage`.

`containsToolCalls()` look much more elegant!

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
Ololoshechkin: Though, the usages all remained the same even now:
```
 val agent = AIAgent(
            mockLLMApi,
            OllamaModels.Meta.LLAMA_3_2,
            strategy = singleRunStrategy(ToolCalls.SEQUENTIAL),
            toolRegistry = testToolRegistry
) {
            install(EventHandler) {
                onToolCall { eventContext -> actualToolCalls += eventContext.toolArgs.toString() }
            }
}
```

or 

```
 val agent = AIAgent(
            mockLLMApi,
            OllamaModels.Meta.LLAMA_3_2,
            strategy = graphStrategy("my-strategy") { ...edge, node .... }
            toolRegistry = testToolRegistry
) {
            install(OpenTelementry) { ... }
            install(Persistency) { ... }
}
```

or 

```
 val agent = AIAgent(
            mockLLMApi,
            OllamaModels.Meta.LLAMA_3_2,
            strategy = simpleStrategy("my-strategy") {  while(true) { ... }  }
            toolRegistry = testToolRegistry
) {
            install(OpenTelementry) { ... }
}
```

Ololoshechkin: So unless you really open and read those type declarations or enable explicit mode -- nothing would change for you)

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/extension/SimpleStrategyDSL.kt
Ololoshechkin: Done :)

EugeneTheDev:
This PR is outdated, since the proposal was implemented in #560 . I'm closing it

PR #436, opened 2025-07-13
Add error handling tests for LLM clients

kpavlov:
- Add error handling tests for LLM clients (openai, anthropic)

#### Type of the change
- [ ] New feature
- [ ] Bug fix
- [ ] Documentation fix
- [x] Tests improvement


integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/llm/AbstractLLMClientErrorsTest.kt
```diff
@@ -0,0 +1,169 @@
+package ai.koog.integration.tests.llm
```
skarpovdev: The tests don't seem to be checking actual integration but are rather mock-based unit tests, so I would move them to `prompt/prompt-executor/prompt-executor-clients` or respective modules.
It's not a strong opinion though.

gradle/libs.versions.toml
```diff
@@ -24,6 +25,9 @@ opentelemetry = "1.51.0"
 aws-sdk-kotlin = "1.4.119"
 
 [libraries]
+aimocks-anthropic-jvm = { module = "me.kpavlov.aimocks:ai-mocks-anthropic-jvm", version.ref = "mokksy" }
+aimocks-core-jvm = { module = "me.kpavlov.aimocks:ai-mocks-core-jvm", version.ref = "mokksy" }
+aimocks-openai-jvm = { module = "me.kpavlov.aimocks:ai-mocks-openai-jvm", version.ref = "mokksy" }
```
skarpovdev: We try to limit the number of external dependencies in our code to a minimum. Any way we can either migrate the code to the koog repo or use another approach?

integration-tests/build.gradle.kts
```diff
@@ -32,6 +32,9 @@ kotlin {
                 implementation(libs.kotlinx.serialization.json)
                 implementation(libs.ktor.client.content.negotiation)
                 runtimeOnly(libs.slf4j.simple)
+                implementation(libs.aimocks.core.jvm)
+                implementation(libs.aimocks.openai.jvm)
+                implementation(libs.aimocks.anthropic.jvm)
```
skarpovdev: Let's sort them -> put before `runtimeOnly`.

skarpovdev (changes requested):

gradle/libs.versions.toml
kpavlov: There are a lot of stuff to migrate...
This is open source project, similar to Wiremok. Contributions are very welcome!

integration-tests/build.gradle.kts
kpavlov: Agree, but can we configure some autoformatter?

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/llm/AbstractLLMClientErrorsTest.kt
kpavlov: This was my original idea. This is definitely a LLMClient level test.  I had some problem with module dependencies and moved tests to integration tests for a while

integration-tests/build.gradle.kts
skarpovdev: I don't think there is one for such purposes :(

kpavlov:
> What about other clients (`GoogleLLMClient`, `BedrockLLMClient`, ...)? Are there any obstacles to cover them as well?

There is [Google Gemini mock](https://mokksy.dev/docs/ai-mocks/gemini/). I did not add a test in this PR.
Sometimes, OpenAI mock can be reused, if the api shape is similar.
I haven't implemented all possible llm mocks yet due to the limited capacity.
There is also possible to write a lower-level SSE response with [Mokksy](https://mokksy.dev/docs/mokksy/)

gradle/libs.versions.toml
skarpovdev: Have you considered using [ktor](https://ktor.io/docs/client-testing.html) libs instead? For example `io.ktor:ktor-client-mock` for the mock part?
We try to keep dependencies limited to libraries that have broad adoption and a strong track record. This helps ensure stability and community support. 

We could generate errors similar to how it's done now:

```
@JvmStatic
fun errors(): Stream<Arguments> = Stream.of(
    Arguments.of("Bad request", HttpStatusCode.BadRequest, IllegalStateException::class.java),
    Arguments.of("Unauthenticated", HttpStatusCode.Unauthorized, IllegalStateException::class.java),
    Arguments.of("Unauthorized", HttpStatusCode.Forbidden, IllegalStateException::class.java),
    Arguments.of("Not found", HttpStatusCode.NotFound, IllegalStateException::class.java),
    Arguments.of("Request entity too large", HttpStatusCode.PayloadTooLarge, IllegalStateException::class.java),
    Arguments.of("Too many requests", HttpStatusCode.TooManyRequests, IllegalStateException::class.java),
    Arguments.of(
        "Internal server error",
        HttpStatusCode.InternalServerError,
        IllegalStateException::class.java
    ),
    Arguments.of("Service unavailable", HttpStatusCode.ServiceUnavailable, IllegalStateException::class.java)
)
```

the test would look something like:

```
@ParameterizedTest(name = "{0}")
@MethodSource("errors")
fun testErrorHandling(
    description: String,
    statusCode: HttpStatusCode,
    expectedExceptionType: Class<out Throwable>
) = runTest {
    val mockEngine = MockEngine { _ ->
        respondError(
            status = statusCode,
            headers = headersOf(HttpHeaders.ContentType, ContentType.Application.Json.toString()),
            content = """{"error": {"message": "Error message for $description", "type": "api_error"}}"""
        )
    }

    val client = AnthropicLLMClient(
        apiKey = "test-api-key",
        settings = AnthropicClientSettings(),
        baseClient = HttpClient(mockEngine),
        clock = Clock.System
    )

    try {
        client.execute(createTestPrompt(), createTestModel(), emptyList())
        fail("Expected ${expectedExceptionType.simpleName} to be thrown")
    } catch (e: Exception) {
        assertTrue(
            expectedExceptionType.isInstance(e),
            "Expected exception of type ${expectedExceptionType.simpleName} but got ${e.javaClass.simpleName}"
        )
        assertTrue(
            e.message?.contains(statusCode.toString()) == true,
            "Exception message should contain status code: ${e.message}"
        )
    }
    }
```
(most of it can be in an abstract class, again similar to how it's done now)

skarpovdev:
> Do you think error handling looks right? IllegalStateException seems to be too generic

I agree it's quite generic right now and could be improved! 💯

aozherelyeva:
As discussed in DM, it's better to transfer these tests into the corresponding executor modules (and to transfer the helper class into agents-test).

aozherelyeva (changes requested):
As discussed in DM, it's better to transfer these tests into the corresponding executor modules (and to transfer the helper class into agents-test).

kpavlov:
Folks, if there are plans to upgrade Kotlin to 2.2.0. It has helped to resolve the compilation issue
cc: @Ololoshechkin

gradle/libs.versions.toml
kpavlov: @skarpovdev, I like how it looks and the idea of using MockEngine is great!
However, handling error scenarios is relatively straightforward, as the input often doesn't significantly impact the outcome. Positive paths, however, require asserting specific request payloads—essentially leading to a partial reimplementation of [WireMock](https://wiremock.org/). That’s the route I had to take. Are you considering going down that path again yourself?

aozherelyeva (commented):
@kpavlov I'd wait for other reviews, but I have some doubts regarding adding `mokksy/ai-mocks` dependency: as it's an open source project, I see you're the only one contributing to it now. 
Although it looks good, there are some doubts from the team when it comes to adding new dependencies to the framework.

devcrocod (commented):
When using mokksy, I see a risk in that the models always need to be kept up to date. Also, the library has a rather narrow scope of application. it’s mainly suitable for those implementing clients for AI providers in Kotlin, such as koog, langchain4j, or specific sdk like anthropic-kotlin-sdk.

I would consider using this library, as long as the potential overhead of abandoning it later is taken into account

PR #438, opened 2025-07-14
[KG-134] add LLModel contextLength and maxOutputTokens

ptitjes:
This PR adds two new attributes to LLModel. cf. [KG-134](https://youtrack.jetbrains.com/issue/KG-134/LLModel-contextLength)

---

#### Type of the change
- [x] New feature
- [ ] Bug fix
- [ ] Documentation fix
- [ ] Tests improvement


Ololoshechkin (approved):
Thank you so much for working on this!

Just a few questions:
1. Why contextLength is nullable? Are there any models that don’t have it?
2. Why some Ollama models (like LLAMA_3_GROK_TOOL_USE_8B and others) have contextLength but do not have maxOutputTokens ? I would assume they should have some output tokens limit?

ptitjes:
> 1. Why contextLength is nullable? Are there any models that don’t have it?

Since the initial push, I found the context length for some that I was missing:

- The Anthropic Claude Instant model in the BedrockModels
- The Ollama embedding models

I will add them soon.

But there are still some models for which I wasn't able to find the context length:

- The OpenAI moderation models
- The OpenAI embedding models

I might need some help with those.

Now, there is the problem of the models pulled directly from the Ollama registry. I defined the `OllamaModelCard.contextLength` as nullable because it is not clear from the go code if it always present. As I never stumbled upon a model pulled from Ollama without a context length, I think I will just throw an error if it is not present. Tell me what you think.

If we resolve those two problems (missing OpenAI context length and the Ollama pulled models context length possibly null), then I will be able to make it non-nullable.

> 2. Why some Ollama models (like LLAMA_3_GROK_TOOL_USE_8B and others) have contextLength but do not have maxOutputTokens ? I would assume they should have some output tokens limit?

Now, that is a whole other story. From my understanding of LLMs, there is nothing that mandates having a max generated output tokens. I believe this is just something done by APIs to limit the output, maybe as some form of security.

I mean, the context length is the size of the context window, and there is nothing preventing to roll that window for ever.

So, for example, there is no max output tokens for any of the Ollama models.

I propose we let that field optional.

Ololoshechkin:
@ptitjes ,

https://cookbook.openai.com/examples/embedding_long_inputs

> The text-embedding-3-small model has a context length of 8191 tokens with the cl100k_base encoding, and we can see that going over that limit causes an error.


As for the 3-large, I couldn't find the official answer, but on 3rd party resources they say:

> 3-large embeds up to 3072 dimensions (compared to ada's 1536) and is about 40% more powerful.
3-large context window is 8191 tokens, same as ada. Approximately 6000 words.


Ao basically it's also 8191

Ololoshechkin:
For models defined in tests please feel free to set everything to some random number I guess, like 1000, wdyt?

Ololoshechkin:
As for the Ollama models -- looks like indeed for 3rd party models it's not required metadata fields (of course Meta or Mistral would include it, but some 3rd party providers might omit this field).

And then it's recommended to either fallback to some default value (2048 /4096) or find the actual base model it derives from....

So we either have to stay with nullable , or we can actually introduce a parameter for the client that would be either a single fallback size, or even a map LLModel -> Int (safer option).

Then we can throw exception in runtime if it's not provided by API and not even manually defined in this fallback map -- ans users would go and define some value consciously.

I still think it would be rather a rare case for some very specific models

WDYT?

ptitjes:
> I still think it would be rather a rare case for some very specific models

Yeah, I agree it would be rare.

I don't having a map will be a good idea, for maintainance. Also, having a default (map) value from constructor might be confusing.

I would prefer that we have a default constant value. We add a warning log line indicating that we fallback to that value for model XYZ. And if ever the user needs to override it, they can always data-copy the LLModel to override it. Would that be OK?

kpavlov:
Let's not forget adding an Open Telemetry span attribute [gen_ai.request.max_tokens](https://opentelemetry.io/docs/specs/semconv/registry/attributes/gen-ai/#gen-ai-request-max-tokens). Could be a separate PR

kpavlov:
Let's not forget to add OpenTelemetry span attribute [gen_ai.request.max_tokens](https://opentelemetry.io/docs/specs/semconv/registry/attributes/gen-ai/#gen-ai-request-max-tokens)

prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/build.gradle.kts
```diff
@@ -15,6 +15,7 @@ kotlin {
                 api(project(":agents:agents-tools"))
                 api(project(":agents:agents-utils"))
                 api(project(":prompt:prompt-executor:prompt-executor-clients"))
+                implementation(project(":prompt:prompt-executor:prompt-executor-clients:prompt-executor-anthropic-client"))
```
EugeneTheDev, line 18: Why did you add it to `commonMain`? It should be not needed here since we provide only JVM implementation for Bedrock, and this dependency is already included in the `jvmMain` block.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-openrouter-client/build.gradle.kts
```diff
@@ -15,6 +15,8 @@ kotlin {
                 api(project(":agents:agents-tools"))
                 api(project(":agents:agents-utils"))
                 api(project(":prompt:prompt-executor:prompt-executor-clients"))
+                implementation(project(":prompt:prompt-executor:prompt-executor-clients:prompt-executor-anthropic-client"))
```
EugeneTheDev, line 18: Why did you add dependencies on Anthropic and Google clients? It should be independent from these clients

EugeneTheDev (changes requested):
Looks good, but I have questions about dependencies

prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/build.gradle.kts
ptitjes: I answered for both your comments in the main thread, because this is the same answer.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-openrouter-client/build.gradle.kts
ptitjes: I answered for both your comments in the main thread, because this is the same answer.

PR #378, opened 2025-07-01
Add support for excluding properties in JSON schema generation

nathanfallet:
Fixes #377 

Proposed changes:
Add an `excludedProperties` parameter like this:
```kotlin
simpleSchemaGenerator.generate(
    "TestClass",
    serializer<TestClass>(),
    excludedProperties = setOf("TestClass.someProperty")
)
```

---

#### Type of the change
- [x] New feature
- [ ] Bug fix
- [ ] Documentation fix

#### Checklist for all pull requests
- [x] The pull request has a description of the proposed change
- [x] I read the [Contributing Guidelines](https://github.com/JetBrains/koog/blob/main/CONTRIBUTING.md) before opening the pull request
- [x] The pull request uses **`develop`** as the base branch
- [x] Tests for the changes have been added
- [x] All new and existing tests passed

##### Additional steps for pull requests adding a new feature
- [x] An issue describing the proposed change exists
- [x] The pull request includes a link to the issue
- [ ] The change was discussed and approved in the issue
- [ ] Docs have been added / updated

Issue #377, opened 2025-07-01
JSON Schema: Exclude properties


nathanfallet:
In some case, we might want to exclude properties when generating a schema.

Here are the example we encountered:
- Our app allows to analyze images and produce a structured description. The user can choose which fields are generated by AI in a long list of properties. The easiest way to implement this is having a data class with optional fields, and dynamically exclude the fields we don't want from the schema. We only query the desired fields.
- Our AI needs to complete missing data from a product. We provide all the data we know about the product, and ask the AI to provide what's missing. The same way as before, we make a data class with optional fields, and exclude fields we already know, so AI only returns missing data.

Proposed changes:
Add an `excludedProperties` parameter like this:
```kotlin
simpleSchemaGenerator.generate(
    "TestClass",
    serializer<TestClass>(),
    excludedProperties = setOf("TestClass.someProperty")
)
```

nathanfallet:
@aozherelyeva Excited to open #638 🥳
This is the last issue we needed to fix for our project, and the new PR is ready, up to date with the new schema generator refactor 😎

PR #443, opened 2025-07-14
Refactor structured output API to make it more flexible, support native structured output for OpenAI and Google.

### LINKED ISSUE #366: Json Schema / Response Schema Support
Hey guys, first of all I want to say - thanks for this great framework, showing great potential and absolutely leveraging all of Kotlin's best features.

---
After my initial exploration of Koog, I noticed that Structured Outputs are being "injected" into the prompt and not provided as a `json_schema response format. See [OpenAI's Structured Output Introduction](https://platform.openai.com/docs/guides/structured-outputs#introduction) and more providers support.

Libraries like `LangChain4j` allow you to essentially give these providers the `response_format`. As far as I see, this isn't the current approach in Koog which seems to be rather error-prone for my understanding? (Thus, the rising need of the `fixingModel` technique in `PromptExecutor.executeStructured`).

---
Is there any specific reason for **not** sending the `response_format` for models that **do** support it? e.g. in via Koog's `OpenAIRequest `? See ([OpenAI response_schema reference](https://platform.openai.com/docs/api-reference/chat/create#chat-create-response_format)). I can only assume it would remove the need for `fixingModel` and similar techniques, allowing a reliable `type-safe` response for models that indeed support it.

Additional info provided by @bobazooba in the Slack channel:
- [OpenAI - Introducing Structured Outputs in the API](https://openai.com/index/introducing-structured-outputs-in-the-api/)

EugeneTheDev:
* Rewrite JSON schema generation from serializable classes to make it generally more extensible and flexible. This allows supporting different JSON schema formats for different LLM providers while still reusing common generation logic.
* Update OpenAI and Google clients to support native structured output.
* Refactor and improve structured output API - `PrompExecutor.executeStructured`, `AIAgentLLMSession.requestLLMStructured`, `nodeLLMRequestStructured` - to make it easier to use with different modes (manual and native)

Fixes #366


nathanfallet:
When this will be merged, it should unlock #377 and its PR #378 (just linking so I can track it and find it later)

EugeneTheDev:
I've added simple method overloads to the structured output API, allowing users to avoid always configuring all parameters manually and to rely on the framework to figure out the best approach instead. It is possible now to write very little code to get the structure in common cases where no fine-grained control is required:
```kotlin
// Prompt executor
promptExecutor.executeStructured<MyStruct>()

// Agent LLM contenxt
llm.writeSession {
  requestLLMStructured<MyStruct>()
}

// Node
val getMyStruct by nodeLLMRequestStructured<MyStruct>()
```

Full example is in the `StructuredOutputExample.kt`. 

@devcrocod can you please check, is it better now?

OfekTeken:
This is perfect, I just have a small suggestion - In my experience, most of the times I'd want to also give a specific system message when generating structured output (instead of relying solely on the history)
Do you think it'd be possible that these methods allow such optional systemMessage parameter to be passed? Or an alternative method, etc..
I think it would be pretty useful, thanks! 👍

prompt/prompt-structure/src/commonMain/kotlin/ai/koog/prompt/structure/PromptExecutorExtensions.kt
```diff
@@ -0,0 +1,271 @@
...
+ * require different schema formats. E.g. for [JsonStructuredData] this means you have to use the appropriate
+ * [JsonSchemaGenerator] implementation for each provider for [StructuredOutput.Native], or fallback to [StructuredOutput.Manual]
+ *
+ * @property fixingParser Optional parser that handles malformed responses by using an auxiliary LLM to
+ * intelligently fix parsing errors. When specified, parsing errors trigger additional
+ * LLM calls with error context to attempt correction of the structure format.
+ */
+public data class StructuredOutputConfig<T>(
+    public val default: StructuredOutput<T>? = null,
+    public val byProvider: Map<LLMProvider, StructuredOutput<T>> = emptyMap(),
```
devcrocod, line 50: Why is it LLMProvider here instead of LLModel? 
As I see, in `PromptExecutor`, where the config will be used, all providers are already present

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/entity/AIAgentSubgraph.kt
```diff
@@ -99,11 +102,15 @@ public open class AIAgentSubgraph<Input, Output>(
             }
 
             val selectedTools = this.requestLLMStructured(
-                structure = JsonStructuredData.createJsonStructure<SelectedTools>(
-                    schemaFormat = JsonSchemaGenerator.SchemaFormat.JsonSchema,
-                    examples = listOf(SelectedTools(listOf()), SelectedTools(tools.map { it.name }.take(3))),
-                ),
-                retries = toolSelectionStrategy.maxRetries,
+                config = StructuredOutputConfig(
```
devcrocod, line 105: In my opinion, this configuration is a bit verbose.
Is there any way to improve it?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/entity/AIAgentSubgraph.kt
```diff
@@ -99,11 +102,15 @@ public open class AIAgentSubgraph<Input, Output>(
...
+                    default = StructuredOutput.Manual(
+                        JsonStructuredData.createJsonStructure<SelectedTools>(
+                            schemaGenerator = FullJsonSchemaGenerator,
```
devcrocod: As I see it, there’s only one generator, so why are we choosing it here?
We could just use `FullJsonSchemaGenerator` by default, and if users want to use a custom one, they can specify it themselves

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/entity/AIAgentSubgraph.kt
```diff
@@ -99,11 +102,15 @@ public open class AIAgentSubgraph<Input, Output>(
...
+                            examples = listOf(SelectedTools(listOf()), SelectedTools(tools.map { it.name }.take(3))),
```
devcrocod, line 109: It feels like there are too many nested levels 
**list -> SelectedTools -> list**
Is there a way to simplify this?

And what if I want to provide an example manually, in md text?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/session/AIAgentLLMSession.kt
```diff
@@ -49,9 +50,7 @@ public sealed class AIAgentLLMSession(
      * Typical usage includes providing input to LLM requests, such as:
      * - [requestLLMWithoutTools]
      * - [requestLLM]
-     * - [requestLLMMultiple]
-     * - [requestLLMStructured]
-     * - [requestLLMStructuredOneShot]
+     * etc.
```
devcrocod, line 53: Why did you skip other requests?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/session/AIAgentLLMSession.kt
```diff
@@ -240,32 +239,81 @@ public sealed class AIAgentLLMSession(
...
+     *
+     * @param T The structure to request.
+     * @param examples Optional list of examples in case manual mode will be used. These examples might help the model to
+     * understand the format better.
+     * @param fixingParser Optional parser that handles malformed responses by using an auxiliary LLM to
+     * intelligently fix parsing errors. When specified, parsing errors trigger additional
+     * LLM calls with error context to attempt correction of the structure format.
+    */
+    public suspend inline fun <reified T> requestLLMStructured(
+        examples: List<T> = emptyList(),
```
devcrocod, line 307: I don’t quite understand how, in this case, the LLM can help with a list of instances of T.
Could you please explain how this works, using an example?

devcrocod: And how are they wrapped in the LLM request?
Where exactly in the prompt are they inserted, and in what format?

build.gradle.kts
```diff
@@ -229,3 +231,28 @@ kover {
     }
 
 }
+
+fun Project.getKotlinCompileTasks(sourceSetName: String): List<Task> {
```
devcrocod, line 242: I don’t quite understand the purpose of these changes

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/DataModel.kt
```diff
@@ -226,28 +226,28 @@ internal class GoogleFunctionDeclaration(
...
- * @property numberOfChoices The number of reply choices to generate.
+ * @property candidateCount The number of reply choices to generate.
  * @property topP The maximum cumulative probability of tokens to consider when sampling.
  * @property topK The maximum number of tokens to consider when sampling.
  */
 @Serializable
 internal class GoogleGenerationConfig(
     val responseMimeType: String? = null,
     val responseSchema: JsonObject? = null,
+    val responseJsonSchema: JsonObject? = null,
```
devcrocod, line 246: There might be issues here, since `responseJsonSchema` is a [union Value](https://protobuf.dev/reference/protobuf/google.protobuf/#value), which means it can be a primitive

prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openai/DataModel.kt
```diff
@@ -232,6 +243,23 @@ internal data class OpenAIAudio(
     val transcript: String? = null,
 )
 
+@Serializable(with = OpenAIResponseFormatSerializer::class)
```
devcrocod, line 261: Why is a custom serializer needed here?  can do without it

Also, the text type is missing here

prompt/prompt-structure/src/commonMain/kotlin/ai/koog/prompt/structure/PromptExecutorExtensions.kt
```diff
@@ -0,0 +1,271 @@
...
+        )
+    }
+}
+
+/**
+ * Default mapping of some providers to their respective known simple JSON schema format generators.
+ * Used to attempt to get a proper generator in the simple version of [executeStructured] (that does not accept [StructuredOutput] explicitly)
+ * to attempt to generate an appropriate schema for the passed [KType].
+ */
+public val DefaultSimpleSchemaGenerators: Map<LLMProvider, JsonSchemaGenerator> = mapOf(
```
devcrocod: In my opinion, it’s not a good idea to use such global variables.

Can we put jsonSchemaGenerator into the companion object for OpenAI and Google respectively?

prompt/prompt-structure/src/commonMain/kotlin/ai/koog/prompt/structure/StructureFixingParser.kt
```diff
@@ -0,0 +1,145 @@
...
+        /**
+         * Default prompt explaining how to fix a malformed structure.
+         */
+        public fun defaultFixingPrompt(
+            builder: PromptBuilder,
+            content: String,
+            structure: StructuredData<*, *>,
+            exception: SerializationException
+        ): PromptBuilder = builder.apply {
+            system {
```
devcrocod, line 106: As I understand it, the user won’t be able to change this prompt?
Where is this prompt taken from?

prompt/prompt-structure/src/commonMain/kotlin/ai/koog/prompt/structure/json/generator/core/GenericJsonSchemaGenerator.kt
```diff
@@ -0,0 +1,194 @@
...
+import kotlinx.serialization.json.*
+
+/**
+ * Generic extensions of [JsonSchemaGenerator] that provides some common base implementations of visit methods.
+ * This class can be used as a base to implement custom generators that share generic schema generation logic.
+ *
+ * Note: it does not handle nullability because these might be different in different schema specs.
+ * Implementations must handle these themselves.
+ */
+public abstract class GenericJsonSchemaGenerator : JsonSchemaGenerator() {
```
devcrocod: Will this work if I have a custom serializer and the descriptor is overridden?

prompt/prompt-structure/src/commonMain/kotlin/ai/koog/prompt/structure/json/generator/core/JsonSchemaGenerator.kt
```diff
@@ -0,0 +1,179 @@
...
+     *
+     * This method processes the provided serial descriptor based on its structural or type-specific characteristics
+     * to produce a JSON object representing the schema fragment. Usually, it matches on [SerialDescriptor.kind] and
+     * invokes a suitable concrete visit implementation from the methods below.
+     *
+     * @return a JSON object representing the schema definition for the given descriptor.
+     */
+    protected abstract fun process(context: GenerationContext): JsonObject
+
+    protected abstract fun processString(context: GenerationContext): JsonObject
```
devcrocod: Why do all the methods return `JsonObject` instead of `JsonElement`?
Should string, boolean, integer, and number really return an object?

prompt/prompt-structure/src/commonMain/kotlin/ai/koog/prompt/structure/json/generator/core/SimpleJsonSchemaGenerator.kt
```diff
@@ -0,0 +1,105 @@
...
+
+/**
+ * Simple implementation of [GenericJsonSchemaGenerator] that produces [LLMParams.Schema.JSON.Simple].
+ * Generates LLM-agnostic schema not tied to any specific provider format.
+ * For LLMs requiring custom formats when using native structured output, consider provider-specific generators instead.
+ *
+ * It is recommended to use [SimpleJsonSchemaGenerator.Default] companion object, which provides a default instance,
+ * instead of creating new instances manually.
+ */
+public open class SimpleJsonSchemaGenerator : GenericJsonSchemaGenerator() {
```
devcrocod: The use of Simple/Full in the naming raises questions, since it’s hard for the user to understand what `Simple` means and why it’s considered simple without additional context

devcrocod (commented):
I have some questions, mostly related to the design. I’d appreciate it if you could clarify them for me

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/entity/AIAgentSubgraph.kt
EugeneTheDev: I thought about it for some time, and I don't see a way to make it less verbose without loosing the flexiblity. After all, this is a purpose of this "advanced" structured output API, to allow controlling different aspects, such as choosing structure generator, choosing structured output strategy (manual/native) per provider and default. We offer a simple API, where you are not required to provide all these verbose parameters and just a class (via reified parameter) is enough. I want to leave it here explicitly like this, since this is part of the core implementation and I want to make it more visible, what is going on. Also, we are using something custom here which is different from simple API - manual mode everywhere. I wanted to leave it like this at least for now, in order not to break accidentaly working approach to selecting tools. But later I think we can indeed replace it just with a simple API.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/entity/AIAgentSubgraph.kt
EugeneTheDev: There are multiple types of generators, and I'm again (as I explained above) chose to be a bit more explicit here for the sake of clarity, to make it easier to understand what is going on here exactly, since it's part of the core implementation. But now that I made `Standard` (ex `Full`) JSON schema generator a default value, this can be removed if you want.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/entity/AIAgentSubgraph.kt
EugeneTheDev: This part is unchanged, it was like this before, because this is what the structure looks like. We are providing a list of examples, and each value in the list has a list as a parameter type too (the list of selected tools, to be more precise). That's why this "list iniside a list" look, but these are actually different things.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/session/AIAgentLLMSession.kt
EugeneTheDev: I didn't want to mention all existing methods here because it felt redundant. I left only a couple just to give a general sense which kind of methods is implied here.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/session/AIAgentLLMSession.kt
EugeneTheDev: Sure. [This is the prompt](https://github.com/JetBrains/koog/blob/3d2962ffceb498746abc78d25596f7b237421728/prompt/prompt-structure/src/commonMain/kotlin/ai/koog/prompt/structure/StructuredPrompts.kt#L15) added as a user message when the manual mode is used. And [this is the prompt](https://github.com/JetBrains/koog/blob/a749563ae1558b210ff2c537adcdbcd35ef15db3/prompt/prompt-structure/src/commonMain/kotlin/ai/koog/prompt/structure/json/JsonStructuredData.kt#L66) with the default JSON structure definition (users can provide custom promp if they want). [Here's how](https://github.com/JetBrains/koog/blob/a749563ae1558b210ff2c537adcdbcd35ef15db3/prompt/prompt-structure/src/commonMain/kotlin/ai/koog/prompt/structure/json/JsonStructuredData.kt#L81) examples are handled in this prompt.

build.gradle.kts
EugeneTheDev: I wanted a single task to check that the project compiles for all targets (JVM, JS, WASM JS, others we may support in the future). Unfortunately, no such task exists out of the box, you have to manually run each one (`compileKotlinJvm`, `compileKotlinJs`, etc). That's why I wrote two small tasks myself. They basically fetch all Kotlin compile tasks for the given source set (main/test) for all subprojects and then run these. Just to save some time during development.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/DataModel.kt
EugeneTheDev: Oh, thank you, I rechecked [the section about JSON schema](https://ai.google.dev/gemini-api/docs/structured-output#json-schema) in documentation, and apparently this is indeed the case. AFAIU you can provide just some primitive type, e.g. `string` or `int` to get it in response. But JSON schema in our `LLMParams` must always be an instance of `JsonObject`, [link](https://github.com/JetBrains/koog/blob/a749563ae1558b210ff2c537adcdbcd35ef15db3/prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/params/LLMParams.kt#L77), so such a case is not possible when using our API. Do you think we should make `schema: JsonElement` instead, to support such cases? TBH this seems like an edge case which would not be very popular and can be easily emulated using a class with a single parameter. WDYT?

prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openai/DataModel.kt
EugeneTheDev: I was inspired by the [GooglePartSerializer](https://github.com/JetBrains/koog/blob/f4f2055769224485960ddf08167df7b6da1d1e64/prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/DataModel.kt#L384), because the situation seems similar here: we have an object in `response_format` with potentially different values for `type`. Since there's a clear `type` property present, it could've been solved using class discriminators, however, we've [disabled these for OpenAI client](https://github.com/JetBrains/koog/blob/a749563ae1558b210ff2c537adcdbcd35ef15db3/prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openai/OpenAILLMClient.kt#L100). And in the current implementation, yes, since there's only a single child for now, [JsonSchema](https://github.com/JetBrains/koog/blob/f4f2055769224485960ddf08167df7b6da1d1e64/prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openai/DataModel.kt#L251), it could've been just a simple class, without inheiritance. I made it like this to potentially support more output formats in the future. Do you think I should make it just a simple class instead? Also, which text type are you talking about? There's no text type for this field.

prompt/prompt-structure/src/commonMain/kotlin/ai/koog/prompt/structure/StructureFixingParser.kt
EugeneTheDev: No, I specifically modified the original implementation to support passing a custom prompt, [fixingPrompt](https://github.com/JetBrains/koog/blob/f4f2055769224485960ddf08167df7b6da1d1e64/prompt/prompt-structure/src/commonMain/kotlin/ai/koog/prompt/structure/StructureFixingParser.kt#L26)

prompt/prompt-structure/src/commonMain/kotlin/ai/koog/prompt/structure/json/generator/core/GenericJsonSchemaGenerator.kt
EugeneTheDev: `generate` method accepts an instance of `Json` that is used to (de)serialize objects, so yes. Everything this json instance knows about (registered subtypes, custom serializers, etc.) should be used properly by the generator.

prompt/prompt-structure/src/commonMain/kotlin/ai/koog/prompt/structure/json/generator/core/JsonSchemaGenerator.kt
EugeneTheDev: Yes, they should, because they return a certain part for the JSON schema definition. And these parts are always json objects even for the primitive types, e.g.:
```json
{ "type": "string" }
```

They can have additional fields too, such as `description` or `nullable`.

prompt/prompt-structure/src/commonMain/kotlin/ai/koog/prompt/structure/json/generator/core/SimpleJsonSchemaGenerator.kt
EugeneTheDev: After some considertation, I decided to rename these as follows:
1. `Simple` -> `Basic`. "Basic" implementation of JSON schema, without additional capabilities such as special functions, nested objects only.
2. `Full` -> `Standard`. "Standard" implementation of JSON schema, with support for object definitions, refs, and certain special functions like `oneOf`/`anyOf`.

Is it better now?

prompt/prompt-structure/src/commonMain/kotlin/ai/koog/prompt/structure/PromptExecutorExtensions.kt
EugeneTheDev: Hm, by using the model as a key we can specify the behavior more granularly, e.g. that Gemini Flash 2.5 Lite should use `Native` output, but Gemini Flash 2.5 Pro should use `Manual`. But I'm not sure if this is a realistic case though. The main idea is that since providers might have differences in JSON schema formats and Natve/No Native support, specifiying the behavior per provider only would be enough. I would say we can leave it like this for now and intrduce per model mapping if there will be a real need for it, wdyt?

prompt/prompt-structure/src/commonMain/kotlin/ai/koog/prompt/structure/PromptExecutorExtensions.kt
EugeneTheDev: After some consideration I decided to use "register pattern" (similar to ServiceLoader) to be able to move LLM provider specific JSON schema generators to their clients' respective modules. So now these LLM clients register these generators (formats) in global maps. This is not an ideal solution, but seems like it achieves nice balance in terms of stability/convenience. I marked these global maps with internal API annotations.

build.gradle.kts
devcrocod: Okay, maybe I’m missing something. Why do we need to check this?

Also, maybe it’s better to add this to https://github.com/JetBrains/koog/blob/develop/buildSrc/src/main/kotlin/ai.kotlin.multiplatform.gradle.kts

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/DataModel.kt
devcrocod: If our API only allows adding `JsonObject` and nothing else, then it’s fine to leave it as is.
If we ever need to support `JsonArray` in the future, we can easily switch to JsonElement without breaking compatibility

prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openai/DataModel.kt
devcrocod: As I see it, [`response_format`](https://platform.openai.com/docs/api-reference/chat/create) is part of the request for OpenAI, so it only needs to be serialized correctly. kotlinx-serialization allows us to do that without a custom serializer

GooglePart, on the other hand, appears in the response from Gemini. That’s why GooglePartSerializer was added. although in theory, `@SerialName` should cover this use case as well

Also, in OpenAI’s documentation, I see three types: `text`, `json_schema` and `json_object`

prompt/prompt-structure/src/commonMain/kotlin/ai/koog/prompt/structure/json/generator/core/SimpleJsonSchemaGenerator.kt
devcrocod: Yes, but that’s just my personal preference :)

I’m curious how we can fill the entire JSON schema for a class? I mean all fields of json schema

Rizzen (approved):

nathanfallet:
Thank you! I'll have a look and work back on #378 soon (if it's still applicable; maybe it's no longer needed, i'll have to inspect the new way)

OfekTeken:
Thanks! Can't wait to use this :)

PR #459, opened 2025-07-18
Add the feedback mechanism for the retry component

antoniibelyshev:

---

[KG-188](https://youtrack.jetbrains.com/issue/KG-188)

Allow to provide condition description and feedback in the retry component.

#### Type of the change
- [ ] New feature
- [x] Bug fix
- [ ] Documentation fix
- [ ] Tests improvement

### DIFF
--- a/agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/SubgraphWithRetry.kt
+++ b/agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/SubgraphWithRetry.kt
@@ -35,8 +35,8 @@ public data class RetrySubgraphResult<Output>(
  */
 public sealed interface ConditionResult {
     /**
-     * Indicates whether the current instance of `ConditionResult` represents a successful state.
-     * Returns `true` if the instance is of type `Approve`, signifying success.
+     * Indicates whether the current instance of [ConditionResult] represents a successful state.
+     * Returns `true` if the instance is of type [Approve], signifying success.
      * Otherwise, returns `false`.
      */
     public val isSuccess: Boolean get() = this is Approve
@@ -53,7 +53,7 @@ public sealed interface ConditionResult {
     /**
      * Represents a condition result indicating rejection without additional feedback.
      *
-     * This implementation of the `ConditionResult` sealed interface signifies that a specific condition
+     * This implementation of the [ConditionResult] sealed interface signifies that a specific condition
      * has been evaluated and rejected within a system or workflow. Unlike other implementations,
      * this result does not include any accompanying feedback or explanation for the rejection.
      *
@@ -64,7 +64,7 @@ public sealed interface ConditionResult {
     /**
      * Represents a condition result that indicates rejection combined with specific feedback.
      *
-     * This implementation of the `ConditionResult` sealed interface signifies that a condition
+     * This implementation of the [ConditionResult] sealed interface signifies that a condition
      * has been evaluated and rejected within a system or workflow. Unlike a standard rejection,
      * this result includes accompanying feedback to provide additional context or guidance
      * regarding the rejection.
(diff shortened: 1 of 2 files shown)


agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/SubgraphWithRetry.kt
```diff
@@ -24,6 +24,66 @@ public data class RetrySubgraphResult<Output>(
...
+ * This sealed interface allows for expressing various outcomes of a condition check.
+ *
+ * Implementations:
+ * - [Approve]: Indicates that the condition was approved.
+ * - [Reject]: Indicates that the condition was rejected without further feedback.
+ * - [RejectWithFeedback]: Indicates that the condition was rejected with additional feedback.
+ */
+public sealed interface ConditionResult {
+    /**
+     * Indicates whether the current instance of `ConditionResult` represents a successful state.
```
EugeneTheDev: nit: When mentioning symbols, please use square brackets instead of backticks: `[ConditionResult]`. This would create a proper reference when rendering docs, making it slightly easier to navigate.

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/SubgraphWithRetry.kt
```diff
@@ -34,26 +94,61 @@ public data class RetrySubgraphResult<Output>(
  * @param name The optional name of the subgraph.
  * @param defineAction A lambda defining the action subgraph to perform within the retry subgraph.
  */
+@Deprecated("Use `condition: suspend AIAgentContextBase.(Output) -> ConditionResult` instead")
```
EugeneTheDev: Let's remove instead of deprecating it. It's fine since we are not stable yet.

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/SubgraphWithRetry.kt
```diff
@@ -124,6 +237,42 @@ public inline fun <reified Input : Any, reified Output> AIAgentSubgraphBuilderBa
  * @param strict If true, an exception is thrown if the condition is not met after the maximum retries.
  * @param name An optional name for the subgraph.
  * @param defineAction A lambda defining the actions within the subgraph.
+ */
+@Deprecated("Use `condition: suspend AIAgentContextBase.(Output) -> ConditionResult` instead")
```
EugeneTheDev: Here as well, let's remove

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/SubgraphWithRetry.kt
```diff
@@ -24,6 +24,66 @@ public data class RetrySubgraphResult<Output>(
...
+     * Represents a condition result that indicates rejection combined with specific feedback.
+     *
+     * This implementation of the `ConditionResult` sealed interface signifies that a condition
+     * has been evaluated and rejected within a system or workflow. Unlike a standard rejection,
+     * this result includes accompanying feedback to provide additional context or guidance
+     * regarding the rejection.
+     *
+     * @property feedback A descriptive message or information explaining the reason for the rejection.
+     */
+    public class RejectWithFeedback(public val feedback: String): ConditionResult
```
EugeneTheDev: I wonder if we need two separate `Reject` classes. Can we simply add nullable `feedback` field to the `Reject` class with default `null` value?

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/SubgraphWithRetry.kt
```diff
@@ -75,11 +180,18 @@ public inline fun <reified Input : Any, reified Output> AIAgentSubgraphBuilderBa
 
         val decide by node<Output, RetrySubgraphResult<Output>> { output ->
             val retries = storage.getValue(retriesKey)
-            val success = condition(output)
+
+            // fork the context before applying the condition
+            // so that user can update prompt and call llm
+            // to determine if the condition is satisfied
+            val conditionResult = fork().condition(output)
```
EugeneTheDev, line 138: Why are you doing `fork` here? There are no parallel executions, so we can just do `this.condition`, right?

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/SubgraphWithRetry.kt
```diff
@@ -64,6 +159,16 @@ public inline fun <reified Input : Any, reified Output> AIAgentSubgraphBuilderBa
             // Increment retries
             storage.set(retriesKey, retries + 1)
 
+            // Add clarification messages to the prompt
+            val clarificationMessages = storage.getValue(clarificationMessagesKey)
+            llm.writeSession {
+                updatePrompt {
```
EugeneTheDev: So on each iteration we are adding all existing clarification messages again? I wonder if this would be an optimal approach, I feel like we would waste tokens this way. How about reworking this approach a bit? We store only the latest clarification message in the storage (not list) and append only it if it's present. All previous messages would be already present in the prompt anyway. And to implement this properly, after every iteration `calrificationMessageKey` in the storage has to be explicitly cleared if there's no actual clarification message after this iteration, to avoid setting message from the previous iteration twice.

EugeneTheDev (changes requested):

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/SubgraphWithRetry.kt
antoniibelyshev: One of the intended use cases is to use [llm-as-a-judje](https://jetbrains.slack.com/archives/C08DFBTTMMW/p1752671203221629?thread_ts=1752669294.000979&cid=C08DFBTTMMW). I thought we might need something like `requestLLMWithoutTools`, e.g.

```kotlin
suspend AIAgentContextBase.(String) -> ConditionResult = { output ->
    llm.writeSession {
        rewritePrompt { prompt ->
            prompt.withMessages {messages ->
                listOf(
                    Message.Assistant(
                        """
                        You are an assistant who identifies weather the retry condition was satisfied.
                        Please analyze weather the responses satisfy the requested condition.
                        Answer only YES or NO.
                        """,
                        ResponseMetaInfo(Clock.System.now())
                    )
                )
            }
        }
        (requestLLMWithoutTools().content == "YES").asConditionResult
    }
}
```

which would change the context. I forked the context, so that the actual prompt is not modified by the condition call. Of course, we can leave that to the user, as it is an advanced use-case, but it would make advanced implementations more complicated. WDYT?

EugeneTheDev: If it's for the future improvements, then I guess yes, it makes sense to leave it like this

PR #466, opened 2025-07-21
Add default value for `totalTokens` field in OpenAIUsage class

proton72:
Fix for missing `totalTokens` in some OpenAI API compartible responses (e.g., qwen).


aozherelyeva:
I think we can close this one as outdated. Please rebase to the `develop` and open a new PR in case the change is still actual 🙂

PR #467, opened 2025-07-22
Add validation for InMemoryPromptCache configurations

aozherelyeva:

---

#### Type of the change
- [ ] New feature
- [ ] Bug fix
- [ ] Documentation fix
- [x] Tests improvement

### DIFF
--- a/prompt/prompt-cache/prompt-cache-model/src/commonMain/kotlin/ai/koog/prompt/cache/memory/InMemoryPromptCache.kt
+++ b/prompt/prompt-cache/prompt-cache-model/src/commonMain/kotlin/ai/koog/prompt/cache/memory/InMemoryPromptCache.kt
@@ -25,7 +25,9 @@ public class InMemoryPromptCache(private val maxEntries: Int?) : PromptCache {
             val limit = when {
                 parts.size == 1 || parts[1].isEmpty() -> null
                 parts[1].equals("unlimited", ignoreCase = true) -> null
-                else -> parts[1].toIntOrNull() ?: error("Invalid memory cache size limit: ${parts[1]}. Expected a number or 'unlimited'.")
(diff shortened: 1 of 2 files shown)


prompt/prompt-cache/prompt-cache-model/src/commonMain/kotlin/ai/koog/prompt/cache/memory/InMemoryPromptCache.kt
```diff
@@ -25,7 +25,9 @@ public class InMemoryPromptCache(private val maxEntries: Int?) : PromptCache {
             val limit = when {
                 parts.size == 1 || parts[1].isEmpty() -> null
                 parts[1].equals("unlimited", ignoreCase = true) -> null
-                else -> parts[1].toIntOrNull() ?: error("Invalid memory cache size limit: ${parts[1]}. Expected a number or 'unlimited'.")
+                parts[1].startsWith("-") -> error("Invalid memory cache size limit: ${parts[1]}. Expected a positive number or 'unlimited'.")
```
skarpovdev, line 28: what if it's 0?
I would allow only > 0 values, unless there is a case for which it's needed. Something like:

`?.takeIf { it > 0 }`

@antoniibelyshev wdyt?

zarechneva-a: It may be done for a case where we technically have a cache but in fact it doesn't hold anything. Not sure if it's intended though :D

antoniibelyshev: Actually, the current implementation will not work properly for the cache size equal to 0: the put method always saves the entry in the cache. I think, having cache size equal to 0 is not a practical use case, so I would suggest allowing only positive cache size limit

Issue #462, opened 2025-07-21
Tool call parameters in opentelemetry


gorpeliates:
Currently, when executing tool calls in verbose mode, only the result of the tool is logged. Could we also include the tool call parameters in the logs? Could also be in the tool call span's tags, but considering that opentelemetry semantic conventions are followed this may option may be infeasible.

Ololoshechkin:
Hi! Please feel free to contribute, we are always welcoming good PRs :) 

@sdubov WDYT?

sdubov:
@gorpeliates, thank you for raising this issue! Indeed, it seems we missed to add parameters to the span. We are very open for a new PRs if you want to make this change. Please note, OpenTelemetry semantic convention specify that tool arguments should be the part of the event body field. Please see `gen_ai.choice` and `gen_ai.assistant.message` sections - https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-events/#event-gen_aichoice.

PR #473, opened 2025-07-24
Tool parameters in opentelemetry traces in verbose mode

gorpeliates:
Adds the arguments and their respective values to the opentelemetry logs when isVerbose(true) . Add ToolParams data class to the EventBodyFields class. Tests in the agents-features-opentelemetry module pass. Relates to issue #462 

---

#### Type of the change
- [x] New feature
- [ ] Bug fix
- [ ] Documentation fix
- [ ] Tests improvement


agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/event/ToolMessageEvent.kt
```diff
@@ -9,6 +11,7 @@ internal class ToolMessageEvent(
     private val provider: LLMProvider,
     private val toolCallId: String?,
     private val toolResult: ToolResult,
+    private val toolArgs: ToolArgs = ToolArgs.Empty(),
```
sdubov: According to [OT Semantic convention events](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-events/) doc, `arguments` is a body field property for [gen_ai.choice](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-events/#event-gen_aichoice) and [gen_ai.assistant.message](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-events/#event-gen_aiassistantmessage). It seems we need to update those two events instead.

`gen_ai.tool.message` is an event with tool response returned back to the model and it does not have args field

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/event/EventBodyFields.kt
```diff
@@ -56,4 +56,13 @@ internal object EventBodyFields {
         override val key: String = "id"
         override val value: String = id
     }
+
+    data class ToolParams(private val varArgs : ai.koog.agents.core.tools.reflect.ToolFromCallable.VarArgs) : EventBodyField() {
```
sdubov: I've checked the[ OT events doc](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-events/) and found that this field is called "arguments". The doc also specifies that it is recommended to pass arguments field as a model response as is:
> Models usually return arguments as a JSON string. In this case, it’s RECOMMENDED to provide arguments as is without attempting to deserialize them. Semantic conventions for individual systems MAY specify a different type for arguments field.

What do you think about such signature that follows the info from the doc?
```
data class Arguments(private val arguments: JsonObject) : EventBodyField()
```

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/event/ToolMessageEvent.kt
gorpeliates: Yes, thanks for pointing it out, I read the documentation as well and will adjust this.

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/event/EventBodyFields.kt
gorpeliates: @sdubov You have the following data class which is also mentioned in the OT conventions in EventBodyFields.kt, however when I search for "tool_calls" in the Test file you mentioned, I see no instance of it and none of the test spans have this attribute. So I believe when we check for the type of the message in (ChoiceEvent for example), The message is never `Message.Tool.Call` so even though you use a tool, you can never see the tool details. Could you give me a bit of a context about this? 
``` kotlin
data class ToolCalls(
        private val tools: List<ai.koog.prompt.message.Message.Tool>
    ) : EventBodyField() {
        override val key: String = "tool_calls"
        override val value: List<Map<String, Any>>
            get() {
                return tools.map { tool ->
                    buildMap {
                        val functionMap = buildMap {
                            put("name", tool.tool)
                            put("arguments", tool.content)
                        }

                        put("function", functionMap)
                        put("id", tool.id ?: "")
                        put("type", "function")
                    }
                }
            }
    }
```
If I understand correctly the tool arguments are also supposed to be there as follows right?:
 
``` kotlin
is Message.Tool.Call -> {
                if (verbose) {
                    add(EventBodyFields.ToolCalls(tools = listOf(message)))
                    add(EventBodyFields.Arguments(arguments = message.contentJson))
                }
            }
```

sdubov: @gorpeliates, sorry for a log reply (due to vacation) and for confusion with "tool_calls". Let me clarify these questions.

The `tool_calls` in the is refers to the [event body for choice and assistant messages](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-events/#event-gen_aiassistantmessage). I've recently updated some related code in scope of [KG-172](https://youtrack.jetbrains.com/issue/KG-172/Log-all-prompt-messages-for-a-particular-LLM-Call-Inference-span). This fix include the changes you describe here as well. Please see the details in the commit [33d7b6260e4945e93984e57f139ea80d249ff89f](https://github.com/JetBrains/koog/commit/33d7b6260e4945e93984e57f139ea80d249ff89f). 

>If I understand correctly the tool arguments are also supposed to be there as follows right?:

Yep. Looks correct. You need to pass `arguments: Arguments` as a parameter to `AssistantMessageEvent` and `ChoiceEvent` events and add them in case `verbose` flag is set to `true`.

sdubov: @gorpeliates, please feel free to also check some drafts that I added to arguments propagation that I created for some internal development and testing. Please check if you find it useful - https://github.com/JetBrains/koog/pull/547.

gorpeliates: @sdubov Thanks for the updates, I will check the drafts out. I think this MR can be closed if you believe the changes here are fixed in the other commit.

PR #475, opened 2025-07-24
KG-164. Fix Tool args serailzation issue for Tracing remote writer

sdubov:
Fix the issue with serialization of Tool Args and Tool Result when using Tracing remote writer ( [KG-164](https://youtrack.jetbrains.com/issue/KG-164/Tool-args-and-result-serialization-fail-when-sending-events-with-Tracing-feature-remote-writer) )

---

#### Type of the change
- [ ] New feature
- [x] Bug fix
- [ ] Documentation fix
- [ ] Tests improvement


agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/model/featureEventTypes.kt
```diff
@@ -286,7 +282,7 @@ public data class ToolCallEvent(
     val runId: String,
     val toolCallId: String?,
     val toolName: String,
```
sproshev, line 285: can this type be changed to `Tool` instead?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/model/featureEventTypes.kt
```diff
@@ -286,7 +282,7 @@
...
-    val toolArgs: ToolArgs,
+    val toolArgs: String,
```
EugeneTheDev: Let's make it `JsonObject` instead, to indicate that it's a Json structure. Setting it just to `String` is too general, and can lead to some problems with deserialization if someone decides to set just some string here. We faced similar problems already when adding persistency support

EugeneTheDev (changes requested):

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/model/featureEventTypes.kt
sdubov: @EugeneTheDev, could you please clarify what kind of problems do you mean? I'll update to `JsonObject`, I'm just interested in knowing about the related issues.

sdubov: Fixed. Updated `ToolArgs` parameter type from `String` to `JsonObject`.

EugeneTheDev: > could you please clarify what kind of problems do you mean?

We were setting some random strings in tests, and then in our actual logic trying to deserialize them as a valid JSON, which failed. Since tool args are always JSON, it's better to communicate this more clearly, to ensure that this parameter will always be a valid JSON and we would be able to deserialize it.

PR #478, opened 2025-07-25
Fix for urls generated from AzureOpenAIClientSettings

ikss:
mergeUrls function function from ktor.DefaultRequest is wrongly merging path segment if `chatCompletionsPath` and `
embeddingsPath` start with slash symbol.

And base url should also end with slash, so last part will not be dropped.

---

#### Type of the change
- [ ] New feature
- [x] Bug fix
- [ ] Documentation fix
- [ ] Tests improvement

### DIFF
--- a/prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openai/azure/Azure.kt
+++ b/prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openai/azure/Azure.kt
@@ -18,7 +18,7 @@ public fun AzureOpenAIClientSettings(
     version: AzureOpenAIServiceVersion,
     timeoutConfig: ConnectionTimeoutConfig = ConnectionTimeoutConfig(),
 ): OpenAIClientSettings = AzureOpenAIClientSettings(
-    baseUrl = "https://$resourceName.openai.azure.com/openai/deployments/$deploymentName",
+    baseUrl = "https://$resourceName.openai.azure.com/openai/deployments/$deploymentName/",
     version = version,
     timeoutConfig = timeoutConfig,
 )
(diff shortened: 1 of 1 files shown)


devcrocod (commented):
There’s no need to add a slash to the `baseUrl`, just as there’s no need to add a `slash` to the path

Take a look at the implementation of [OpenaiClient](https://github.com/JetBrains/koog/blob/8d834b80859b42033e24d785e51b5878d8ae7015/prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openai/OpenAILLMClient.kt#L52)
ktor is responsible for constructing the full url path, so these changes aren’t necessary

ikss:
> ktor is responsible for constructing the full url path, so these changes aren’t necessary

I thought the same, but my attept to use koog with Azure deployment failed in this case. Please take a look on screenshots
Here I put [breakpoint on line 124](https://github.com/ktorio/ktor/blob/92266d53ebcefe9c4caf033a41b41f0b2a81060f/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/DefaultRequest.kt#L124) where ktor is concatenating paths.
completions url is "/chat/completions" and deployment name is "sk-gpt-4o".

Before calling `concat`:
image

After `concat`
image

So encoded path segments became just  "/chat/completions" without deployment name. And actual call was made to 
`https://{azure_user_name}.openai.azure.com/chat/completions?api-version=2025-01-01-preview` without mandatory `openai/deployments/$deploymentName` and this result in 404.

After my change same `encodedPathSegments` parameters in `concat` became these:
image

And after concat it properly became this without loosing any parts:
image

aozherelyeva (approved):
LGTM. I thought of suggesting using the mockk that we already use, but found that it has issues when mocking an HTTP client, so no problem with the new dependency.

aozherelyeva:
> @aozherelyeva do you know what the qodana error is related to? and does it affect merge changes?

It's quite expected because the forked repo doesn't have Qodana token, therefore it can't run usual PR checks. We can't so anything about it for now, so I think we can just merge the PR if no other issues are present.

PR #501, opened 2025-07-30
GH 498 makes it possible to use AWS profiles for Bedrock authentication

### LINKED ISSUE #498: Make it possible to use aws config profiles for Bedrock authentication
Currently, it uses only [the static config](https://github.com/JetBrains/koog/blob/d3fe0ba99a44c0229b8af0229bbc339956e8448d/prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/bedrock/BedrockLLMClient.kt#L108)

It should be possible to use the local [AWS config profile](https://docs.aws.amazon.com/cli/v1/userguide/cli-configure-files.html) provided by the Kotlin SDK Bedrock class `ProfileCredentialsProvider`

Workaround and example improvement:

```kt
    val bedrockRuntimeClient = BedrockRuntimeClient {
        region = "eu-central-1"
        credentialsProvider = ProfileCredentialsProvider(profileName = dev)
    }
    val bedrockExecutor = SingleLLMPromptExecutor(BedrockLLMClient(bedrockRuntimeClient))
```

I would try to propose an MR if it's okay

artemptushkin:
It closes https://github.com/JetBrains/koog/issues/498

I propose to expose a new constructor `BedrockLLMClient` so it's possible to use AWS config profiles.

It could be done better, respecting and other `CredentialsProvider` classes or having a nice builder DSL, but I keep it out of the scope so far

### DIFF
--- a/prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/bedrock/BedrockLLMClient.kt
+++ b/prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/bedrock/BedrockLLMClient.kt
@@ -157,6 +157,27 @@ public class BedrockLLMClient(
         clock = clock
     )
 
+    /**
+     * Creates a new Bedrock LLM client configured with the specified AWS profile.
+     *
+     * @param credentialsProvider any [CredentialsProvider] implementation to use for authentication
+     * @param settings Configuration settings for the Bedrock client, such as region and endpoint
+     * @param clock A clock used for time-based operations
+     * @return A configured [LLMClient] instance for Bedrock
+     */
+    public constructor(
+        credentialsProvider: CredentialsProvider,
+        settings: BedrockClientSettings = BedrockClientSettings(),
+        clock: Clock = Clock.System,
+    ) : this(
+        bedrockClient = bedrockClient(
+            settings = settings,
+            credentialsProvider = credentialsProvider
+        ),
+        moderationGuardrailsSettings = settings.moderationGuardrailsSettings,
+        clock = clock
+    )
+
     private val json = Json {
         ignoreUnknownKeys = true
         isLenient = true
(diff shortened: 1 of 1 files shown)


prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/bedrock/BedrockLLMClient.kt
```diff
@@ -105,24 +123,35 @@ public class BedrockLLMClient(
...
-        },
+    /**
+     * Creates a new Bedrock LLM client configured with the specified AWS profile.
+     *
+     * @param awsProfileName The name of the AWS config profile defined in ~/.aws/config
+     * @param settings Configuration settings for the Bedrock client, such as region and endpoint
+     * @param clock A clock used for time-based operations
+     * @return A configured [LLMClient] instance for Bedrock
+     */
+    public constructor(
```
bay73, line 157: Can we have a constructor which accepts CredentialsProvider directly? In this case we can also use DefaultCredcentialsProvider to allow AWS SDK choose authentication settings

artemptushkin: yes, sure. to be updated

artemptushkin: updated

EugeneTheDev:
Hi, can you please elaborate why a separate contructor/function is needed? We already have a primary constructor that accepts `BedrockRuntimeClient` to allow custom configuration. Shouldn't one just use it to pass such info?

EugeneTheDev (approved):

artemptushkin:
@Ololoshechkin if you could review please, Eugene approved this and I just resolved conflicts

EugeneTheDev:
Thanks for the idea, I see that it has been hanging for some time, and there are merge conflicts now. I'll create a small PR myself, based on this. I decided to just update the constructor that accepts keys and token to accept `CredentialProvider` instead. It should solve the same problem, while providing more flexibility regarding choosing convenient authentication method.
https://youtrack.jetbrains.com/issue/KG-442/

PR #506, opened 2025-07-30
Use Inference Profiles on AWS Bedrock Models

### LINKED ISSUE #488: Support for Inference Profiles on AWS Bedrock Models
I encountered an issue while attempting to call the Claude Sonnet 4 model on AWS Bedrock:

```
Invocation of model ID anthropic.claude-sonnet-4-20250514-v1:0 with on-demand throughput isn’t supported. Retry your request with the ID or ARN of an inference profile that contains this model.
ValidationException(message=Invocation of model ID anthropic.claude-sonnet-4-20250514-v1:0 with on-demand throughput isn’t supported. Retry your request with the ID or ARN of an inference profile that contains this model.)
```

However, modifying the model ID to include the inference profile prefix causes a failure in the following check in `BedrockClient`:

```
internal fun getBedrockModelFamily(model: LLModel): BedrockModelFamilies {
    require(model.provider == LLMProvider.Bedrock) { "Model ${model.id} is not a Bedrock model" }
    return when {
        model.id.startsWith("anthropic.claude") -> BedrockModelFamilies.AnthropicClaude
        model.id.startsWith("amazon.nova") -> BedrockModelFamilies.AmazonNova
        model.id.startsWith("ai21.jamba") -> BedrockModelFamilies.AI21Jamba
        model.id.startsWith("meta.llama") -> BedrockModelFamilies.Meta
        else -> throw IllegalArgumentException("Model ${model.id} is not a supported Bedrock model")
    }
}
```

Is it possible to use AWS Bedrock inference profiles in Koog currently? Or is this something that needs to be addressed?

Thanks!

aozherelyeva:
Fixes #488

---

#### Type of the change
- [ ] New feature
- [x] Bug fix
- [ ] Documentation fix
- [ ] Tests improvement

- [] Tests for the changes have been added
- [] All new and existing tests passed


aozherelyeva:
Heavy tests aren't fixed yet, but I'm on my way to do it.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/bedrock/BedrockLLMClient.kt
```diff
@@ -136,10 +136,10 @@ public class BedrockLLMClient(
...
         require(model.provider == LLMProvider.Bedrock) { "Model ${model.id} is not a Bedrock model" }
         return when {
-            model.id.startsWith("anthropic.claude") -> BedrockModelFamilies.AnthropicClaude
-            model.id.startsWith("amazon.nova") -> BedrockModelFamilies.AmazonNova
-            model.id.startsWith("ai21.jamba") -> BedrockModelFamilies.AI21Jamba
-            model.id.startsWith("meta.llama") -> BedrockModelFamilies.Meta
+            model.id.startsWith("us.anthropic.claude") -> BedrockModelFamilies.AnthropicClaude
+            model.id.startsWith("us.amazon.nova") -> BedrockModelFamilies.AmazonNova
+            model.id.startsWith("us.ai21.jamba") -> BedrockModelFamilies.AI21Jamba
+            model.id.startsWith("us.meta.llama") -> BedrockModelFamilies.Meta
```
samDobsonDev: Thank you for working on this :) Just had a look now, and what about Inference Profiles that route requests from EU regions? So the model ID in that case would be something like `eu.anthropic.claude-sonnet-4-20250514-v1:0`

zarechneva-a: That's a good point. Let me think of how to make it more flexible! I'm thinking of introducing an env variable for the region name, and then parsing its beginning and using it here as a prefix.

samDobsonDev: Could do something like checking if the modelId contains specific substrings that identify the model family, regardless of their position in the string:

```
internal fun getBedrockModelFamily(model: LLModel): BedrockModelFamilies {
    require(model.provider == LLMProvider.Bedrock) { "Model ${model.id} is not a Bedrock model" }
    return when {
        model.id.contains("anthropic.claude") -> BedrockModelFamilies.AnthropicClaude
        model.id.contains("amazon.nova") -> BedrockModelFamilies.AmazonNova
        model.id.contains("ai21.jamba") -> BedrockModelFamilies.AI21Jamba
        model.id.contains("meta.llama") -> BedrockModelFamilies.Meta
        else -> throw IllegalArgumentException("Model ${model.id} is not a supported Bedrock model")
    }
}
```

So we'd use `.contains` rather than `.startsWith`

However, this would require people setting up their own `LLModel` implementations that include custom inference profile prefixes, which isn't ideal. Having some configuration somewhere is a good idea, as you suggest. For a given model, here are the different formats that the model ID can be in:

```
Standard: anthropic.claude-sonnet-4-20250514-v1:0
Inference Profile ID Prefix: eu.anthropic.claude-sonnet-4-20250514-v1:0
Inference Profile ARN Prefix: arn:aws:bedrock:eu-west-1:<aws_account_id>:inference-profile/eu.anthropic.claude-sonnet-4-20250514-v1:0
```

What if we allowed the user to specify the inference profile prefix when constructing a `BedrockClientSettings` object? Then we can pass the inference profile prefix here and manually amend the model ID to include said prefix?

However, this would raise another issue. Not _all_ models require inference profiles and support on-demand throughput out of the box. This means having an inference profile prefix attached to the `BedrockLLMClient` is not ideal, as every model would use it where it might not be necessary to do so. Instead, it would be beneficial to provide said prefix when the user decides which model to use.

So instead of this:

```
AIAgentConfig = AIAgentConfig(
  prompt = ...,
  model = BedrockModels.AnthropicClaude4Sonnet,
  maxAgentIterations = ...
)
```

It could be something like this:

```
AIAgentConfig(
  prompt = ...,
  model = BedrockModel(model = BedrockModels.AnthropicClaude4Sonnet, inferenceProfilePrefix = inferenceProfilePrefix),
  maxAgentIterations = ...
)
```

Where the `inferenceProfilePrefix` is an optional argument.

zarechneva-a: Sounds good!

prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/bedrock/BedrockLLMClient.kt
```diff
@@ -51,7 +51,7 @@ import kotlinx.serialization.json.Json
  * @property moderationGuardrailsSettings Optional settings of the AWS bedrock Guardrails (see [AWS documentation](https://docs.aws.amazon.com/bedrock/latest/userguide/guardrails-use-independent-api.html) ) that would be used for the [LLMClient.moderate] request
  */
 public class BedrockClientSettings(
-    internal val region: String = "us-east-1",
+    internal val region: String = "us-west-2",
```
devcrocod: As far as I know, AWS regions are a set, why not make it an enum?

devcrocod (commented):
I think we should look into how models are provided in AWS

As far as I understand, the structure is:
**Region** → **Model Family** → **Model Name**

We could wrap this in enums or classes to avoid using raw strings. It would also simplify our code by removing string-based matching

prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/bedrock/BedrockLLMClient.kt
zarechneva-a: Great idea!

agents/agents-features/agents-features-memory/src/jvmMain/kotlin/ai/koog/agents/memory/storage/Aes256GCMStorageEncryptor.kt
```diff
@@ -1,7 +1,7 @@
 package ai.koog.agents.memory.storage
 
 import java.security.SecureRandom
-import java.util.*
+import java.util.Base64
```
devcrocod, line 4: A good practice would be to use the Kotlin api where possible. Although in this case it doesn’t change anything, since kotlin simply wraps Java `Base64`. This approach will reduce potential issues if the code ever needs to be moved to common

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/MultipleLLMPromptExecutorIntegrationTest.kt
```diff
@@ -108,6 +112,7 @@ class MultipleLLMPromptExecutorIntegrationTest {
     private val geminiApiKey: String get() = readTestGoogleAIKeyFromEnv()
     private val openAIApiKey: String get() = readTestOpenAIKeyFromEnv()
     private val anthropicApiKey: String get() = readTestAnthropicKeyFromEnv()
+    private val googleApiKey: String get() = readTestGoogleAIKeyFromEnv()
```
devcrocod, line 114: What's the difference between `geminiApiKey` and `googleApiKey`?

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/utils/RetryUtils.kt
```diff
@@ -36,20 +37,18 @@ object RetryUtils {
...
     }
 
-    suspend fun <T> withRetry(
+    fun <T> withRetry(
         times: Int = 3,
         delayMs: Long = 1000,
         testName: String = "test",
         action: suspend () -> T
-    ): T {
+    ): T = runBlocking {
```
devcrocod, line 45: Should we use `runTest`?

prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/bedrock/BedrockModels.kt
```diff
@@ -5,6 +5,97 @@ import ai.koog.prompt.executor.clients.anthropic.AnthropicModels
...
+     */
+    val effectiveModel: LLModel = model.copy(provider = LLMProvider.Bedrock, id = effectiveModelId)
+}
+
+/**
+ * AWS Bedrock regions.
+ *
+ * Represents all available AWS regions where Bedrock service is supported.
+ */
+@Serializable
```
devcrocod, line 43: Pay attention to the behavior of the Serializable enum
https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/json.md#decoding-enums-in-a-case-insensitive-manner

devcrocod (approved):
Overall, PR looks good, but please take a look at my questions and comments

I think in this case it would be better to create an issue and a separate PR for fixing the Bedrock client

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/utils/RetryUtils.kt
zarechneva-a: Nope, that's intended because we use a delay inside the method, and the `delay()` is not applied when called inside `runTest()`.

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/MultipleLLMPromptExecutorIntegrationTest.kt
zarechneva-a: No difference, just duplication :) Thanks for the catch!

PR #511, opened 2025-07-31
abragin/JBAI-15990-LLM-message-tracing

agbragin:
Adjust OpenTelemetry spans so they are properly displayed in Langfuse and Weave.

---

#### Type of the change
- [ ] New feature
- [X] Bug fix
- [ ] Documentation fix
- [ ] Tests improvement

### DIFF
--- a/agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetry.kt
+++ b/agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetry.kt
@@ -6,6 +6,8 @@ import ai.koog.agents.core.agent.entity.AIAgentStorageKey
 import ai.koog.agents.core.feature.AIAgentFeature
 import ai.koog.agents.core.feature.AIAgentPipeline
 import ai.koog.agents.core.feature.InterceptContext
+import ai.koog.agents.core.feature.handler.AfterLLMCallContext
+import ai.koog.agents.core.feature.handler.BeforeLLMCallContext
 import ai.koog.agents.features.opentelemetry.attribute.CommonAttributes
 import ai.koog.agents.features.opentelemetry.attribute.SpanAttributes
 import ai.koog.agents.features.opentelemetry.event.AssistantMessageEvent
@@ -24,6 +26,7 @@ import ai.koog.agents.features.opentelemetry.span.SpanEndStatus
 import ai.koog.agents.features.opentelemetry.span.SpanProcessor
 import ai.koog.prompt.message.Message
 import io.github.oshai.kotlinlogging.KotlinLogging
+import io.opentelemetry.api.trace.Span
 import io.opentelemetry.api.trace.StatusCode
 import kotlinx.coroutines.currentCoroutineContext
 
@@ -49,7 +52,7 @@ public class OpenTelemetry {
      * correctly and resources are properly released.
      */
     public companion object Feature : AIAgentFeature<OpenTelemetryConfig, OpenTelemetry> {
-
+        private const val MESSAGE_CONTENT_PLACEHOLDER = "Message content is not traced. Use verbose mode to trace."
         private val logger = KotlinLogging.logger { }
 
         override val key: AIAgentStorageKey<OpenTelemetry> = AIAgentStorageKey("agents-features-opentelemetry")
@@ -260,28 +263,13 @@ public class OpenTelemetry {
                 // Start span
                 spanProcessor.startSpan(inferenceSpan)
 
-                // Add events to the InferenceSpan after the span is created
-                val lastMessage = eventContext.prompt.messages.lastOrNull()
-
-                val events: List<GenAIAgentEvent> = lastMessage?.let { message ->
-                    buildList {
-                        when (message) {
-                            is Message.User -> add(UserMessageEvent(provider, message, verbose = config.isVerbose))
-                            is Message.System -> add(SystemMessageEvent(provider, message, verbose = config.isVerbose))
-                            is Message.Assistant -> add(
-                                AssistantMessageEvent(
-                                    provider,
-                                    message,
-                                    verbose = config.isVerbose
-                                )
-                            )
-
-                            else -> {}
-                        }
+                if (config.useEvents) {
+                    addBeforeLLMCallEvents(inferenceSpan, eventContext, config)
+                } else {
+                    eventContext.prompt.messages.forEachIndexed { index, message ->
+                        setMessageAttributes(inferenceSpan.span, index, message, config, "prompt")
                     }
-                } ?: emptyList()
-
-                inferenceSpan.addEvents(events)
+                }
             }
 
             pipeline.interceptAfterLLMCall(interceptContext) { eventContext ->
@@ -302,28 +290,23 @@ public class OpenTelemetry {
 
                 val inferenceSpan = spanProcessor.getSpanOrThrow<InferenceSpan>(inferenceSpanId)
 
-                val provider = eventContext.model.provider
-
-                // Add events to the InferenceSpan before finishing the span
-                val lastMessage = eventContext.responses.lastOrNull()
-
-                val moderationResult = eventContext.moderationResponse
-
-                val events: List<GenAIAgentEvent> = lastMessage?.let { message ->
-                    buildList {
-                        when (message) {
-                            is Message.Assistant -> add(ChoiceEvent(provider, message, config.isVerbose))
-                            else -> {}
+                if (config.useEvents) {
+                    addAfterLLMCallEvents(inferenceSpan, eventContext, config)
+                } else {
+                    eventContext.prompt.messages
+                        .filter {
+                            // TODO: check tool message here
+                            it.role == Message.Role.Assistant
                         }
-                    }
-                } ?: moderationResult?.let {
-                    buildList { add(ModerationResponseEvent(provider, it, config.isVerbose)) }
-                } ?: emptyList()
+                        .forEachIndexed { index, message ->
+                            setMessageAttributes(inferenceSpan.span, index, message, config, "completion")
+                        }
+                }
 
-                inferenceSpan.addEvents(events)
-
-                // Stop InferenceSpan
-                spanProcessor.endSpan(inferenceSpanId)
+                spanProcessor.endSpan(
+                    inferenceSpanId,
+                    listOf(SpanAttributes.Response.FinishReasons(listOf(SpanAttributes.Response.FinishReasonType.Stop)))
+                )
             }
 
             //endregion LLM Call
@@ -463,6 +446,60 @@ public class OpenTelemetry {
             }
 
             //endregion Tool Call
+        }
+
+        private fun setMessageAttributes(span: Span, index: Int, message: Message, config: OpenTelemetryConfig, type: String) {
+            span.setAttribute("gen_ai.$type.$index.role", message.role.name.lowercase())
+            span.setAttribute(
+                "gen_ai.$type.$index.content",
+                if (config.isVerbose) message.content else MESSAGE_CONTENT_PLACEHOLDER
+            )
+        }
+
+        private fun addBeforeLLMCallEvents(inferenceSpan: InferenceSpan, eventContext: BeforeLLMCallContext, config: OpenTelemetryConfig) {
+            // Add events to the InferenceSpan after the span is created
+            val lastMessage = eventContext.prompt.messages.lastOrNull()
+            val provider = eventContext.model.provider
+
+            val events: List<GenAIAgentEvent> = lastMessage?.let { message ->
+                buildList {
+                    when (message) {
+                        is Message.User -> add(UserMessageEvent(provider, message, verbose = config.isVerbose))
+                        is Message.System -> add(SystemMessageEvent(provider, message, verbose = config.isVerbose))
+                        is Message.Assistant -> add(
+                            AssistantMessageEvent(
+                                provider,
+                                message,
+                                verbose = config.isVerbose
+                            )
+                        )
+
+                        else -> {}
+                    }
+                }
+            } ?: emptyList()
+
+            inferenceSpan.addEvents(events)
+        }
+
+        private fun addAfterLLMCallEvents(inferenceSpan: InferenceSpan, eventContext: AfterLLMCallContext, config: OpenTelemetryConfig) {
+            // Add events to the InferenceSpan before finishing the span
+            val lastMessage = eventContext.responses.lastOrNull()
+
+            val moderationResult = eventContext.moderationResponse
+
+            val events: List<GenAIAgentEvent> = lastMessage?.let { message ->
+                buildList {
+                    when (message) {
+                        is Message.Assistant -> add(ChoiceEvent(eventContext.model.provider, message, config.isVerbose))
+                        else -> {}
+                    }
+                }
+            } ?: moderationResult?.let {
+                buildList { add(ModerationResponseEvent(eventContext.model.provider, it, config.isVerbose)) }
+            } ?: emptyList()
+
+            inferenceSpan.addEvents(events)
         }
     }
 }
--- a/agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetry.kt
+++ b/agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetry.kt
@@ -6,17 +6,9 @@ import ai.koog.agents.core.agent.entity.AIAgentStorageKey
 import ai.koog.agents.core.feature.AIAgentFeature
 import ai.koog.agents.core.feature.AIAgentPipeline
 import ai.koog.agents.core.feature.InterceptContext
-import ai.koog.agents.core.feature.handler.AfterLLMCallContext
-import ai.koog.agents.core.feature.handler.BeforeLLMCallContext
 import ai.koog.agents.features.opentelemetry.attribute.CommonAttributes
 import ai.koog.agents.features.opentelemetry.attribute.SpanAttributes
-import ai.koog.agents.features.opentelemetry.event.AssistantMessageEvent
-import ai.koog.agents.features.opentelemetry.event.ChoiceEvent
-import ai.koog.agents.features.opentelemetry.event.GenAIAgentEvent
-import ai.koog.agents.features.opentelemetry.event.ModerationResponseEvent
-import ai.koog.agents.features.opentelemetry.event.SystemMessageEvent
 import ai.koog.agents.features.opentelemetry.event.ToolMessageEvent
-import ai.koog.agents.features.opentelemetry.event.UserMessageEvent
 import ai.koog.agents.features.opentelemetry.span.CreateAgentSpan
 import ai.koog.agents.features.opentelemetry.span.ExecuteToolSpan
 import ai.koog.agents.features.opentelemetry.span.InferenceSpan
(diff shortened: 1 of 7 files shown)


agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/attribute/CommonAttributes.kt
```diff
@@ -10,6 +11,16 @@ internal object CommonAttributes {
         override val value: String = provider.id
     }
 
+    data class Role(private val message: Message) : GenAIAttribute {
```
kpavlov: Does it make sense storing the Message reference as field? How would it affect garbage collection?

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/attribute/CommonAttributes.kt
```diff
@@ -10,6 +11,16 @@ internal object CommonAttributes {
...
+        override val key: String = "role"
+        override val value: String = message.role.name
+    }
+
+    data class Content(private val message: Message) : GenAIAttribute {
```
kpavlov: Same here

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/attribute/CommonAttributes.kt
agbragin: What potential issues with the garbage collection do you expect here? How would it differ from the case when we store `String`?

Using `Message` instead of `String` adds more flexibility in how the name values are stored and retrieved.

agents/agents-features/agents-features-opentelemetry/src/jvmTest/kotlin/ai/koog/agents/features/opentelemetry/langfuse/TracingTest.kt
```diff
@@ -0,0 +1,307 @@
...
+import java.time.Duration
+import java.util.concurrent.TimeUnit
+import kotlin.test.AfterTest
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+
+/**
+ * Test traces OpenTelemetry structure conformance to Langfuse and Weave data models.
+ */
+class TracingTest {
```
sdubov: Could you please rename the test class to be more descriptive? I would also suggest to split this class into two to not make this class very big. Since we have two different providers, I would suggest to have something like "LangfuseTracingTest" and "WeaveTracingTest". Wdyt?

agents/agents-features/agents-features-opentelemetry/src/jvmTest/kotlin/ai/koog/agents/features/opentelemetry/langfuse/TracingTest.kt
```diff
@@ -0,0 +1,307 @@
...
+        }
+
+        agent.use {
+            it.run(userPrompt)
+        }
+    }
+}
+
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+abstract class BaseOpenTelemetryTracingTest {
```
sdubov: @agbragin, could you please clarify why do we need this class? It seems that nothing changes if I delete it. In tests you use `InMemorySpanExporter` and do not relay on the configuration from this class. Please let me know if I miss anything here.

agents/agents-features/agents-features-opentelemetry/src/jvmTest/kotlin/ai/koog/agents/features/opentelemetry/langfuse/TracingTest.kt
```diff
@@ -0,0 +1,307 @@
...
+import java.util.concurrent.TimeUnit
+import kotlin.test.AfterTest
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+
+/**
+ * Test traces OpenTelemetry structure conformance to Langfuse and Weave data models.
+ */
+class TracingTest {
+    val inMemorySpanExporter: InMemorySpanExporter = InMemorySpanExporter.create()
```
sdubov, line 28: Here and below in other tests you use `InMemorySpanExporter` and verify spans collected inside this exporter. It seems that test verify this case only. I feel that I miss the point here. Could you please clarify how do we assert the Langfuse and Weave integration by tests?

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetryConfig.kt
```diff
@@ -81,6 +83,18 @@ public class OpenTelemetryConfig : FeatureConfig() {
...
+    /**
+     * A boolean property that determines whether OpenTelemetry generative AI data should be recorded using events as
+     * described in OpenTelemetry semantic conventions.
+     *
+     * Defaults to `false` since major OpenTelemetry backends do not support this
+     * and rely on span attributes instead.
+     *
+     * @see <a href="https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-events/">OpenTelemetry GenAI Events Specification</a>
+     */
+    public val useEvents: Boolean
```
sdubov: I found this API a bit confusing. It seems for me as the internal implementation details and should not be exposed to an end user.

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetry.kt
```diff
@@ -260,28 +263,13 @@ public class OpenTelemetry {
...
-                                AssistantMessageEvent(
-                                    provider,
-                                    message,
-                                    verbose = config.isVerbose
-                                )
-                            )
-
-                            else -> {}
-                        }
+                if (config.useEvents) {
```
sdubov: major: @agbragin, as far as I remember from our discussion, we agreed to keep current approach unchanged and append the attributes (that are used by Langfuse and Weave) to process LLM input/output. So, at the end it should contain both (until the OT docs are not final).

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetry.kt
```diff
@@ -302,27 +290,20 @@ public class OpenTelemetry {
...
-                    buildList {
-                        when (message) {
-                            is Message.Assistant -> add(ChoiceEvent(provider, message, config.isVerbose))
-                            else -> {}
+                if (config.useEvents) {
+                    addAfterLLMCallEvents(inferenceSpan, eventContext, config)
+                } else {
+                    eventContext.prompt.messages
+                        .filter {
+                            // TODO: check tool message here
```
sdubov: major: Similar comment as I posted in the other review - the `prompt` object might contain previous assistant messages. It feels an error prone to process all history all on each LLM Call.

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetry.kt
```diff
@@ -302,27 +290,20 @@ public class OpenTelemetry {
...
                         }
-                    }
-                } ?: moderationResult?.let {
-                    buildList { add(ModerationResponseEvent(provider, it, config.isVerbose)) }
-                } ?: emptyList()
-
-                inferenceSpan.addEvents(events)
+                        .forEachIndexed { index, message ->
+                            setMessageAttributes(inferenceSpan.span, index, message, config, "completion")
+                            inferenceSpan.span.setAttribute("gen_ai.completion.$index.finish_reason", "stop")
```
sdubov: Please use `endSpan` call `attributes` parameter to provider list of attributes for updating the attributes in the span. I think we will deprecate `span: Span` visibility in future to keep the clean API.

sdubov: Please do not use raw attribute keys. For "finish_reason", we have `SpanAttributes.Response.FinishReason` that get an available options as a parameter of type `FinishReasonType`. I also do not see the example for `gen_ai.completion.$index.finish_reason` in OT Semantic convention. As far as I can see the "finish_reason" is only applicable as an attribute with key `gen_ai.response.finish_reason`. Could you please point me to the doc if I miss this?

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetry.kt
```diff
@@ -464,5 +445,59 @@ public class OpenTelemetry {
 
             //endregion Tool Call
         }
+
+        private fun setMessageAttributes(span: Span, index: Int, message: Message, config: OpenTelemetryConfig, type: String) {
+            span.setAttribute("gen_ai.$type.$index.role", message.role.name.lowercase())
+            span.setAttribute(
+                "gen_ai.$type.$index.content",
```
sdubov, line 434: Same here. Please, do not use raw attributes.

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetry.kt
```diff
@@ -464,5 +445,59 @@ public class OpenTelemetry {
 
             //endregion Tool Call
         }
+
+        private fun setMessageAttributes(span: Span, index: Int, message: Message, config: OpenTelemetryConfig, type: String) {
```
sdubov, line 431: major: @agbragin, could you please clarify if it is expected to set the same set of attributes in both "before..." and "after..." handlers? Why do we need to duplicate this logic here?

sdubov (changes requested):

agents/agents-features/agents-features-opentelemetry/src/jvmTest/kotlin/ai/koog/agents/features/opentelemetry/langfuse/TracingTest.kt
agbragin: At the moment all checks we have here are not related to any particular provider, so there is a little sense to split this (but it would be required later with agent graph implementation, so I would add a specific test on that stage).

These tests can be moved to `OpenTelemetryTest`, but I would consider replacing `MockSpanExporter` with `InMemorySpanExporter` that is provided by OpenTelemetry.

agents/agents-features/agents-features-opentelemetry/src/jvmTest/kotlin/ai/koog/agents/features/opentelemetry/langfuse/TracingTest.kt
agbragin: You are right, it's outdated. Deleted.

agents/agents-features/agents-features-opentelemetry/src/jvmTest/kotlin/ai/koog/agents/features/opentelemetry/langfuse/TracingTest.kt
agbragin: Renamed to `TraceStructureTest`.

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetry.kt
agbragin: Having both with the current events implementation blocks proper input/output rendering in Langfuse:
image

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetry.kt
agbragin: It's the same logic that is applied to [logging](https://youtrack.jetbrains.com/issue/KG-172/Log-all-prompt-messages-for-a-particular-LLM-Call-Inference-span#focus=Comments-27-12446842.0-0), and I believe we need to have a consistency between both. 
Also, we must log _all messages that are sent to the LLM_ – that's the tracing purpose, and that how it's implemented by other libraries and frameworks.
If there is some inconsistency between `eventContext.prompt.messages` and the set of messages sent to the LLM on a particular step, I believe it must be fixed in the upstream code.

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetry.kt
agbragin: The idea to is have both client (user, system) and server (assistant) messages traced, but now I understand that I likely had a wrong expectations regarding `eventContext.prompt.messages`. 
It looks like `interceptAfterLLMCall` `eventContext.prompt.messages` contains all the messages from `interceptBeforeLLMCall` `eventContext.prompt.messages` **plus** messages added by the assistant, right?
Then we need to substract messages from `interceptBeforeLLMCall` on `interceptAfterLLMCall` step to avoid duplication.

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetry.kt
agbragin: Fixed.

agents/agents-features/agents-features-opentelemetry/src/jvmTest/kotlin/ai/koog/agents/features/opentelemetry/langfuse/TracingTest.kt
agbragin: Here we verify the trace structure is the one that is properly visualized in Langfuse/Weave.
I see no straightforward ways to verify it directly using physical Langfuse/Weave services.

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetry.kt
```diff
@@ -464,5 +447,59 @@ public class OpenTelemetry {
...
+            span.setAttribute("gen_ai.$type.$index.role", message.role.name.lowercase())
```
sdubov, line 432: This approach lock us with one provider structure. `gen_ai.$type.$index.role` is not documented as far as I can see. This might be not good for future frameworks support.

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetry.kt
sdubov: Isn't it the problem with the "body" field for events? So, if we extract all body fields into event attributes, would not it fix the problem?

agents/agents-features/agents-features-opentelemetry/src/jvmTest/kotlin/ai/koog/agents/features/opentelemetry/langfuse/TracingTest.kt
sdubov: How does these tests differ from ones in `OpenTelemetryTest` class? 

I would think over a pure integration tests to report somewhere in Langfuse server that is running in Docker and get reported spans from the running server if it has the API like that. However, it might be quite time consiming and can be fixed separately. Could you please create a ticket for such test if you do not plan to add them in this commit, to not miss it?

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetry.kt
sdubov: Ok. I got it. I will work on [KG-172](https://youtrack.jetbrains.com/issue/KG-172/) to fix this.

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetry.kt
sdubov: >It looks like interceptAfterLLMCall eventContext.prompt.messages contains all the messages from interceptBeforeLLMCall eventContext.prompt.messages plus messages added by the assistant, right?

That's correct. But after llm call you also have `responses` property that should contain only responses from the current request.

sdubov (changes requested):

agents/agents-features/agents-features-opentelemetry/src/jvmTest/kotlin/ai/koog/agents/features/opentelemetry/langfuse/TracingTest.kt
agbragin: It makes a little sense to create such test since traces downloaded from the Langfuse will be the same that we uploaded. The thing we want to test is that traces are properly visualized in the UI, and there is no easy way to test that.

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetry.kt
agbragin: It's possible taking into account that:
1. All messages must be traced
1. Message order, role and indices must be preserved

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetry.kt
agbragin: Is documented [here](https://docs.smith.langchain.com/observability/how_to_guides/trace_with_opentelemetry#genai-standard-attributes) with a reference to GenAI standard attributes (wasn't successful in searching in the standard itself).

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetry.kt
agbragin: Fixed, thanks!

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetry.kt
agbragin: But it wouldn't work for `interceptBeforeLLMCall` where we need to add attributes without closing the span, would it?

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetryConfig.kt
agbragin: I removed it from the configuration.

PR #512, opened 2025-07-31
Add support for iOS targets

### LINKED ISSUE #177: Support iOS target
Currently in the readme its mentioned that the supported targets are JVM and JS.
> Multiplatform: Run agents on both JVM and JS targets with Kotlin Multiplatform.

Are there any plans to add support for iOS target if its not there already?

### LINKED ISSUE #382: Support for Kotlin/Native (at least Apple platforms)
> If Kotlin/Native is planned (not sure about this), we could also implement [Foundation Models](https://developer.apple.com/documentation/foundationmodels) for Apple platforms. So we would have full support for Kotlin Multiplatform Mobile apps.
> 
> The library is already KMP first (since it uses Ktor etc...). So I guess it could make sense in the future. 

 _Originally posted by @nathanfallet in [#379](https://github.com/JetBrains/koog/issues/379#issuecomment-3027315060)_

Related to #381

nathanfallet:
I added iOS native targets

Fixes #177 

---

#### Type of the change
- [ ] New feature
- [ ] Bug fix
- [ ] Documentation fix
- [ ] Tests improvement
- [x] New KMP platforms


kagg886:
wow，that's good.

if we have multi engine in context(like okhttp),does we have some api to choose it?

nathanfallet:
There is the engineFactory function for Ollama, but for other things I think it chooses in what's available (it works with Js engine already so it should work the same with other platforms)

README.md
```diff
@@ -29,7 +29,7 @@ Key features of Koog include:
 - **Flexible graph workflows**: Design complex agent behaviors using intuitive graph-based workflows.
 - **Modular feature system**: Customize agent capabilities through a composable architecture.
 - **Scalable architecture**: Handle workloads from simple chatbots to enterprise applications.
-- **Multiplatform**: Run agents on JVM, JS, WasmJS targets with Kotlin Multiplatform.
+- **Multiplatform**: Run agents on JVM, JS, WasmJS, Apple, Linux, Windows targets with Kotlin Multiplatform.
```
devcrocod: Why are Linux and Windows targets needed?
Adding a large number of targets makes maintenance more difficult, especially Windows

nathanfallet: Linux is particularly useful to use the library with Ktor Native on any Linux or Docker environment. Because Ktor Native is way smaller in binary size and memory consumption than Ktor JVM.

nathanfallet:
The more target we support, the more users can use the library everywhere. I always try to support as much targets as possible to be the more KMP possible.
I remember when I started to work with KMP, I was sometimes limited because some libraries were not supporting all iOS or watchOS targets, so I could not use it in the app. I don't want anyone to experience this again.

Also, there is not much work since Ktor does all the work of multiplatform abstraction, so we maintain only common code.

nathanfallet:
The latest version of Kotlin Logging was published with the Kotlin 2.2.0 compiler so it does not work. Do I rollback and skip some watchOS targets? Or do I upgrade the project to use Kotlin 2.2.0 too?

devcrocod:
@nathanfallet 
I understand and support the idea of adding more targets. However, Koog is a framework, and with every new target comes a certain level of responsibility and maintenance overhead

> Also, there is not much work since Ktor does all the work of multiplatform abstraction, so we maintain only common code.

That’s not entirely accurate, we rely on other dependencies as well. In addition, supporting more targets complicates the `@Tool`, which currently has a JVM-only implementation

> Linux is particularly useful to use the library with Ktor Native on any Linux or Docker environment. Because Ktor Native is way smaller in binary size and memory consumption than Ktor JVM.

> Imagine I want to build a Desktop Native Windows app. Why shouldn't I be allowed to use Koog?

In practice, all of these use cases are already covered by the JVM target. Also, Compose doesn’t support these native targets

Another important point is that Kotlin officially treats [Linux as Tier 2](https://kotlinlang.org/docs/native-target-support.html#tier-2) and [Windows as Tier 3](https://kotlinlang.org/docs/native-target-support.html#tier-3) platforms. This means Kotlin does not test mingwX64 on CI, and we might encounter compiler issues when building Koog for Windows. To manage this risk, we would need to introduce a tiered support model within Koog as well. So that users have a clear understanding of what’s officially supported and what may be unstable. Otherwise, we’d tie our release stability to the state of Kotlin’s native targets, which could negatively impact our main JVM releases

Therefore, I suggest that in this PR we proceed only with iOS support. For the other native targets, we can open a separate issue to discuss support levels, the process for adding and releasing them, and associated maintenance implications

nathanfallet:
Okay, I'll comment out the other targets, so we can uncomment them later when we want to (but they will be ready)

nathanfallet:
I will take a look about adding tests for iOS (configuring GH Actions to run on macOS)

.github/workflows/checks.yml
```diff
@@ -98,6 +98,10 @@ jobs:
         if: matrix.os != 'ubuntu-latest'
         run: ./gradlew jvmTest --continue
 
+      - name: iosX64Test with Gradle Wrapper
```
devcrocod: I think the `iosSimulatorArm64` tests should also be added, right?

@aozherelyeva How will this affect our current testing?

agents/agents-features/agents-features-common/build.gradle.kts
```diff
@@ -35,6 +35,24 @@ kotlin {
...
+            dependencies {
+                api(libs.ktor.client.curl)
+            }
+        }
+
+        mingwMain {
+            dependencies {
+                api(libs.ktor.client.winhttp)
+            }
+        }
```
devcrocod: need to remove all such places

It’s strange that this doesn’t lead to an exception

agents/agents-utils/src/appleMain/kotlin/ai/koog/agents/utils/Coroutines.apple.kt
devcrocod, line 1: If use `native/`, it will cover all cases with Kotlin/Native

Or do you think we’ll need to split it into specific targets in the future?

buildSrc/src/main/kotlin/ai.kotlin.multiplatform.gradle.kts
```diff
@@ -16,6 +16,30 @@ plugins {
...
+    //watchosArm32()
+    //watchosArm64()
+    //tvosSimulatorArm64()
+    //tvosX64()
+    //tvosArm64()
+    iosArm64()
+
+    // Tier 3
+    //mingwX64()
+    //watchosDeviceArm64()
```
devcrocod: I don’t think leaving commented-out code is a good practice
At the very least, adding the targets to the build script in the future won’t be a difficult task

prompt/prompt-executor/prompt-executor-clients/prompt-executor-ollama-client/build.gradle.kts
```diff
@@ -31,6 +31,24 @@ kotlin {
...
+            dependencies {
+                api(libs.ktor.client.curl)
+            }
+        }
+
+        mingwMain {
+            dependencies {
+                api(libs.ktor.client.winhttp)
+            }
+        }
```
devcrocod: please drop it

prompt/prompt-executor/prompt-executor-clients/prompt-executor-ollama-client/src/appleMain/kotlin/ai/koog/prompt/executor/ollama/client/OllamaClient.apple.kt
devcrocod, line 1: It would be good to get rid of the explicit/actual here, since ktor can choose the required engine automatically
However, if you decide to keep it, you’ll also need to remove it for linux and mingw, or place it under `native/`

.github/workflows/checks.yml
zarechneva-a: The only failing test is `OpenAIEmbedderTest`, but I'm not sure whether it's a critical issue, as JVM target is running for it without a problem.

zarechneva-a: I'm not even sure we should run them for the iOS target. @nathanfallet WDYT?

devcrocod: > The only failing test is OpenAIEmbedderTest

In my opinion, this is an issue because the provider code has no platform-specific dependencies and should run on any target

I also checked that macos-latest refers to [macOS 14 on arm64](https://github.com/actions/runner-images?tab=readme-ov-file#available-images), so `iosX64Test` won’t run on it. gradle will just skip it, since for arm devices it should be `iosSimulatorArm64`

zarechneva-a: Hm, yes, I also see that `iosX64Test` is being skipped... good point! 
@nathanfallet could you take a look at the tests?

nathanfallet:
@aozherelyeva I think we're good! I only kept changes related to iOS (linux and mingw changes were totally removed) and I also updated CI to run iOS tests for arm instead of previous ones.

devcrocod:
I found the problem:
the `OpenAIEmbedderTest` tests are failing because in these tests the `MockOpenAIEmbedderClient` inherits from `OpenAIClient("")`. In Kotlin/Native this causes an error due to top-level initialization of the httpClient. If we change the inheritance to `LLMEmbeddingProvider`, like in the case of Ollama, it fixes the issue. However, in that case I’m not entirely sure about the correctness of the tests.

@aozherelyeva could you please advise?

aozherelyeva:
Thanks to @devcrocod, the problem is found and already solved. So now the `iosSimulatorArm64Test` target shall fully pass! 😃

devcrocod (approved):
finally 🎉

nicolasf:
Hi!

I can't build Koog on my Mac with these changes. I'm building from 8b313b05.
I didn't have time to debug this yet, but commenting out the ios targets fixes the issue for me.

Here's the build error:
```BRANCH_KOOG_IS_RELEASING_FROM=develop TC_BUILD_COUNTER=1 ./gradlew publishToMavenLocal
> Task :integration-tests:generateMetadataFileForIosArm64Publication FAILED

[Incubating] Problems report is available at: file:///Users/nicolas/Projects/koog/build/reports/problems/problems-report.html

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':integration-tests:generateMetadataFileForIosArm64Publication'.
> java.io.FileNotFoundException: /Users/nicolas/Projects/koog/integration-tests/build/libs/integration-tests-iosArm64Main-0.3.0.1.klib (No such file or directory)

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.
> Get more help at https://help.gradle.org.

Deprecated Gradle features were used in this build, making it incompatible with Gradle 9.0.

You can use '--warning-mode all' to show the individual deprecation warnings and determine if they come from your own scripts or plugins.

For more on this, please refer to https://docs.gradle.org/8.14.1/userguide/command_line_interface.html#sec:command_line_warnings in the Gradle documentation.

BUILD FAILED in 4s
125 actionable tasks: 8 executed, 4 from cache, 113 up-to-date
```

Might be some issue on my side, but I thought it was worth sharing.

nicolasf:
@nathanfallet 
The error happens when building koog itself locally from the develop branch.
I forgot to add the command I'm using, which explains why it shows as 0.3.0.1:
```
BRANCH_KOOG_IS_RELEASING_FROM=develop TC_BUILD_COUNTER=1 ./gradlew publishToMavenLocal
```
The error will also happen without the env vars, but for 0.3.0 instead.

If no one else has issues building it, it might be something on my environment.

Thanks.

PR #539, opened 2025-08-04
KG-187. Add Debugger feature

sdubov:
- Debugger feature is used to provide functionality for debugging AIAgent and collect metadata during agent execution. Currently, the feature provides a way to set up a local server and stream events (similar to the Tracing feature). It should be extended in the future for debugging purposes;
- Add tests to verify Debugger feature.
- 
---

#### Type of the change
- [x] New feature
- [ ] Bug fix
- [ ] Documentation fix
- [ ] Tests improvement

### DIFF
--- a/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/AIAgentPipeline.kt
+++ b/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/AIAgentPipeline.kt
@@ -158,7 +158,7 @@ public class AIAgentPipeline {
     internal suspend fun prepareFeatures() {
         withContext(featurePrepareDispatcher) {
             registeredFeatures.values.forEach { featureConfig ->
-                featureConfig.messageProcessor.map { processor ->
+                featureConfig.messageProcessors.map { processor ->
                     launch {
                         logger.debug { "Start preparing processor: ${processor::class.simpleName}" }
                         processor.initialize()
@@ -176,7 +176,7 @@ public class AIAgentPipeline {
      * ensuring resources are released appropriately.
      */
     internal suspend fun closeFeaturesStreamProviders() {
-        registeredFeatures.values.forEach { config -> config.messageProcessor.forEach { provider -> provider.close() } }
+        registeredFeatures.values.forEach { config -> config.messageProcessors.forEach { provider -> provider.close() } }
     }
 
     //region Trigger Agent Handlers
--- a/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/config/FeatureConfig.kt
+++ b/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/config/FeatureConfig.kt
@@ -11,18 +11,18 @@ import ai.koog.agents.core.feature.message.FeatureMessageProcessor
  */
 public abstract class FeatureConfig {
 
-    private val _messageProcessor = mutableListOf<FeatureMessageProcessor>()
+    private val _messageProcessors = mutableListOf<FeatureMessageProcessor>()
 
     /**
      * Provides a read-only list of `FeatureMessageProcessor` instances registered with the feature configuration.
      */
-    public val messageProcessor: List<FeatureMessageProcessor>
-        get() = _messageProcessor.toList()
+    public val messageProcessors: List<FeatureMessageProcessor>
+        get() = _messageProcessors.toList()
 
     /**
      * Adds a message processor to the configuration.
      */
     public fun addMessageProcessor(processor: FeatureMessageProcessor) {
-        _messageProcessor.add(processor)
+        _messageProcessors.add(processor)
     }
 }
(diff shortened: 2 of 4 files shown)


agents/agents-features/agents-features-debugger/src/jsMain/kotlin/ai/koog/agents/features/debugger/EnvironmentVariablesReader.js.kt
```diff
@@ -0,0 +1,18 @@
...
+
+@Suppress("EXPECT_ACTUAL_CLASSIFIERS_ARE_IN_BETA_WARNING")
+internal actual object EnvironmentVariablesReader {
+
+    private val logger = KotlinLogging.logger { }
+
+    internal actual fun getEnvironmentVariable(name: String): String? {
+        logger.debug { "Getting environment variable '$name'" }
+
+        val envVarValue = js("process.env[name]")
```
devcrocod: I’m not sure this will work in kotlin/js

agents/agents-features/agents-features-debugger/src/commonMain/kotlin/ai/koog/agents/features/debugger/feature/Debugger.kt
```diff
@@ -0,0 +1,265 @@
...
+     * This companion object provides functionality for configuring and enabling a debugging system
+     * in the AI agent framework. It logs debugging events, connects to a debugging server, and handles
+     * various stages of an agent's lifecycle such as start, finish, and error events.
+     * Debugger Feature also tracks strategy executions, node executions, LLM calls, and tool operation events.
+     *
+     * The feature can be customized using the `DebuggerConfig` and works in tandem with the `AIAgentPipeline` infrastructure
+     * to intercept various events and log them to a remote writer connected to a debugging server. The port for the debugger
+     * server can either be explicitly set in the configuration or derived from environment variables.
+     */
+    public companion object Feature : AIAgentFeature<DebuggerConfig, Debugger> {
```
devcrocod, line 56: Why is this construct used? Why not make Debugger an object in that case?

agents/agents-features/agents-features-debugger/src/commonMain/kotlin/ai/koog/agents/features/debugger/feature/DebuggerConfig.kt
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Configuration class for managing debugger-specific settings.
+ *
+ * This class extends the base `FeatureConfig` to enable the configuration of
+ * debugger-related parameters. It allows setting and retrieving the port
+ * number used by the debugger.
+ */
+public class DebuggerConfig : FeatureConfig() {
+
+    private var _port: Int? = null
```
devcrocod, line 14: Here, `port` can be made public.

devcrocod (approved):
lgtm
I have a couple of questions, could you please take a look

agents/agents-features/agents-features-debugger/src/jsMain/kotlin/ai/koog/agents/features/debugger/EnvironmentVariablesReader.js.kt
sdubov: Good point. I did some local testing and it seems it does not work as expected in JS. For now we need support for jvm. I've updated the code to throw the exception and will add this support when needed ([KG-194](https://youtrack.jetbrains.com/issue/KG-194/Add-support-for-Debugger-feature-for-JS-and-WASM-targets)).

agents/agents-features/agents-features-debugger/src/commonMain/kotlin/ai/koog/agents/features/debugger/feature/Debugger.kt
sdubov: This pattern is used widely for other features as well. Feature class is used as a state container, `Feature` companion object contains the installation logic for a particular. With this, you can use constructions like
```
AIAgent(...) {
    install(YourFeature) {
        // Feature configuration
    }
}

agents/agents-features/agents-features-debugger/src/commonMain/kotlin/ai/koog/agents/features/debugger/feature/DebuggerConfig.kt
sdubov: Thank you for the advise. I would prefer to set variables more explicitly via dedicated methods. I think the API is a bit cleaner.

devcrocod: In my opinion, this is exactly what Kotlin tried to get rid of, unnecessary get/set methods, especially for primitive types

agents/agents-features/agents-features-debugger/src/commonMain/kotlin/ai/koog/agents/features/debugger/feature/Debugger.kt
devcrocod: But wouldn’t it be possible to do that if we were using objects? Then we would have lazy initialization

To be honest, I don’t understand why static is used here

PR #550, opened 2025-08-06
Extract content metadata into a separate method

skarpovdev:
1. Extracted `content` into a separate method to make `metadata` and `isFile/isDirectory` faster.
2. Renamed `FileContent` to `FileContentType` to better reflect its purpose.
3. Covered the new method with tests.
4. Fixed typo in `MockFileSystemProvider` name.

~~**QUESTION 1** Please take a look at `getFileContentType` in `FilteredFileSystemProvider` and in~~ ~~`FileSystemProvider`. Should we make the return type nullable and use the same check as for `metadata`?~~
~~```~~
~~return if (filter.show(path, fs)) {~~
~~    fs.getFileContentType(path)~~
~~} else {~~
~~    null~~
~~}~~
~~```~~
~~Now it uses `requireAllowed(path)`.~~

~~**QUESTION 2** Should I just get rid of `Path.contentType()` and inline its logic in `getFileContentType` since it's~~ ~~the~~ only place where it's used and move `isFileHeadTextBased` to `Select`?~~
~~UPD: done in the latest commit to allow easy revert.~~

---

#### Type of the change
- [x] New feature
- [ ] Bug fix
- [ ] Documentation fix
- [x] Tests improvement

### DIFF
--- a/rag/rag-base/src/jvmTest/kotlin/ai/koog/rag/base/files/JVMFileSystemProviderTest.kt
+++ b/rag/rag-base/src/jvmTest/kotlin/ai/koog/rag/base/files/JVMFileSystemProviderTest.kt
@@ -181,8 +181,7 @@ class JVMFileSystemProviderTest : KoogTestBase() {
 
     @Test
     fun `test ReadOnly metadata`() = runBlocking {
-        val metadata =
-            FileMetadata(FileMetadata.FileType.File, hidden = false, content = FileMetadata.FileContent.Text)
+        val metadata = FileMetadata(FileMetadata.FileType.File, hidden = false)
         val testMetadata = readOnly.metadata(file1)
         assertEquals(metadata, testMetadata)
     }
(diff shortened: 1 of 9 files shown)


sproshev (changes requested):
see the answer for the first question

rag/rag-base/src/jvmTest/kotlin/ai/koog/rag/base/files/JVMFileSystemProviderTest.kt
```diff
@@ -189,17 +189,40 @@ class JVMFileSystemProviderTest : KoogTestBase() {
...
+    fun `test getFileContentType with directory throws IllegalArgumentException`() {
+        assertThrows<IllegalArgumentException> {
+            runBlocking { select.getFileContentType(dirEmpty) }
+        }
+    }
+
+    @Test
+    fun `test getFileContentType with Binary type`() = runBlocking {
+        val bytes = byteArrayOf(0xFF.toByte())
+        val binaryFile = dirEmpty.resolve("binary.bin").apply {
```
sproshev: I'd suggest using testArchive.zip or testImage.jpg from resources as a real-life examples

sproshev (approved):
thank you!

PR #560, opened 2025-08-08
Non-graph API

Rizzen:
This PR introduces new kind of agents API - ActAgent, allowing users to create simpler agents without working with graphs.

---

#### Type of the change
- [x] New feature
- [ ] Bug fix
- [ ] Documentation fix
- [ ] Tests improvement
- [ ] Refactoring


agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/ActAIAgent.kt
```diff
@@ -0,0 +1,211 @@
...
+
+    private val runningMutex = Mutex()
+
+    private fun <Config : FeatureConfig, Feature : Any> install(
+        feature: AIAgentNonGraphFeature<Config, Feature>,
+        configure: Config.() -> Unit
+    ) {
+        pipeline.install(feature, configure)
+    }
+
```
CaelumF:
```suggestion
    init {
        FeatureContext(this).featureContext()
    }
```

CaelumF: I might be missing something important/intentional but the features work for me when I add this :)

side note, I really like these interfaces!

Rizzen: @CaelumF Good catch, thank you!

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgents.kt
```diff
@@ -0,0 +1,96 @@
...
+ * @property strategy Strategy defining the local behavior of the agent.
+ * @property agentConfig Configuration details for the local agent that define its operational parameters.
+ * @property toolRegistry Registry of tools the agent can interact with, defaulting to an empty registry.
+ * @property installFeatures Lambda for installing additional features within the agent environment.
+ * @property clock The clock used to calculate message timestamps
+ *
+ * @see [AIAgent] class
+ */
+@OptIn(ExperimentalUuidApi::class)
+public inline fun <reified Input, reified Output> AIAgent(
```
Ololoshechkin, line 31: Now, as it's an external function it would be even harder to instantiate from Java.

Maybe let's keep everything inside `AIAgent` and make them separate constructors (and in Java they would be factory methods) ?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/ActAIAgent.kt
```diff
@@ -0,0 +1,213 @@
...
+ * tool registry, and strategy for handling operations.
+ *
+ * @param promptExecutor The `PromptExecutor` responsible for processing language model prompts and managing interactions with the AI model.
+ * @param agentConfig The `AIAgentConfigBase` defining the configuration for the AI agent, including prompts, model details, and iteration limits.
+ * @param toolRegistry A `ToolRegistry` specifying the tools available to the agent. If no tools are provided, it defaults to an empty registry.
+ * @param loop A suspendable lambda function representing the custom loop execution strategy. It takes an input of type `Input` and a context of type `AIAgentLoopContext`, and returns
+ *  an output of type `Output`.
+ * @return A `LoopAIAgent` instance initialized with the specified prompt executor, configuration, tool registry, and loop strategy.
+ */
+public fun <Input, Output> actAIAgent(
```
Ololoshechkin: Let's make the creation of AIAgent uniform -- so that the switching is easy, and learning curve is smoother:

You would always write:
```
AIAgent(
        promptExecutor = executor,
        strategy = <some strategy>,
        agentConfig = agentConfig,
        toolRegistry = toolRegistry
) 
```
(or the other constructor of `AIAgent` class)


And then -- only the strategy type would determine how the agent works. 

It would also allow experimenting and switching strategies much easier

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgentLoopStrategy.kt
```diff
@@ -0,0 +1,42 @@
...
+ * using the provided `name` and `loop` parameters. The `loop` function specifies the behavior
+ * of the agent within its execution loop. It is executed with the given input and the
+ * `AIAgentLoopContext` to produce an output.
+ *
+ * @param name The name of the strategy, describing its purpose or behavior.
+ * @param loop A suspending function representing the execution behavior of the agent in the loop.
+ *             It takes an input of type `Any?` and an `AIAgentLoopContext`, and produces an output of type `Any?`.
+ * @return An `AIAgentLoopStrategy` configured with the provided name and loop function.
+ */
+public fun <Input, Output> loopStrategy(name: String = "loopStrategy", loop: suspend AIAgentLoopContext.(input: Input) -> Output): AIAgentLoopStrategy<Input, Output> =
```
Ololoshechkin: Why is it a loop strategy?

Can't agents have non-loop design if written without graphs?

Ololoshechkin: Let's consider supporting the general case for non-graph strategies

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/ActAIAgent.kt
```diff
@@ -0,0 +1,213 @@
...
+ *
+ * @param Input The type of input data expected by the agent.
+ * @param Output The type of output data produced by the agent.
+ * @param id The unique identifier for the agent instance.
+ * @param promptExecutor The executor responsible for processing prompts and interacting with language models.
+ * @param agentConfig The configuration for the agent, including the prompt structure and execution parameters.
+ * @param toolRegistry The registry of tools available for the agent. Defaults to an empty registry if not specified.
+ */
+@ExperimentalUuidApi
+public class ActAIAgent<Input, Output>(
```
Ololoshechkin: Also, why is it an `ActAIAgent` ? Again, if it's a non-graph generic agent that can do anything it wants inside the strategy, it's not Act or ReAct -- so we probably need some more general name here

Ololoshechkin (changes requested):

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/ActAIAgent.kt
Ololoshechkin: We already got rid of `singleRunAgent` in favor of `AIAgent` being a single entry point. So let's probably keep up with that approach :)

Ololoshechkin:
```
AIAgent(
        promptExecutor = executor,
        strategy = loopStrategy { ... } , // or let's probably think about some better name. I don't have any good options, though
        agentConfig = agentConfig,
        toolRegistry = toolRegistry
) 
```

Ololoshechkin: And let's also make sure that all overrides of existing AIAgent constructor are supported

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/ActAIAgent.kt
Rizzen: Any suggestions on the name? We didn't have better name then ActAgent (NonGraphAgent sounds horrible for me)

Rizzen: @kpavlov I believe we need some fresh perspective here, may be you have better suggestions :)

docs/docs/act-ai-agent.md
```diff
@@ -0,0 +1,202 @@
...
+
+val result = agent.run("Say hi in one sentence")
+println(result)
+```
+
+What happens?
+- requestLLMMultiple(input) sends the user input and receives one or more assistant messages.
+- We return the only message’s content (typical one‑shot flow).
+
+Tip: If you want to return structured data, parse the content or use the Structured Data API.
```
sdubov, line 52: It would be nice to have a link to Structured Data API md file here.

docs/docs/act-ai-agent.md
```diff
@@ -0,0 +1,202 @@
...
+```kotlin
+class Switch {
+    private var on = false
+    fun on() { on = true }
+    fun off() { on = false }
+    fun isOn() = on
+}
+
+class SwitchTools(private val sw: Switch) {
+    fun turn_on() = run { sw.on(); "ok" }
```
sdubov, line 67: @Rizzen, no judgment, but why do you return "ok" here? :)

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/entity/AIAgentGraphStrategy.kt
```diff
@@ -0,0 +1,102 @@
+@file:Suppress("MissingKDocForPublicAPI")
```
sdubov: Leftover? Why do we need to suppress "missing KDock" warning? Maybe, it would be better to add some docs if needed?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/entity/AIAgentStrategy.kt
```diff
@@ -1,100 +1,44 @@
...
-    public val nodeFinish: FinishNode<Output>,
-    toolSelectionStrategy: ToolSelectionStrategy,
-    private val serializer: Json = Json { prettyPrint = true }
-) : AIAgentSubgraph<Input, Output>(
-    name,
-    nodeStart,
-    nodeFinish,
-    toolSelectionStrategy
-) {
+public interface AIAgentStrategy<Input, Output, TContext : AIAgentContext> {
```
sdubov: It might worth fixing this out of scope of this ticket, but I think it would be cleaner to have a common type naming pattern: 
`AIAgentStrategy<Input, Output, TContext : AIAgentContext>` -> `AIAgentStrategy<TInput, TOutput, TContext : AIAgentContext>`

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/ActAIAgent.kt
```diff
@@ -0,0 +1,213 @@
...
+    public val promptExecutor: PromptExecutor,
+    public val agentConfig: AIAgentConfigBase,
+    override val id: String = Uuid.random().toString(),
```
sdubov: Please see the comment and fix here https://github.com/JetBrains/koog/pull/644
The suggestion here is the same as provided in the PR, move override for `id` property out from constructor and use lazy evaluation:
```
public class ActAIAgent<Input, Output>(
    ...
    id: String? = null,
    ...
)

override val id: String by lazy { id ?: Uuid.random().toString() }
```

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
```diff
@@ -1,521 +1,21 @@
...
-    override val id: String = Uuid.random().toString(),
-    public val toolRegistry: ToolRegistry = ToolRegistry.EMPTY,
-    public val clock: Clock = Clock.System,
-    private val installFeatures: FeatureContext.() -> Unit = {},
-) : AIAgentBase<Input, Output>, AIAgentEnvironment, Closeable {
-
-    private companion object {
-        private val logger = KotlinLogging.logger {}
-    }
+public interface AIAgent<Input, Output> : Closeable {
```
sdubov, line 8: Love this change (`AIAgent` is an interface)!

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgentLoopContext.kt
```diff
@@ -0,0 +1,73 @@
...
+ * as well as operational constraints like iteration limits.
+ * @property llm The context for interacting with the large language model used by the agent, enabling message history
+ * retrieval and processing.
+ * @property stateManager The state management component responsible for tracking and updating the agent's state during its execution.
+ * @property storage A storage interface providing persistent storage capabilities for the agent's data.
+ * @property strategyName The name of the agent's strategic approach or operational method, determining its behavior
+ * during execution.
+ */
+@Suppress("UNCHECKED_CAST")
+public class AIAgentLoopContext(
```
sdubov: Why is it named "loop"? We do not run anything in the loop here as far as I understand. We just call `agent.run()` that execute a defined strategy. So, could you please clarify why "loop"?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgentLoopStrategy.kt
sdubov: I will be second here. The strategy for non-graph agent might not include any loops as far as I understand. Maybe, we can name it without connection to a loop (since you use `NonGraph` in some entities naming, maybe use this one, or... any that we consider appropriate. I would prefer to follow similarity in naming for entities that are split by some general common logic).

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgents.kt
sdubov, line 31: Should not we have one for `NonGraphAgent` here as well?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgentSimpleStrategies.kt
```diff
@@ -30,7 +30,7 @@ import ai.koog.agents.core.dsl.extension.onToolCall
  *                - SingleRunMode.PARALLEL: Executes multiple tool calls in parallel.
  * @return An instance of AIAgentStrategy configured according to the specified single-run mode.
  */
-public fun singleRunStrategy(runMode: ToolCalls = ToolCalls.SINGLE_RUN_SEQUENTIAL): AIAgentStrategy<String, String> =
+public fun singleRunStrategy(runMode: ToolCalls = ToolCalls.SINGLE_RUN_SEQUENTIAL): AIAgentGraphStrategy<String, String> =
```
sdubov, line 33: Should not we have the non-graph version for the same API as well now? Maybe, it might be useful to have API for non-graph API as well to keep the similarity across the library for both graph and non-graph options (whatever users prefer).

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/EssentialAIAgentContext.kt
```diff
@@ -0,0 +1,53 @@
...
+import ai.koog.agents.core.agent.context.AIAgentContext
+import ai.koog.agents.core.agent.context.AIAgentLLMContext
+import ai.koog.agents.core.agent.entity.AIAgentStateManager
+import ai.koog.agents.core.agent.entity.AIAgentStorage
+import ai.koog.agents.core.agent.entity.AIAgentStorageKey
+import ai.koog.agents.core.environment.AIAgentEnvironment
+import ai.koog.agents.core.feature.AIAgentFeature
+import ai.koog.prompt.message.Message
+
+/**
```
sdubov: Missing KDoc.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/EssentialAIAgentContext.kt
```diff
@@ -0,0 +1,53 @@
...
+ */
+public class EssentialAIAgentContext(
```
sdubov: Sorry, what does "essential" mean here in reference to graph and non-graph contexts? I would again refer to some similarity in entities naming.

agents/agents-features/agents-features-debugger/src/commonMain/kotlin/ai/koog/agents/features/debugger/feature/Debugger.kt
```diff
@@ -53,7 +53,7 @@ public class Debugger {
      * to intercept various events and log them to a remote writer connected to a debugging server. The port for the debugger
      * server can either be explicitly set in the configuration or derived from environment variables.
      */
-    public companion object Feature : AIAgentFeature<DebuggerConfig, Debugger> {
+    public companion object Feature : AIAgentGraphFeature<DebuggerConfig, Debugger> {
```
sdubov, line 55: Why is it a graph feature? I would expect to be able to use debugger feature for non-graph agents as well.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/builder/AIAgentEdgeBuilder.kt
```diff
@@ -69,7 +69,7 @@ public class AIAgentEdgeBuilderIntermediate<IncomingOutput, IntermediateOutput,
      */
     @EdgeTransformationDslMarker
     public infix fun onCondition(
-        block: suspend AIAgentContextBase.(output: IntermediateOutput) -> Boolean
+        block: suspend AIAgentContext.(output: IntermediateOutput) -> Boolean
```
sdubov: Here and below, since `AIAgentEdgeBuilder` is the graph-related class, maybe it worth using `AIAgentGraphContext` instead?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/builder/AIAgentStrategyBuilder.kt
```diff
@@ -20,12 +20,12 @@ public class AIAgentStrategyBuilder<Input, Output>(
     inputType: KType,
     outputType: KType,
     private val toolSelectionStrategy: ToolSelectionStrategy,
-) : AIAgentSubgraphBuilderBase<Input, Output>(), BaseBuilder<AIAgentStrategy<Input, Output>> {
+) : AIAgentSubgraphBuilderBase<Input, Output>(), BaseBuilder<AIAgentGraphStrategy<Input, Output>> {
```
sdubov: Same naming question to discuss - should not we name it `AIAgentGraphStrategyBuilder` now?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/environment/GenericAgentEnvironment.kt
```diff
@@ -0,0 +1,178 @@
...
+private class DirectToolCallsEnablerImpl : DirectToolCallsEnabler
+
+@OptIn(InternalAgentToolsApi::class)
+private class AllowDirectToolCallsContext(val toolEnabler: DirectToolCallsEnabler)
+
+@OptIn(InternalAgentToolsApi::class)
+private suspend inline fun <T> allowToolCalls(block: suspend AllowDirectToolCallsContext.() -> T) =
+    AllowDirectToolCallsContext(DirectToolCallsEnablerImpl()).block()
+
+internal class GenericAgentEnvironment(
```
sdubov, line 35: Again about naming, maybe it worth choose the same pattern in naming and use if through entities? (for now, I've faced options: "essential" and "generic"). Please let me know if I miss something.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/AIAgentFeature.kt
```diff
@@ -22,11 +22,22 @@ public interface AIAgentFeature<Config : FeatureConfig, TFeature : Any> {
...
+ *
+ * @param Config The type of configuration required for the feature, extending [FeatureConfig].
+ * @param TFeature The type representing the concrete implementation of the feature.
+ */
+public interface AIAgentGraphFeature<Config : FeatureConfig, TFeature : Any> : AIAgentFeature<Config, TFeature> {
     /**
      * Installs the feature into the specified [AIAgentPipeline].
      */
-    public fun install(config: Config, pipeline: AIAgentPipeline)
+    public fun install(config: Config, pipeline: AIAgentGraphPipeline)
```
sdubov, line 40: Here and below - this is not necessarily a graph-related feature. I would like to be able to use `EventHandler` feature with both graph and non-graph agents.

agents/agents-features/agents-features-tokenizer/src/commonMain/kotlin/ai/koog/agents/features/tokenizer/feature/MessageTokenizer.kt
```diff
@@ -164,7 +167,7 @@ public class MessageTokenizer(public val promptTokenizer: PromptTokenizer) {
      * This feature integrates a message tokenizer into the agent pipeline, allowing for tokenization
      * of input messages. It supports both caching and non-caching tokenization strategies based on the configuration.
      */
-    public companion object Feature : AIAgentFeature<MessageTokenizerConfig, MessageTokenizer> {
+    public companion object Feature : AIAgentGraphFeature<MessageTokenizerConfig, MessageTokenizer>, AIAgentNonGraphFeature<MessageTokenizerConfig, MessageTokenizer> {
```
sdubov, line 170: You implement both interfaces here Graph and non-graph, but you use the `AIAgentGraphPipeline` when install it. It feels not correct.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/model/message/Environment.kt
```diff
@@ -31,7 +31,7 @@ public data class AIAgentEnvironmentInitializeMessageContent(
 public data class AIAgentEnvironmentToAgentInitializeMessage(
     override val content: AIAgentEnvironmentInitializeMessageContent,
     val toolsForStages: Map<String, List<ToolDescriptor>>,
-    val agent: AIAgentStrategy<*, *>,
+    val agent: AIAgentGraphStrategy<*, *>,
```
sdubov: Could you please clarify why does the `AIAgentEnvironmentToAgentInitializeMessage` use `AIAgentGraphStrategy`? Can we use it only with graph-agents?

agents/agents-core/src/commonTest/kotlin/ai/koog/agents/core/agent/ActAgentTests.kt
```diff
@@ -0,0 +1,146 @@
...
+import ai.koog.agents.features.eventHandler.feature.EventHandler
+import ai.koog.agents.testing.tools.getMockExecutor
+import ai.koog.agents.testing.tools.mockLLMAnswer
+import ai.koog.prompt.llm.OllamaModels
+import kotlinx.coroutines.test.runTest
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+@OptIn(kotlin.uuid.ExperimentalUuidApi::class)
+class ActAgentTests {
```
sdubov: Do we expect to have more tests here? It seems that we need to check all general operations for non-graph agents as well. I would expect to see a number of tests that cover these cases. Not sure if we need to add this in the same PR, but it might be useful for testing.

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentStrategies.kt
```diff
@@ -16,12 +16,12 @@ import ai.koog.prompt.message.Message
...
 
 /**
- * Creates and configures a [ai.koog.agents.core.agent.entity.AIAgentStrategy] for executing a chat interaction process.
+ * Creates and configures a [ai.koog.agents.core.agent.entity.AIAgentGraphStrategy] for executing a chat interaction process.
  * The agent orchestrates interactions between different stages, nodes, and tools to
  * handle user input, execute tools, and provide responses.
  * Allows the agent to interact with the user in a chat-like manner.
  */
-public fun chatAgentStrategy(): AIAgentStrategy<String, String> = strategy("chat") {
+public fun chatAgentStrategy(): AIAgentGraphStrategy<String, String> = strategy("chat") {
```
sdubov, line 24: Should not we have an extra functions for non-graph strategy?

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt
```diff
@@ -204,7 +204,7 @@ public inline fun <reified Input, reified ProvidedResult : SubgraphResult> AIAge
     finishTool: ProvideSubgraphResult<ProvidedResult>,
     llmModel: LLModel? = null,
     llmParams: LLMParams? = null,
-    noinline defineTask: suspend AIAgentContextBase.(input: Input) -> String
+    noinline defineTask: suspend AIAgentContext.(input: Input) -> String
```
sdubov: Do we expect subgraph be a common entity or a graph-agents entity only? From the naming, I would expect to see it in graph agents only. So, here and below, can it be `AIAgentGraphContext`?

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/SubgraphWithRetry.kt
```diff
@@ -82,7 +82,7 @@ public val Boolean.asConditionResult: ConditionResult
  */
 @AIAgentBuilderDslMarker
 public inline fun <reified Input : Any, reified Output> AIAgentSubgraphBuilderBase<*, *>.subgraphWithRetry(
-    noinline condition: suspend AIAgentContextBase.(Output) -> ConditionResult,
+    noinline condition: suspend AIAgentContext.(Output) -> ConditionResult,
```
sdubov: Here, the same question, can we use `AIAgentGraphContext` instead?

agents/agents-features/agents-features-event-handler/src/commonMain/kotlin/ai/koog/agents/features/eventHandler/feature/EventHandler.kt
```diff
@@ -71,19 +75,45 @@ public class EventHandler {
...
             config: EventHandlerConfig,
-            pipeline: AIAgentPipeline,
+            pipeline: AIAgentGraphPipeline,
         ) {
             logger.info { "Start installing feature: ${EventHandler::class.simpleName}" }
 
             val featureImpl = EventHandler()
-            val interceptContext = InterceptContext(this, featureImpl)
-
-            //region Intercept Agent Events
```
sdubov, line 81: Sorry, why did you drop regions? :) We have a number of events now, it is quite hard to manage that we override everything now. Regions help here.

prompt/prompt-llm/src/commonMain/kotlin/ai/koog/prompt/llm/LLModel.kt
```diff
@@ -16,7 +16,7 @@ public data class LLModel(
     val provider: LLMProvider,
     val id: String,
     val capabilities: List<LLMCapability>,
-    val contextLength: Long,
+    val contextLength: Long = 0,
```
sdubov: Related change? Why do we have now a default value here?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/AIAgentGraphPipeline.kt
```diff
@@ -0,0 +1,260 @@
...
+
+    /**
+     * Notifies all registered handlers that an agent has started execution.
+     *
+     * @param runId The unique identifier for the agent run
+     * @param agent The agent instance for which the execution has started
+     * @param strategy The strategy being executed by the agent
+     */
+    @OptIn(InternalAgentsApi::class)
+    public suspend fun onBeforeAgentStarted(
```
sdubov, line 191: I think Agent specific events are quite common for both graph and non-graph agents.

sdubov (changes requested):
@Rizzen, I like the API in general. Looks really cool! 

I have some general concerns about the changes in features. Currently, all features need to implement both interfaces and register handlers twice. Even more, each feature should have a duplicated set of handlers if you what to use it in both graph and non-graph agents. This might be a bit overcomplicated. 

I would propose to use a bit different approach if you like it. In scope of events, the difference between graph and non-graph agents are in the events that agent execution triggers. For graph agents we have all these before/after node, etc. For non-graph API you have the same, but no node-related events. So, I would expect that if you install the same feature in a non-graph agent you will receive all events registered inside a feature, except nodes-related (as they are not fired on agent execution).

Taking this, you can leave current implementation for all feature as it is right now without any change. The same is true for Agent pipeline. You keep it the same with ability to register all events (graph and non-graph). So, feature can also register them all. On a `AIAgentPipeline` side you can add some logic (it can be a flag in constructor or use some other checks for context type for each handler) that will define what type of agent is that - graph or non-graph. Based on flag you can ignore some handlers that are not needed. This approach make Feature development as simple as it is right now (without duplicate handlers registration), but delegate this graph/non-graph responsibility to the pipeline class (which I think much easier to control).

Please let me know what do you think about proposed solution.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/ActAIAgent.kt
ptitjes: My take from highest to lowest preference ProceduralAgent, ProgramaticAgent, CodeBasedAgent, FunctionalAgent.

Rizzen:
@sdubov I don't really like the idea of "universal registration" you suggest.

The idea behind the current design is clear separation of concerns. When you develop a feature for either a graph or non-graph agent, you only deal with the relevant handlers. This keeps the mental model simple — you know exactly which events your feature can react to.

With a universal interface, you end up in situations like: “Okay, I have something called `afterNode`, let's try it out”, then you install the feature on a non-graph agent and... nothing happens. After some time digging through source code, you finally discover that some handlers in the interface you implemented are silently ignored because of an internal flag in the pipeline.
That kind of invisible behavior would be very frustrating to users — it's unpredictable and hard to debug.

This also opens the door to other confusing scenarios. For example, it would technically allow installing something like the `Persistency` feature into a non-graph agent. It would appear to install successfully but would silently do nothing, which is extremely unclear for users and could easily lead to subtle bugs.

From an implementation perspective, this would also require the non-graph agent to maintain a graph-like pipeline to satisfy all those extra handlers, even if they’re never actually used. That adds unnecessary complexity and coupling in places where it doesn’t belong.

The current design avoids these problems by using clear, explicit interfaces:

If you’re writing a `NonGraphFeature`, you only see the relevant events, and everything you implement will always fire.

The type system enforces compatibility — you simply cannot install a non-graph feature into a graph agent or vice versa. This makes incorrect usage impossible at compile time, rather than a silent runtime no-op.

This makes the system much more predictable and maintainable. A developer always knows exactly what events are available, what will trigger, and what won’t. It also keeps the boundaries between graph and non-graph agents crisp, without forcing one to emulate the internals of the other just to support a universal abstraction. Yes, it requires to impement two `installs` in case of features that should work on both types of agents, but it's lesser price for clear interfaces and type safety.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/ActAIAgent.kt
sdubov: It seems to be quite hard naming question 🙂 I will also give some examples from my side:
- `ExecAgent` (execution agent; I know, graph is also executed)`
- `FlatAgent`
- `SequentialAgent`
- `CommandsAgent` (vs. `GraphAgent` )

I like `ProceduralAgent` by @ptitjes from the meaning side, but I found it is quite hard to pronounce.

and suggestion from @prokashev: `LinearAgent` sounds really cool as well, but it might be hard to understand from the scratch what does it mean.

ptitjes:
Just a note to say that the current persistency feature stores not only the current graph node but also the history and the agent state. I would hope that the persistency feature would allow for explicit checkpoint creation/recall from a programatic agent. If it's necessary to decouple the addition of the current graph node to the checkpoint, through an extension point, then it's the way to go IMO.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/ActAIAgent.kt
Rizzen: Well among all the variants I would prefer `FunctionalAgent` - because it's actually what it is - simple a function passed into the agent. And it matches with Kotlin `fun` (for function) keyword.

docs/docs/act-ai-agent.md
Rizzen: Just to message llm about success of operation :)

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgentSimpleStrategies.kt
Rizzen: I'm not sure if we really want it, singleRunStrategy is a simple graph strategy and we don't have non-graph equivalent for it.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/EssentialAIAgentContext.kt
Rizzen: it's a leftover, to delete

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/builder/AIAgentStrategyBuilder.kt
Rizzen: I think we should, yeah.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/environment/GenericAgentEnvironment.kt
Rizzen: `Essential` is a leftower, `GenericAgentEnvironment` is a default implementation for AgentEnvironment, I didn't come up with better name so far (`CommonAgentEnvironment` is also option, but it's pretty much the same)

agents/agents-features/agents-features-tokenizer/src/commonMain/kotlin/ai/koog/agents/features/tokenizer/feature/MessageTokenizer.kt
Rizzen: It has two installs - for Graph and Non Graph pipelines

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/AIAgentFeature.kt
Rizzen: Current `EventHandler` implementation is closely tied to graph, we need to port it to non-graph agent. Yes, we want to do it, but I would prefer to do it separately

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/model/message/Environment.kt
Rizzen: Well, turns out that that class is unused at all :D

Rizzen: Ill remove it

agents/agents-core/src/commonTest/kotlin/ai/koog/agents/core/agent/ActAgentTests.kt
Rizzen: I expect a number of integration tests to be written separately

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentStrategies.kt
Rizzen: We don't have plans to have some generic non-graph chat agent strategy at the moment

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/AIAgentFeature.kt
Rizzen: EventHandler also has two separate installs

agents/agents-features/agents-features-event-handler/src/commonMain/kotlin/ai/koog/agents/features/eventHandler/feature/EventHandler.kt
Rizzen: Well, I consider the regions as necessary evil when you have REALLY HUGE class and cannot do better then introduce regions. But here we have only 200 lines, it's not a lot

prompt/prompt-llm/src/commonMain/kotlin/ai/koog/prompt/llm/LLModel.kt
Rizzen: Unrelated

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/AIAgentGraphPipeline.kt
Rizzen: Unfortunately, `onBeforeAgentStarted` knows a lot about internals of agent, such as existence of graph, so it was moved to graph specific pipeline. I believe we need separate onBeforeAgentStarted in non-graph pipeline

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgentFunctionalStrategy.kt
```diff
@@ -0,0 +1,42 @@
...
+    ): TOutput = context.loop(input)
+}
+
+/**
+ * Creates an instance of a loop strategy for an AI agent.
+ *
+ * This function constructs and returns a specific implementation of `AIAgentLoopStrategy`
+ * using the provided `name` and `loop` parameters. The `loop` function specifies the behavior
+ * of the agent within its execution loop. It is executed with the given input and the
+ * `AIAgentLoopContext` to produce an output.
```
sdubov: Here and in other places, do not forget to update KDoc as well with a new naming.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgentFunctionalStrategy.kt
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * @param name The name of the strategy, describing its purpose or behavior.
+ * @param loop A suspending function representing the execution behavior of the agent in the loop.
+ *             It takes an input of type `Any?` and an `AIAgentLoopContext`, and produces an output of type `Any?`.
+ * @return An `AIAgentLoopStrategy` configured with the provided name and loop function.
+ */
+public fun <Input, Output> functionalStrategy(name: String = "funStrategy", loop: suspend AIAgentFunctionalContext.(input: Input) -> Output): AIAgentFunctionalStrategy<Input, Output> =
+    AIAgentFunctionalStrategy(name, loop)
```
sdubov: Parameter name should be updated.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/FunctionalAIAgent.kt
```diff
@@ -0,0 +1,218 @@
...
+ * @param agentConfig The configuration for the agent, including the prompt structure and execution parameters.
+ * @param toolRegistry The registry of tools available for the agent. Defaults to an empty registry if not specified.
+ */
+@ExperimentalUuidApi
+public class FunctionalAIAgent<Input, Output>(
+    public val promptExecutor: PromptExecutor,
+    public val agentConfig: AIAgentConfigBase,
+    public val toolRegistry: ToolRegistry = ToolRegistry.EMPTY,
+    public val strategy: AIAgentFunctionalStrategy<Input, Output>,
+    private val _id: String? = null,
```
sdubov: Nit: you do not need to define it as a class field as we do not want to use the `_id` anywhere inside the class. We rather should use `id` only. You can just pass it as a parameter with name `id`:
```
public val strategy: AIAgentFunctionalStrategy<Input, Output>,
id: String? = null,
public val clock: Clock = Clock.System,
...
...
```

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/GraphAIAgent.kt
```diff
@@ -0,0 +1,230 @@
...
+ */
+@OptIn(ExperimentalUuidApi::class)
+public open class GraphAIAgent<Input, Output>(
+    public val inputType: KType,
+    public val outputType: KType,
+    public val promptExecutor: PromptExecutor,
+    public val agentConfig: AIAgentConfigBase,
+    public val toolRegistry: ToolRegistry = ToolRegistry.EMPTY,
+    private val strategy: AIAgentGraphStrategy<Input, Output>,
+    override val id: String = Uuid.random().toString(),
```
sdubov: Since you add this `GraphAIAgent` as a new file, I assume that you need to add the fix for nullable String `id` constructor parameter here as well. Please see the [comment](https://github.com/JetBrains/koog/pull/560#discussion_r2318935726)

agents/agents-core/src/commonTest/kotlin/ai/koog/agents/core/agent/ActAgentTests.kt
sdubov: Could you please rename the test class accordingly?
`ActAgentTests` -> `FunctionalAIAgentTest`

agents/agents-core/src/commonTest/kotlin/ai/koog/agents/core/agent/ActAgentTests.kt
```diff
@@ -0,0 +1,146 @@
...
+    @Test
+    fun test_ActAgent_MixedResults() = runTest {
```
sdubov: ... also, test names would be nice to update to have a easy-to-read mapping between business logic and tests.

examples/src/main/kotlin/ai/koog/agents/example/nongraph/ActAgentWithTools.kt
```diff
@@ -0,0 +1,59 @@
+package ai.koog.agents.example.nongraph
```
sdubov: File need to be renamed as well.
`ActAgentWithTools` -> `FunctionalAgentWIthTools`

examples/src/main/kotlin/ai/koog/agents/example/nongraph/SimpleActAgent.kt
```diff
@@ -0,0 +1,23 @@
+package ai.koog.agents.example.nongraph
```
sdubov: Could you please update the file according to the new naming?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/entity/AIAgentStrategy.kt
```diff
@@ -1,100 +1,44 @@
...
-    public val nodeFinish: FinishNode<Output>,
-    toolSelectionStrategy: ToolSelectionStrategy,
-    private val serializer: Json = Json { prettyPrint = true }
-) : AIAgentSubgraph<Input, Output>(
-    name,
-    nodeStart,
-    nodeFinish,
-    toolSelectionStrategy
-) {
+public interface AIAgentStrategy<TInput, TOutput, TContext : AIAgentContext> {
```
sdubov, line 17: Thanks for updating type names! ❤️

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt
```diff
@@ -287,7 +288,7 @@ public inline fun <reified Input, reified ProvidedResult : SubgraphResult> AIAge
...
```
sdubov: Here and below, does it make sense to update to `AIAgentGraphContextBase` as well?

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/SubgraphWithRetry.kt
```diff
@@ -94,7 +95,7 @@ public inline fun <reified Input : Any, reified Output> AIAgentSubgraphBuilderBa
     return subgraph(name = name) {
         val retriesKey = createStorageKey<Int>("${name}_retires")
         val initialInputKey = createStorageKey<Any>("${name}_initial_input")
-        val initialContextKey = createStorageKey<AIAgentContextBase>("${name}_initial_context")
+        val initialContextKey = createStorageKey<AIAgentContext>("${name}_initial_context")
```
sdubov: Here and below, `AIAgentGraphContextBase` ?

agents/agents-ext/src/commonTest/kotlin/ai/koog/agents/ext/agent/SubgraphWithRetryTest.kt
```diff
@@ -22,7 +22,7 @@ import kotlin.test.assertTrue
 
 private const val MAX_AGENT_ITERATIONS = 20
 private const val SUCCESS = "success"
-private val TEST_CONDITION: AIAgentContextBase.(String) -> ConditionResult = { (it == SUCCESS).asConditionResult }
+private val TEST_CONDITION: AIAgentContext.(String) -> ConditionResult = { (it == SUCCESS).asConditionResult }
```
sdubov: `AIAgentGraphContextBase` ?

docs/docs/act-ai-agent.md
```diff
@@ -0,0 +1,202 @@
+# ActAIAgent: How to build a single‑run agent step by step
```
sdubov: I assume this file name need to be updated to `functional-ai-agent.md`. As well as all examples should use updated classes.

examples/src/main/kotlin/ai/koog/agents/example/chat/ActAgentChat.kt
```diff
@@ -0,0 +1,28 @@
+package ai.koog.agents.example.chat
```
sdubov: Could you please update the file name according to the new naming?

sdubov (changes requested):
There are some outdated naming left in a codebase. Other than that, I think it can go to the branch from my vision. Thank you!

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/ActAIAgent.kt
ptitjes: Yeah, I had put FunctionalAgent at the end of my list because I was afraid it might imply to some users that they are "pure", i.e. without side-effect, which they obviously are not. But functions in Kotlin are not pure, so yeah it's not the worst name.

Ololoshechkin (approved):

Ololoshechkin:
Let's make follow-up cosmetic changes in the future PRs :)
Conceptually, this one looks good.

examples/src/main/kotlin/ai/koog/agents/example/funApi/FunAgentExample.kt
```diff
@@ -0,0 +1,23 @@
...
+import ai.koog.agents.core.agent.functionalAIAgent
+import ai.koog.agents.core.agent.requestLLMMultiple
+import ai.koog.prompt.executor.llms.all.simpleOllamaAIExecutor
+import ai.koog.prompt.llm.OllamaModels
+import kotlinx.coroutines.runBlocking
+import kotlin.uuid.ExperimentalUuidApi
+
+@OptIn(ExperimentalUuidApi::class)
+fun main(): Unit = runBlocking {
+    val actAgent = functionalAIAgent<String, String>(
```
sdubov: `actAgent` -> `functionalAgent`

examples/src/main/kotlin/ai/koog/agents/example/funApi/FunAgentWithTools.kt
```diff
@@ -0,0 +1,59 @@
...
+
+@OptIn(ExperimentalUuidApi::class)
+fun main(): Unit = runBlocking {
+    val promptExec = simpleOllamaAIExecutor()
+    val switch = Switch()
+    val toolRegistry = ToolRegistry {
+        tools(SwitchTools(switch).asTools())
+    }
+
+    val actAgent = functionalAIAgent<String, String>(
```
sdubov: `actAgent` -> `functionalAgent`

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/FunctionalAIAgent.kt
```diff
@@ -0,0 +1,218 @@
...
+ * @param toolRegistry A `ToolRegistry` specifying the tools available to the agent. If no tools are provided, it defaults to an empty registry.
+ * @param loop A suspendable lambda function representing the custom loop execution strategy. It takes an input of type `Input` and a context of type [AIAgentFunctionalStrategy], and returns
+ *  an output of type `Output`.
+ * @return A [FunctionalAIAgent] instance initialized with the specified prompt executor, configuration, tool registry, and loop strategy.
+ */
+public fun <Input, Output> functionalAIAgent(
+    promptExecutor: PromptExecutor,
+    agentConfig: AIAgentConfigBase,
+    toolRegistry: ToolRegistry = ToolRegistry.EMPTY,
+    loop: suspend AIAgentFunctionalContext.(input: Input) -> Output
```
sdubov: `loop` -> `func`

examples/src/main/kotlin/ai/koog/agents/example/chat/FunctionalAgentChat.kt
```diff
@@ -0,0 +1,28 @@
...
+import ai.koog.agents.core.agent.functionalAIAgent
+import ai.koog.agents.core.agent.requestLLM
+import ai.koog.prompt.executor.llms.all.simpleOllamaAIExecutor
+import ai.koog.prompt.llm.OllamaModels
+import kotlinx.coroutines.runBlocking
+import kotlin.uuid.ExperimentalUuidApi
+
+@OptIn(ExperimentalUuidApi::class)
+fun main(): Unit = runBlocking {
+    val loopAgent = functionalAIAgent<String, Unit>(
```
sdubov: `loopAgent` -> `funcAgent`.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/FunctionalAIAgent.kt
```diff
@@ -0,0 +1,218 @@
...
+        agentConfig = agentConfig,
+        toolRegistry = toolRegistry,
+        strategy = functionalStrategy(
+            func = loop
+        )
+    )
+}
+
+/**
+ * Creates a `LoopAIAgent` that continuously performs actions based on the input, a context, and a defined loop logic.
```
sdubov: Fix KDoc.

sdubov (approved):
All major comments are fixed. There are some other places with outdated comments and KDoc that needs to be updated as well. I've added several places that need to be fixed. @Rizzen you can just search for "loop" or "actAgent" text and fix it here or later. Other that that, LGTM!

Issue #589, opened 2025-08-15
[Feature Request] Enhance Structured Outputs with JSON Schema Constraints (minimum/maximum, minLength/maxLength)


OcelusPRO:
Hello,

I'm using the structured output feature with language models, and it's incredibly powerful. Currently, the library automatically determines the type and name of each field, allowing us to add a `description`. This is a great start!

However, I'd like to propose an enhancement to make this feature even more robust by incorporating more advanced JSON Schema validation constraints. Specifically, it would be very beneficial to be able to define constraints for numbers and strings.

### Proposed Enhancement

**For Numbers**
It would be fantastic to specify a valid range for numeric values using the `minimum` and `maximum` properties of JSON Schema. This could be elegantly implemented using an annotation.

**For Strings**
Similarly, for strings, it's often crucial to control the length of the generated text. This can be achieved with the `minLength` and `maxLength` properties.

Proposed Syntax:

```kotlin
@Serializable
@LLMDescription("The sample function")
fun sampleFun(
    @LLMRange(min = 0, max = 5)  val score: Int,
    @LLMLength(minLength= 250, maxLength = 500) val explanation: String
)
```

This would translate to the following JSON Schema:

```json
{
  "name": "sampleFun",
  "description": "The sample function",
  "parameters": {
    "type": "object",
    "properties": {
      "score": {
        "type": "integer",
        "minimum": 0,
        "maximum": 5
      },
      "explanation": {
        "type": "string",
        "minLength": 250,
        "maxLength": 500
      }
    },
    "required": ["score", "explanation"]
  }
}
```

## Why is this important?
Adding these constraints offers several significant advantages:

1. Improved Reliability: It provides a stronger guarantee on the shape of the data returned by the model. For example, when asking an AI to rate a customer's review on a scale of 0 to 5, providing this constraint directly in the schema prevents the model from "hallucinating" and returning a value outside this range (e.g., 8 or 50). It enforces the output format much more strictly than just describing it in the prompt.

2. Practical Use Cases: This is very useful for real-world applications where the output needs to fit specific constraints, such as:
  - Ensuring a generated summary fits within a UI dialog box (maxLength).
  - Guaranteeing a numerical value corresponds to a valid range in a system (e.g., a percentage from 0 to 100).

3. Better Model Guidance: Modern language models are trained to understand and respect JSON Schema constraints. By providing a more detailed schema, we can guide the model more effectively, leading to more accurate and predictable results from the very first generation.

[...]

PR #590, opened 2025-08-16
Enhance Structured Outputs with JSON Schema Constraints (minimum/maximum, minLength/maxLength)

OcelusPRO:
As it was described in #589 , the goal is to add some additional constraints in the Json schemas, all the main information is indicated in the issue

This PR adds support for the following constraints:

- `minimum`/`maximum` for numbers
- `minLength`/`maxLength` for strings
- `minContains`/`maxContains` for arrays

---

#### Type of the change
- [x] New feature
- [ ] Bug fix
- [ ] Documentation fix
- [ ] Tests improvement
- [ ] Refactoring

- [x] The pull request includes a link to the issue -> Related to #589 


agents/agents-tools/src/commonMain/kotlin/ai/koog/agents/core/tools/annotations/LLMMax.kt
```diff
@@ -0,0 +1,20 @@
...
+ * to enforce a constraint that the value must not exceed the specified maximum value.
+ *
+ * @property max The maximum allowable value.
+ */
+@SerialInfo
+@Target(
+    AnnotationTarget.PROPERTY,
+    AnnotationTarget.VALUE_PARAMETER,
+)
+public annotation class LLMMax(val max: Int)
```
EugeneTheDev, line 20: For this and other annotations:

1. Since they are applicable only for structured output (unlike `@LLMDescription`, which is also used for tools), let's not put these in tools module. There's already an `annotations` package in `prompt-structure` module, let's put it there. 
2. Also, the naming is a bit confusing, since it doesn't state it explicitly that it's for a JSON generation only. Can we maybe expand a bit, something like: `LLMStructureMax`

prompt/prompt-structure/src/commonMain/kotlin/ai/koog/prompt/structure/json/generator/JsonSchemaConsts.kt
```diff
@@ -20,6 +20,12 @@ public object JsonSchemaConsts {
         public const val REQUIRED: String = "required"
         public const val NULLABLE: String = "nullable"
         public const val DESCRIPTION: String = "description"
+        public const val MIN: String = "minimum"
```
EugeneTheDev, line 23: Did you check support for such properties in [OpenAI](https://platform.openai.com/docs/guides/structured-outputs) and [Google AI](https://ai.google.dev/gemini-api/docs/structured-output) APIs? They usually list properties from JSON Schema they support. If these are not supported (which I think might be the case at least for some), let's handle it explicitly in Google and OpenAI schema generators and throw the exception explicitly, stating that these properties are not supported? You can check respective LLM clients' modules for these generators and use existing implementation for reference, to add these exceptions

prompt/prompt-structure/src/commonMain/kotlin/ai/koog/prompt/structure/json/generator/JsonSchemaGenerator.kt
```diff
@@ -78,6 +81,112 @@ public abstract class JsonSchemaGenerator {
...
+        /**
+         * Retrieves the maximum allowable numeric value for the descriptor type based on the `@LLMMax`
+         * or `@LLMRange` annotations. If both annotations are present, an exception is thrown.
+         * If neither annotation is found, returns null.
+         *
+         * @return The maximum numeric value defined by the `@LLMMax` or `@LLMRange` annotation,
+         * or null if no maximum value is specified.
+         * @throws IllegalStateException if both `@LLMMax` and `@LLMRange` annotations are present.
+         */
+        public fun getTypeMax(): Int? {
```
EugeneTheDev, line 95: You don't need `getType` methods for annotations like min/max, since these are applicable to strings/numbers, and you can't really annotate existing library data types. You would use these annotations on class elements only.

prompt/prompt-structure/src/commonMain/kotlin/ai/koog/prompt/structure/json/generator/GenericJsonSchemaGenerator.kt
```diff
@@ -68,9 +68,49 @@ public abstract class GenericJsonSchemaGenerator : JsonSchemaGenerator() {
         }
     }
 
+    protected fun JsonObjectBuilder.putMax(max: Int, type: String) {
```
EugeneTheDev, line 71: This method is actually not needed (see comments below)

prompt/prompt-structure/src/commonMain/kotlin/ai/koog/prompt/structure/json/generator/GenericJsonSchemaGenerator.kt
```diff
@@ -68,9 +68,49 @@ public abstract class GenericJsonSchemaGenerator : JsonSchemaGenerator() {
...
+        when (type) {
+            JsonSchemaConsts.Types.STRING -> put(JsonSchemaConsts.Keys.MAX_LENGTH, max)
+            JsonSchemaConsts.Types.NUMBER -> put(JsonSchemaConsts.Keys.MAX, max)
+            JsonSchemaConsts.Types.INTEGER -> put(JsonSchemaConsts.Keys.MAX, max)
+            JsonSchemaConsts.Types.ARRAY -> put(JsonSchemaConsts.Keys.MAX_ITEMS, max)
+            else -> throw IllegalArgumentException("Unsupported type for max: $type")
+        }
+    }
+    protected fun JsonObjectBuilder.putMin(min: Int, type: String) {
```
EugeneTheDev, line 80: Also not needed (see below)

prompt/prompt-structure/src/commonMain/kotlin/ai/koog/prompt/structure/json/generator/GenericJsonSchemaGenerator.kt
```diff
@@ -68,9 +68,49 @@ public abstract class GenericJsonSchemaGenerator : JsonSchemaGenerator() {
...
+        when (type) {
+            JsonSchemaConsts.Types.STRING -> put(JsonSchemaConsts.Keys.MIN_LENGTH, min)
+            JsonSchemaConsts.Types.NUMBER -> put(JsonSchemaConsts.Keys.MIN, min)
+            JsonSchemaConsts.Types.INTEGER -> put(JsonSchemaConsts.Keys.MIN, min)
+            JsonSchemaConsts.Types.ARRAY -> put(JsonSchemaConsts.Keys.MIN_ITEMS, min)
+            else -> throw IllegalArgumentException("Unsupported type for min: $type")
+        }
+    }
+
+    private fun minMaxCheck(min: Int?, max: Int?, rejectNegative: Boolean): Error? {
```
EugeneTheDev, line 90: Also not needed (see below)

prompt/prompt-structure/src/commonMain/kotlin/ai/koog/prompt/structure/json/generator/GenericJsonSchemaGenerator.kt
```diff
@@ -125,6 +189,43 @@ public abstract class GenericJsonSchemaGenerator : JsonSchemaGenerator() {
         putDescription(context.currentDescription)
     }
 
+    private fun applyConstraints(
```
EugeneTheDev, line 192: Also not needed (see below)

prompt/prompt-structure/src/commonMain/kotlin/ai/koog/prompt/structure/json/generator/GenericJsonSchemaGenerator.kt
```diff
@@ -150,18 +251,24 @@ public abstract class GenericJsonSchemaGenerator : JsonSchemaGenerator() {
...
+                    val elementMax = context.getElementMax(i)
+
+                    val propertyContext = context.copy(
+                        descriptor = propertyDescriptor,
+                        currentDefPath = context.currentDefPath + context.descriptor,
+                        currentDescription = context.getElementDescription(i)
+                            ?: context.copy(descriptor = propertyDescriptor).getTypeDescription()
                     )
+
+                    val processedProperty = if (elementMin != null || elementMax != null) {
```
EugeneTheDev, line 264: I think the same idea can be implemented in a cleaner manner:
1. Introduce optional fields, something like `minimum` and `maximum`, to the `GenerationContext` class, just like it is done now with `currentDescription`. 
2. Now, the check that these values are of a correct range can be done only here, since this is the only place where we extract these values (hence separate method for checking is not needed)
3. Also, these separate `applyConstraints` and `putMin`/ `putMax` can be avoided too, because now, with expanded `GenerationContext`, we can get these values directly from it and check and apply them in place, similar to what you do with `context.getTypeMin()` in these `put...` methods now (only using getType is wrong, for the reasons I described in the respective comment above)

TL:DR - follow the `currentDescription` approach, since it is similar to what you want to achieve here (it's also an annotation value that we want to apply as an additional property in the final schema for each object property). Then the implementation should look cleaner, shorter and without additional methods.

prompt/prompt-structure/src/commonTest/kotlin/ai/koog/prompt/structure/json/generator/JsonSchemaGeneratorTest.kt
```diff
@@ -162,6 +167,181 @@ class JsonSchemaGeneratorTest {
...
+        val array: List<String>
+    )
+
+
+    @Test
+    fun testIntMin() {
+        val result = basicGenerator.generate(json, "IntMinTest", serializer<IntMinTest>(), emptyMap())
+        val schema = json.encodeToString(result.schema)
+
+        assertContains(schema, "\"minimum\": 5")
```
EugeneTheDev, line 257: I would prefer a full resulting JSON check in asserts with `assertEquals`, not just contains. Just to be sure and to make it more visible, what the final result is supposed to look like

EugeneTheDev (changes requested):
Thanks, the idea is interesting, however, the implementation still requries some polishing and clarifications.

PR #593, opened 2025-08-18
KG-259. Open Telemetry should mask sensitive event fields and attributes instead of skip them

sdubov:
Open Telemetry sensitive data should not be skipped when creating evens and attributes. We rather need to mask them to not lose important data. Replace all sensitive data with the special HiddenString type.

---

#### Type of the change
- [ ] New feature
- [x] Bug fix
- [ ] Documentation fix
- [ ] Tests improvement
- [ ] Refactoring

### DIFF
--- a/agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/span/SpanProcessor.kt
+++ b/agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/span/SpanProcessor.kt
@@ -14,7 +14,10 @@ import java.util.concurrent.locks.ReentrantReadWriteLock
 import kotlin.concurrent.read
 import kotlin.concurrent.write
 
-internal class SpanProcessor(private val tracer: Tracer) {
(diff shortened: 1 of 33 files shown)


agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/span/SpanProcessor.kt
```diff
@@ -14,7 +14,10 @@ import java.util.concurrent.locks.ReentrantReadWriteLock
 import kotlin.concurrent.read
 import kotlin.concurrent.write
 
-internal class SpanProcessor(private val tracer: Tracer) {
+internal class SpanProcessor(
+    private val tracer: Tracer,
+    private val verbose: Boolean
```
tiginamaria: Maybe set `true` by default?

sdubov: Good point. I would only set to `false` value to make sure we do not expose any sensitive information by default.

PR #601, opened 2025-08-19
KG-49: Fix enum parsing when type is not mentioned

tiginamaria:
This pull request fixes enum tool parsing for Intellij MCP server tools where the tool argument have enum type and does not mention the type of the enums assuming string by default

Fixes: https://youtrack.jetbrains.com/issue/KG-244

## Breaking Changes

---

#### Type of the changes
- [ ] Documentation update
- [ ] Tests improvement
- [ ] Refactoring

#### Checklist

### DIFF
--- a/agents/agents-mcp/src/jvmMain/kotlin/ai/koog/agents/mcp/McpToolDefinitionParser.kt
+++ b/agents/agents-mcp/src/jvmMain/kotlin/ai/koog/agents/mcp/McpToolDefinitionParser.kt
@@ -91,6 +91,25 @@ public object DefaultMcpToolDescriptorParser : McpToolDescriptorParser {
                     return parseParameterType(nonNullType, depth + 1)
                 }
             }
+
+            /**
+             * Special case for enum string types.
+             * Schema example:
+             * {
+             *   "enumParam": {
+             *     "enum": [
+             *       "value1",
+             *       "value2"
+             *     ],
+             *     "title": "Enum string parameter"
+             *   }
+             * }
+             */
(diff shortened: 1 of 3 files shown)


agents/agents-mcp/src/jvmMain/kotlin/ai/koog/agents/mcp/McpToolDefinitionParser.kt
```diff
@@ -91,6 +91,25 @@ public object DefaultMcpToolDescriptorParser : McpToolDescriptorParser {
...
+             *   "enumParam": {
+             *     "enum": [
+             *       "value1",
+             *       "value2"
+             *     ],
+             *     "title": "Enum string parameter"
+             *   }
+             * }
+             */
+            val enum = element["enum"]?.jsonArray
```
devcrocod, line 108: Sorry, maybe I don’t quite understand. But is this example actually valid? Because usually an enum looks like this:

```json
{
  "type": "string",
  "enum": ["red", "green", "blue"]
}
```

devcrocod: As far as I know, title is an optional field, but type is required

devcrocod (commented):
Okay, I read the comment. It clarified a bit of what’s going on here.

In that case, I don’t think we should make a workaround for this case, since it goes against JSON Schema. If everything else is correct, then the issue is in how the enum is being formed. So this needs to be fixed either on the mcp-sdk side or on the server side

PR #652, opened 2025-08-25
KG-159: Make `parts` field nullable because Gemini sometimes doesn't return it

Ololoshechkin:
## Breaking Changes

---

#### Type of the changes
- [ ] Documentation update
- [ ] Tests improvement
- [ ] Refactoring

#### Checklist


antoniibelyshev (approved):
Thank you! I would suggest dropping a warning message in case of an empty `parts` (and empty ` content` too).

AFAIU, missing parts field is not a good sign: for example, it could result in an empty list of responses, which is not a supposed behavior.

Ololoshechkin:
@antoniibelyshev thanks! I'll add a warning.
Indeed, it's not a supposed behavior, but apparently, that's what Gemini Flash 2.5 and Pro 2.5 do from time to time (and not on all accounts). There's a thread about that -- https://discuss.ai.google.dev/t/gemini-2-5-pro-with-empty-response-text/81175/219 -- the first problem happened in April, and the last question was from 4 hour ago -- and still not fixed.

So, on our end we can only support this and put a warning

PR #681, opened 2025-08-28
Fix compilation and publication for iOS

EugeneTheDev:


buildSrc/src/main/kotlin/ai.kotlin.configuration.gradle.kts
```diff
@@ -36,3 +36,13 @@ tasks.withType<KotlinJvmCompile>().configureEach {
         freeCompilerArgs.add("-Xjvm-default=all")
     }
 }
+
+configurations.all {
+    resolutionStrategy {
+        eachDependency {
+            if (requested.group == "org.jetbrains.kotlin") {
+                useVersion("2.1.21")
```
EugeneTheDev, line 44: We should use kotlin-bom or a non-hardcoded param

build.gradle.kts
```diff
@@ -13,7 +13,7 @@ group = "ai.koog"
 version = run {
     // our version follows the semver specification
 
-    val main = "0.4.0"
+    val main = "0.4.0.1"
```
kpavlov:
```suggestion
    val main = "0.4.1"
```

kpavlov (approved):

agents/agents-mcp/build.gradle.kts
```diff
@@ -21,9 +21,9 @@ kotlin {
                 api(project(":prompt:prompt-executor:prompt-executor-llms-all"))
 
                 api(libs.mcp)
+                api(libs.kotlinx.serialization.json)
```
sdubov, line 24: Unnecessary change.

buildSrc/src/main/kotlin/ai.kotlin.configuration.gradle.kts
sdubov: @EugeneTheDev , do you expect to fix this in current commit? I would at least take a version from `libs` is possible or from some extension function.

sdubov (changes requested):
Please also fix the version in the root `build.gradle.kts`.

buildSrc/src/main/kotlin/ai.kotlin.configuration.gradle.kts
kpavlov: It was fixed

PR #683, opened 2025-08-28
Improve Kotlin version management in convention gradle plugin

EugeneTheDev:
Make Kotlin version management more stable and aligned

Follow-up after #681

### DIFF
--- a/buildSrc/src/main/kotlin/ai.kotlin.configuration.gradle.kts
+++ b/buildSrc/src/main/kotlin/ai.kotlin.configuration.gradle.kts
@@ -14,6 +14,7 @@ val libs = the<LibrariesForLibs>()
  */
 val kotlinLanguageVersion = KotlinVersion.KOTLIN_2_1
 val kotlinApiVersion = KotlinVersion.KOTLIN_2_1
+val kotlinBomVersion = requireNotNull(libs.kotlin.bom.get().version)
 
 extensions.getByType<KotlinProjectExtension>().apply {
     jvmToolchain {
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/kotlin/ai.kotlin.configuration.gradle.kts
```diff
@@ -1,12 +1,20 @@
...
 import org.jetbrains.kotlin.gradle.dsl.KotlinProjectExtension
 import org.jetbrains.kotlin.gradle.dsl.KotlinVersion
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompilationTask
 import org.jetbrains.kotlin.gradle.tasks.KotlinJvmCompile
 
+val libs = the<LibrariesForLibs>()
+
+val kotlinLanguageVersion = KotlinVersion.KOTLIN_2_1
+val kotlinApiVersion = KotlinVersion.KOTLIN_2_1
+val kotlinCoreLibrariesVersion = "2.1.0"
```
kpavlov: Now library catalog can be referenced as well:
```suggestion
val kotlinCoreLibrariesVersion = libs.versions.kotlin.get()
```

EugeneTheDev: This was done intentionally, and it's a bit more complicated. The idea was, that setting lower `coreLibrariesVersion` explicitly would allow us to use newer version of Kotlin Gradle Plugin while remaining compatible with older Kotlin versions by adding older kotlin-stdlib version on our `compileClasspath`. So that we get newer compiler features without the need to upgrade Kotlin version. This works for Kotlin/JVM, but for most of the KMP targets `languageLevel` is not supported, and your library ABI version is your KGP version. That means there's actually almost no sense in this setup, and we still have to stick to the right KGP. And also yes, it looks strage considering we have a higher version in the kotlin-bom that we use to set Kotlin versions. So I'll remove this additional core libraries cofiguration entirely, since it doesn't do anything for us

kpavlov (approved):

PR #481, opened 2025-07-25
SQL Persistence Providers with JetBrains Exposed ORM

ariawisp:
This PR introduces SQL-based persistence storage providers for agent checkpoints, extending the persistence framework with support for relational databases through [JetBrains Exposed ORM](https://github.com/JetBrains/Exposed). This complements the Redis provider (PR #480) to provide comprehensive persistence options for both embedded and server-side deployments.

## Overview

Building on the persistence framework established in PR #480, this implementation adds SQL database support with the same architectural principles but optimized for different deployment scenarios:

- **Redis (PR #480)**: Optimized for distributed servers, ephemeral state, and mid-execution checkpoints
- **SQL (This PR)**: Ideal for durable session persistence, crash recovery, and long-term checkpoint storage

### Modular Architecture
- **Separate module**: SQL dependencies isolated in `agents-features-sql` module to keep them optional
- **Configurable cleanup**: TTL cleanup behavior is fully configurable with `CleanupConfig`
- **Resource management**: All providers implement `AutoCloseable` for proper cleanup
- **Performance optimizations**: Indexed TTL columns and configurable cleanup intervals

### Two-Tier Cleanup Configuration
- **Cleanup intervals**: Configurable minimum time between cleanup operations (default: 1 minute)
- **Cleanup enabling**: Can be completely disabled for high-performance scenarios where TTL is not needed
- **Efficient indexing**: TTL timestamp column is indexed for fast cleanup queries
- **Conditional execution**: Cleanup only runs when enabled AND TTL is configured

### Multi-Database Support
- **PostgreSQL**: Production-grade with JSONB support and HikariCP pooling
- **MySQL**: Enterprise-ready with JSON column support (5.7+)
- **H2**: Perfect for testing and embedded applications
- **SQLite**: Zero-configuration for desktop and mobile applications

### Architecture Highlights
- **Generic SQL abstraction**: `SQLPersistencyStorageProvider` base class in commonMain
- **Exposed ORM integration**: Type-safe SQL operations with `ExposedPersistencyStorageProvider`
- **Database-specific optimizations**: Vendor-specific performance tuning using Exposed's dialect types
- **Connection pooling**: HikariCP integration for PostgreSQL and MySQL
- **TTL support**: Automatic checkpoint expiration with efficient, configurable cleanup

### Checkpoint Persistence Use Cases

This PR focuses on **checkpoint persistence** (execution state), not semantic memory. The providers enable:

1. **Session Resumption** (Postgres recommended)
[...]

PR #705, opened 2025-09-01
Ariawisp/postgres persistence provider

Rizzen:
rebased and fixed version of https://github.com/JetBrains/koog/pull/481 by @ariawisp


agents/agents-features/agents-features-sql/src/jvmMain/kotlin/ai/koog/agents/features/sql/providers/H2PersistencyStorageProvider.kt
```diff
@@ -0,0 +1,254 @@
...
+     */
+    public data class PoolStats(
+        val activeConnections: Int,
+        val idleConnections: Int,
+        val totalConnections: Int,
+        val threadsAwaitingConnection: Int,
+        val maxPoolSize: Int
+    ) {
+        val utilizationPercent: Double = (activeConnections.toDouble() / maxPoolSize) * 100
+        val isHighUtilization: Boolean = utilizationPercent > 80.0
```
tiginamaria: nit: define constants?

agents/agents-features/agents-features-sql/src/jvmMain/kotlin/ai/koog/agents/features/sql/providers/MySQLPersistencyStorageProvider.kt
```diff
@@ -0,0 +1,167 @@
...
+     */
+    public data class PoolStats(
+        val activeConnections: Int,
+        val idleConnections: Int,
+        val totalConnections: Int,
+        val threadsAwaitingConnection: Int,
+        val maxPoolSize: Int
+    ) {
+        val utilizationPercent: Double = (activeConnections.toDouble() / maxPoolSize) * 100
+        val isHighUtilization: Boolean = utilizationPercent > 80.0
```
tiginamaria: nit same: define constants?

agents/agents-features/agents-features-sql/src/jvmMain/kotlin/ai/koog/agents/features/sql/providers/PostgresPersistencyStorageProvider.kt
```diff
@@ -0,0 +1,216 @@
...
+                threadsAwaitingConnection = ds.hikariPoolMXBean?.threadsAwaitingConnection ?: 0,
+                maxPoolSize = ds.maximumPoolSize
+            )
+        }
+    }
+    
+    /**
+     * Connection pool statistics for monitoring.
+     */
+    public data class PoolStats(
```
tiginamaria: Looks like code duplication?

gradle/libs.versions.toml
```diff
@@ -60,6 +66,17 @@ mcp = { module = "io.modelcontextprotocol:kotlin-sdk", version.ref = "mcp" }
 slf4j-simple = { module = "org.slf4j:slf4j-simple", version.ref = "slf4j" }
 jetsign-gradle-plugin = { module = "com.jetbrains:jet-sign", version.ref = "jetsign" }
 testcontainers = { module = "org.testcontainers:testcontainers", version.ref = "testcontainers" }
+testcontainers-postgresql = { module = "org.testcontainers:postgresql", version.ref = "testcontainers" }
+exposed-core = { module = "org.jetbrains.exposed:exposed-core", version.ref = "exposed" }
```
EugeneTheDev, line 79: nit: Let's use [exposed-bom](https://mvnrepository.com/artifact/org.jetbrains.exposed/exposed-bom) to set all exposed-related versions

agents/agents-features/agents-features-sql/src/commonMain/kotlin/ai/koog/agents/features/sql/providers/SQLPersistencyStorageProvider.kt
```diff
@@ -0,0 +1,99 @@
+package ai.koog.agents.features.sql.providers
+
+import ai.koog.agents.snapshot.feature.AgentCheckpointData
```
EugeneTheDev: nit: Unused import

agents/agents-features/agents-features-sql/src/commonMain/kotlin/ai/koog/agents/features/sql/providers/SQLPersistencyStorageProvider.kt
```diff
@@ -0,0 +1,99 @@
...
+ *
+ * @constructor Initializes the SQL persistence provider.
+ * @param persistenceId Unique identifier for this agent's persistence data
+ * @param tableName Name of the table to store checkpoints (default: "agent_checkpoints")
+ * @param ttlSeconds Optional TTL for checkpoint entries in seconds (null = no expiration)
+ */
+public abstract class SQLPersistencyStorageProvider(
+    protected val persistenceId: String,
+    protected val tableName: String = "agent_checkpoints",
+    protected val ttlSeconds: Long? = null
```
EugeneTheDev: Or we can use `kotlin.time.Duration` here,for better API

agents/agents-features/agents-features-sql/src/commonMain/kotlin/ai/koog/agents/features/sql/providers/SQLPersistencyStorageProvider.kt
```diff
@@ -0,0 +1,99 @@
...
+        return ttlSeconds?.let { 
+            timestamp.toEpochMilliseconds() + (it * 1000)
+        }
+    }
+    
+    /**
+     * Validates that the persistence ID contains only safe characters for SQL.
+     * This helps prevent SQL injection when persistence ID is used in queries.
+     */
+    protected fun validatePersistenceId() {
```
EugeneTheDev: I don't think this is needed here. SQLi prevention is already handled by all modern SQL libraries, and it should be an implementation concern.

agents/agents-features/agents-features-sql/src/jvmMain/kotlin/ai/koog/agents/features/sql/providers/ExposedPersistencyStorageProvider.kt
```diff
@@ -0,0 +1,302 @@
...
+ * Configuration for TTL cleanup behavior
+ *
+ * @property enabled Whether TTL cleanup should be performed automatically
+ * @property intervalMs Minimum interval between cleanup operations in milliseconds (default: 1 minute)
+ */
+public data class CleanupConfig(
+    val enabled: Boolean = true,
+    val intervalMs: Long = 60_000L // 1 minute default
+) {
+    public companion object {
```
EugeneTheDev, line 30: MIssing KDoc for public symbols

agents/agents-features/agents-features-sql/src/jvmMain/kotlin/ai/koog/agents/features/sql/providers/ExposedPersistencyStorageProvider.kt
```diff
@@ -0,0 +1,302 @@
...
+     * Exposed table definition for storing agent checkpoints.
+     * 
+     * Schema:
+     * - Composite primary key: (persistence_id, checkpoint_id)
+     * - Timestamp for ordering and querying
+     * - JSON column for flexible checkpoint data storage
+     * - Optional TTL timestamp for expiration (indexed for efficient cleanup)
+     */
+    public open class CheckpointsTable(tableName: String) : Table(tableName) {
+        public val persistenceId: Column<String> = varchar("persistence_id", 255)
```
EugeneTheDev: Pubilc symbols are not documented in KDoc

agents/agents-features/agents-features-sql/src/jvmMain/kotlin/ai/koog/agents/features/sql/providers/ExposedPersistencyStorageProvider.kt
```diff
@@ -0,0 +1,302 @@
...
+            applyDialectOptimizations()
+            block()
+        }
+    }
+    
+    /**
+     * Applies database-specific optimizations based on the current dialect.
+     * This method is called at the beginning of each transaction.
+     */
+    protected open fun Transaction.applyDialectOptimizations() {
```
EugeneTheDev: Given that there are already dedicated implementation classes for all these dialects, it breaks separation of concerns a bit that this when is put here, in the base class. I would suggest removing this method, making `transaction` method abstract and letting each dialect implementation implement this `transaction` in its own way.

agents/agents-features/agents-features-sql/src/jvmMain/kotlin/ai/koog/agents/features/sql/providers/ExposedPersistencyStorageProvider.kt
```diff
@@ -0,0 +1,302 @@
...
+    /**
+     * Exposed table definition for storing agent checkpoints.
+     * 
+     * Schema:
+     * - Composite primary key: (persistence_id, checkpoint_id)
+     * - Timestamp for ordering and querying
+     * - JSON column for flexible checkpoint data storage
+     * - Optional TTL timestamp for expiration (indexed for efficient cleanup)
+     */
+    public open class CheckpointsTable(tableName: String) : Table(tableName) {
```
EugeneTheDev: Let's move it to hte top level, for better readability?

agents/agents-features/agents-features-sql/src/jvmMain/kotlin/ai/koog/agents/features/sql/providers/ExposedPersistencyStorageProvider.kt
```diff
@@ -0,0 +1,302 @@
...
+                checkpointsTable.persistenceId eq this@ExposedPersistencyStorageProvider.persistenceId
+            }.count()
+        }
+    }
+    
+    /**
+     * Closes any resources associated with this provider.
+     * Concrete implementations should override this to close connection pools.
+     */
+    override fun close() {
```
EugeneTheDev: Let's not override it, if it does nothing. This might lead to problems if implementations forget to implement it.

agents/agents-features/agents-features-sql/src/commonMain/kotlin/ai/koog/agents/features/sql/providers/SQLPersistencyStorageProvider.kt
```diff
@@ -0,0 +1,99 @@
...
+     * Executes a database transaction with the given operations.
+     * Implementations should ensure proper transaction isolation and rollback on failure.
+     */
+    protected abstract suspend fun <T> transaction(block: suspend () -> T): T
+    
+    /**
+     * Cleans up expired checkpoints based on TTL.
+     * This should be called periodically or before operations to maintain database hygiene.
+     */
+    protected abstract suspend fun cleanupExpired()
```
EugeneTheDev: Let's make it public, to allow users to call it manually for greater flexibility

agents/agents-features/agents-features-sql/src/commonMain/kotlin/ai/koog/agents/features/sql/providers/SQLPersistencyStorageProvider.kt
```diff
@@ -0,0 +1,99 @@
...
+    protected val json: Json = Json { 
+        prettyPrint = true
+        ignoreUnknownKeys = true
+    }
+    
+    /**
+     * Initializes the database schema if it doesn't exist.
+     * This should be called once during provider initialization.
+     */
+    public abstract suspend fun initializeSchema()
```
EugeneTheDev: Proper schema management is a complicated topic, there might be some nuances users would want to control (indexes, permission management, etc.). Also, database user that will be reading/writing checkpoints might not even have such permissions (create/drop tables). This is a common best practice. There are dedicated tools that handle migrations, e.g. Flyway. Given that, I would suggest leaving it out of scope and removing this method. Users will configure the tables themselves in a way they want, this class will only focus on interacting with this data, without altering the database.

agents/agents-features/agents-features-sql/src/jvmMain/kotlin/ai/koog/agents/features/sql/providers/ExposedPersistencyStorageProvider.kt
```diff
@@ -0,0 +1,302 @@
...
+        
+        override val primaryKey: Table.PrimaryKey = PrimaryKey(persistenceId, checkpointId)
+        
+        init {
+            // Create composite index for efficient queries
+            index(isUnique = false, persistenceId, createdAt)
+        }
+    }
+    
+    public override suspend fun initializeSchema() {
```
EugeneTheDev: See the comment regarding removal of this method. Even Exposed itself deprecated `SchemaUtils.createMissingTablesAndColumn` in favor of external proper migration management tools.

agents/agents-features/agents-features-sql/src/jvmMain/kotlin/ai/koog/agents/features/sql/providers/ExposedPersistencyStorageProvider.kt
```diff
@@ -0,0 +1,302 @@
...
+            }
+            if (deletedCount > 0) {
+                lastCleanupTime = now
+            }
+        }
+    }
+    
+    override suspend fun getCheckpoints(): List<AgentCheckpointData> {
+        validatePersistenceId()
+        conditionalCleanup()
```
EugeneTheDev: Here and in other `get` methods - the fact that a `get` operation might remove some data from the table is an unexpected side-effect, IMHO. Let's avoid such implicit side-effects and instead expose `cleanupExpired` as a public method, allowing the users themselves to decide when they want to perform the cleanup.

agents/agents-features/agents-features-sql/src/jvmMain/kotlin/ai/koog/agents/features/sql/providers/ExposedPersistencyStorageProvider.kt
```diff
@@ -0,0 +1,302 @@
...
+        }
+    }
+    
+    public override suspend fun initializeSchema() {
+        newSuspendedTransaction(Dispatchers.IO, database) {
+            SchemaUtils.createMissingTablesAndColumns(checkpointsTable)
+        }
+    }
+    
+    override suspend fun <T> transaction(block: suspend () -> T): T {
```
EugeneTheDev: Let's keep it abstract and let implementations define it, to potentially apply dialect-specific optimizations (see related comment above regarding the removal of the `applyDialectOptimizations` method)

agents/agents-features/agents-features-sql/src/jvmMain/kotlin/ai/koog/agents/features/sql/providers/ExposedPersistencyStorageProvider.kt
```diff
@@ -0,0 +1,302 @@
...
+    }
+    
+    /**
+     * Conditionally performs cleanup based on configuration and TTL settings.
+     * Only runs cleanup if:
+     * 1. Cleanup is enabled in config
+     * 2. TTL is configured (ttlSeconds is not null)
+     * 3. Enough time has passed since last cleanup
+     */
+    private suspend fun conditionalCleanup() {
```
EugeneTheDev: Let's remove it, and make it a user responsiblitiy to perform a cleanup via explicit `cleanupExpired` call.

agents/agents-features/agents-features-sql/src/jvmMain/kotlin/ai/koog/agents/features/sql/providers/ExposedPersistencyStorageProvider.kt
```diff
@@ -0,0 +1,302 @@
...
+ * @param tableName Name of the table to store checkpoints (default: "agent_checkpoints")
+ * @param ttlSeconds Optional TTL for checkpoint entries in seconds (null = no expiration)
+ * @param cleanupConfig Configuration for TTL cleanup behavior
+ */
+public abstract class ExposedPersistencyStorageProvider(
+    persistenceId: String,
+    protected val database: Database,
+    tableName: String = "agent_checkpoints",
+    ttlSeconds: Long? = null,
+    private val cleanupConfig: CleanupConfig = CleanupConfig.default()
```
EugeneTheDev: Also, this means this parameter would not be needed anymore

agents/agents-features/agents-features-sql/src/jvmMain/kotlin/ai/koog/agents/features/sql/providers/ExposedPersistencyStorageProvider.kt
```diff
@@ -0,0 +1,302 @@
...
+     * - Optional TTL timestamp for expiration (indexed for efficient cleanup)
+     */
+    public open class CheckpointsTable(tableName: String) : Table(tableName) {
+        public val persistenceId: Column<String> = varchar("persistence_id", 255)
+        public val checkpointId: Column<String> = varchar("checkpoint_id", 255)
+        public val createdAt: Column<Long> = long("created_at").index()
+        public val checkpointJson: Column<String> = text("checkpoint_json")
+        public val ttlTimestamp: Column<Long?> = long("ttl_timestamp").nullable().index()
+        
+        override val primaryKey: Table.PrimaryKey = PrimaryKey(persistenceId, checkpointId)
```
EugeneTheDev: Based on my suggestions regarding leaving migration management out of scope, let's remove `primaryKey` and all `index` invocations. Again, we can specify in the comment that it's recommended to create such indexes when creating a table, but the responsibility to do so will be on the user.

agents/agents-features/agents-features-sql/src/jvmMain/kotlin/ai/koog/agents/features/sql/providers/H2PersistencyStorageProvider.kt
```diff
@@ -0,0 +1,254 @@
...
+
+import com.zaxxer.hikari.HikariConfig
+import com.zaxxer.hikari.HikariDataSource
+import org.jetbrains.exposed.sql.Database
+
+/**
+ * H2 Database-specific implementation of [ExposedPersistencyStorageProvider] for managing
+ * agent checkpoints in H2 databases.
+ *
+ * H2 is a lightweight, embeddable Java SQL database that's perfect for:
```
EugeneTheDev: Here and in other dialect implementations: I believe we don't need to tell about the database features. That makes KDoc bloated, and does not relate directly to the storage provider class itself. I believe that if someone chooses a particlar debase, they already know its strength and weaknesses and already choose it for a reason, so we don't need to repeat it here. As an alternative, we can briefly highlight it in our docs.

agents/agents-features/agents-features-sql/src/jvmMain/kotlin/ai/koog/agents/features/sql/providers/H2PersistencyStorageProvider.kt
```diff
@@ -0,0 +1,254 @@
...
+     * Creates a provider with a JDBC URL and credentials.
+     *
+     * @param persistenceId Unique identifier for this agent's persistence data
+     * @param jdbcUrl H2 JDBC connection URL
+     * @param username Database username (default: "sa")
+     * @param password Database password (default: empty)
+     * @param tableName Name of the table to store checkpoints (default: "agent_checkpoints")
+     * @param ttlSeconds Optional TTL for checkpoint entries in seconds
+     */
+    public constructor(
```
EugeneTheDev: Here and in other dialect-specific provider classes. Why do we need several similar constructors? They complicate things, and make the class look more bloated. We wouldn't be able to covere all cases anyways. Let's remove them, and simply keep the primary constructor that accepts the `database: Databse` parameter. Again, if someone picks Exposed-based storage provider, I would assume they are either already familiar with Exposed, or can figure out how to connect to a database.

agents/agents-features/agents-features-sql/src/jvmMain/kotlin/ai/koog/agents/features/sql/providers/H2PersistencyStorageProvider.kt
```diff
@@ -0,0 +1,254 @@
...
+     */
+    override fun close() {
+        dataSource?.close()
+    }
+    
+    /**
+     * Returns connection pool statistics if using HikariCP.
+     * Useful for monitoring connection usage and performance.
+     */
+    public fun getPoolStats(): PoolStats? {
```
EugeneTheDev: Here and in other dialect specific implementations: this is not the concern of the storage provider. It complicates things unnecessary. If one uses connection pools to connect to dabases, I believe that can monitor them themselves directly. This should be out of scope for this provider. I beleive it should not concern itself with connection or database management. It should be simple CRUD repo.

agents/agents-features/agents-features-sql/src/jvmMain/kotlin/ai/koog/agents/features/sql/providers/H2PersistencyStorageProvider.kt
```diff
@@ -0,0 +1,254 @@
...
+        database = Database.connect(dataSource),
+        tableName = tableName,
+        ttlSeconds = ttlSeconds
+    ) {
+        this.dataSource = dataSource
+    }
+    
+    private var dataSource: HikariDataSource? = null
+    
+    public companion object {
```
EugeneTheDev, line 26: nit: Let's suppress missing KDoc declaration warning on public companion objects.

agents/agents-features/agents-features-sql/src/jvmTest/kotlin/ai/koog/agents/features/sql/providers/SQLPersistenceProvidersTest.kt
```diff
@@ -0,0 +1,167 @@
...
+import kotlinx.serialization.json.JsonPrimitive
+import kotlin.test.*
+
+/**
+ * Tests for SQL persistence providers.
+ * 
+ * Focuses on H2 in-memory database for actual functionality tests
+ * and compilation verification for other providers.
+ */
+class SQLPersistenceProvidersTest {
```
EugeneTheDev, line 23: I see testcontainers in test dependencies, but I don't see them being used. Let's create a number of test suites, one per each dialect provider. This should be relatively simple to do with testcontainers. 

We can extract common test logic with the tests themsevles to the abstract class, and then implementing test suites would only need to provide the correct storage provider implementation (Postgres, MySQL, etc.) and handle it's creation and shutdown.

EugeneTheDev (changes requested):
I like the idea, however, it requires some polishing

agents/agents-features/agents-features-sql/src/commonMain/kotlin/ai/koog/agents/features/sql/providers/SQLPersistencyStorageProvider.kt
Rizzen: That's no way we require users to create proper tables when setting up persistence. It's beyond user-unfriendly, nobody will use that API. It's better to require users to provide Koog with database which will allow to create tables and drop them.

EugeneTheDev: I would disagree here. This is what enterprises generally do, because they have stricter requirements regarding database management and control. And, as I already highlighted, it introduces more complexity for this class. As a middleground, we can provide some default initialization utility class, e.g. using Flyway. So that if user does not have any special requirements or just wants to prototype something, they can use this default preset.

Rizzen: We can provide the option to disable automatic table creation for sure, but as default - it should handle it automatically.

When user setting up persistence with particular provider - the only mandatory required input should and might be the connection string only. All the other options are advanced scenarios and should not be necessary for simple setup.

agents/agents-features/agents-features-sql/src/jvmMain/kotlin/ai/koog/agents/features/sql/providers/ExposedSQLMigrator.kt
```diff
@@ -0,0 +1,34 @@
...
+ * A no-operation implementation of the [ExposedSQLMigrator] interface.
+ *
+ * This class is designed to be used in scenarios where schema migrations are not required
+ * or are managed externally. It provides an empty implementation of the [migrate] method,
+ * effectively acting as a placeholder.
+ *
+ * Use this class if you need to satisfy the dependency on [ExposedSQLMigrator] but do not
+ * want to perform any migration actions.
+ */
+public class NoOpExposedSQLMigrator : ExposedSQLMigrator {
```
EugeneTheDev: Can it be an `object` then?

agents/agents-features/agents-features-sql/src/jvmMain/kotlin/ai/koog/agents/features/sql/providers/ExposedSQLMigrator.kt
```diff
@@ -0,0 +1,34 @@
...
+ * This interface is designed to be implemented by classes that need to handle schema migrations
+ * for databases, ensuring compatibility and flexibility in schema evolution.
+ *
+ * ExposedSQLMigrator provides an abstraction for executing migrations asynchronously, allowing
+ * for better control and management of database schema changes as part of the application's lifecycle.
+ *
+ * The primary function to be implemented is [migrate], which encapsulates the details of performing
+ * the required schema update or adjustments based on the application's requirements.
+ */
+public interface ExposedSQLMigrator {
```
EugeneTheDev: Since it doesn't actually depend on Exposed, can we call it just `SQLMigrator` and make it a property in the SQLPersistencyStorageProvider abstract class?

agents/agents-features/agents-features-sql/src/commonMain/kotlin/ai/koog/agents/features/sql/providers/SQLPersistencyStorageProvider.kt
```diff
@@ -0,0 +1,88 @@
...
+    protected val json: Json = Json {
+        prettyPrint = true
+        ignoreUnknownKeys = true
+    }
+
+    /**
+     * Initializes the database schema if it doesn't exist.
+     * This should be called once during provider initialization.
+     */
+    public abstract suspend fun migrate()
```
EugeneTheDev: Following up on the comment above regarding introducing migrator constructor parameter here with default NO-OP migrator, we then won't have to duplicate the code by overriding `migrate` in all dialect implementations. We can turn it into a simple final method:

```kt
public suspend fun migrate() {
    migrator.migrate()
}
```

agents/agents-features/agents-features-sql/src/jvmTest/kotlin/ai/koog/agents/features/sql/providers/H2PersistencyStorageProviderTest.kt
```diff
@@ -0,0 +1,103 @@
...
+        return H2PersistencyStorageProvider.inMemory(
+            persistenceId = "h2-agent",
+            databaseName = "h2_test_db",
+            tableName = "agent_checkpoints_test",
+            ttlSeconds = ttlSeconds
+        )
+    }
+
+    @Test
+    fun `migrate and basic CRUD against H2 in-memory`() = runBlocking {
```
EugeneTheDev, line 34: Since the tests themselves are similar accross different dialects, can we reduce code duplication and extract them into a base class? Actual dialect test suites would then extend from it and would have to override only the actual storage provider instance creation and cleanup.

EugeneTheDev (approved):
Thanks for improving it, looks good now, I left a couple more minor suggestion how we can make it slighlty better, please check them.

agents/agents-features/agents-features-sql/src/jvmMain/kotlin/ai/koog/agents/features/sql/providers/ExposedSQLMigrator.kt
Rizzen: It doesn't depend (yet) but used only there, so I would prefer it to stay that local

EugeneTheDev: In the current setup yes, it makes sense, but if we move it up to the base abstract class (to make `migrate` there non-abstract), then it would make sense, no?

PR #707, opened 2025-09-01
[WIP] KG-294 Support for Non-Ktor http clients. Part 1

kpavlov:
## Motivation and Context

KG-294 Support non-Ktor clients in public API
> Currently, all LLM Clients accept HttpClient from Ktor (optional). It will confuse Java users who don't have Ktor dependency but already have a client.

First step its to introduce KtorHttpClient interface and migrate all the client code from io.ktor.client.HttpClient to it.

## Breaking Changes

---

#### Type of the changes
- [ ] Documentation update
- [ ] Tests improvement
- [x] Refactoring

#### Checklist


agents/agents-utils/src/commonMain/kotlin/ai/koog/agents/utils/KtorHttpClient.kt
```diff
@@ -0,0 +1,135 @@
...
+ *
+ * @property clientName The name of the client, used for logging and traceability.
+ * @property logger A logging instance of type KLogger for recording client-related events and errors.
+ * @constructor Creates a KtorHttpClient instance with an optional base Ktor HttpClient and configuration block.
+ *
+ * @param baseClient The base Ktor HttpClient instance to be used. Default is a newly created instance.
+ * @param configurer A lambda function to configure the base Ktor HttpClient instance.
+ * The configuration is applied using the Ktor `HttpClient.config` method.
+ */
+public class KtorHttpClient(
```
Ololoshechkin: Such name might be confusing, because it's a wrapper around HttpClient from Ktor. What if someone actually would use `KtorHttpClient` thinking it's from Ktor?

Maybe let's make it internal, and add a public extension to `KoogHttpClient`, so that one would use `KoogHttpClient.fromKtor(...)` (or `fromKtorClient`)  -- and the next one would be `KoogHttpClient.fromSpring()`(or `fromSpringClient`), etc.

agents/agents-utils/src/commonMain/kotlin/ai/koog/agents/utils/KoogHttpClient.kt
```diff
@@ -0,0 +1,73 @@
...
+import kotlinx.coroutines.flow.Flow
+import kotlin.reflect.KClass
+
+/**
+ * Abstract interfaces defining a contract for HTTP client implementations.
+ * Provides methods for making HTTP POST requests and handling Server-Sent Events (SSE) streams.
+ *
+ * Implementations are supposed to use a particular library or framework.
+ */
+public interface KoogHttpClient {
```
Ololoshechkin, line 14: This class is only used in protected field of OpenAI base client (and apparently will be added to other LLM clients the same way).

Should we really make it public right now? 

I think we should provide constuctor overrides (for Kotlin) and factory methods (for Java) for the LLM Clients, so that they would take Ktor, Spring, and other HttpClients directly -- and under the hood we would construct the correct implementation of `KoogHttpClient` silently , and then use it uniformly inside. Should we even expose this to the users?

Ololoshechkin (changes requested):

agents/agents-utils/src/commonMain/kotlin/ai/koog/agents/utils/KoogHttpClient.kt
kpavlov: I am not very big fan of keeping interfaces private in the integration points to other frameworks. I think we should keep this interface public for dependency injection and to let other people implement their custom http clients. This approach keeps Koog open for extension, like Spring Framework. 

Also, it can't be internal due to cross-module dependencies in OpenAI LLM clients

agents/agents-utils/src/commonMain/kotlin/ai/koog/agents/utils/KtorHttpClient.kt
kpavlov: I was not sure about the name. `KoogKtorHttpClient` sounds better to me

prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client-base/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openai/base/AbstractOpenAILLMClient.kt
```diff
@@ -90,15 +77,17 @@ public abstract class OpenAIBasedSettings(
  * @param baseClient The HTTP client to use for API requests. Defaults to a new HttpClient instance.
  * @param clock Clock instance used for tracking response metadata timestamps. Defaults to Clock.System.
  */
-public abstract class AbstractOpenAILLMClient<TResponse : OpenAIBaseLLMResponse, TStreamResponse : OpenAIBaseLLMStreamResponse>(
+public abstract class AbstractOpenAILLMClient<
+    TResponse : OpenAIBaseLLMResponse,
+    TStreamResponse : OpenAIBaseLLMStreamResponse
+    >(
```
Ololoshechkin: nit: the formating looks non-standard to me. Let's fix it to improve readability :)

PR #723, opened 2025-09-03
KG-374. Allow setting all config properties for a Remote feature writer

sdubov:
[KG-375](https://youtrack.jetbrains.com/issue/KG-374/Remote-Feature-configuration-classes-hides-configuration-properties-in-base-class)

- Remote feature writer receives a configuration property with a number of properties (port, host, timeouts, etc.). These properties should be accessible from any inherited configuration class.

## Motivation and Context
Allow to configure connection properties from any Config class provided by `koog`.

## Breaking Changes
yes

---

#### Type of the changes
- [ ] Documentation update
- [ ] Tests improvement
- [ ] Refactoring

#### Checklist


agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/remote/client/config/ClientConnectionConfig.kt
```diff
@@ -3,45 +3,81 @@ package ai.koog.agents.core.feature.remote.client.config
...
+ * @property requestTimeout The timeout duration for requests, defaulting to null.
+ * @property connectTimeout The timeout duration for establishing a connection, defaulting to null.
  */
 public abstract class ClientConnectionConfig(
     public val host: String,
     port: Int? = null,
-    public val protocol: URLProtocol = URLProtocol.HTTPS,
-    public val headers: Map<String, String> = emptyMap(),
+    protocol: URLProtocol? = null,
+    headers: Map<String, String>? = null,
```
kpavlov: String to array of strings (multi-value map)

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/remote/client/config/DefaultClientConnectionConfig.kt
```diff
@@ -1,31 +1,72 @@
...
+
+        /**
+         * Represents the default connection timeout used for client connections.
+         */
+        public val defaultConnectionTimeout: Duration = 15.seconds
+    }
 
-        val defaultProtocol: URLProtocol = URLProtocol.HTTPS
+    init {
+        appendSerializersModule(agentFeatureMessageSerializersModule)
```
kpavlov: Nit: not sure if config holder class should have any logic

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/remote/server/config/DefaultServerConnectionConfig.kt
```diff
@@ -44,6 +50,17 @@ public class DefaultServerConnectionConfig(
...
+         *
+         * This constant is set to `false` by default, meaning the server does not wait for
+         * an initial connection and continues with its execution. It is used as a default value
+         * for the `waitConnection` parameter in the `DefaultServerConnectionConfig` class.
+         */
+        public const val DEFAULT_WAIT_CONNECTION: Boolean = false
+    }
+
+    init {
+        appendSerializersModule(agentFeatureMessageSerializersModule)
```
kpavlov: Same here

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/remote/client/config/DefaultClientConnectionConfig.kt
```diff
@@ -1,31 +1,72 @@
...
+         * This constant defines the default hostname or IP address as "127.0.0.1",
+         * typically representing the local machine (localhost).
+         */
+        public const val DEFAULT_HOST: String = "127.0.0.1"
 
-        const val DEFAULT_PORT: Int = 50881
+        /**
+         * Represents the default port number used for client connections.
+         */
+        public const val DEFAULT_PORT: Int = 50881
```
kpavlov, line 56: Side question: what is the story behind this strange port number?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/remote/client/config/ClientConnectionConfig.kt
sdubov: @kpavlov, could you please provide some examples when we need a "key to collection" for custom headers? Currently, we apply headers through `append()` ktor API function that receives `key: String, value: String` parameters.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/remote/client/config/DefaultClientConnectionConfig.kt
sdubov: We need to apply `SerializersModule` that provide serialization details for out internal classes. Otherwise, you will have errors on deserialization of you need to delegate this operation to a user. Please feel free to share if you have some good design in mind.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/remote/server/config/DefaultServerConnectionConfig.kt
sdubov: Please see my answer above.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/remote/client/config/DefaultClientConnectionConfig.kt
sdubov: No story :) Random number from ports available range.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/remote/client/config/ClientConnectionConfig.kt
kpavlov: According to HTTP specification, headers might have multiple values, e.g. [Accept](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Accept) header. While we didn't have such case yet, in general it's better to follow the spec.
It might not be a blocker right now, just a betterment suggestion

sdubov: Fixed

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/remote/client/config/DefaultClientConnectionConfig.kt
kpavlov: I woul call it explicitly where DefaultServerConnectionConfig is created.
There is a potential issue [here](https://github.com/JetBrains/koog/blob/ea30cd9cf3609c5e912f6647e908fe578515f3c0/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/writer/FeatureMessageRemoteWriter.kt#L36): if custom connection config will be provided, then the SerializationModule will not be applied, because `init{}` is not called. Keeping ConnectionConfig pure value object without side-effects would make things more straightforward.

How would you like this? :
```kotlin
public class FeatureMessageRemoteServer(
    public val connectionConfig: ServerConnectionConfig,
) : FeatureMessageServer {

    init {
        connectionConfig.appendSerializersModule(connectionConfig.jsonConfig.serializersModule)
    }
    ...
```

sdubov: True, in the case you provided, the `init` for the default config is not called by design. The idea here is to provide a more flexible solution for a user and delegate this responsibility to an end-user. So, they provide any config they want with any configured modules.

Let me think about the provided approach. I have some concerns that now this logic lives outside of config. I will try to update and will see if this would work better.

sdubov: @kpavlov, I looked through the logic for default configuration initialisation and found the root. We do not need to specifically configure serializers modules in that case. ConnectionConfig uses the default set of serailzers modules that are applied for each config. Some time ago, when the logic was split between modules, we defied only a base classes. Further, when I moved features core logic into agents core module, we can configure this inside a base `ConnectionConfig` class. I deleted unnecessary classes and merged it with a base logic. 

Thank you for raising this concern that leads to a forgotten code cleanup. Please see the last commit.

Issue #731, opened 2025-09-04
Make `LLMClient.executeStreaming` expose more information


OKatrych:
Currently, when using `LLMClient.executeStreaming`, it returns only a flow of content chunks `Flow<String>`. This is sufficient for simple cases, but not enough for production applications that require displaying more information to the user. 
Things that I missed:
- finish reason (in the same way as `LLMClient.execute` has), we don't know a reason why streaming suddenly stopped (some error, or it was the expected ending)
- All the errors that happen during the stream are skipped. Take, for example, `AnthropicLLMClient`:
```kotlin
event
    .takeIf { it.event == "content_block_delta" } // "error" event is skipped
    ?.data?.trim()?.let { json.decodeFromString<AnthropicStreamResponse>(it) }
    ?.delta?.text?.let { emit(it) }
```
- I think it was already mentioned [here](https://youtrack.jetbrains.com/issue/KG-142/Add-Message-ID-to-all-Messages), we don't save the message ID received from the provider

Currently, in my project, I use this class to represent the stream result from any provider:
```kotlin
sealed interface StreamResult {

    data class Chunk(
        val id: String,
        val content: String,
    ) : StreamResult

    data class Finished(
        val id: String,
        val reason: FinishReason?,
    ) : StreamResult

    data class Error(val error: String?) : StreamResult
}
```
It would be nice to see something similar in Koog.

Issue #624, opened 2025-08-20
Missing `tools` argument in `executeStreaming()` in `LLMClient` and `PromptExecutor`


hiking93:
I'm using the default `singleRunModeStrategy` and it makes the tool calls work. But I found it difficult to make the change to streaming for the strategy, just to get partial results of messages.

Furthermore, I tried comparing `execute()` and `executeStreaming()` in `LLMClient`.
Taking `OllamaClient` for example, in `execute()`:

https://github.com/JetBrains/koog/blob/0.3.0/prompt/prompt-executor/prompt-executor-clients/prompt-executor-ollama-client/src/commonMain/kotlin/ai/koog/prompt/executor/ollama/client/OllamaClient.kt#L143

```kotlin
override suspend fun execute(
    prompt: Prompt, model: LLModel, tools: List<ToolDescriptor>
): List<Message.Response> {
    ...
    return parseResponse(response.body<OllamaChatResponseDTO>())
    ...
}
```

It executes with `tools: List<ToolDescriptor>`, and parses the response as `OllamaChatResponseDTO`, extracting not only `messages.content` but also `messages.toolCalls` and other stuff:

```kotlin
private fun parseResponse(response: OllamaChatResponseDTO): List<Message.Response> {
    val messages = response.message ?: return emptyList()
    val content = messages.content
    val toolCalls = messages.toolCalls ?: emptyList()

    // Get token counts from the response, or use null if not available
    val promptTokenCount = response.promptEvalCount
    val responseTokenCount = response.evalCount
    ...
}
```

But in `executeStreaming()`, the `tools: List<ToolDescriptor>` is missing, which means it always passes `null` for `tools` into `OllamaChatRequestDTO`; and it streams only `chunk.message.content` from the response, dropping all other information of `chunk: OllamaChatResponseDTO`.

```kotlin
override fun executeStreaming(
    prompt: Prompt, model: LLModel
): Flow<String> = flow {
    ...
    val chunk = ollamaJson.decodeFromString<OllamaChatResponseDTO>(line)
    chunk.message?.content?.let { content ->
        if (content.isNotEmpty()) {
            emit(content)
        }
    }
    ...
}
```

hiking93:
Looks like the issue has been fixed in https://github.com/JetBrains/koog/pull/747, so I'm closing it, thanks!

Issue #513, opened 2025-07-31
Any plans to support the stream tool call functionality?


z-yn:
When using Koog to build applications, I noticed that `LLMClient.executeStreaming` does not support Tool Calls. The same applies to `PromptExecutor.executeStreaming` and `requestLLMStreaming`.
 Are there any plans to support ​**stream tool call​** functionality in the future?

aozherelyeva:
Hi @z-yn, thank you for the request!
I've added the request to [our tracker](https://youtrack.jetbrains.com/issue/KG-206/Support-the-stream-tool-call-functionality), please feel free to follow and upvote it.
We'll get back to you once we have any updates on this one.

Rizzen:
@z-yn Hello!

In the example from @aozherelyeva, our streaming API is used to take advantage of structured output, enabling you to parse streaming responses on the fly and trigger tool calls immediately.

When you mention "truly streaming tool calls", I’m not entirely sure what you mean. In the LangGraph example you provided, the LLM output is streamed, and the tool calls are split out during that process - but you still need to wait until the end of each tool call’s streaming before properly invoking that tool, because while streaming - the tool is incomplete.

With our API, you can achieve the same pattern, plus go a step further:
- You can stream and parse structured data (e.g., a list of entities) as it’s generated.
- This lets you call tools incrementally, without waiting for the entire tool call to finish.
- The result is a more responsive UX, especially in scenarios where partial data can be acted on immediately.

hiking93:
I'm new to this library, so let me know if I misunderstood something :)

I'm using the default `singleRunModeStrategy` and it has made tool calls working. But I find it difficult to make the change to streaming for the strategy, just to get partial results of messages.

Furthermore, I tried comparing `execute()` and `executeStreaming()` in LLM clients.
Taking Ollama for example, in `execute()`:

https://github.com/JetBrains/koog/blob/0.3.0/prompt/prompt-executor/prompt-executor-clients/prompt-executor-ollama-client/src/commonMain/kotlin/ai/koog/prompt/executor/ollama/client/OllamaClient.kt#L143

```kotlin
override suspend fun execute(
    prompt: Prompt, model: LLModel, tools: List<ToolDescriptor>
): List<Message.Response> {
    ...
    return parseResponse(response.body<OllamaChatResponseDTO>())
    ...
}
```

Here we parse the response as `OllamaChatResponseDTO`, getting not only `messages.content` but also `messages.toolCalls` and other stuff:

```kotlin
[...]

PR #747, opened 2025-09-05
Refactor streaming api to support tool calls

dosier:
## Motivation and Context
Before this change it was not possible to use tool-calling when interacting with a LLM through the streaming interface. 
This is useful to have as many front-end integrated agents may want to incrementally update the UI but still have tool calling support.

Relevant issues:
- https://github.com/JetBrains/koog/issues/731
- https://github.com/JetBrains/koog/issues/624
- https://github.com/JetBrains/koog/issues/513

## Breaking Changes
Streaming related functions that returned a `Flow<String>` now return a `Flow<StreamFrame>`.

---

#### Type of the changes
- [x] Documentation update
- [x] Tests improvement
- [x] Refactoring

#### Checklist


prompt/prompt-executor/prompt-executor-model/src/commonMain/kotlin/ai/koog/prompt/executor/model/PromptExecutor.kt
```diff
@@ -39,7 +42,24 @@ public interface PromptExecutor {
...
+
+    /**
+     * Executes a given prompt using the specified language model and returns a stream of output as a flow of `StreamFrame` objects.
+     *
+     * @param prompt The prompt containing input messages and parameters to guide the language model execution.
+     * @param model The language model to be used for processing the prompt.
+     * @param tools A list of `ToolDescriptor` objects that define the tools available for the execution.
+     * @return A flow emitting `StreamFrame` objects that represent the streaming output of the language model.
+     */
+    public fun executeStreamingWithTools(
```
kpavlov: I see tools is a new parameter with default value. It's a compatible change, especially with `@JvmOverloads`. What if `executeStreaming(...)` will be updated without introducing new method?

rubencagnie:
@dosier - I was looking a bit into your work and was able to get it running for OpenAI, but I had to do some changes. I created a PR [on your repo](https://github.com/dosier/koog/pull/1) with my proposed changes

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/streaming/StreamFrame.kt
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    /**
+     * Represents a frame of a streaming response from a LLM that signals the end of the stream.
+     *
+     * @property finishReason The reason for the stream to end.
+     */
+    @Serializable
+    public data class End(
+        val finishReason: String? = null
```
OKatrych: It would be nice to add a `metaInfo` field here to expose information about token usage.
```kotlin
val metaInfo: ResponseMetaInfo
```
We already do this for regular responses, so all data models are already there.

dosier:
I did some more work on it with @rubencagnie , will soon merge https://github.com/rubencagnie/koog/pull/3 into this branch and fix the docs and it should be good for review :)

examples/src/main/kotlin/ai/koog/agents/example/streaming/StreamingAgentWithTools.kt
```diff
@@ -0,0 +1,116 @@
...
+        // Tell me if the switch if on or off. Elaborate on how you will determine that. After that, if it was off, turn it on. Be very verbose in all the steps
+
+        agent.run(input)
+
+        println()
+        println("Enter your message:")
+    }
+}
+
+fun streamingWithToolsStrategy() = strategy("streaming_loop") {
```
rubencagnie, line 93: I was wondering if we should have this part of `AIAgentNodes.kt` or not. This strategy works, but it might be too specific for this implementation? Overall, it's a strategy that takes in a string, streams the response - then applies the response and goes into the agent loop again as long as there are tools to execute...

examples/src/main/kotlin/ai/koog/agents/example/streaming/StreamingAgentWithTools.kt
```diff
@@ -0,0 +1,116 @@
...
+    val executeMultipleTools by nodeExecuteMultipleTools(parallelTools = true)
+    val nodeStreaming by nodeLLMRequestStreamingAndSendResults()
+
+    val mapStringToRequests by node<String, List<Message.Request>> { input ->
```
rubencagnie, line 97: I am personally a big fan of this - because it allows a more robust agent loop where the input of a node is `List<Message.Request>` vs a `String`. I would like some thoughts from the community

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/streaming/StreamFrame.kt
dosier: Added this now @OKatrych

dosier: TODO: add support for this in AWS models

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/streaming/StreamFrameExt.kt
```diff
@@ -0,0 +1,77 @@
...
+        when (frame) {
+            is StreamFrame.Append -> content = content?.plus(frame.text) ?: frame.text
+            is StreamFrame.ToolCall -> Unit
+            is StreamFrame.End -> finishReason = frame.finishReason
+        }
+    }
+    return if (content.isNullOrBlank() && finishReason.isNullOrBlank()) {
+        null
+    } else {
+        Message.Assistant(
```
rubencagnie: I was just thinking - if the result is purely a `finish-reason` that is tool call, then it's not really relevant to return a `Message.Assistant` with an empty string... Because it wasn't really an assistant message in that case

dosier: Maybe we should just change it to 
```kt
    return  Message.Assistant(
        content = content?:return null,
```
Also getting rid of the `isBlank` check, assume streamframe doesn't emit blank frames, otherwise we might as well filter it out at the flow builder level.

OKatrych:
Anthropic logic seems broken, added initial MR to start fixing it.
https://github.com/dosier/koog/pull/2

examples/src/main/kotlin/ai/koog/agents/example/streaming/StreamingAgentWithTools.kt
dosier: @Rizzen do u have any thoughts on this?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/handler/StreamHandlerContext.kt
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * Represents the context for handling a before-stream event.
+ * This context is provided when streaming is about to begin.
+ *
+ * @property runId The unique identifier for this streaming session.
+ * @property prompt The prompt that will be sent to the language model for streaming.
+ * @property model The language model instance being used for streaming.
+ * @property tools The list of tool descriptors available for the streaming call.
+ */
+public data class BeforeStreamContext(
```
kpavlov, line 22: nit: BeforeStreamContext and AfterStreamContext are very similar. Did you consider combining them into something StreamEventContext? On the other hand, having explicit types makes API safer

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/handler/StreamHandler.kt
```diff
@@ -0,0 +1,94 @@
...
+
+/**
+ * A functional interface implemented to handle logic that occurs before streaming from a large language model (LLM) begins.
+ * It allows preprocessing steps or validation based on the provided prompt, available tools, targeted LLM model,
+ * and a unique run identifier.
+ *
+ * This can be particularly useful for custom input manipulation, logging, validation, or applying
+ * configurations to the streaming request based on external context.
+ */
+public fun interface BeforeStreamHandler {
```
kpavlov, line 72: nit: BeforeStreamHandler and AfterStreamHandler have the same signature. Would it make sense to combine them or use inheritance?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/AIAgentPipeline.kt
```diff
@@ -202,6 +216,57 @@ public abstract class AIAgentPipeline {
...
+     * allowing them to perform preprocessing or logging operations.
+     *
+     * @param runId The unique identifier for this streaming session
+     * @param prompt The prompt being sent to the language model
+     * @param model The language model being used for streaming
+     * @param tools The list of available tool descriptors for this streaming session
+     */
+    public suspend fun onBeforeStream(runId: String, prompt: Prompt, model: LLModel, tools: List<ToolDescriptor>) {
+        val eventContext = BeforeStreamContext(runId, prompt, model, tools)
+        streamHandlers.values.forEach { handler -> handler.beforeStreamHandler.handle(eventContext) }
```
kpavlov, line 234: It makes sense to add logging around the handler invocation to have better visibility in case of an exception

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/AIAgentPipeline.kt
```diff
@@ -202,6 +216,57 @@ public abstract class AIAgentPipeline {
...
+     *
+     * This method notifies all registered stream handlers about each incoming stream frame,
+     * allowing them to process, transform, or aggregate the streaming content in real-time.
+     *
+     * @param runId The unique identifier for this streaming session
+     * @param streamFrame The individual stream frame containing partial response data
+     */
+    public suspend fun onStreamFrame(runId: String, streamFrame: StreamFrame) {
+        val eventContext = StreamFrameContext(runId, streamFrame)
+        streamHandlers.values.forEach { handler -> handler.streamFrameHandler.handle(eventContext) }
```
kpavlov, line 248: It makes sense to add logging around the handler invocation to have better visibility in case of an exception.

Failing when one handler throws an exception is a design choice, isn't it?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/AIAgentPipeline.kt
```diff
@@ -202,6 +216,57 @@ public abstract class AIAgentPipeline {
...
+     * @param tools The list of tool descriptors that were available for this streaming session
+     */
+    public suspend fun onAfterStream(
+        runId: String,
+        prompt: Prompt,
+        model: LLModel,
+        tools: List<ToolDescriptor>
+    ) {
+        val eventContext = AfterStreamContext(runId, prompt, model, tools)
+        streamHandlers.values.forEach { handler -> handler.afterStreamHandler.handle(eventContext) }
```
kpavlov, line 283: It makes sense to add logging around the handler invocation to have better visibility in case of an exception

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/PromptExecutorProxy.kt
```diff
@@ -39,11 +43,35 @@ public class PromptExecutorProxy(
...
+        model: LLModel,
+        tools: List<ToolDescriptor>
+    ): Flow<StreamFrame> {
+        logger.debug { "Executing LLM streaming call (prompt: $prompt, tools: [${tools.joinToString { it.name }}])" }
+        return executor.executeStreaming(prompt, model, tools)
+            .onStart {
+                pipeline.onBeforeStream(runId, prompt, model, tools)
+            }
+            .onEach {
+                pipeline.onStreamFrame(runId, it)
```
kpavlov, line 73: What if an exception happens here? Would `onCompletion` handlers ever be called? Would it make sense to add onFailure handlers as well?

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/streaming/StreamFrame.kt
```diff
@@ -0,0 +1,57 @@
...
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.json.Json
+import kotlinx.serialization.json.JsonObject
+import kotlinx.serialization.json.jsonObject
+
+/**
+ * Represents a frame of a streaming response from a LLM.
+ */
+@Serializable
+public sealed interface StreamFrame {
```
kpavlov, line 13: Would it be good to define a type for StreamFrame content: `StreamFrame<T>` and then use it in the `Append<T>(val content: T)`

prompt/prompt-executor/prompt-executor-model/src/commonMain/kotlin/ai/koog/prompt/executor/model/PromptExecutor.kt
```diff
@@ -33,13 +34,18 @@ public interface PromptExecutor {
...
-     * @return A flow emitting strings that represent the streaming output of the language model.
+     * @param tools A list of `ToolDescriptor` objects that define the tools available for the execution.
+     * @return A flow emitting `StreamFrame` objects that represent the streaming output of the language model.
      */
-    public fun executeStreaming(prompt: Prompt, model: LLModel): Flow<String>
+    public fun executeStreaming(
+        prompt: Prompt,
+        model: LLModel,
+        tools: List<ToolDescriptor> = emptyList()
+    ): Flow<StreamFrame>
```
kpavlov, line 48: This is a backwards-incompatible change

Did you consider parameterizing `PromptExecutor<S>`  with a streaming result type? I am even thinking about extracting `StreamingPromptExceutor<S>` as a separate interface, but it's a different discussion and a bigger change.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/PromptExecutorProxy.kt
EugeneTheDev: `onCompletion` is sort of like `finally` in try catch, so it will always be called, regardless of the completion (successful/with error)

kpavlov (approved):
LGTM if signature change is not the concern

Rizzen:
I believe it's fine to introduce a breaking change here since we only have `Flow<String>` right now, and we know it needs to be changed eventually — better to do it sooner rather than later.

prompt/prompt-executor/prompt-executor-model/src/commonMain/kotlin/ai/koog/prompt/executor/model/PromptExecutor.kt
dosier: This is outdated

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/handler/StreamHandlerContext.kt
dosier: Tbh @rubencagnie may be able to answer this better since the whole stream handler section was their contribution. Though thinking about it,  I feel like we'd only need one type of hook `onStreaming` that just returns the flow, then ppl can just use flow operators for `onStart`, `onCompletion`, error handling, etc...

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/streaming/StreamFrame.kt
dosier: I don't like this, since the variants don't share payload type. `Append` is text, `ToolCall` has arguments, `End` has finish reason/meta. So I don't really see what generics would add other than complicate the streaming API. But maybe I am missing a use case...

prompt/prompt-executor/prompt-executor-model/src/commonMain/kotlin/ai/koog/prompt/executor/model/PromptExecutor.kt
dosier: I did consider keeping changes backwards compatible but @Rizzen said it was fine to introduce breaking changes and preferred it in this case (https://kotlinlang.slack.com/archives/C08SLB97W23/p1757445646431719?thread_ts=1757104129.000169&cid=C08SLB97W23).

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/handler/StreamHandlerContext.kt
dosier: @kpavlov After inspecting the code and similar implementations in the framework, I think the way it is right now matches the rest of the API better. Yes we could move common properties up to `StreamHandlerContext` but that's also not done in similar implementations e.g. [StrategyEventHandlerContext](https://github.com/JetBrains/koog/blob/develop/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/handler/StrategyEventHandlerContext.kt)

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/handler/StreamHandler.kt
dosier: I would agree with you that this seems kinda redundant, but it is in-line with the rest of the api (look at any of the other Handler implementations). 
Imo these should just be type aliases instead of explicitly declared fun interfaces, but that seems out of scope for this PR :D.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/handler/StreamHandlerContext.kt
rubencagnie: Yes, I followed how the framework was already doing the events

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/AIAgentPipeline.kt
dosier: I agree, however, this is currently matching the rest of the api, I'd say this should be it's own issue, improving the logging / error handling of all event handlers.

dosier: This currently matches the rest of the api, we could add logging for each handler invocation, but I don't see why we would do it only for the stream handlers, so perhaps this should be addressed in a separate PR?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/AIAgentPipeline.kt
dosier: Failing when one handler throws is how the rest of the api seems to be designed yes.

(Also regarding the logging, [see previous comment](https://github.com/JetBrains/koog/pull/747#discussion_r2356958888))

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/AIAgentPipeline.kt
dosier: [see previous comment](https://github.com/JetBrains/koog/pull/747#discussion_r2356958888)

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/PromptExecutorProxy.kt
dosier: In my last commits, I added:
```
            .catch { error ->
                logger.debug(error) { "Error in LLM streaming call" }
                pipeline.onStreamError(runId, error)
                throw error
            }
```

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/AIAgentPipeline.kt
dosier: I did add more logging in general:
```kt
    override fun executeStreaming(
        prompt: Prompt,
        model: LLModel,
        tools: List<ToolDescriptor>
    ): Flow<StreamFrame> {
        logger.debug { "Executing LLM streaming call (prompt: $prompt, tools: [${tools.joinToString { it.name }}])" }
        return executor.executeStreaming(prompt, model, tools)
            .onStart {
                logger.debug { "Starting LLM streaming call" }
                pipeline.onBeforeStream(runId, prompt, model, tools)
            }
            .onEach {
                logger.debug { "Received frame from LLM streaming call: $it" }
                pipeline.onStreamFrame(runId, it)
            }
            .catch { error ->
                logger.debug(error) { "Error in LLM streaming call" }
                pipeline.onStreamError(runId, error)
                throw error
            }
            .onCompletion { error ->
                logger.debug(error) { "Finished LLM streaming call" }
                pipeline.onAfterStream(runId, prompt, model, tools)
            }
    }
```

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/streaming/StreamFrame.kt
kpavlov: I am thinking about image and audio models that might return Audio or video, or models returning structured data or JSON, so `Append` might have different data types.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/AIAgentPipeline.kt
kpavlov: Separate PR is fine. This one is big enough

kpavlov:
A script to check the knit part:
```
./gradlew dokkaGeneratePublicationHtml
echo "Starting knit generation..."
FILES_BEFORE_KNIT=$(find docs/src/ -name "*.kt" 2>/dev/null | wc -l || echo "0")
echo "Files before knit: $FILES_BEFORE_KNIT"

./gradlew :docs:knit

FILES_AFTER_KNIT=$(find docs/src/ -name "*.kt" 2>/dev/null | wc -l || echo "0")
echo "Files after knit: $FILES_AFTER_KNIT"

KNIT_GENERATED_FILES=$((FILES_AFTER_KNIT - FILES_BEFORE_KNIT))
echo "Knit generated $KNIT_GENERATED_FILES files"

echo "Starting assemble..."
./gradlew :docs:assemble
```

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/streaming/StreamFrameFlowBuilderError.kt
```diff
@@ -0,0 +1,26 @@
...
+ */
+public sealed class StreamFrameFlowBuilderError(message: String) : Throwable(message) {
+
+    /**
+     * Occurs when trying to upsert partial tool call data (e.g., parts of the tool arguments contents)
+     * when there is no pre-existing partial tool call data to append to.
+     *
+     * The first partial tool call data for a new tool call should always include the `id`.
+     */
+    public object NoPartialToolCallToComplete :
```
EugeneTheDev, line 14: nit: exceptions shouldn't be object

EugeneTheDev (approved):
Looks good to me, thanks! Only a minor style comment (we can merge it as it is now and fix it later)

PR #750, opened 2025-09-06
Make `objectType` nullable in `OpenAIChatCompletion`.

### LINKED ISSUE #749: Field 'object' is required for type with serial name 'ai.koog.prompt.executor.clients.openai.models.OpenAIChatCompletionResponse'
```
Caused by: kotlinx.serialization.MissingFieldException: Field 'object' is required for type with serial name 'ai.koog.prompt.executor.clients.openai.models.OpenAIChatCompletionResponse', but it was missing
```

adamglin0:
Add nullability to the object in OpenAIChatCompletion.

Because in the OpenAI compatibility interface of the BigModel platform/Z.ai platform, this is null.

Fix #749 

## Breaking Changes

---

#### Type of the changes
- [ ] Documentation update
- [ ] Tests improvement
- [ ] Refactoring

#### Checklist

### DIFF
--- a/prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openai/models/OpenAIChatCompletion.kt
+++ b/prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openai/models/OpenAIChatCompletion.kt
@@ -227,7 +227,7 @@ public class OpenAIChatCompletionResponse(
     public val serviceTier: String? = null,
     public val systemFingerprint: String? = null,
     @SerialName("object")
-    public val objectType: String,
+    public val objectType: String? = null,
     public val usage: OpenAIUsage? = null,
 ) : OpenAIBaseLLMResponse
 
--- a/prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openai/models/OpenAIChatCompletion.kt
+++ b/prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openai/models/OpenAIChatCompletion.kt
@@ -227,7 +227,7 @@ public class OpenAIChatCompletionResponse(
     public val serviceTier: String? = null,
     public val systemFingerprint: String? = null,
     @SerialName("object")
-    public val objectType: String = "text_completion",
+    public val objectType: String?,
     public val usage: OpenAIUsage? = null,
 ) : OpenAIBaseLLMResponse
 
@@ -270,6 +270,6 @@ public class OpenAIChatCompletionStreamResponse(
     public val serviceTier: String? = null,
     public val systemFingerprint: String? = null,
     @SerialName("object")
-    public val objectType: String = "text_completion",
+    public val objectType: String?,
     public val usage: OpenAIUsage? = null,
 ) : OpenAIBaseLLMStreamResponse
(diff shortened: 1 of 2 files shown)


kpavlov (changes requested):
Hi, @adamglin0 , thank you very much for the fix.
Would it be possible to add deserialization test to avoid regressions in the future. Is there any known json response that could be used for testing?

Also, would it be more correct to define a default value for the `objectType` instead of making it optional? It would keep compatibility with calling code.

There is a [helper](https://github.com/JetBrains/koog/blob/develop/test-utils/src/commonMain/kotlin/ai/koog/test/utils/SerializationHelpers.kt#L10) you might want to use for deserialization testing.

adamglin0:
> Hi, @adamglin0 , thank you very much for the fix. Would it be possible to add deserialization test to avoid regressions in the future. Is there any known json response that could be used for testing?
> 
> Also, would it be more correct to define a default value for the `objectType` instead of making it optional? It would keep compatibility with calling code.
> 
> There is a [helper](https://github.com/JetBrains/koog/blob/develop/test-utils/src/commonMain/kotlin/ai/koog/test/utils/SerializationHelpers.kt#L10) you might want to use for deserialization testing.

Thank you for your reply!

If I set a default value for objectType, it cannot meet the verifyDeserialization requirement when the`object` is absent from the payload json . Do you have any good suggestions?

prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openai/models/OpenAIChatCompletion.kt
```diff
@@ -227,7 +227,7 @@ public class OpenAIChatCompletionResponse(
     public val serviceTier: String? = null,
     public val systemFingerprint: String? = null,
     @SerialName("object")
-    public val objectType: String,
+    public val objectType: String?,
```
kpavlov:
```suggestion
    @SerialName("object")
    @EncodeDefault
    public val objectType: String = "chat.completion",
```
https://platform.openai.com/docs/api-reference/chat/object

prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openai/models/OpenAIChatCompletion.kt
```diff
@@ -270,6 +270,6 @@ public class OpenAIChatCompletionStreamResponse(
...
```
kpavlov:
```suggestion
    @SerialName("object")
    @EncodeDefault
    public val objectType: String = "chat.completion.chunk",
```
https://platform.openai.com/docs/api-reference/chat_streaming/streaming

kpavlov (commented):
Thank you @adamglin0 for adding test!
According to protocol, the `obectType` is mandatory. What if we provide the default value? I hope the new tests should still pass, but the returned object will be still valid.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openai/models/OpenAIChatCompletion.kt
adamglin0: If we use @EncodeDefault
1. ✅
{object: "chat.completion"}  -> objectType: "chat.completion" -> {object: "chat.completion"}

2. ❌
{}  -> objectType: "chat.completion" -> {object: "chat.completion"}

The test cannot be passed.

aozherelyeva:
Hi @adamglin0, I see the PR is hanging a while. Would you mind updating it or closing it if you don't want to proceed to work on it? Thanks!

adamglin0:
> Hi @adamglin0, I see the PR is hanging a while. Would you mind updating it or closing it if you don't want to proceed to work on it? Thanks!

I will close it

PR #773, opened 2025-09-10
KG-385 [docs] Set up uv for docs project

EugeneTheDev:
Set up a proper uv project in `docs` to make it more stable and easier to run locally and on CI


docs/pyproject.toml
```diff
@@ -0,0 +1,8 @@
+[project]
+name = "koog-docs"
+version = "0.1.0"
```
Rizzen: adjust it with project version

PR #821, opened 2025-09-18
Update build configuration and dependencies

kpavlov:
# Update build configuration and dependencies

- Change versioning logic to use UTC time in `yyyyMMdd-HHmm` format for nightly builds
- Adjust default version suffix to `-SNAPSHOT`
- Bump `okhttp` version from `4.12.0` to `5.1.0`

## Motivation and Context

- Version should include minutes to allow multiple development versions to be published per day
- Default version should include "SNAPSHOT" suffix to avoid collisions with release versions

---

#### Type of the changes
- [ ] Documentation update
- [ ] Tests improvement
- [x] Refactoring

#### Checklist

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -10,13 +10,15 @@ import org.jetbrains.kotlin.gradle.targets.js.ir.KotlinJsIrLink
 import org.jetbrains.kotlin.gradle.tasks.BaseKotlinCompile
 import org.jlleitschuh.gradle.ktlint.KtlintExtension
 import java.time.Clock
+import java.time.ZoneId
+import java.time.format.DateTimeFormatter
 import java.util.Base64
 
 group = "ai.koog"
 version = run {
     // our version follows the semver specification
 
-    val main = "0.4.1"
+    val main = "0.4.3"
 
     val feat = run {
         val releaseBuild = !System.getenv("BRANCH_KOOG_IS_RELEASING_FROM").isNullOrBlank()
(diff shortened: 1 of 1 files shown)


EugeneTheDev (approved):

build.gradle.kts
```diff
@@ -29,8 +31,9 @@ version = run {
             if (branch != "develop") {
                 throw GradleException("Nightly builds are allowed only from the develop branch")
             }
-            val date = Clock.systemUTC().instant().atZone(java.time.ZoneId.of("CET"))
-                .format(java.time.format.DateTimeFormatter.ISO_LOCAL_DATE)
+            val date = Clock.systemUTC().instant()
+                .atZone(ZoneId.of("UTC"))
+                .format(DateTimeFormatter.ofPattern("yyyyMMdd-HHmm"))
```
tiginamaria, line 36: Why we need `HHmm`? It could make the version update harder for those who using nightly

build.gradle.kts
```diff
@@ -66,7 +69,7 @@ version = run {
             }
         } else {
             // do not care
-            ""
+            "-SNAPSHOT"
```
tiginamaria, line 72: I'd personally leave it empty, this code is not reachable I guess, as in teamcity we have only two types of builds: release and nighlty

build.gradle.kts
```diff
@@ -29,8 +31,9 @@ version = run {
             if (branch != "develop") {
                 throw GradleException("Nightly builds are allowed only from the develop branch")
             }
-            val date = Clock.systemUTC().instant().atZone(java.time.ZoneId.of("CET"))
-                .format(java.time.format.DateTimeFormatter.ISO_LOCAL_DATE)
+            val date = Clock.systemUTC().instant()
+                .atZone(ZoneId.of("UTC"))
```
tiginamaria, line 35: Why UTC, should I change the timer in teamcity then?

build.gradle.kts
kpavlov: Because we might want to have multiple builds a day

build.gradle.kts
kpavlov: It is confusing and might overwrite released version in local maven if I do publishToMavenLocal

build.gradle.kts
kpavlov: UTC is a timezone by default, unless timezone is explicitly stated. Do we want to deal with daylight saving times?
If teamcity runs at midnight - then yes. It might be cool if TeamCity could send build number as a parameter instead of having the logic here

PR #824, opened 2025-09-18
Make `NoPartialToolCallToComplete` error a class

dosier:
This is a nitpick change from #747


Ololoshechkin (commented):
@dosier could you please provide some reasoning? Why do you want to have it as class?

dosier:
> @dosier could you please provide some reasoning? Why do you want to have it as class?

@EugeneTheDev mentioned it (https://github.com/JetBrains/koog/pull/747#discussion_r2358644004).

I can think of 2 reasons maybe:
1. data objects persist in memory
2.  maybe for tracing the error it can be nice if it's a class instance since it will have a distinct address

EugeneTheDev:
Thanks for the fix @dosier 
@Ololoshechkin here's a pretty good explanation why it shouldn't be an object. TL;DR the main problem is that it might mess up the stack trace
https://stackoverflow.com/a/71951322

EugeneTheDev (approved):

PR #829, opened 2025-09-20
[agents] Add structured output with tools integration strategy

### LINKED ISSUE #400: Allow tool calls in the structured LLM output nodes
Currently, we have APIs like `nodeLLMRequest` and `AIAgentLLMWriteSession.requestLLM` which allow the LLM to either produce a raw text output or a tool call. We may then honor this tool call, plug the tool result back into the LLM, run the LLM again to produce the final raw text response that incorporates the tool result.

On the other hand, we have APIs like `nodeLLMRequestStructured` and `AIAgentLLMWriteSession.requestLLMStructured` which force the LLM to produce an output that conforms to a given structure. However, these APIs are not flexible enough to allow the LLM to first make tool calls in order to be able to produce the final structured result.

It would be great if tool calls could also work pretty much in the same way when producing structured outputs as when producing raw text outputs.

rubencagnie:
## Motivation and Context
Add new strategy `structuredOutputWithToolsStrategy` that combines structured output capabilities with tool execution, enabling agents to generate structured responses while leveraging external tools.

This implementation includes:
- New `nodeSetStructuredOutput` node for configuring structured output in agent graphs
- `parseResponseToStructuredResponse` method in AIAgentLLMSession for parsing structured responses
- Support for both sequential and parallel tool execution
- Example demonstrating weather forecast generation with tools and structured output

The strategy follows a pipeline: set structured output → execute input → call LLM → handle tool calls or transform structured response → return final output.

Fixes #400

## Breaking Changes
N/A
---

#### Type of the changes
- [ ] Documentation update
- [ ] Tests improvement
- [ ] Refactoring

#### Checklist


examples/simple-examples/src/main/kotlin/ai/koog/agents/example/structuredoutput/tools/WeatherTools.kt
```diff
@@ -0,0 +1,26 @@
...
+        +"The weather forecast for "
+        +city
+        +" is "
+        +"Cloudy"
+        +"temperature = 18"
+        +"precipitation = 30"
+        +"lat = 34.0522, lon = -118.2437"
+        +"pollution = medium"
+        +"alert = Moderate. Possible thunderstorms in the evening, windSpeed = 45.5"
+        +"news = Local news: https://example.com/news, Global news: https://example.com/global-news"
```
devcrocod, line 24: You can use a multiline string here or markdown

prompt/prompt-structure/src/commonMain/kotlin/ai/koog/prompt/structure/PromptExecutorExtensions.kt
```diff
@@ -48,7 +48,65 @@ public data class StructuredOutputConfig<T>(
...
+) {
+    /**
+     * Updates a given prompt to configure structured output using the specified large language model (LLM).
+     * Depending on the model's support for structured outputs, the prompt is updated either manually or natively.
+     *
+     * @param model The large language model (LLModel) used to determine the structured output configuration.
+     * @param prompt The original prompt to be updated with the structured output configuration.
+     * @return A new prompt reflecting the updated structured output configuration.
+     */
+    public fun updatePrompt(model: LLModel, prompt: Prompt): Prompt {
```
devcrocod, line 60: It seems to me that this is not the most obvious place for this method. Also, its behavior when it is called and applied

devcrocod: Maybe it should be made private?

EugeneTheDev (approved):
Thank you for your contribution, the feature looks useful and is cleanly implemented.

The only minor comment, or rather suggestion, that I have is to maybe avoid relying extensively on AI autogenerated KDocs. They are quite often too long and vague and don't help with quickly understanding the main idea of a function/class/property that much.

PR #857, opened 2025-09-24
History compression improvement

### LINKED ISSUE #706: compressing history strips system message off
```kotlin
  val nodeCompressHistory by nodeLLMCompressHistory<String>(
                "compress_history",
                strategy = HistoryCompressionStrategy.FromLastNMessages(1)
            )
```

I intend to compress history with only last user/agent message, but this leads to empty system message.

tiginamaria:
* Refine history compression strategies to preserve system messages in history
* Add history compression strategy for history with multiple system prompts

Fixes #706

## Breaking Changes

---

#### Type of the changes
- [ ] Documentation update
- [ ] Tests improvement
- [ ] Refactoring

#### Checklist

### DIFF
--- a/docs/docs/history-compression.md
+++ b/docs/docs/history-compression.md
@@ -406,13 +406,15 @@ import ai.koog.prompt.message.Message
 class MyCustomCompressionStrategy : HistoryCompressionStrategy() {
     override suspend fun compress(
         llmSession: AIAgentLLMWriteSession,
-        preserveMemory: Boolean,
         memoryMessages: List<Message>
     ) {
         // 1. Process the current history in llmSession.prompt.messages
         // 2. Create new compressed messages
         // 3. Update the prompt with the compressed messages
 
+        // Save original messages to preserve them
+        val originalMessages = llmSession.prompt.messages
+        
         // Example implementation:
         val importantMessages = llmSession.prompt.messages.filter {
             // Your custom filtering logic
@@ -423,10 +425,9 @@ class MyCustomCompressionStrategy : HistoryCompressionStrategy() {
         // Or you can change the current model: `llmSession.model = AnthropicModels.Sonnet_3_7` and ask some other LLM model -- but don't forget to change it back after
 
         // Compose the prompt with the filtered messages
-        composePromptWithRequiredMessages(
-            llmSession,
+        val compressedMessages = composeMessageHistory(
+            originalMessages,
             importantMessages,
-            preserveMemory,
             memoryMessages
         )
     }


agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/extension/HistoryCompressionStrategies.kt
```diff
@@ -41,6 +39,7 @@ public abstract class HistoryCompressionStrategy {
      */
     protected suspend fun compressPromptIntoTLDR(llmSession: AIAgentLLMWriteSession): List<Message.Response> {
         return with(llmSession) {
+            // If there are any tool calls left in a history, we are not allowed to send a user message back
```
antoniibelyshev, line 42: I think, it should be more like "unmatchedToolCalls". Theoretically, we could have a history like [call1, user, call2]. Then both of these calls would be a problem when sending a request to an llm

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/extension/HistoryCompressionStrategies.kt
```diff
@@ -41,6 +39,7 @@ public abstract class HistoryCompressionStrategy {
...
             dropTrailingToolCalls()
```
antoniibelyshev, line 43: There could also be another issue with dropping the tool calls. Let's say the graph is llmCallOnlyTools -> historyCompression -> executeTool -> llmCall, so that when compressing the history, we would have a trailing tool call. However, we'll need it when we make the second llmCall, otherwise we would have a tool result with no tool. If it is hard to workaround, we should at least mention this in the docs, and recommend using the history compression just before an llmCall, so that the message history at that point is valid for sending to the llm

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/extension/HistoryCompressionStrategies.kt
```diff
@@ -182,23 +294,26 @@ public abstract class HistoryCompressionStrategy {
...
             llmSession: AIAgentLLMWriteSession,
-            preserveMemory: Boolean,
             memoryMessages: List<Message>
         ) {
-            val chunkedTLDR = llmSession.prompt.messages.chunked(chunkSize).flatMap { chunk ->
-                llmSession.clearHistory()
-
-                llmSession.prompt = llmSession.prompt.withMessages { chunk }
+            val originalMessages = llmSession.prompt.messages
+            val tldrMessageChunks = llmSession.prompt.messages.chunked(chunkSize).flatMap { messageChunk ->
```
antoniibelyshev, line 304: It is possible that a boundary between two chunks would go in-between a tool call and its result. In that case, the first of these chunks would work fine due to the dropTrailingToolCalls. However, I think that the second chunk would fail since there is a toolResult and there is no tool call

PR #871, opened 2025-09-26
(KG-212) Introduce ResponseProcessor to fix tool call messages from weak models

antoniibelyshev:
[KG-212](https://youtrack.jetbrains.com/issue/KG-212) Add validation and retry logic for tool calling from local models

PR introduces a generic `ResponseProcessor` which can process the messages received from the llm. A few out-of-the-box implementations aim to fix tool call messages, so that the agent follows the expected flow.

## Motivation and Context

Small models (e.g. most ollama models) often fail to properly generate a tool call messages. Some problems are
- incorrectly generated jsons
- Incorrect tool name / parameter names
- Message "I will call this tool" instead of a direct tool call

`ResponseProcessor` and its implementations are designed to deal with such issues.

## Breaking Changes

---

#### Type of the changes
- [ ] Documentation update
- [ ] Tests improvement
- [ ] Refactoring

#### Checklist


agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/processor/ToolCallFixLLMAsAJudje.kt
```diff
@@ -0,0 +1,122 @@
...
+ * @property intentSystemMessage Definition of the system message content used to assess tool call intent.
+ * @property fixSystemMessage Definition of the system message content used to fix incorrect tool calls.
+ * @property getFeedback Feedback generated based on the language model's response.
+ * @property fallback A function to be executed as a fallback when retries have been exhausted.
+ * @property extractJsonToolCall A processing chain responsible for extracting tool call information from responses.
+ * @property maxRetries The number of retries allowed when attempting to fix a tool call.
+ * @property showHistory Determines whether to preserve and showcase conversation history during processing.
+ */
+@ResponseProcessorApi
+public class ToolCallFixLLMAsAJudge(
```
Ololoshechkin: Let's think about this name 🤔 
@tiginamaria @Rizzen WDYT? :)

agents/agents-test/src/jvmTest/kotlin/ai/koog/agents/test/SimpleAgentMockedTest.kt
```diff
@@ -67,42 +70,48 @@ class SimpleAgentMockedTest {
     }
 
     val eventHandlerConfig: EventHandlerConfig.() -> Unit = {
-        onToolCall { eventContext ->
+        onToolExecutionStarting { eventContext ->
```
Ololoshechkin: Why do we need to rename this?

It would break backward compatibility + it's much longer name

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/processor/ExtractToolCall.kt
```diff
@@ -0,0 +1,127 @@
...
+import ai.koog.prompt.message.ResponseMetaInfo
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.json.Json
+import kotlinx.serialization.json.JsonNames
+import kotlinx.serialization.json.JsonObject
+import kotlinx.serialization.json.jsonObject
+
+@Serializable
+private data class ToolCallMessage(
+    @JsonNames("id", "tool_call_id")
```
Ololoshechkin: these names -- is it something that you received in practice from LLMs?

Let's probably allow passing them as configuration parameters? Otherwise if some user would receive yet another name from yet another LLM -- it wouldn't work.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/session/AIAgentLLMWriteSession.kt
```diff
@@ -396,34 +397,55 @@ public class AIAgentLLMWriteSession internal constructor(
...
-     * with the response received from the LLM.
+     * Makes an asynchronous request to a Large Language Model (LLM), processes the response using the provided
+     * responseProcessor, and updates the current prompt with the processed response.
      *
+     * @param responseProcessor The processor to apply to the LLM response.
      * @return A [Message.Response] object containing the response from the LLM.
      */
-    override suspend fun requestLLM(): Message.Response {
-        return super.requestLLM().also { response ->
+    public suspend fun requestLLM(responseProcessor: ResponseProcessor): Message.Response {
```
Ololoshechkin: Let's add a default value here

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/processor/ToolCallFixLLMAsAJudje.kt
```diff
@@ -0,0 +1,123 @@
...
+    private val intentSystemMessage: TextContentBuilderBase<*>.() -> Unit = { assessToolCallIntent() },
+    private val fixSystemMessage: TextContentBuilderBase<*>.() -> Unit = { fixToolCall() },
```
Ololoshechkin: Why builder lambdas, and not just strings?

User can create their own `TextContentBuilderBase` if they want:

`fun f(arg: String) {...}`

`f(markdown { ... })` / `f(text { ... })` -- I'm calling them by myself. 
Or when I don't need  builders -- `f("just text")` is more convenient

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/extension/AIAgentLLMActions.kt
```diff
@@ -132,3 +132,20 @@ public suspend fun AIAgentLLMWriteSession.replaceHistoryWithTLDR(
 public fun AIAgentLLMWriteSession.dropTrailingToolCalls() {
     rewritePrompt { prompt -> prompt.withMessages { messages -> messages.dropLastWhile { it is Message.Tool.Call } } }
 }
+
+/**
+ * Runs provided block on a copy of the context, without changing the original context.
+ */
+public suspend fun <T> AIAgentLLMWriteSession.withTemporaryContext(
```
Ololoshechkin: You can also just call `copy()` on `AIAgentContext` and create an inner context. And then -- call anything on this new context without messing up the initial one.

Ololoshechkin (changes requested):

agents/agents-test/src/jvmTest/kotlin/ai/koog/agents/test/SimpleAgentMockedTest.kt
antoniibelyshev: `onToolCall` is a deprecated version. I did not introduce the updated naming, I just updated the names to the new ones in this test, and it shouldn't break backward compatibility. Would it be better to leave the deprecated version here? I can change it back :slightly_smiling_face:

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/processor/ToolCallFixLLMAsAJudje.kt
antoniibelyshev: Agree

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/extension/AIAgentLLMActions.kt
antoniibelyshev: There is no access to the `AIAgentContext` inside the `AIAgentLLMWriteSession`, but I agree that `copy` functionality would be better. I will add such method for the `AIAgentLLMWriteSession`

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/processor/ToolCallFixLLMAsAJudje.kt
antoniibelyshev: WDYT of `LLMBasedToolCallExtractor`?

ptitjes:
@antoniibelyshev some models also produce Python-formatted tool calls. Do you also handle this case?

I had to make a small Python method-call-expression-with-literal-arguments parser for that.
cf. https://github.com/ptitjes/konvo/blob/develop/konvo-core/src/commonMain/kotlin/io/github/ptitjes/konvo/core/agents/toolkit/CallFixingPromptExecutor.kt#L34
and https://github.com/ptitjes/konvo/tree/develop/konvo-core/src/commonMain/kotlin/io/github/ptitjes/konvo/core/agents/toolkit/python
You can take/reuse/modify that code it if you want to.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/processor/ExtractToolCall.kt
antoniibelyshev: Yes, these names are actual json keys generated by LLMs in different combinations)) I will update the code to allow customization

agents/agents-test/src/jvmTest/kotlin/ai/koog/agents/test/SimpleAgentMockedTest.kt
Ololoshechkin: Thanks for noticing! I've asked @sdubov about the new naming :)

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/processor/ResponseProcessor.kt
```diff
@@ -0,0 +1,75 @@
...
+import kotlin.RequiresOptIn
+import kotlin.jvm.JvmStatic
+
+/**
+ * Opt-in annotation for ResponseProcessor API.
+ */
+@RequiresOptIn(
+    message = "ResponseProcessor API can update LLM responses and make additional calls to LLM. Please opt-in only when you reviewed the api and understand the risks."
+)
+public annotation class ResponseProcessorApi
```
EugeneTheDev: Don't we need to apply the annotation on the base `ResponseProcessor` class too?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/processor/ResponseProcessor.kt
```diff
@@ -0,0 +1,75 @@
...
+        process(session, listOf(message)).first()
+
+    /**
+     * Updates a list of messages.
+     *
+     * @param session The session to use for updating.
+     * @param messages The messages to update.
+     * @return The updated messages.
+     */
+    protected abstract suspend fun updateMessages(
```
EugeneTheDev: nit: this is actually immutable operation, it doesn't update any state. I would suggest `processMessages` instead, it communicates the idea a bit better - transformation from `A` to `B`

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/processor/JsonUtils.kt
```diff
@@ -0,0 +1,173 @@
...
+            decodedToolCall.tool,
+            decodedToolCall.args.toString(),
+            metaInfo
+        )
+    }
+
+    val toolName = getToolName(messageContent, toolKeys) ?: return null
+
+    val params = runCatching {
+        session.toolRegistry.getTool(toolName).descriptor.requiredParameters
```
EugeneTheDev: AFAIU you don't need the whole `AIAgentLLMWriteSession`, only the `ToolRegistry`, right? In this case it seems cleaner to decouple this whole file from the `agents-core` and move it to the `agents-tool`, since it contains some tool-related logic that doesn't need the whole AI agent mechanism. Wdyt?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/processor/ExtractToolCall.kt
```diff
@@ -0,0 +1,87 @@
...
+        ignoreUnknownKeys = true
+        isLenient = true
+    },
+    private val idKeys: List<String> = defaultIdKeys,
+    private val toolKeys: List<String> = defaultToolKeys,
+    private val argsKeys: List<String> = defaultArgsKeys,
+    private val allowEscapedBackslash: Boolean = false,
+) : ResponseProcessor() {
+    override suspend fun updateMessages(
+        session: AIAgentLLMWriteSession,
```
EugeneTheDev: Same here, should be possible to decouple and move this whole file if the previous one is moved

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/processor/ResponseProcessor.kt
```diff
@@ -0,0 +1,75 @@
...
+        session: AIAgentLLMWriteSession,
```
EugeneTheDev: Do we need to create such a close connection with the AI agent session? You can also get tool call messages without AI agent, using only prompt executor. On top of that, all the messages that are sent this way, through the agent session, might be appended to the agent's message history, which might no be something that we want, if these are just some utility calls. Take `StructureFixingParser` for example, it does not depend on the agent, doesn't preserve it's utility requests in the message history and operates using direct requests on prompt executor. I suggest using a similar approach here, and also potentially moving this whole component out of the `agent-core` if there won't be any tight coupling anymore

EugeneTheDev (changes requested):
Generally looks good, but I have some architectural comments and suggestion, please check them and let me know what do you think.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/processor/ResponseProcessor.kt
Ololoshechkin: Or `processResponses`

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/processor/ResponseProcessor.kt
Ololoshechkin: +1 to that! Let's just create `List<Response> -> List<Response>` functional interface which is clear and can be created separately from agents

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/processor/ResponseProcessor.kt
```diff
@@ -0,0 +1,75 @@
...
+            session: AIAgentLLMWriteSession,
+            messages: List<Message.Response>
+        ): List<Message.Response> {
+            var result = messages
+            for (processor in processors) {
+                result = processor.process(session, result)
+            }
+            return result
+        }
+    }
```
Ololoshechkin: Maybe just override operator `plus` ?

Ololoshechkin: instead of `Chain(processor1, processor2, processor3)` -- will be just `processor1 + processor2 + processor3`

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/prompt/Prompts.kt
```diff
@@ -56,4 +56,120 @@ internal object Prompts {
             +"This summary will be the ONLY context available for continuing this conversation, along with the system message."
             +"Ensure it contains ALL essential information needed to proceed effectively."
         }
+
+    fun TextContentBuilderBase<*>.assessToolCallIntent(): Unit =
```
Ololoshechkin: Do you actually need the `TextContentBuilderBase<*>` receiver here?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/processor/LLMBasedToolCallExtractor.kt
```diff
@@ -0,0 +1,133 @@
...
+
+/**
+ * A response processor that fixes incorrectly communicated tool calls.
+ *
+ * Applies LLM-As-A-Judge approach to assess if a tool call was intended.
+ * If a tool call was intended, LLM is asked to do a proper llm call.
+ */
+@ResponseProcessorApi
+public class LLMBasedToolCallExtractor(
+    private val intentSystemMessage: String = text { assessToolCallIntent() },
```
Ololoshechkin: `text { assessToolCallIntent() }` looks redundant. Maybe just `ToolCallExtractor.Prompts.ASSES_TOOL_CALL_INTENT` which is a companion object static String?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/processor/LLMBasedToolCallExtractor.kt
```diff
@@ -0,0 +1,133 @@
...
+    private val fixSystemMessage: String = text { fixToolCall() },
```
Ololoshechkin: same here

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/processor/LLMBasedToolCallExtractor.kt
```diff
@@ -0,0 +1,133 @@
...
+    private val getFeedback: AIAgentLLMWriteSession.(Message.Response, List<String>) -> String? =
+        { message, toolKeys -> defaultGetFeedback(message, toolKeys) },
+    idKeys: List<String> = defaultIdKeys,
```
Ololoshechkin: Please add all parameter descriptions (ideally -- with usage examples) to KDocs here. What is `idKeys` and `toolKeys` is not clear without reading the code :)

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/processor/ExtractToolCall.kt
```diff
@@ -0,0 +1,87 @@
...
+@ResponseProcessorApi
+public class ExtractJsonToolCall(
+    private val json: Json = Json {
+        ignoreUnknownKeys = true
+        isLenient = true
+    },
+    private val idKeys: List<String> = defaultIdKeys,
+    private val toolKeys: List<String> = defaultToolKeys,
+    private val argsKeys: List<String> = defaultArgsKeys,
+    private val allowEscapedBackslash: Boolean = false
```
Ololoshechkin: Please add parameter KDocs here too

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/processor/LLMBasedToolCallExtractor.kt
Ololoshechkin: Upd: maybe let's rename:
- `idKeys` to `toolIdJsonKeys`
- `toolKeys` to `toolNameJsonKeys`

or same but without "Json":
- `idKeys` to `toolIdKeys`
- `toolKeys` to `toolNameKeys`

otherwise, what is 'id key' ? :)

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/processor/LLMBasedToolCallExtractor.kt
```diff
@@ -0,0 +1,133 @@
...
+    private val toolKeys: List<String> = defaultToolKeys,
+    argsKeys: List<String> = defaultArgsKeys,
+    allowEscapedBackslash: Boolean = false,
+    private val fallback: suspend AIAgentLLMWriteSession.(Message.Response) -> Message.Response = { it },
```
Ololoshechkin: How is fallback used? Ideally the name should explain the meaning/action. Currently it doesn't :)
Also, please don't forget KDoc

Ololoshechkin: Additionally, let's please not introduce `AIAgentLLMWriteSession` as a receiver in the lambda parameter here and above unless absolutely neccesary. FIrst of all, context receiver lambda parameters require lots of work to support from Java, second -- it's actually almost not used here -- everything that is implemented so far can be done without it (maybe an additional `toolRegistry` field will be required instead, though).

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/prompt/Prompts.kt
Ololoshechkin: Let's just keep it as a String value inside some object?

Ololoshechkin (dismissed):
Also agree with @EugeneTheDev that it's better to go without the agent session at all. It will be also much easier to support (especially in Java API)

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/processor/ResponseProcessor.kt
antoniibelyshev: I agree that it would be a better solution -- to have this component outside of the agents module. However, there is a problem with the access to the tools (not only the tool descriptors, since some implementations need to parse the arguments). Currently, the only way to do that is to pass the tool registry to the response processor. To achieve that, I see a few solutions:

1. Define the `ResponseProcessor` in the `prompt` module, and if an implementation needs the tool registry -- implement it in the `agents` module and require the tool registry in the constructor.

2. Make the ResponseProcessor.processResponses to accept List<Tool> (and not the List<ToolDescriptor>>, but that would mean the LLMSession.tools should also be List<Tool>, not List<ToolDescriptor>.

What do you think will be better?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/processor/ResponseProcessor.kt
antoniibelyshev: I thought of that, but then the default implementation would also require the `OptIn`, which is not cool. However, I decided to use nullable processor everywhere, with null being the default (meaning no processing), thus no default implementation needed and I can apply the annotation to the base `ResponseProcessor`

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/processor/ResponseProcessor.kt
antoniibelyshev: I think, just `process` should be good since the class name already has `response` in it

prompt/prompt-processor/src/commonMain/kotlin/ai/koog/prompt/processor/Messages.kt
```diff
@@ -0,0 +1,133 @@
+package ai.koog.prompt.processor
+
+import ai.koog.agents.core.tools.ToolDescriptor
+import ai.koog.prompt.markdown.markdown
+
+internal object Messages {
```
EugeneTheDev: Let's call it `Prompts`, like we do it in other places.

prompt/prompt-processor/src/commonMain/kotlin/ai/koog/prompt/processor/ResponseProcessor.kt
```diff
@@ -0,0 +1,74 @@
...
+ */
+@RequiresOptIn(
+    message = "ResponseProcessor API can update LLM responses and make additional calls to LLM. Please opt-in only when you reviewed the api and understand the risks."
+)
+public annotation class ResponseProcessorApi
+
+/**
+ * A processor for handling and modifying LLM responses.
+ */
+public abstract class ResponseProcessor @ResponseProcessorApi constructor() {
```
EugeneTheDev: Why did you put the annotation only on constructor? Can we just set it on the whole class?

prompt/prompt-processor/src/commonMain/kotlin/ai/koog/prompt/processor/ToolCallJsonFixConfig.kt
```diff
@@ -0,0 +1,195 @@
...
+
+/**
+ * Configuration for parsing and fixing tool call json
+ *
+ * @param json The json parser to use for parsing tool call json
+ * @param idJsonKeys The keys used by various models for tool ID in tool call json
+ * @param nameJsonKeys The keys used by various models for tool name in tool call json
+ * @param argsJsonKeys The keys used by various models for tool arguments in tool call json
+ */
+public class ToolCallJsonFixConfig(
```
EugeneTheDev: There are a couple of things that look a bit off if this is inteded as a config class. Configs are usually just structures, without complicated logic inside, so:
1. Let's make configured properties public, not private
2. The logic inside looks quite complex and should not belong to configs class. Let's move it out somewhere?

EugeneTheDev (changes requested):
Looks mostly good now, a couple of minor comments

prompt/prompt-processor/src/commonMain/kotlin/ai/koog/prompt/processor/ResponseProcessor.kt
antoniibelyshev: If I put the annotation on the class, then the functions in the `LLMSession` and in the built-in nodes also require the opt-in. The annotation on the constructor makes the opt-in required only for the non-default options (when the user creates an instance)

EugeneTheDev: AFAIU usually we want to mark the whole component with OptIn annotation, since all its members are part of the API, no?

prompt/prompt-processor/src/commonMain/kotlin/ai/koog/prompt/processor/Messages.kt
antoniibelyshev: I thought that `Messages` is more precise since in koog `prompt` is a list of messages, and here are individual messages. I will rename it to `Prompts` for now, but I would suggest discussing this naming at some point (and probably creating a single place for all messages/prompts)

prompt/prompt-processor/src/commonMain/kotlin/ai/koog/prompt/processor/ResponseProcessor.kt
EugeneTheDev: You don't have to propagate OptIn if you don't want to. We also use this pattern a lot in our built-in components that use `Internal...` marked components. Just put `@OptIn(<your opt in annotation>)` where you use this component if you want to aknowledge the OptIn and not propagate it further

prompt/prompt-processor/src/commonMain/kotlin/ai/koog/prompt/processor/Messages.kt
EugeneTheDev: Ah, I see where the confusion comes from. Here we call it prompt in its "original" sense - i.e., a text that is send to LLM to complete it

EugeneTheDev: That's why prompts reflects the idea a bit better, IMHO

prompt/prompt-processor/src/commonMain/kotlin/ai/koog/prompt/processor/ToolCallJsonFixConfig.kt
antoniibelyshev: I tried to move the functionality out of the config, and it seems to be an unnecessary separation. Probably, it just should not be a config, but rather a heuristic for example? Then the change would be to just rename `ToolCallJsonFixConfig` -> `ToolCallJsonFixHeuristic`. And probably making the values public. WDYT?

prompt/prompt-processor/src/commonMain/kotlin/ai/koog/prompt/processor/ResponseProcessor.kt
```diff
@@ -0,0 +1,74 @@
...
+
+    /**
+     * Processes the given LLM responses.
+     * These responses were received using [executor], [prompt], [model], [tools].
+     */
+    internal abstract suspend fun process(
```
EugeneTheDev: Missed it the first time, why is this method `internal`? The class itself is `public` meaning anyone can extend it. Let's make this `process` method public too

prompt/prompt-processor/src/commonMain/kotlin/ai/koog/prompt/processor/ResponseProcessor.kt
```diff
@@ -0,0 +1,74 @@
...
+        executor: PromptExecutor,
+        prompt: Prompt,
+        model: LLModel,
+        tools: List<ToolDescriptor>,
+        responses: List<Message.Response>
+    ): List<Message.Response>
+
+    internal suspend fun process(
```
EugeneTheDev: And this one `public` as well

EugeneTheDev (changes requested):
Sorry, a few more comments, just noticed a few more things

prompt/prompt-processor/src/commonMain/kotlin/ai/koog/prompt/processor/ToolCallJsonFixConfig.kt
EugeneTheDev: BTW, why methods are `internal`, they look useful enough to be used in custom user implementations? I took a closer look at this class and how it is used, and here's what I can suggest. AFAIK it contains a set of useful functions both for LLM-based and manual tool call fixing. How about we reuse similar approach from Structured Output and turn it into an abstract class: `ToolFixProcessor`. It would extend `ResponseProcessor` and contains these useful utility methods as `protected`. Then these two existing (LLM-based and manual) processors would extend this new class and use some of its utility methods needed. BTW x2, lets also adjust naming for these existing processors a bit, to make it more consistent: `ManualToolFixProcessor`, `LLMToolFixProcessor`, something like this. This way hierarchy should look clearer. WDYT?

prompt/prompt-processor/src/commonMain/kotlin/ai/koog/prompt/processor/ResponseProcessor.kt
antoniibelyshev: I tried to annotate the class, not the constructor, and added the opt-in where the ResponseProcessor is used. However, the functions which use ResponseProcessor as a parameter, still require the opt-in for their usage

EugeneTheDev (approved):

PR #879, opened 2025-09-29
Add support for executing shell commands with user confirmation and timeout handling

Faanbaria:
This PR adds an ExecuteShellCommandTool with JVM executor for running system commands through AI agents. Includes user confirmation before execution, timeouts, cross-platform shell support, and captures stdout, stderr, and exit codes. The JVM implementation uses ProcessBuilder.
https://youtrack.jetbrains.com/issue/KG-323/Move-ExecTool-to-Koog

#### Type of the changes
- [ ] Documentation update
- [ ] Tests improvement
- [ ] Refactoring

#### Checklist

### DIFF
--- a/agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/shell/shell/ShellCommandConfirmation.kt
+++ b/agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/shell/shell/ShellCommandConfirmation.kt
@@ -1,0 +1,36 @@
+package ai.koog.agents.ext.shell.shell
+
+/**
+ * User's decision on whether to execute a shell command.
+ */
+public sealed class ShellCommandConfirmation {
+    /** Command execution approved. */
+    public data object Approved : ShellCommandConfirmation()
+
+    /** Command execution denied. */
+    public data object Denied : ShellCommandConfirmation()
+
+    /**
+     * Command execution denied with explanation.
+     *
+     * @property reason Explanation for denial
+     */
+    public data class DeniedWithReason(val reason: String) : ShellCommandConfirmation()
+}
+
+/**
+ * Strategy for obtaining user confirmation before executing shell commands.
+ */
+public interface ShellCommandConfirmationHandler {
+    /**
+     * Requests confirmation to execute a command.
+     *
+     * @param command Command to execute
+     * @param workingDirectory Working directory, or null to use the current directory
+     * @return Approval or denial decision
+     */
+    public suspend fun requestConfirmation(
+        command: String,
+        workingDirectory: String?
+    ): ShellCommandConfirmation
+}
--- a/agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/shell/shell/ShellCommandConfirmation.kt
+++ b/agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/shell/shell/ShellCommandConfirmation.kt
@@ -7,15 +7,12 @@ public sealed class ShellCommandConfirmation {
     /** Command execution approved. */
     public data object Approved : ShellCommandConfirmation()
 
-    /** Command execution denied. */
-    public data object Denied : ShellCommandConfirmation()
-
     /**
      * Command execution denied with explanation.
      *
-     * @property reason Explanation for denial
+     * @property userResponse A free-form user reply when not approving; may be a reason or simply a “no”.
      */
-    public data class DeniedWithReason(val reason: String) : ShellCommandConfirmation()
+    public data class Denied(val userResponse: String) : ShellCommandConfirmation()
 }
 
 /**
--- a/agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/shell/shell/ShellCommandConfirmationHandlers.kt
+++ b/agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/shell/shell/ShellCommandConfirmationHandlers.kt
@@ -17,7 +17,7 @@ public class AlwaysDenyConfirmationHandler : ShellCommandConfirmationHandler {
     override suspend fun requestConfirmation(
         command: String,
         workingDirectory: String?
-    ): ShellCommandConfirmation = ShellCommandConfirmation.Denied
+    ): ShellCommandConfirmation = ShellCommandConfirmation.Denied("User not available, all commands will be denied")
 }
 
 /**
@@ -30,13 +30,14 @@ public class PrintShellCommandConfirmationHandler : ShellCommandConfirmationHand
         command: String,
         workingDirectory: String?
     ): ShellCommandConfirmation {
-        println("Execute: $command")
+        println("Agent wants to execute: $command")
         workingDirectory?.let { println("In: $it") }
-        print("Confirm (y/n): ")
+        print("Confirm (y / n / reason-for-denying): ")
 
-        return when (readln().lowercase()) {
+        val userResponse = readln().lowercase()
+        return when (userResponse) {
             "y", "yes" -> ShellCommandConfirmation.Approved
-            else -> ShellCommandConfirmation.Denied
+            else -> ShellCommandConfirmation.Denied(userResponse)
         }
     }
 }
--- a/agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/shell/shell/ShellCommandConfirmationHandlers.kt
+++ b/agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/shell/shell/ShellCommandConfirmationHandlers.kt
@@ -1,0 +1,42 @@
+package ai.koog.agents.ext.shell.shell
+
+/**
+ * Confirmation handler that always approves commands.
+ */
+public class AlwaysApproveConfirmationHandler : ShellCommandConfirmationHandler {
+    override suspend fun requestConfirmation(
+        command: String,
+        workingDirectory: String?
+    ): ShellCommandConfirmation = ShellCommandConfirmation.Approved
+}
+
+/**
+ * Confirmation handler that always denies commands.
+ */
+public class AlwaysDenyConfirmationHandler : ShellCommandConfirmationHandler {
+    override suspend fun requestConfirmation(
+        command: String,
+        workingDirectory: String?
+    ): ShellCommandConfirmation = ShellCommandConfirmation.Denied
+}
+
+/**
+ * Confirmation handler that prompts user via console input.
+ *
+ * Prints command details and waits for y/yes to approve, any other input denies.
+ */
+public class PrintShellCommandConfirmationHandler : ShellCommandConfirmationHandler {
+    override suspend fun requestConfirmation(
+        command: String,
+        workingDirectory: String?
+    ): ShellCommandConfirmation {
+        println("Execute: $command")
+        workingDirectory?.let { println("In: $it") }
+        print("Confirm (y/n): ")
+
+        return when (readln().lowercase()) {
+            "y", "yes" -> ShellCommandConfirmation.Approved
+            else -> ShellCommandConfirmation.Denied
+        }
+    }
+}
--- a/agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/tool/shell/ExecuteShellCommandTool.kt
+++ b/agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/tool/shell/ExecuteShellCommandTool.kt
@@ -4,6 +4,7 @@ import ai.koog.agents.core.tools.Tool
 import ai.koog.agents.core.tools.annotations.LLMDescription
 import kotlinx.serialization.KSerializer
 import kotlinx.serialization.Serializable
+import kotlin.coroutines.cancellation.CancellationException
 
 /**
  * Executes shell commands with user approval and automatic timeout.
@@ -104,14 +105,17 @@ public class ExecuteShellCommandTool(
      *
      * @param args The command string, timeout duration, optional working directory
      * @return Result containing the command output and exit code, or an error message explaining why it didn't run
-     */
+     * @throws CancellationException if canceled while the command is executing
+     **/
     override suspend fun execute(args: Args): Result = when (
         val confirmation = confirmationHandler.requestConfirmation(args)
     ) {
-        is ShellCommandConfirmation.Approved -> runCatching {
+        is ShellCommandConfirmation.Approved -> try {
             val result = executor.execute(args.command, args.workingDirectory, args.timeoutSeconds)
             Result(args.command, result.exitCode, result.output)
-        }.getOrElse { e ->
+        } catch (e: CancellationException) {
+            throw e
+        } catch (e: Exception) {
             Result(args.command, null, "Failed to execute command: ${e.message}")
         }
 
--- a/agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/tool/shell/ExecuteShellCommandTool.kt
+++ b/agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/tool/shell/ExecuteShellCommandTool.kt
@@ -35,15 +35,18 @@ public class ExecuteShellCommandTool(
         )
         val command: String,
         @property:LLMDescription(
+            "Maximum execution time, in seconds. " +
+                "Commands that exceed this limit are terminated; " +
+                "commands that finish sooner return immediately. " +
+                "Choose a value that balances long-running work (to avoid cutting it off near completion) " +
+                "against the risk of hangs, infinite loops, or waiting indefinitely."
+        )
+        val timeoutSeconds: Int,
+        @property:LLMDescription(
             "An absolute filesystem path where the command runs. Must exist and be accessible; otherwise execution will fail immediately." +
                 "Optional. Default: uses the current working directory when null."
         )
-        val workingDirectory: String? = null,
-        @property:LLMDescription(
-            "Maximum execution time in seconds. If the command runs longer, it will be terminated." +
-                "Optional. Default: 60 seconds."
-        )
-        val timeoutSeconds: Int = 60
+        val workingDirectory: String? = null
     )
 
     /**
@@ -99,7 +102,7 @@ public class ExecuteShellCommandTool(
      * with the specified timeout. If denied, timed out, or crashes, returns a result explaining
      * what happened.
      *
-     * @param args The command string, optional working directory, and timeout duration
+     * @param args The command string, timeout duration, optional working directory
      * @return Result containing the command output and exit code, or an error message explaining why it didn't run
      */
     override suspend fun execute(args: Args): Result = when (
--- a/agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/tool/shell/ExecuteShellCommandTool.kt
+++ b/agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/tool/shell/ExecuteShellCommandTool.kt
@@ -28,11 +28,21 @@ public class ExecuteShellCommandTool(
      */
     @Serializable
     public data class Args(
-        @property:LLMDescription("Shell command to execute (e.g., 'ls -la', 'git status')")
+        @property:LLMDescription(
+            "The exact shell command line to execute." +
+                "- Examples: 'git status', './gradlew assemble', 'ls -la'" +
+                "- Do not include 'cd' commands; use workingDirectory instead."
+        )
         val command: String,
-        @property:LLMDescription("Working directory for execution, or null for current directory")
+        @property:LLMDescription(
+            "An absolute filesystem path where the command runs. Must exist and be accessible; otherwise execution will fail immediately." +
+                "Optional. Default: uses the current working directory when null."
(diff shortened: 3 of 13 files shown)


agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/shell/shell/ShellCommandConfirmation.kt
```diff
@@ -0,0 +1,33 @@
+package ai.koog.agents.ext.shell.shell
```
BLannoo: Should this maybe be:

package ai.koog.agents.ext.tool.shell

And also moved to that folder?

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/tool/shell/ShellCommandTool.kt
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    override val argsSerializer: KSerializer<Args> = Args.serializer()
+    override val resultSerializer: KSerializer<Result> = Result.serializer()
+    override val name: String = "__execute_shell_command__"
+    override val description: String = """
+        Executes shell commands on the host system with user confirmation.
+        
+        Captures all text printed by the command along with the exit code.
+        Requires user approval before execution. Commands run with configurable timeout (default: 60 seconds).
```
BLannoo: Not entirely sure of this wording, it doesn't work so well in the case we give auto-approval. But maybe we don't need to consider that case?

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/tool/shell/ShellCommandTool.kt
```diff
@@ -0,0 +1,120 @@
...
+        
+        Do not use to change system configuration, install software, or escalate privileges.
```
BLannoo: Does pip install count as installing software and do we want the agent to refrain from that?

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/shell/shell/ShellCommandConfirmationHandlers.kt
```diff
@@ -0,0 +1,43 @@
...
+    override suspend fun requestConfirmation(
+        command: String,
+        workingDirectory: String?
+    ): ShellCommandConfirmation = ShellCommandConfirmation.Approved
+}
+
+/**
+ * Confirmation handler that always denies commands.
+ */
+public class AlwaysDenyConfirmationHandler : ShellCommandConfirmationHandler {
```
BLannoo: I am writing out the article and realise on a functional level this confirmation handler does not make much sense.

Like why provide the tool, if you will always deny execution, this is the same as not adding the tool to your LLM. But not adding the tool is less confusing.

I suggest dropping this one.

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/tool/shell/ExecuteShellCommandTool.kt
```diff
@@ -0,0 +1,120 @@
...
+         * - The exit code (omitted if the command was denied or timed out)
+         *
+         * @return Formatted string ready for display or logging
+         */
+        public fun textForLLM(): String = buildString {
+            appendLine("Command: $command")
+            if (output.isNotEmpty()) {
+                append(output)
+                if (!output.endsWith('\n')) appendLine()
+            }
```
BLannoo, line 82: Otherwise state output was empty

Faanbaria: I added the message (no output) in that case.

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/tool/shell/ShellCommandTool.kt
```diff
@@ -0,0 +1,122 @@
...
+    ) {
+        is ShellCommandConfirmation.Approved -> runCatching {
+            withTimeout(args.timeoutSeconds.seconds) {
+                val result = executor.execute(args.command, args.workingDirectory)
+                Result(args.command, result.exitCode, result.output)
+            }
+        }.getOrElse { e ->
+            when (e) {
+                is TimeoutCancellationException ->
+                    Result(args.command, null, "Command timed out after ${args.timeoutSeconds} seconds")
```
BLannoo: Would be nice if we can have the partial logs in case of time-out or failure

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/tool/shell/ShellCommandTool.kt
Faanbaria: I changed it to use `Depending on configuration, the command may run immediately or ask the user before running.`

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/tool/shell/ExecuteShellCommandTool.kt
```diff
@@ -0,0 +1,118 @@
...
+     * @property command The command to run (e.g., "ls -la", "git status", "npm install")
+     * @property workingDirectory Directory to run the command in, or null to use the current directory
+     * @property timeoutSeconds Cancel the command if it runs longer than this many seconds (default: 60)
+     */
+    @Serializable
+    public data class Args(
+        @property:LLMDescription(
+            "The exact shell command line to execute." +
+                "- Examples: 'git status', './gradlew assemble', 'ls -la'" +
+                "- Do not include 'cd' commands; use workingDirectory instead."
```
sdubov: >"- Do not include 'cd' commands; use workingDirectory instead."

Is it due to a performance reasons or any other? Would not LLM perform better without this limitation and call `cd` directly when it needs?

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/tool/shell/ExecuteShellCommandTool.kt
```diff
@@ -0,0 +1,118 @@
...
+        @property:LLMDescription(
+            "An absolute filesystem path where the command runs. Must exist and be accessible; otherwise execution will fail immediately." +
+                "Optional. Default: uses the current working directory when null."
+        )
+        val workingDirectory: String? = null,
+        @property:LLMDescription(
+            "Maximum execution time in seconds. If the command runs longer, it will be terminated." +
+                "Optional. Default: 60 seconds."
+        )
+        val timeoutSeconds: Int = 60
```
sdubov: Could you please clarify why is it 60 seconds? it sounds like a long timeout for a command execution.

agents/agents-ext/src/jvmMain/kotlin/ai/koog/agents/ext/tool/shell/JvmShellCommandExecutor.kt
```diff
@@ -0,0 +1,109 @@
...
+     * @param timeoutSeconds Maximum execution time in seconds, or null for no timeout
+     * @return [ExecutionResult] containing combined stdout/stderr output and process exit code
+     */
+    override suspend fun execute(
+        command: String,
+        workingDirectory: String?,
+        timeoutSeconds: Int?
+    ): ExecutionResult = withContext(Dispatchers.IO) {
+        val shellCommand = if (IS_WINDOWS) {
+            listOf("cmd.exe", "/c", command)
```
sdubov: I would probably specify the full path here to make sure we invoke a system `cmd.exe` tool: 
```
val systemRoot = System.getenv("SystemRoot") ?: "C:\\Windows"
val cmdTool = "$systemRoot\\System32\\cmd.exe"
```

agents/agents-ext/src/jvmMain/kotlin/ai/koog/agents/ext/tool/shell/JvmShellCommandExecutor.kt
```diff
@@ -0,0 +1,109 @@
...
+        } else {
+            listOf("bash", "-c", command)
```
sdubov: I would update here for "/bin/bash -c" as well.

Do you need to keep compatibility between different shells? Alternative more universal approach would be using "/bin/sh -c" here instead.

agents/agents-ext/src/jvmMain/kotlin/ai/koog/agents/ext/tool/shell/JvmShellCommandExecutor.kt
```diff
@@ -0,0 +1,109 @@
...
+        }
+
+        val stderrJob = launch(Dispatchers.IO) {
+            process.errorStream.bufferedReader().useLines { lines ->
+                lines.forEach { stderrBuilder.append(it).append('\n') }
+            }
+        }
+
+        val completed = if (timeoutSeconds != null) {
+            process.waitFor(timeoutSeconds.toLong(), TimeUnit.SECONDS)
```
sdubov: If I do not miss anything, `process.waitFor()` is a synchronous call. It will block the thread event when the coroutine is cancelled, e.g. imagine the case when you run an agent that call a command that is running 60 seconds. The agent is terminated by a user and it needs to cancel the execution. However, we will be waiting for a `waitFor()` to finish before terminating the process.

agents/agents-ext/src/jvmMain/kotlin/ai/koog/agents/ext/tool/shell/JvmShellCommandExecutor.kt
```diff
@@ -0,0 +1,109 @@
...
+                appendLine(timeoutMessage)
+            }.trimEnd()
+
+            return@withContext ExecutionResult(
+                output = combinedOutput,
+                exitCode = null
+            )
+        }
+
+        withTimeout(1.seconds) {
```
sdubov: Could you please clarify why is it timeout 1 second here?

agents/agents-ext/src/jvmMain/kotlin/ai/koog/agents/ext/tool/shell/JvmShellCommandExecutor.kt
```diff
@@ -0,0 +1,109 @@
...
+            stdoutJob.join()
+            stderrJob.join()
+        }
+
+        val stdoutResult = stdoutBuilder.toString().replace("\r\n", "\n").trimEnd()
```
sdubov: Interesting... Could you please clarify why do you override the ending here?

agents/agents-ext/src/jvmMain/kotlin/ai/koog/agents/ext/tool/shell/JvmShellCommandExecutor.kt
```diff
@@ -0,0 +1,109 @@
...
+        val stderrResult = stderrBuilder.toString().replace("\r\n", "\n").trimEnd()
+
+        val combinedOutput = buildString {
+            if (stdoutResult.isNotEmpty()) appendLine(stdoutResult)
```
sdubov: It seems that you do not need to have a check for `isEmpty()` here. It should add an empty is empty string when either is empty,

sdubov (changes requested):

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/tool/shell/ExecuteShellCommandTool.kt
BLannoo: Maybe we need the word this, but we are mostly trying to communicate to the LLM that each execTool call is executed in a separate shell, so doing cd in one command and than ls in the next might not have the expected result.

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/tool/shell/ExecuteShellCommandTool.kt
Faanbaria: No particular reason. Sixty seconds just felt like a reasonable default. Do you think it should be lower?

agents/agents-ext/src/jvmMain/kotlin/ai/koog/agents/ext/tool/shell/JvmShellCommandExecutor.kt
Faanbaria: When I run the tests on Windows, they fail if I don't make this change.

agents/agents-ext/src/jvmMain/kotlin/ai/koog/agents/ext/tool/shell/JvmShellCommandExecutor.kt
Faanbaria: The issue here is that calling appendLine on empty strings will create extra newlines. For example, when a command fails, this would add a blank line before the error output and the command.

Faanbaria: Here is an example from the test `command fails with error message`.
The result would be:
```
Command: grep nonexistent /nonexistent/file.txt

grep: /nonexistent/file.txt: No such file or directory
Exit code: 2
```

agents/agents-ext/src/jvmMain/kotlin/ai/koog/agents/ext/tool/shell/JvmShellCommandExecutor.kt
Faanbaria: I copied the 1-second timeout from [code-engine](https://github.com/JetBrains/code-engine/blob/ba4cf1a8c8e7040b2d2f1e006639a9f2c6ededc4/code-exec/code-exec-jvm/src/main/kotlin/ai/grazie/code/exec/jvm/JVMSpawnProcessExecutionProvider.kt#L58). Since they have more experience with the agent running this tool and it’s enough to get the output. Do you have any other recommendations?

agents/agents-ext/src/jvmMain/kotlin/ai/koog/agents/ext/tool/shell/JvmShellCommandExecutor.kt
Faanbaria: I changed it to make `timeoutSeconds` mandatory in `ShellCommandExecutor, since it's already is in the tool.

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/tool/shell/ExecuteShellCommandTool.kt
sdubov: So, you mean that executing one command is absolutely isolated from the other? Just curious, what benefits do you get from that? 

I would assume that you also can mention to LLM to combine commands with `;`? So, it can execute them all if it helps somehow. My potential concern here is that I suspect that `cd` should be pretty natural command for LLM. Note: It is just a guess. No facts that I can relay on from my side.

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/tool/shell/ExecuteShellCommandTool.kt
sdubov: It depends on a command. If you execute `find` command it might be longer, depending on the disk space that you specify for a search, or a command like `top` that holds the output by default (can be configured, but who known how LLM will execute it). Other commands should be quite fast.

Let me rephrase this question to: "what does this 60 seconds mean?" :) I'm fine with that number as long as it reflects the purpose of this timeout.

agents/agents-ext/src/jvmMain/kotlin/ai/koog/agents/ext/tool/shell/JvmShellCommandExecutor.kt
sdubov: I just do not completely follow the logic here. From my vision:
1. We close the process after it finish with execution (or after the timeout defined for the `waitFor` call). At that point, streams should be closed as well. 
2. We read the out/err stream content from the buffer, but add a timeout here for some reason. 

What are we protect ourselves from by adding a timeout?  
What will happen if a stream read takes more than a second? We will just loose some output. I'm fine with the timeout if it is what is expected here.

agents/agents-ext/src/jvmMain/kotlin/ai/koog/agents/ext/tool/shell/JvmShellCommandExecutor.kt
sdubov: Oh, wait. You use `StringBuilder`. Why don't you just use `appendLine(...)` instead of `append(...).append('\n')` when add a new line of out/err stream above (line 52, 58)? Should not it fix the problem? I would expect `StreamBuilder` uses the platform specific line endings.

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/tool/shell/ExecuteShellCommandTool.kt
Faanbaria: So it's not about benefits for the LLM, it's about simplicity of the tool. Supporting statefull command execution would be a lot more complex.

We did update the description to make this clearer for the LLM

agents/agents-ext/src/jvmMain/kotlin/ai/koog/agents/ext/tool/shell/JvmShellCommandExecutor.kt
Faanbaria: Yeah maybe we don't need this, we removed it

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/tool/shell/ExecuteShellCommandTool.kt
Faanbaria: Me and Bruno discussed it and we think we don't really need a default here, it might make more sense to have the LLM provide a value every time. What do you think?

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/tool/shell/ExecuteShellCommandTool.kt
sdubov: Ok. Sounds good for me. Thank you for clarifying

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/tool/shell/ExecuteShellCommandTool.kt
sdubov: I'm not sure how it will work with LLM. Maybe a parameter with some default value would be fine. I think it can be tuned after if needed for some cases. It is not a blocker for a commit, for sure. Rather, a thing to think about for future.

agents/agents-ext/src/jvmMain/kotlin/ai/koog/agents/ext/tool/shell/JvmShellCommandExecutor.kt
```diff
@@ -0,0 +1,104 @@
...
+
+        val stdoutJob = launch(Dispatchers.IO) {
+            process.inputStream.bufferedReader().useLines { lines ->
+                lines.forEach { stdoutBuilder.append(it).appendLine() }
+            }
+        }
+
+        val stderrJob = launch(Dispatchers.IO) {
+            process.errorStream.bufferedReader().useLines { lines ->
+                lines.forEach { stderrBuilder.append(it).appendLine() }
```
sdubov: Can we just wrap this: `stderrBuilder.append(it).appendLine()` into `stderrBuilder.appendLine(it)` ?

agents/agents-ext/src/jvmMain/kotlin/ai/koog/agents/ext/tool/shell/JvmShellCommandExecutor.kt
sdubov: Thank you fort the change, but it seems not solve the issue with blocking thread. You still can get a very long timeout here that you will not be able to cancel by the outer events (e.g. coroutine termination/cancellation). 

I would suggest here to get rid of `waitFor()` API and use some custom logic, e.g.:
```
val isCompleted = withTimeoutOrNull(timeoutSeconds * 1000) {
    process.onExit().await()
} != null

if (!isCompleted) {
    logger.warn("Command '$shellCommand' was not able to finish in ${timeoutSeconds}s.")
    process.destroyForcibly()
    ...
}
```

sdubov (changes requested):
There is one thing I would suggest to fix before merge related to blocking `waitFor()` API. Other than that, LGTM!

sdubov (approved):
LGTM!

PR #923, opened 2025-10-03
(KG-442) [prompt] update BedrockLLMClient constructor to accept more generic CredentialsProvider instead of just keys

### LINKED ISSUE #498: Make it possible to use aws config profiles for Bedrock authentication

EugeneTheDev:
Update secondary constructor in `BedrockLLMClient` to accept `CredentialsProvider` interface type instead of just keys. This would give more flexibility to choose convenient auth method without having to use primary constructor and constructing the whole `BedrockRuntimeClient` by hand

Initiated by #501 , fixes #498

### DIFF
--- a/docs/docs/prompt-api.md
+++ b/docs/docs/prompt-api.md
@@ -678,6 +678,7 @@ import ai.koog.prompt.executor.clients.retry.RetryingLLMClient
 import ai.koog.prompt.executor.llms.MultiLLMPromptExecutor
 import ai.koog.prompt.executor.llms.SingleLLMPromptExecutor
 import ai.koog.prompt.llm.LLMProvider
+import aws.sdk.kotlin.runtime.auth.credentials.StaticCredentialsProvider
 
 -->
 ```kotlin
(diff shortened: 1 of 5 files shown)


kpavlov (approved):
Thank you @EugeneTheDev! This change makes it possible to use the Bedrock client on mobile platforms, where environment variables are not present

PR #925, opened 2025-10-04
[prompt] Throw exception on nullable types in OpenAI JSON Schema generators explicitly

EugeneTheDev:
OpenAI structured output doesn't support nullable types. Currently, in this case you get an error from the API about an invalid schema, which isn't helpful because it doesn't explain why the schema is invalid exactly. Update our schema generators to throw descriptive exception explicitly during generation.

### DIFF
--- a/prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openai/structure/OpenAIBasicJsonSchemaGenerator.kt
+++ b/prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openai/structure/OpenAIBasicJsonSchemaGenerator.kt
@@ -20,6 +20,14 @@ public open class OpenAIBasicJsonSchemaGenerator : BasicJsonSchemaGenerator() {
      */
     public companion object Default : OpenAIBasicJsonSchemaGenerator()
 
+    override fun process(context: GenerationContext): JsonObject {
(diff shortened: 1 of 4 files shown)


kpavlov (commented):
The nullable type is represented in the schema like this:
```json
 "unit": {
    "type": ["string", "null"],
    "description": "The unit to return the temperature in",
    "enum": ["F", "C"]
}
```
Is it possible to output it as an array?

https://platform.openai.com/docs/guides/structured-outputs/supported-schemas.#all-fields-must-be-required

EugeneTheDev:
That's what we were doing already for Standard schema generator, and apparently OpenAI still failed. Let me recheck, I didn't notice they claimed nullability support with unions in their docs

EugeneTheDev:
I see, the problem was exactly with arrays - it fails when array type is specified as nullable, like this
```json
"suggestions": {
    "type": [
        "array",
        "null"
    ],
    "items": {
        "type": "string"
    },
    "description": "Suggestions for possible values for this node"
},
```

So we don't need to prohibit nullables completely, just for the arrays, it seems. Or, alternatively, we can try the same trick with `anyOf`. I'll check

EugeneTheDev:
Ok, after a further investigation looks like it was not a problem either. The root cause was a combination of edge cases - it fails when you put your nullable array inside another array. E.g. with a setup like this:

```kotlin
data class Result(val elements: List<Element>)

data class Element(val suggestions: List<String>?)
```

This really sounds like an edge case and more of a weird limitation on the provider (OpenAI) side that we can't do much about. I'm closing this PR

PR #929, opened 2025-10-06
KG-440. Handle Assistant message in subgraphWithTask method

sdubov:
[KG-440](https://youtrack.jetbrains.com/issue/KG-440/Update-the-subgraphWithTask-to-use-giveFeedback-node-with-assistant-response). SubgraphWithTask API supports models with no "tool_choice" capability.

- Add ability to use `subgraphWithTask` API with models that does not have `tool_choice` capability;
- Update mock test framework logic to be able to set conditions on tool call mocks.

## Breaking Changes
No

---

#### Type of the changes
- [ ] Documentation update
- [x] Tests improvement
- [ ] Refactoring

#### Checklist

### DIFF
--- a/agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt
+++ b/agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt
@@ -11,11 +11,13 @@ import ai.koog.agents.core.dsl.builder.AIAgentSubgraphDelegate
 import ai.koog.agents.core.dsl.builder.forwardTo
 import ai.koog.agents.core.dsl.extension.nodeLLMRequest
 import ai.koog.agents.core.dsl.extension.nodeLLMSendToolResult
+import ai.koog.agents.core.dsl.extension.onIsInstance
 import ai.koog.agents.core.dsl.extension.onToolCall
 import ai.koog.agents.core.dsl.extension.setToolChoiceRequired
 import ai.koog.agents.core.environment.ReceivedToolResult
 import ai.koog.agents.core.environment.executeTool
 import ai.koog.agents.core.environment.toSafeResult
+import ai.koog.agents.core.tools.DirectToolCallsEnabler
 import ai.koog.agents.core.tools.Tool
 import ai.koog.agents.core.tools.ToolDescriptor
 import ai.koog.agents.core.tools.annotations.InternalAgentToolsApi
@@ -50,6 +52,20 @@ public object SubgraphWithTaskUtils {
      */
     @InternalAgentToolsApi
     public const val FINALIZE_SUBGRAPH_TOOL_NAME: String = "finalize_task_result"
+
+    /**
+     * A constant string describing the purpose and usage of a tool within the agent framework.
+     *
+     * This constant is intended to represent the action of finalizing a subgraph process and providing
+     * the final result. It is specifically used internally to indicate when a process is considered
+     * complete and the output of that process should be returned.
+     *
+     * Marked with `@InternalAgentToolsApi`, this value is primarily designed for internal use within
+     * the agent tools API and subject to change in future releases. Its usage in external
+     * implementations should be approached with caution.
+     */
+    @InternalAgentToolsApi
+    public const val FINALIZE_SUBGRAPH_TOOL_DESCRIPTION: String = "Call this tool when finish and provide final result"
 
     /**
      * Creates and returns a `Tool` instance with serializers and a descriptor for processing.
@@ -137,16 +153,27 @@ public inline fun <reified Input, reified Output> AIAgentSubgraphBuilderBase<*, 
     toolSelectionStrategy: ToolSelectionStrategy,
     llmModel: LLModel? = null,
     llmParams: LLMParams? = null,
+    assistantResponseRepeatMax: Int? = null,
     noinline defineTask: suspend AIAgentGraphContextBase.(input: Input) -> String
 ): AIAgentSubgraphDelegate<Input, Output> = subgraph(
     toolSelectionStrategy = toolSelectionStrategy,
     llmModel = llmModel,
     llmParams = llmParams,
 ) {
-    val finishToolDescriptor =
-        serializer<Output>().descriptor.asToolDescriptor(toolName = SubgraphWithTaskUtils.FINALIZE_SUBGRAPH_TOOL_NAME)
+    // An identity tool that provides arguments as a tool result without changes.
+    val finishTool = object : Tool<Output, Output>() {
+        override val argsSerializer: KSerializer<Output> = serializer()
+        override val resultSerializer: KSerializer<Output> = serializer()
+        override val name: String = SubgraphWithTaskUtils.FINALIZE_SUBGRAPH_TOOL_NAME
+        override val description: String = SubgraphWithTaskUtils.FINALIZE_SUBGRAPH_TOOL_DESCRIPTION
+        override suspend fun execute(args: Output): Output = args
+    }
 
-    setupSubgraphWithTask<Input, Output, Output>(finishToolDescriptor, defineTask)
+    setupSubgraphWithTask<Input, Output, Output>(
+        finishTool = finishTool,
+        assistantResponseRepeatMax = assistantResponseRepeatMax,
+        defineTask = defineTask
+    )
 }
 
 /**
@@ -193,14 +220,18 @@ public inline fun <reified Input, reified Output, reified OutputTransformed> AIA
     finishTool: Tool<Output, OutputTransformed>,
     llmModel: LLModel? = null,
     llmParams: LLMParams? = null,
+    assistantResponseRepeatMax: Int? = null,
     noinline defineTask: suspend AIAgentGraphContextBase.(input: Input) -> String
 ): AIAgentSubgraphDelegate<Input, OutputTransformed> = subgraph(
     toolSelectionStrategy = toolSelectionStrategy,
     llmModel = llmModel,
     llmParams = llmParams,
 ) {
-    val finishToolDescriptor = finishTool.descriptor
-    setupSubgraphWithTask<Input, Output, OutputTransformed>(finishToolDescriptor, defineTask)
+    setupSubgraphWithTask<Input, Output, OutputTransformed>(
+        finishTool = finishTool,
+        assistantResponseRepeatMax = assistantResponseRepeatMax,
+        defineTask = defineTask
+    )
 }
 
 /**
@@ -223,14 +254,18 @@ public inline fun <reified Input, reified Output, reified OutputTransformed> AIA
     finishTool: Tool<Output, OutputTransformed>,
     llmModel: LLModel? = null,
     llmParams: LLMParams? = null,
+    assistantResponseRepeatMax: Int? = null,
     noinline defineTask: suspend AIAgentGraphContextBase.(input: Input) -> String
 ): AIAgentSubgraphDelegate<Input, OutputTransformed> = subgraph(
     toolSelectionStrategy = ToolSelectionStrategy.Tools(tools.map { it.descriptor }),
     llmModel = llmModel,
     llmParams = llmParams,
 ) {
-    val finishToolDescriptor = finishTool.descriptor
-    setupSubgraphWithTask<Input, Output, OutputTransformed>(finishToolDescriptor, defineTask)
+    setupSubgraphWithTask<Input, Output, OutputTransformed>(
+        finishTool = finishTool,
+        assistantResponseRepeatMax = assistantResponseRepeatMax,
+        defineTask = defineTask
+    )
 }
 
 /**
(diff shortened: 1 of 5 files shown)


agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt
```diff
@@ -310,33 +316,37 @@ public inline fun <reified Input, reified Output, reified OutputTransformed> AIA
     finishToolDescriptor: ToolDescriptor,
     noinline defineTask: suspend AIAgentGraphContextBase.(Input) -> String
 ) {
+    val askAssistantToFinishMaximum = SubgraphWithTaskUtils.ASSISTANT_RESPONSE_REPEAT_MAX
+    var askAssistantToFinishCounter = 1
```
antoniibelyshev: It's better to put it in the storage during the setup and update there. Otherwise it could cause concurrency issues: for example, when using parallel node execution.

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt
```diff
@@ -364,24 +376,60 @@ public inline fun <reified Input, reified Output, reified OutputTransformed> AIA
...
             environment.executeTool(toolCall)
         }
     }
+
     val sendToolResult by nodeLLMSendToolResult()
 
+    @OptIn(DetachedPromptExecutorAPI::class)
+    val handleAssistantMessage by node<Message.Assistant, Message.Response> { response ->
+        if (llm.model.capabilities.contains(LLMCapability.ToolChoice)) {
+            error(
```
antoniibelyshev, line 435: Let's not not throw an error, and just give a warning in the logs. I would say, there could be a scenario when a model with `LLMCapability.ToolChoice` does not generate a tool. We should give it a chance to recover.

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt
```diff
@@ -364,24 +376,60 @@ public inline fun <reified Input, reified Output, reified OutputTransformed> AIA
...
+                    "and was not able to call `${finishToolDescriptor.name}` tool after <$askAssistantToFinishMaximum> attempts."
+            )
+        }
+
+        llm.writeSession {
+            // append a new message to the history with feedback:
+            updatePrompt {
+                user {
+                    markdown {
+                        h1("DO NOT CHAT WITH ME DIRECTLY! CALL TOOLS, INSTEAD.")
```
antoniibelyshev, line 457: Can we allow to customize the feedback, make it a parameter with default value? The models which are weak at tool calling require more accurate prompting, which could also be model-specific, so we can let the users tweak this by themselves. WDYT?

antoniibelyshev (approved):
Thank you, looks good! Just pointed out a few minor issues

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt
sdubov: Thank you for pointing this out. It was the original logic. My expectation here was: if a model has `tool_choice` capability and we ask it to call tools, there is no way it return anything else, but a tool call. That's why we throw here. 

I've discussed this originally with @tiginamaria. @antoniibelyshev, could you please provide some examples or confirm that a model can return an Assistant message back instead?

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt
sdubov: @antoniibelyshev, what other feedback prompts do you have in mind? Maybe we can generalise it to keep it up to the library? If it is not the case, I can add a parameter as well.

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt
sdubov: @antoniibelyshev, could you please explain your concern here a bit? When running nodes in parallel, I would like the code to be isolated in each thread. E.g. when I run two `subgraphWithTasks` in parallel, I would like them to handle assistant messages independently as I can run them on different models, for example.

antoniibelyshev: As I understand it, one could create an instance of the `subgraphWithTask` node, and then pass it twice to the parallel node:

```
val task by nodeDoNothing<Unit>()
val parallelNode by parallel(task, task) { ... }
```

then both branches will access the same instance of `askAssistantToFinishCounter`.

As far as I remember, this concern was initially raised by @EugeneTheDev when I worked on the retry subgraph. Probably, he could help here.

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt
antoniibelyshev: Generally, I also would expect the model with tool_choice to always produce tools when asked. However, I am not totally sure it is the case for all providers. Anyway, if we got to this point, the response was not a tool call, meaning that the model does not produce only tool calls

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt
antoniibelyshev: I do not have specific prompts in mind

skarpovdev (approved):

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt
```diff
@@ -310,33 +316,37 @@ public inline fun <reified Input, reified Output, reified OutputTransformed> AIA
     finishToolDescriptor: ToolDescriptor,
     noinline defineTask: suspend AIAgentGraphContextBase.(Input) -> String
 ) {
+    val askAssistantToFinishMaximum = SubgraphWithTaskUtils.ASSISTANT_RESPONSE_REPEAT_MAX
```
EugeneTheDev: Can this be moved to function parameters with default value, to make it configurable?

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt
EugeneTheDev: Yes, this is an easy to make mistake, you can't use variables inside your strategy just like this. The reason is that strategy instance can be shared between different agents, since this is a sort of a blueprint, and this would make this counter state shared too. If you want to store some state during strategy execution, you have to use `storage` map, since it is tied to the execution context, not to the strategy itself. 

But it won't protect from concurrency issues, since this is not the concern of this `storage` component. It sounds like you just need to avoid passing the same node several times if you want to do a parallel run, just create two copies.

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt
EugeneTheDev: +1 here and in general for making all prompts customizable with default value (you can check e.g. how it is done for `JsonStructuredData`)

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt
EugeneTheDev: Can we handle the case when the model doesn't have tool calling capability early on, doing this validation as a first step in this function? If we know that model would never produce tool calls, we can avoid spending tokens on requesting it to provide assistant message, if we are going to throw exception later on anyways, without using this assistant message in any meaningful way.

EugeneTheDev (changes requested):

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt
antoniibelyshev: If the model does not have ToolChoice capability, it still might call tools, so it makes sense to correct it only when it didn't call a tool.

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt
sdubov: @EugeneTheDev, sure we can move it to parameters. However, my point here was that it is the internal implementation that I would explicitly want to hide from a user. From my perspective, it is not something that a user should configure. It is a parameter that we need to define internally to make sure our `subgraphWithTask` works well.

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt
sdubov: @EugeneTheDev, that was the whole point for this change. `subgraphWithTask` API should support any model that a user would like to use.

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt
```diff
@@ -364,24 +376,60 @@ public inline fun <reified Input, reified Output, reified OutputTransformed> AIA
...
+            }
+
+            requestLLM()
+        }
+    }
+
     nodeStart then setupTask then nodeCallLLM then nodeDecide
 
     edge(nodeDecide forwardTo callToolHacked onToolCall { true })
+    edge(nodeDecide forwardTo handleAssistantMessage onCondition { it is Message.Assistant } transformed { it as Message.Assistant })
```
EugeneTheDev: nit: You can combine condition and transformation using `onIsInstance`

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt
sdubov: Discussed this in person. Agreed with @EugeneTheDev  to keep it internal and update it later if we need to customise it from any perspective.

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt
sdubov: Fixed.

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt
sdubov: Fixed. Please see the last commit with the fix.

sdubov:
@EugeneTheDev, fixed all comments based on our coding session today. Could you please check the last commit in the PR?

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt
```diff
@@ -301,42 +341,46 @@
...
- * @param finishToolDescriptor A descriptor for the tool that determines the condition to finalize the subgraph's operation.
+ * @param finishTool A descriptor for the tool that determines the condition to finalize the subgraph's operation.
  * @param defineTask A suspending lambda that defines the main task of the subgraph, producing a task description based on the input.
  */
-@InternalAgentToolsApi
 @OptIn(InternalAgentToolsApi::class)
 public inline fun <reified Input, reified Output, reified OutputTransformed> AIAgentSubgraphBuilderBase<Input, OutputTransformed>.setupSubgraphWithTask(
-    finishToolDescriptor: ToolDescriptor,
+    finishTool: Tool<Output, OutputTransformed>,
+    assistantResponseRepeatMax: Int? = null,
```
EugeneTheDev, line 352: nit: Why do we make it nullable and use constants as a fallback (ASSISTANT_RESPONSE_REPEAT_MAX). Why can we just get rid of the constant and nullable parameter and make it just a default value - `assistantResponseRepeatMax: Int = 3`? Should look more idiomatic

EugeneTheDev (approved):
LGTM, thanks, minor comment only

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt
sdubov: Fixed. Thanks for there nit!

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt
sdubov: I prefer to have one point when I resolve the default parameters. If you have multiple invocations where you need to pass an optional parameters, then you need to specify the default parameter in every method:
```
fun a(foo: Int = 3) = b(foo)
fun b(foo: Int = 3) { ... }
```
I think it is easier and less error prone when you have a logic that resolve the default parameters in one place (root function).

EugeneTheDev: Good point, then you can still have a const that you just assign as a default parameter everywhere - no need to have nullable parameters if the actual value is always not null (number of iterations)

EugeneTheDev: AFAIK this is a common practice, e.g.`kotlinx.coroutines` do that too. You can take a look at how e.g. `Channel` (or other similar functions) are defined

PR #936, opened 2025-10-07
Added support for filters in PersistenceProvider

Rizzen:
## Breaking Changes

---

#### Type of the changes
- [ ] Documentation update
- [ ] Tests improvement
- [ ] Refactoring

#### Checklist

### DIFF
--- a/agents/agents-features/agents-features-snapshot/src/commonMain/kotlin/ai/koog/agents/snapshot/providers/PersistencyStorageProvider.kt
+++ b/agents/agents-features/agents-features-snapshot/src/commonMain/kotlin/ai/koog/agents/snapshot/providers/PersistencyStorageProvider.kt
@@ -11,10 +11,10 @@ import ai.koog.agents.snapshot.feature.AgentCheckpointData
         "ai.koog.agents.snapshot.feature.PersistenceStorageProvider"
     )
 )
-public typealias PersistencyStorageProvider = PersistenceStorageProvider
+public typealias PersistencyStorageProvider<Filter> = PersistenceStorageProvider<Filter>
 
-public interface PersistenceStorageProvider {
-    public suspend fun getCheckpoints(agentId: String): List<AgentCheckpointData>
+public interface PersistenceStorageProvider<Filter> {
+    public suspend fun getCheckpoints(agentId: String, filter: Filter? = null): List<AgentCheckpointData>
     public suspend fun saveCheckpoint(agentId: String, agentCheckpointData: AgentCheckpointData)
-    public suspend fun getLatestCheckpoint(agentId: String): AgentCheckpointData?
+    public suspend fun getLatestCheckpoint(agentId: String, filter: Filter? = null): AgentCheckpointData?
 }
--- a/agents/agents-features/agents-features-snapshot/src/commonMain/kotlin/ai/koog/agents/snapshot/providers/filters/AgentCheckpointFilter.kt
+++ b/agents/agents-features/agents-features-snapshot/src/commonMain/kotlin/ai/koog/agents/snapshot/providers/filters/AgentCheckpointFilter.kt
@@ -3,12 +3,12 @@ package ai.koog.agents.snapshot.providers.filters
 import ai.koog.agents.snapshot.feature.AgentCheckpointData
 
 /**
- * An interface that extends [AgentCheckpointFilter] to provide a mechanism for evaluating
- * whether a specific agent checkpoint meets the predicate-defined conditions.
+ * Interface for defining predicates to filter agent checkpoint data.
  *
- * This filter is used to determine, based on certain criteria, whether an agent's checkpoint
- * should be acted upon, retained, or utilized in a specific context. The evaluation is
- * performed using the [check] function, which implements the predicate logic.
+ * Implementations of this interface are utilized in storage providers to determine
+ * whether specific agent checkpoints meet given conditions. This allows for selective
+ * retrieval and processing of checkpoints based on custom logic, such as filtering by
+ * node, time, or specific checkpoint properties.
  */
 public interface AgentCheckpointPredicateFilter {
     /**
--- a/agents/agents-features/agents-features-snapshot/src/commonMain/kotlin/ai/koog/agents/snapshot/providers/filters/AgentCheckpointFilter.kt
+++ b/agents/agents-features/agents-features-snapshot/src/commonMain/kotlin/ai/koog/agents/snapshot/providers/filters/AgentCheckpointFilter.kt
@@ -1,0 +1,27 @@
+package ai.koog.agents.snapshot.providers.filters
+
+import ai.koog.agents.snapshot.feature.AgentCheckpointData
+
+/**
+ * Interface for filtering agent checkpoints.
+ */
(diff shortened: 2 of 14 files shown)


agents/agents-features/agents-features-sql/src/jvmMain/kotlin/ai/koog/agents/features/sql/providers/CheckpointsTable.kt
```diff
@@ -0,0 +1,70 @@
...
+     *
+     * This column stores serialized checkpoint data as a JSON string. It is a required field,
+     * ensuring the persistence of critical state information for agent checkpoints.
+     */
+    public val checkpointJson: Column<String> = text("checkpoint_json")
+
+    /**
+     * Represents the TTL (Time-To-Live) timestamp for a checkpoint.
+     *
+     * This column stores an optional expiration timestamp, in milliseconds since the epoch,
```
kpavlov, line 55: qq: could it be Instant?

agents/agents-features/agents-features-sql/src/jvmMain/kotlin/ai/koog/agents/features/sql/providers/ExposedPersistencyStorageProvider.kt
```diff
@@ -167,23 +140,29 @@ public abstract class ExposedPersistenceStorageProvider(
 
         transaction {
```
kpavlov, line 142: nit: It's ok to be eventually consistent. Maybe it is possible to relax the transaction requirement?

agents/agents-features/agents-features-sql/src/jvmMain/kotlin/ai/koog/agents/features/sql/providers/ExposedPersistencyStorageProvider.kt
```diff
@@ -167,23 +140,29 @@ public abstract class ExposedPersistenceStorageProvider(
...
-        return transaction {
-            checkpointsTable
-                .select(checkpointsTable.checkpointJson)
-                .where {
+    override suspend fun getCheckpoints(agentId: String, filter: ExposedPersistenceFilter?): List<AgentCheckpointData> {
+        if (filter == null) {
+            return transaction {
+                checkpointsTable.select(checkpointsTable.checkpointJson).where {
                     checkpointsTable.persistenceId eq agentId
+                }.orderBy(checkpointsTable.createdAt to SortOrder.ASC).mapNotNull { row ->
```
kpavlov, line 159: plus `AND ttl_timestamp >= now()`
Is there any index by `agentId`, `createdAt` and `ttl_timestamp` defined?

kpavlov (approved):
Thank you, @Rizzen
Comments might be addressed separately

agents/agents-features/agents-features-snapshot/src/commonMain/kotlin/ai/koog/agents/snapshot/providers/filters/AgentCheckpointFilter.kt
```diff
@@ -0,0 +1,27 @@
+package ai.koog.agents.snapshot.providers.filters
+
+import ai.koog.agents.snapshot.feature.AgentCheckpointData
+
+/**
+ * Interface for filtering agent checkpoints.
+ */
+public interface AgentCheckpointFilter
```
EugeneTheDev: Why do we need two separate interfaces here? Why can't `AgentCheckpointFilter` have `check` method? What was the idea behind marker interface?

agents/agents-features/agents-features-sql/src/jvmMain/kotlin/ai/koog/agents/features/sql/providers/ExposedPersistencyStorageProvider.kt
Rizzen: What specific relaxations are you suggesting? If we remove the explicit transaction wrapper, we would effectively be relying on the database’s own delete operation handling behavior.

agents/agents-features/agents-features-snapshot/src/commonMain/kotlin/ai/koog/agents/snapshot/providers/filters/AgentCheckpointFilter.kt
Rizzen: 1. The original purpose of the marker interface was to identify entities that should be treated as Filters. However, since the base Provider interface does not define generic type boundaries, this may no longer be necessary. I’ll revisit this decision.

2. The AgentCheckpointFilter cannot include a check method because a sufficiently generic implementation is not feasible. For instance, when using a Postgres provider, it would be inefficient to retrieve all checkpoints and filter them in memory. Instead, it’s preferable to enable the construction of an appropriate query. The same principle applies to other persistence providers.

agents/agents-features/agents-features-snapshot/src/commonMain/kotlin/ai/koog/agents/snapshot/providers/PersistencyStorageProvider.kt
```diff
@@ -11,10 +11,10 @@ import ai.koog.agents.snapshot.feature.AgentCheckpointData
         "ai.koog.agents.snapshot.feature.PersistenceStorageProvider"
     )
 )
-public typealias PersistencyStorageProvider = PersistenceStorageProvider
+public typealias PersistencyStorageProvider<Filter> = PersistenceStorageProvider<Filter>
 
-public interface PersistenceStorageProvider {
-    public suspend fun getCheckpoints(agentId: String): List<AgentCheckpointData>
+public interface PersistenceStorageProvider<Filter> {
```
EugeneTheDev, line 16: Sorry, I fail to understand how this generic `Filter` can be used when you only know the most basic type, this `PersistenceStorageProvider` interface, and don't know the concrete implementation. If I only have an instance of this type, how can I pass a filter here, if It's essentially `Any?`? I think we need to redesign this part, to provide some generic `Filter` interface that wouldn't be tied to a particular in-memory filter operation, so it could be used in-memory, with a database, etc.

EugeneTheDev (changes requested):

agents/agents-features/agents-features-snapshot/src/commonMain/kotlin/ai/koog/agents/snapshot/providers/PersistencyStorageProvider.kt
EugeneTheDev: Discussed this, it's more clear to me now, it's ok to leave generic parameter here.

EugeneTheDev (approved):

agents/agents-features/agents-features-sql/src/jvmMain/kotlin/ai/koog/agents/features/sql/providers/CheckpointsTable.kt
Rizzen: Yes, it could, but I have another PR with schema change on the way - so I'll change it to Time together with these upcoming changes

PR #943, opened 2025-10-09
Add support for `Message.Reasoning` interactions across clients

devcrocod:
[KG-432](https://youtrack.jetbrains.com/issue/KG-432)
[KG-297](https://youtrack.jetbrains.com/issue/KG-297)

## Motivation and Context
A new message type has been added: `Reasoning`

## Breaking Changes
Yes,
The implementation of Anthropic messages has been modified for future support and schema compliance

---

#### Type of the changes
- [ ] Documentation update
- [ ] Tests improvement
- [x] Refactoring

#### Checklist


prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/message/Message.kt
```diff
@@ -138,6 +139,30 @@ public sealed interface Message {
...
+     * @property content The content of the message as a string.
+     * @property role The [Role] of the message, indicating its source or function in the chat (e.g., assistant, user).
+     *                Defaults to [Role.Assistant].
+     * @property metaInfo Metadata associated with the reasoning, captured as a [ResponseMetaInfo] instance.
+     *                    Provides details such as token usage and timestamps.
+     */
+    @Serializable
+    public data class Reasoning(
+        @Transient
+        public val original: Thinking? = null,
```
EugeneTheDev: I have some questions regarding the chosen design. Why do we have a reference to the "original" message from the API response here, that is not serialized? This would lead to lost information when we persist and then unpersist message history. On top of that, I don't see "thought signature" parameter here, that OpenAI, Anthropic and Google use to encode internal model thinking state and to allow the model to continue reasoning across mutliple interactions. I suppose the idea was to put it in the `Thinking?` reference, but then again, we would lose this information when persisting the state.


Instead, how about we implement a simpler solution? Since all major providers have this "signature" thing that is used to encode internal thinking state, we can remove this `Thikning?` interface and replace it with optional `signature: String? = null`, property. The `content` of the message would be the actual message from the model, if it provides textual representation of its thought process, and empty otherwise (e.g. Google models do not provide textual thoughts unless explicitly asked, or not at all, AFAIK). This would make the API simpler and persistence-friendly, so we can continue reasoning not across multiple interactions, but even across multiple agent launches and app terminations.

EugeneTheDev (changes requested):

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/message/Message.kt
devcrocod: Yes, you’re right. We’ll lose that information during serialization. The choice of this approach was originally made for simplicity. Messages from different providers vary too much. The specified signature from Anthropic doesn’t exist in others, and the documentation doesn’t clearly explain what it represents. Google and OpenAI also have encrypted, which significantly reduces the number of tokens when sending responses back. OpenAI additionally has summary. We could move all such properties into a dedicated class, but they would be rarely used by other providers. Alternatively, we could design a dynamic serialization mechanism for thinking and provide a dedicated serialization module

EugeneTheDev: Wait, but AFAIK after checking Anthropic, OpenAI and Google API docs, they all have thought signatures, and some of them have optional textual thought. That's enough of a similarity, right? Looks like we can introduce this common signature field, since all major providers support it. Hence my comment. If they are already so similar, why not make it simpler. I suspect the math behind this "thought process" is similar too, that's why the APIs look similar

EugeneTheDev: And we don't actually _care_, what is the format of this signature, right? Because it will be different across providers, but we are never meant to decode it anyway. It exists solely to keep the thinking process going across interactions, so we are just passing it around, right?

I see where the problem might be, though. Since we can pass the prompt containing reasoning messages from one model to the other reasoning model, and it will fail because the signature format is different. In this case, can we turn signature into an object, that would contain the signature itself and LLModel identifier that produced this signature. And then we can fail in clients if the signature model is different from the model we're trying to call now. Wdyt?

sdubov (approved):
Looks good for me. Open Telemetry does not have any special handling for such cases. The rest looks good from my side as well.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/bedrock/modelfamilies/anthropic/BedrockAnthropicClaudeSerialization.kt
```diff
@@ -55,19 +53,32 @@ internal object BedrockAnthropicClaudeSerialization {
...
                     }
                 }
 
+                is Message.Reasoning -> {
+                    if (msg.content.isNotEmpty()) {
+                        messages.add(
+                            BedrockAnthropicInvokeModelMessage.Assistant(
+                                content = listOf(
+                                    BedrockAnthropicInvokeModelContent.Thinking(
+                                        signature = msg.encrypted!!,
```
EugeneTheDev: nit: It might be a bit better to throw a more descriptive exception explicitly in this case, instead of using double-bang that would produce generic NPE without additional context.

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/message/Message.kt
```diff
@@ -138,6 +138,29 @@ public sealed interface Message {
...
+     * @property role The [Role] of the message, indicating its source or function in the chat (e.g., assistant, user).
+     *                Defaults to [Role.Assistant].
+     * @property metaInfo Metadata associated with the reasoning, captured as a [ResponseMetaInfo] instance.
+     *                    Provides details such as token usage and timestamps.
+     */
+    @Serializable
+    public data class Reasoning(
+        public val id: String? = null,
+        public val encrypted: String? = null,
+        override val content: String,
```
EugeneTheDev: Please rebase on `develop`. This part was recently updated, now we have `parts: List<ContentPart>` instead. And `content` is a read-only property now, that returns concatenated textual parts. So you would likely need to update to `override val parts: List<ContentPart.Text>` here

EugeneTheDev (approved):
Looks good now, thank you. A couple of small comments

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/message/Message.kt
```diff
@@ -138,6 +138,29 @@ public sealed interface Message {
...
+        override val role: Role = Role.Assistant,
+        override val metaInfo: ResponseMetaInfo
+    ) : Message, Response {
```
sdubov: `Response` is a subclass of `Message` class. You can drop the `Message` here.

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/message/Message.kt
```diff
@@ -155,6 +160,37 @@ public sealed interface Message {
...
+     * @property role The [Role] of the message, indicating its source or function in the chat (e.g., assistant, user).
+     *                Defaults to [Role.Assistant].
+     * @property metaInfo Metadata associated with the reasoning, captured as a [ResponseMetaInfo] instance.
+     *                    Provides details such as token usage and timestamps.
+     */
+    @Serializable
+    public data class Reasoning(
+        public val id: String? = null,
+        public val encrypted: String? = null,
+        override val parts: List<ContentPart>,
```
EugeneTheDev: Let's limit it to the `List<ContentPart.Text>` since only text content is supported here

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClient.kt
```diff
@@ -589,11 +605,21 @@ public open class GoogleLLMClient(
         val parts = candidate.content?.parts.orEmpty()
         val responses = parts.map { part ->
             when (part) {
-                is GooglePart.Text -> Message.Assistant(
-                    content = part.text,
-                    finishReason = candidate.finishReason,
-                    metaInfo = metaInfo
-                )
+                is GooglePart.Text -> {
```
denis-domanskii: @devcrocod do we support thoughtSignature only for text responses? 

Gemini API states that:
> Signatures are returned from the model within other parts in the response, for example function calling or text parts. Return the entire response with all parts back to the model in subsequent turns.
https://ai.google.dev/gemini-api/docs/thinking#signatures

And thoughtSignature is a part of "Part" entity: https://ai.google.dev/api/caching#Part 

But maybe I misinterpreted your code.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClient.kt
```diff
@@ -327,6 +327,22 @@ public open class GoogleLLMClient(
                     )
                 }
 
+                is Message.Reasoning -> {
+                    flushCalls()
+                    contents.add(
+                        GoogleContent(
+                            role = "assistant",
```
denis-domanskii, line 334: If I understand right, here we add thoughts and signature as a separate message. It seems that Google asking to be careful with the thoughts and signatures:

> Other usage limitations to consider with function calling include:
> - Signatures are returned from the model within other parts in the response, for example function calling or text parts. Return the entire response with all parts back to the model in subsequent turns.
> - Don't concatenate parts with signatures together.
> - Don't merge one part with a signature with another part without a signature. 
> https://ai.google.dev/gemini-api/docs/thinking#signatures

So maybe it would be safer to reproduce the original content with all parts with their specific thoughts and signatures?

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClient.kt
devcrocod: That’s right
I added `thoughtSignature` to the interface for all Parts

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClient.kt
devcrocod: Unfortunately, google api is strange, so it’s not easy to abstract this into a shared messages.

But I think we can just store the original messages and keep a reference to them
@EugeneTheDev wdyt?

EugeneTheDev: This is a very good question. It raises a concern regarding the current design around messages. This got me thinking. 

Previously, when we didn't have `ContentPart` parts in messages, it was clear why we needed multiple separate messages - each "part" in the API response was treated as a separate response message. But now that we actually have proper parts, messages look a bit strange - why do we treat parts from the API as standalone messages? So it seems it would be more logical to have a single `Message.Assistant` / `Message.User` now with a list of parts - tool calls, tool responses, textual responses, images, thoughts, etc. It would align more with actual API schemas. And then it would've been easier to add proper support for this Google-specific schema by adding optional `thought` parameter to all parts. 

This would be a significant breaking change, unfortunately, because it would modify a very core and very popular part of our API. Unfortunately, we likely would need something like this, since it's already getting somewhat confusing. 

As a workaround right now, we can probably add this thought object (`Thought(thought, thoughtSignature)`) to all response messages, to avoid having unserializable references. It then would make sense to reuse the same `Thought` object in the `Message.Reasoning` btw too.

devcrocod: In my opinion, we should introduce another level of abstraction `Content`, which would encapsulate all of this.
`ContentPart` essentially represents a type of content: `text`, `image`, or `document`.

Note that this situation only applies to Google. For example, `thoughtSignature` includes inside `FunctionCall`, which is not the case for other providers

Also, merging them would break the logic used by OpenAI and Anthropic, where these parts must remain as separate items. By combining them, we risk losing fields like content, summary, and id from the reasoning messages

PR #948, opened 2025-10-09
Revise Getting started, reorganize Overview

innateteniuk:
- Reorganize the **Overview** page and transform it into a section:
    - Introduce cards that highlight important topics
    - Move **Key features** and **LLM providers** to separate pages within the Overview section
    - Rename Key concepts to Glossary and move it to the Overview section
    - Move **Installation** and **Quickstart example** to a new **Getting started** guide
- Transform the **Getting started** section into a single-page guide
- Add a new **Agent types** section with guides for single-run agents, functional agents, and complex workflow agents

## Motivation and Context

To simplify navigation, highlight important topics, and improve consistency.

## Breaking Changes

---

#### Type of the changes
- [x] Documentation update
- [ ] Tests improvement
- [ ] Refactoring

#### Checklist


docs/mkdocs.yml
```diff
@@ -93,14 +100,15 @@ plugins:
...
+        'key-concepts.md': 'glossary.md'
   - llmstxt:
       markdown_description: Koog is a Kotlin-based framework designed to build and run AI agents entirely in idiomatic Kotlin.
       full_output: llms-full.txt
       # List of sections and pages with descriptions, to be included in llms.txt
       sections:
         Documentation:
           - index.md: This page provides a general overview of the Koog framework, its key features and setup.
-          - key-concepts.md: This page provides explanations of key terms and concepts related to Koog and agentic development.
+          - glossary.md: This page provides explanations of key terms and concepts related to Koog and agentic development.
```
marinkovicmarko, line 118: Maybe also add new entries for `key-features.md` and `llm-providers.md`?

marinkovicmarko (approved):

docs/docs/index.md
```diff
@@ -3,120 +3,147 @@
...
-- OpenAI
-- Anthropic
-- DeepSeek
-- OpenRouter
-- Ollama
-- Bedrock
+</div>
 
-For detailed guidance on using these providers with dedicated LLM clients, refer to [Runnning prompts with LLM clients](prompt-api.md#running-prompts-with-llm-clients).
+## Agents
```
azhebel: Agent types?

docs/docs/index.md
```diff
@@ -3,120 +3,147 @@
...
+</div>
 
-2. Make sure that you have `mavenCentral()` in the list of repositories.
+## Core functionality
 
-### Maven
+<div class="grid cards" markdown>
 
-1. Add dependencies to the `pom.xml` file:
+-   :material-chat-processing-outline:{ .lg .middle } [Prompt API](prompt-api.md)
```
azhebel: Align all names on cards maybe?
- Prompts
- Tools
- Strategies
- Events

docs/docs/index.md
```diff
@@ -3,120 +3,147 @@
...
+
+    Generate responses in structured formats
+
+-   :material-waves:{ .lg .middle } [Streaming API](streaming-api.md)
+
+    ---
+
+    Process responses in real-time with streaming support and parallel tool calls
+
+-   :material-puzzle-outline:{ .lg .middle } [Features](features-overview.md)
```
azhebel: I know it's a term, but it looks weird that a section about features is mentioned as an "advanced feature") Not sure what to do with it though...

docs/docs/key-features.md
```diff
@@ -0,0 +1,18 @@
+# Key features
```
azhebel, line 1: Same here. There are custom agent features to extend an agent's functionality, and then these generic "features" of the framework...

docs/docs/llm-providers.md
```diff
@@ -0,0 +1,19 @@
+# LLM providers
+
+Koog integrates with major LLM providers and supports local models through integration with [Ollama](https://ollama.com/).
+The full list of supported providers, platforms, and their LLM client implementations is provided below.
```
azhebel: The table contains providers and corresponding clients, but the phrase mentions "platforms". Is something missing? And I am not sure there is a need for a lengthy intro sentence before the table that does not add any info.

docs/docs/getting-started.md
```diff
@@ -0,0 +1,119 @@
...
+
+This guide will help you install Koog and create your first AI agent.
+
+## Prerequisites
+
+Before you start, make sure that you have the following:
+
+- A working Kotlin/JVM project with Gradle or Maven.
+- Java 17+ installed.
+- A valid [OpenAI](https://platform.openai.com/api-keys) API key.
```
azhebel: Perhaps it makes sense to add a note that the example here assumes using OpenAI, but really it can be adopted to any of the supported LLM providers (with a link to the corresponding page)?

docs/docs/getting-started.md
```diff
@@ -0,0 +1,119 @@
...
+
+<!--- INCLUDE
+import ai.koog.agents.core.agent.AIAgent
+import ai.koog.prompt.executor.clients.openai.OpenAIModels
+import ai.koog.prompt.executor.llms.all.simpleOpenAIExecutor
+import kotlinx.coroutines.runBlocking
+-->
+```kotlin
+fun main() = runBlocking {
+    // Get an API key from the OPENAI_API_KEY environment variable
```
azhebel: Maybe add a step before this set the OpenAI API key as the value of `OPENAI_API_KEY` environment variable? Or another LLM provider key.

Even better, I like how you used tabs for Gradle(Kotlin/Groovy)/Maven. We can do similar tabs here as well, but with different agent examples for different LLM providers. And then the prerequisites will not mention OpenAI key specifically but just that some LLM provider API key is necessary. And there is no key for ollama at all.

This also lead me to an idea that on the supported providers page, the table can also contain links to where each one lets you get the key, similar to how you provided a link to [OpenAI ](https://platform.openai.com/api-keys)

azhebel:
@innateteniuk you were also planning to add some generic page about different agent types, right? This will be in a separate task?

docs/docs/index.md
innateteniuk:
```suggestion
## Agent types
```

docs/docs/index.md
innateteniuk:
```suggestion
-   :material-chat-processing-outline:{ .lg .middle } [Prompts](prompt-api.md)
```

docs/docs/index.md
innateteniuk: Let’s rename "Advanced features" to "Advanced usage".

docs/docs/key-features.md
innateteniuk: Moreover, some features (such as Agent persistence) are highlighted separately. 
And we also treat OpenTelemetry support as an integration rather than a feature.

I would suggest moving the pages about agent features (Memory, Tracing, OpenTelemetry, etc.) out of the "Features" section and removing the "Features" section from the TOC.

docs/docs/index.md
innateteniuk: The Features card can look like this:

```
-   :material-puzzle-outline:{ .lg .middle } [**Other agent capabilities**](features-overview.md)

    ---

    Customize your agent capabilities using [agent memory](agent-memory.md), [tracing](tracing.md), and other features
```

docs/docs/llm-providers.md
innateteniuk: > The table contains providers and corresponding clients, but the phrase mentions "platforms". Is something missing?

That's because Amazon Bedrock is a platform. But I'm not sure if we need to be that precise here. So maybe let's stick to "LLM providers" and remove "platforms".

innateteniuk:
> @innateteniuk you were also planning to add some generic page about different agent types, right? This will be in a separate task?

@azhebel Yes, I plan to do it as a separate task.

docs/docs/key-features.md
azhebel: The problem is they are still referred to as "features" in the code, like the `AIAgentFeature` interface and its implementations. And the `installFeatures` lambda. And the whole `agents-features` module. So we won't be able to not mention them and describe them as features. Ideally, we should even describe how users can extend the API to create their own features. IMO it is better to rename this specific list of "Key features" to... what?) capabilities?

azhebel (approved):

docs/docs/key-features.md
innateteniuk: Yes, I agree that my previous suggestion doesn’t fully address the issue.

Regarding renaming, I tried to find other options, these were:

* Key capabilities
* Key highlights
* Highlights

To me, “Key capabilities” seems like the best replacement.

How about this plan:

1. Rename "Key features" to "Key capabilities".
2. Move the pages about agent features (Memory, Tracing, OpenTelemetry, etc.) out of the "Features" section.
3. Remove the "Features" section from the TOC for now.
4. Keep in mind (and create an issue) that we need to document the concept of “feature” and explain how users can extend the API to create their own features.

WDYT?

azhebel: I agree that we need to separately document the concept of features, but why not leave features (Memory, Tracing, OpenTelemetry, etc.) in the Features section in this case? If we rename the "Key features" page then there shouldn't be a conflict.

aozherelyeva (approved):
✅ Verified the samples:
- OpenAI
- Anthropic
- Google
- DeepSeek
- OpenRouter
- Bedrock
- Ollama

PR #949, opened 2025-10-10
[agents] (KG-459, KG-445) Refactor agent features: move the scope from context to agent pipeline (agent) and simplify agent pipeline

EugeneTheDev:
Refactor agent features to properly scope them to the agent pipeline (agent), make the logic more straightforward and remove unneeded entities:
- `AIAgentPipeline` is now a central point of feature interactions, including feature storage and retrieval.
- Features are installed on the agent pipeline (agent) level, not on the context level. That means they are instantiated once per agent, not once per each context copy.
- `install` methods in features now must return an instance of a feature implementation.
- `interceptContextAgentFeature` was removed since it's not needed anymore - features are installed in agent (pipeline), not in the specific context.
- `InterceptContext` was removed, it is not actually needed.

Additonally, related refactorings were performed:
- Make agent type generic in `AIAgentService` for greater flexibility to better communicate the actual agent type each service implementation is creating
- Move pipeline classes to a dedicated `pipeline` package for better structure.
- Move AIAgentFunctionalContextExt.kt to respecitve package in `dsl`
- Remove `UnexpectedAIAgentMessageException` and `UnexpectedDoubleInitializationException`.

Fixes [KG-459](https://youtrack.jetbrains.com/issue/KG-459). Also fixes [KG-445](https://youtrack.jetbrains.com/issue/KG-445), making agent memory work properly.


Ololoshechkin (changes requested):
Please keep the "withPersistence" and ability to rollback to checkpoint.

EugeneTheDev:
@Ololoshechkin Yes, we discussed this with @Rizzen briefly already, and I will be refactoring it in another direction. I will lift features back to the agent level, because it looks kinda strange that we have to recreate all features on each context copy. I reviewed the actual features that we have, and the only feature that actually needs `AIAgentContext` is `AgentMemory`. So it seems the correct direction would be to refactor `AgentPipeline` and features by removing `interceptContext` to prevent bugs like this with features grabbing expiring context instace this way. If the feature would require current context, it should be provided explicitly through some method parameters when interacting with the feature. 

Currently it affects only `AgentMemory`, so I will try to refactor it to not use the context implicitly and to remove `interceptContext` altogether. It is not needed anywhere else AFAIK (e.g. for observability), so if the features won't use it, we can remove it. It is broken currently anyways.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
```diff
@@ -181,17 +161,18 @@ public interface AIAgent<Input, Output> : Closeable {
...
          * @return An instance of an AI agent configured with the specified parameters and capable of executing its logic.
          */
+        @OptIn(ExperimentalUuidApi::class)
         public inline operator fun <reified Input, reified Output> invoke(
             promptExecutor: PromptExecutor,
             agentConfig: AIAgentConfig,
             strategy: AIAgentGraphStrategy<Input, Output>,
             toolRegistry: ToolRegistry = ToolRegistry.EMPTY,
-            id: String? = null,
+            id: String = Uuid.random().toString(),
```
sdubov: Could you please revert this and below changes back? It was done here due to the following comment from a user that uses native builds. Please see the [comment](https://github.com/JetBrains/koog/pull/560#discussion_r2318935726).

Ololoshechkin (approved):

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/AIAgentFeature.kt
```diff
@@ -1,74 +1,64 @@
...
+     * Installs the feature into the specified [pipeline] associated with the [agent].
+     * @return The implementation of the feature.
      */
-    @Suppress("UNCHECKED_CAST")
-    @InternalAgentsApi
-    public fun installUnsafe(config: Any?, pipeline: AIAgentGraphPipeline): Unit = install(config as Config, pipeline)
+    public fun install(
+        config: TConfig,
+        pipeline: AIAgentGraphPipeline,
+        agent: GraphAIAgent<*, *>
```
sdubov: As we discussed, the `agent` might be not necessary here. It is used only for one particular case for the `AgentMemory` feature when we set an agent name. The logic can be replaced with setting an unknown name inside a strategy interceptor.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
EugeneTheDev: Ok. I'm not fond of unnecessary nullability, but I didn't come up with a clean solution off the top of my head, so reverted it to nullable (not really fond of it, but what can we do...)

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/context/AIAgentContext.kt
```diff
@@ -180,3 +141,48 @@ public interface AIAgentContext {
...
+ * Retrieves a feature from the [AIAgentContext.pipeline] associated with this context using the specified key.
+ *
+ * @param TFeature A feature implementation type.
+ * @param feature A feature to fetch.
+ * @param featureKlass The [KClass] of the feature to be retrieved.
+ * @return The feature associated with the provided key, or null if no matching feature is found.
+ * @throws IllegalArgumentException if the specified [featureKlass] does not correspond to a registered feature.
+ */
+public fun <TFeature : Any> AIAgentContext.feature(
+    featureKlass: KClass<TFeature>,
```
sdubov: Sorry, is this naming intended? `featureKlass` instead of `featureClass`?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/config/FeatureConfig.kt
```diff
@@ -12,29 +12,25 @@ import ai.koog.agents.core.feature.message.FeatureMessageProcessor
...
-    public val messageProcessors: List<FeatureMessageProcessor>
-        get() = _messageProcessors.toList()
-
     /**
-     * A filter for events to be processed by a feature.
+     * A list of [FeatureMessageProcessor] instances registered with the feature configuration.
      */
-    public val eventFilter: (AgentLifecycleEventContext) -> Boolean
-        get() = _eventFilter
+    public var messageProcessors: List<FeatureMessageProcessor> = emptyList()
```
sdubov: `messageProcessor` should return you a list of existing processors. The only public api here was `addMessageProcessor()` to append to the list. Could you please clarify why do you need to override the whole list here?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/config/FeatureConfig.kt
```diff
@@ -12,29 +12,25 @@ import ai.koog.agents.core.feature.message.FeatureMessageProcessor
...
      */
     public fun addMessageProcessor(processor: FeatureMessageProcessor) {
-        _messageProcessors.add(processor)
+        messageProcessors = messageProcessors + processor
     }
 
+    /**
+     * A filter for events to be processed by a feature.
+     */
+    public var eventFilter: (AgentLifecycleEventContext) -> Boolean = { true }
```
sdubov: We briefly discussed this topic already... Does it make sense to deprecate the explicit functions in case you change this to `var`, and make a public setter?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/message/FeatureMessageProcessor.kt
```diff
@@ -29,18 +26,8 @@ public abstract class FeatureMessageProcessor : Closeable {
...
-    /**
-     * A [StateFlow] representing the current open state of the processor.
-     */
-    public abstract val isOpen: StateFlow<Boolean>
-
-    /**
-     * Initializes the feature output stream provider to ensure it is ready for use.
-     */
-    public open suspend fun initialize() { }
+    public var messageFilter: (FeatureMessage) -> Boolean = { true }
```
sdubov, line 29: Does it make sense to deprecate a property set through public method?

sdubov (approved):
@EugeneTheDev, looks good for me. 
There are two comments that I would point out:
1. `agent` parameter can be dropped in Pipeline `install()` methods (we discussed that this can be deleted)
2. check some minor comments about message processors

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/context/AIAgentContext.kt
EugeneTheDev: It is, because it's `KClass`. But since we don't have to work around `class` keyword in this case, I can rename it to `featureClass`, np

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/config/FeatureConfig.kt
EugeneTheDev: Reverted

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/message/FeatureMessageProcessor.kt
EugeneTheDev: Reverted

PR #956, opened 2025-10-13
[utils, http-client] (KG-294, KG-451) Introduce http-client modules for Koog HTTP client wrappers, add implementations for OkHttp and Java HttpClient

EugeneTheDev:
- Move more global util classes from agent-specific `agent-utils` module to higher-level general purpose `utils` module
- Introduce a new set of modules for Koog HTTP client wrappers: `http-client`
- Implement test infrastructure for HTTP client wrappers
- Implement two more clients, in addition to existing Ktor: OkHttp and Java HttpClient

**TODO:**
These clients are not actually used properly right now. As a next step, we need to update all LLM clients to work with an instance of `KoogHttpClient`, not Ktor's `HttpClient`. This would also mean updating and extending `KoogHttpClient` itself, to support necessary configuration options and behavior.


agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
```diff
@@ -11,11 +11,11 @@ import ai.koog.agents.core.agent.entity.AIAgentStorageKey
 import ai.koog.agents.core.annotation.InternalAgentsApi
 import ai.koog.agents.core.feature.AIAgentFeature
 import ai.koog.agents.core.tools.ToolRegistry
-import ai.koog.agents.utils.Closeable
 import ai.koog.prompt.dsl.prompt
 import ai.koog.prompt.executor.model.PromptExecutor
 import ai.koog.prompt.llm.LLModel
 import ai.koog.prompt.params.LLMParams
+import ai.koog.utils.Closeable
```
kpavlov: Just an idea: maybe move Closeable, SuitableForIo, and "use" to "ai.koog.utils.io" package?

http-client/http-client-java/src/main/kotlin/ai/koog/http/client/java/JavaKoogHttpClient.kt
```diff
@@ -0,0 +1,181 @@
...
+@Experimental
+internal class JavaKoogHttpClient internal constructor(
+    private val clientName: String,
+    private val logger: KLogger,
+    private val httpClient: HttpClient,
+    private val json: Json
+) : KoogHttpClient {
+
+    private data class RequestBody(
+        val body: String,
```
kpavlov, line 40: 🎗️for the future: to support IO stream as source (multimodal usecase)

http-client/http-client-java/src/main/kotlin/ai/koog/http/client/java/JavaKoogHttpClient.kt
```diff
@@ -0,0 +1,181 @@
...
+        request: T,
+        requestBodyType: KClass<T>,
+        responseType: KClass<R>
+    ): R = withContext(Dispatchers.SuitableForIO) {
+        val requestBody = prepareRequestBody(request, requestBodyType)
+
+        val httpRequest = HttpRequest.newBuilder()
+            .uri(URI.create(path))
+            .POST(HttpRequest.BodyPublishers.ofString(requestBody.body))
+            .header("Content-Type", requestBody.contentType)
```
kpavlov, line 55: Nit: Would be nice to add "User-Agent" and cache control headers

http-client/http-client-java/src/main/kotlin/ai/koog/http/client/java/JavaKoogHttpClient.kt
```diff
@@ -0,0 +1,181 @@
...
+                @Suppress("UNCHECKED_CAST")
+                responseBody as R
+            } else {
+                val serializer = serializer(responseType.java)
+                @Suppress("UNCHECKED_CAST")
+                json.decodeFromString(serializer, responseBody) as R
+            }
+        } else {
+            val errorBody = response.body()
+            logger.error { "Error from $clientName API: ${response.statusCode()}: $errorBody" }
```
kpavlov: Potentially, error body may expose PII if server isn't properly implemented. Maybe, the raw body should be printed in only trace log level.

http-client/http-client-java/src/main/kotlin/ai/koog/http/client/java/JavaKoogHttpClient.kt
```diff
@@ -0,0 +1,181 @@
...
+            error("Error from $clientName API: ${response.statusCode()}: $errorBody")
```
kpavlov: Let's exclude raw body from exception message. Reasons:
- The message should be one line to keep logs readable 
- It might expose PII
- it might be tooooo looooong

Would it made sense to introduce KoogHttpException with status code (not sure about abbreviated error body)?

http-client/http-client-java/src/main/kotlin/ai/koog/http/client/java/JavaKoogHttpClient.kt
```diff
@@ -0,0 +1,181 @@
...
+        }
+    }
+
+    override fun <T : Any, R : Any, O : Any> sse(
```
kpavlov, line 81: Nit: In general, it might be not only sse, but any chunked response. SSE is an extension of chunking with the specific decoder

http-client/http-client-java/src/main/kotlin/ai/koog/http/client/java/JavaKoogHttpClient.kt
```diff
@@ -0,0 +1,181 @@
...
+            logger.debug { "SSE connection opened for $clientName" }
+
+            // Process the stream of lines
+            response.body().forEach { line ->
+                try {
+                    val dataPrefix = "data: "
+                    // SSE format: "data: <content>"
+                    val data = if (line.startsWith(dataPrefix)) {
+                        line.substring(dataPrefix.length)
+                    } else if (line.isNotEmpty() && !line.startsWith(":")) {
```
kpavlov, line 119: Are "id", "event" and "comment" fields currently supported?

kpavlov, line 119: Are "id", "event" and "comment" fields currently supported?

http-client/http-client-java/src/main/kotlin/ai/koog/http/client/java/JavaKoogHttpClient.kt
```diff
@@ -0,0 +1,181 @@
...
+                } catch (e: Exception) {
+                    logger.error(e) { "Error processing SSE event from $clientName: ${e.message}" }
+                    close(e)
+                }
+            }
+
+            logger.debug { "SSE connection closed for $clientName" }
+            close()
+        } catch (e: Exception) {
+            val errorMessage = "Exception during streaming from $clientName: ${e.message}"
```
kpavlov, line 140: Nit: Some exceptions related to stream/connection termination might not be useful => might be suppressed

http-client/http-client-java/src/main/kotlin/ai/koog/http/client/java/JavaKoogHttpClient.kt
```diff
@@ -0,0 +1,181 @@
...
+    private fun <T : Any> prepareRequestBody(
+        request: T,
+        requestBodyType: KClass<T>,
+    ): RequestBody {
+        return if (requestBodyType == String::class) {
+            @Suppress("UNCHECKED_CAST")
+            RequestBody(body = request as String, contentType = "text/plain")
+        } else {
+            val serializer = serializer(requestBodyType.java)
+            RequestBody(body = json.encodeToString(serializer, request), contentType = "application/json")
```
kpavlov, line 162: JSON encoder/decoder might be passed as lambda properties to the class with a default to Kotlinx serialization serde

http-client/http-client-java/src/main/kotlin/ai/koog/http/client/java/JavaKoogHttpClient.kt
```diff
@@ -0,0 +1,181 @@
...
+    private val httpClient: HttpClient,
+    private val json: Json
+) : KoogHttpClient {
+
+    private data class RequestBody(
+        val body: String,
+        val contentType: String
+    )
+
+    override suspend fun <T : Any, R : Any> post(
```
kpavlov, line 44: Let's not forget to create also Java-friendly methods with CompletionStage as a result

http-client/http-client-java/src/main/kotlin/ai/koog/http/client/java/JavaKoogHttpClient.kt
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    override fun <T : Any, R : Any, O : Any> sse(
+        path: String,
+        request: T,
+        requestBodyType: KClass<T>,
+        dataFilter: (String?) -> Boolean,
+        decodeStreamingResponse: (String) -> R,
+        processStreamingChunk: (R) -> O?
+    ): Flow<O> = callbackFlow {
```
kpavlov, line 88: Let's create another wrapper method with Java Stream as a result

http-client/http-client-okhttp/src/main/kotlin/ai/koog/http/client/okhttp/OkHttpKoogHttpClient.kt
```diff
@@ -0,0 +1,176 @@
...
+                    @Suppress("UNCHECKED_CAST")
+                    responseBody as R
+                } else {
+                    val serializer = serializer(responseType.java)
+                    @Suppress("UNCHECKED_CAST")
+                    json.decodeFromString(serializer, responseBody) as R
+                }
+            } else {
+                val errorBody = response.body.string()
+                logger.error { "Error from $clientName API: ${response.code}: $errorBody" }
```
kpavlov: Please check the ktor client comments

http-client/http-client-okhttp/src/main/kotlin/ai/koog/http/client/okhttp/OkHttpKoogHttpClient.kt
```diff
@@ -0,0 +1,176 @@
...
+ * flexibility, and customization in HTTP interactions.
+ *
+ * @param clientName The name of the client instance, used for identifying or logging client operations.
+ * @param logger A `KLogger` instance used for logging client events and errors.
+ * @param okHttpClient The OkHttp client instance to be used. Defaults to a new OkHttpClient instance.
+ * @param json The Json instance used for serialization/deserialization. Defaults to a default Json instance.
+ * @return An instance of `KoogHttpClient` configured with the provided parameters.
+ */
+@Experimental
+public fun KoogHttpClient.Companion.fromOkHttpClient(
```
kpavlov, line 178: Should we add "@JvmStatic" to static methods for better Java interoperability?

http-client/http-client-test/src/main/kotlin/ai/koog/http/client/test/KoogHttpClientTestSuite.kt
```diff
@@ -0,0 +1,191 @@
...
+import kotlin.test.assertNotNull
+import kotlin.test.fail
+
+/**
+ * Abstract test suite for KoogHttpClient implementations.
+ * Provides common test scenarios that can be reused across different HTTP client implementations.
+ *
+ * Subclasses must implement [createClient] to provide the specific client implementation to test.
+ */
+abstract class KoogHttpClientTestSuite {
```
kpavlov: Nit: Test suite is usually a union of tests. Maybe call this just AbstractKoogHttpClientTest?

http-client/http-client-test/src/main/kotlin/ai/koog/http/client/test/MockWebServer.kt
```diff
@@ -0,0 +1,135 @@
...
+import kotlinx.coroutines.runBlocking
+import kotlin.concurrent.atomics.AtomicInt
+import kotlin.concurrent.atomics.ExperimentalAtomicApi
+
+/**
+ * MockWebServer provides a simple test server using Ktor with CIO engine.
+ * This server can be used to test HTTP clients with both regular POST requests and SSE streaming.
+ */
+@OptIn(ExperimentalAtomicApi::class)
+class MockWebServer {
```
kpavlov, line 30: DRY: use wiremock (Kotlin extension) or [Mokksy](https://moksy.dev) (for true streaming/SSE)

kpavlov (approved):
Thank you, @EugeneTheDev 🙏🏻 
Very valuable. Comment might be addressed separately

http-client/http-client-java/src/main/kotlin/ai/koog/http/client/java/JavaKoogHttpClient.kt
EugeneTheDev: Will probably do in the next PR with further developments of KoogHttpClient

http-client/http-client-java/src/main/kotlin/ai/koog/http/client/java/JavaKoogHttpClient.kt
EugeneTheDev: Will probably do in the next PR with further developments of KoogHttpClient. I plan to turn the base `KoogHttpClient` into an abstract class to allow for some common configuration options. Inspired by HttpClient from Ktor.

http-client/http-client-java/src/main/kotlin/ai/koog/http/client/java/JavaKoogHttpClient.kt
EugeneTheDev: Good point about error body, will exclude. I'll think about dedicated exception in the next PR, thanks

http-client/http-client-java/src/main/kotlin/ai/koog/http/client/java/JavaKoogHttpClient.kt
EugeneTheDev: I would say we are following YAGNI here, since we are not implemeting a general-purpose HTTP client, but rather the one for our specific purposes. So I suggest adding what's actually needed for now. And we can always add the rest later

http-client/http-client-java/src/main/kotlin/ai/koog/http/client/java/JavaKoogHttpClient.kt
EugeneTheDev: Again, YAGNI. We're only using the `data` parts. We can add others if needed later, shouldn't be a big problem. But I'll think in the next PR, maybe we can introduce some simple class here, that will contain chunk type and its raw content, and return a flow of these classses, since it should be not that hard to do.

EugeneTheDev: The same as above

http-client/http-client-java/src/main/kotlin/ai/koog/http/client/java/JavaKoogHttpClient.kt
EugeneTheDev: Maybe, but I would say connection exceptions are also informative, since they indicate that stream has finished in a non-normal manner. I would prefer keeping it as it is for now, and then adjusting based on users'/our own feedback.

http-client/http-client-java/src/main/kotlin/ai/koog/http/client/java/JavaKoogHttpClient.kt
EugeneTheDev: That's actually what I've been thinking to do as the next step, when I'll turn `KoogHttpClient` in an abstract class. Then it will be able to have configuration settings and presets.

http-client/http-client-java/src/main/kotlin/ai/koog/http/client/java/JavaKoogHttpClient.kt
EugeneTheDev: We would need to support this in the future, agreed. However, I propose not doing it right now, since it sounds like part of a broader and more global task of making our API Java-friendly. I would like to focus first on actually bringing Koog HTTP client into LLM clients, to finish this migration from hardcoded Ktor clients in our LLM clients. But your point is definitely valid.

http-client/http-client-okhttp/src/main/kotlin/ai/koog/http/client/okhttp/OkHttpKoogHttpClient.kt
EugeneTheDev: This is an extension method, it's static by default, in the scope of the current file, i.e. in the class that will be created from the current file. We have to make it Java-friendly though, since it has default parameters, so we need to mark it with `@JvmOverloads`.  But again, we have a lot of methods wtih default parameters across our codebase, so that's also part of a broader Java API task.

http-client/http-client-test/src/main/kotlin/ai/koog/http/client/test/MockWebServer.kt
EugeneTheDev: Sorry, wdym regarding DRY here? I thought that the simplest and closest to reality test would be just to implement a small real server with Ktor. It's really small, and allows to create all sorts of real HTTP responses, so you can test the clients reliably and easily. Doesn't look like repetition to me, but maybe I'm missing something.

PR #972, opened 2025-10-14
(KG-463) Fix integration tests and related bugs and problems. Improve test setup to speed up tests and make them more robust and clean

EugeneTheDev:
This PR introduces a number of improvements related to our test setup and tests themselves, along with several related bugfixes:

### Parallel test execution

* All tests now run in parallel, except for Ollama tests
* This significantly speeds up Heavy Tests: from ~40 minutes down to ~15 minutes
* While this doesn't speed up regular unit tests (Checks), it helps uncover concurrency problems and sub-optimal test setup, mostly relating to global state or improper synchronization techniques. Several bugs were found and fixed in this PR thanks to this change. Despite not providing a speed increase for Checks, the improved test quality and additional concurrency bugs detection makes it beneficial, so I propose keeping it enabled for regular Checks as well

### Ollama test setup updates

* Updated Ollama test setup to make it easy to use a local Ollama server
* Introduced proper downloaded model caching when running Ollama with Docker using volumes
* Updated Ollama Tests CI config to download and use the Ollama server directly instead of using Docker
* This avoids Docker overhead and leads to a noticeable speed improvement: from ~40 minutes down to ~25 minutes

### Test refactoring and fixes

* Refactored and fixed all integration and Ollama tests to make them pass
* A few related bugs and problems were found and fixed along the way

PR #990, opened 2025-10-17
KG-475: Make Ollama tests faster

fazlerahmanejazi:
## Motivation and Context
[KG-475](https://youtrack.jetbrains.com/issue/KG-475/Make-Ollama-tests-faster)
- cache the shared Ollama Testcontainers fixture so every integration test reuses the same container
- rely on Testcontainers’ default image pull policy instead of forcing alwaysPull

## Breaking Changes
No.

---

#### Type of the changes
- [ ] Documentation update
- [x] Tests improvement
- [ ] Refactoring

#### Checklist


fazlerahmanejazi:
@aozherelyeva suppose you're the expert for this(tagging since I can’t set reviewers)

kpavlov (approved):
LGTM

kpavlov:
I suggest starting the Ollama container once and not stopping it at all. Let's create a holder object like [TestEnvironment](https://github.com/kpavlov/sample-banking-api/tree/main/src/it/kotlin/me/kpavlov/sample/banking/api/it/fixture) and start the Ollama container during its initialization. This will ensure that the container is started once, before all the tests

kpavlov (changes requested):
I suggest starting the Ollama container once and not stopping it at all until JVM shutdown. Testcontainers should probably take care of it. It would make the setup simple and straightforward

fazlerahmanejazi:
> I suggest starting the Ollama container once and not stopping it at all until JVM shutdown. Testcontainers should probably take care of it. It would make the setup simple and straightforward

@kpavlov It is exactly what we already implemented: `OllamaTestFixtureExtension` already holds a singleton fixture `OllamaTestFixture`. The first test class starts the container, subsequent ones reuse it, and it stays running until JVM shutdown. We only stop it when the JVM exits (or when the very last consumer releases it), so we get the “start once, use everywhere” behavior.

Update: Thanks for the suggestion — I get your intent now. Simplified so the container stays up until JVM shutdown.

fazlerahmanejazi:
@kpavlov Dropped the reference counting and release path. The fixture now spins up once and sticks around until JVM shutdown (or a setup failure).

aozherelyeva (approved):
Thank you for the enhancement. Let's give it a try!

aozherelyeva:
Some [recent changes](https://github.com/JetBrains/koog/pull/972) place Ollama tests out of a Docker image, so I believe the things changed a bit. Sorry, but it seems that with the newest changes this PR becomes obsolete 😢

EugeneTheDev:
A bit more info on what will change after #972 regarding Ollama tests: I decided to ditch Docker to avoid container overhead and instead install and run Ollama directly on CI instances. And this actually gave a nice speed boost, [check this run](https://github.com/JetBrains/koog/actions/runs/18637505466) - it only took 23 minutes to complete compared to regular ~40 minutes with Docker. I also kept an option to still run Ollama tests by hosting Ollama in Docker container, and added models cache by having a dedicated volume, so that you don't have to pull models from scratch on every run. However, now that you can run Ollama tests with local Ollama both on CI and on your own machine, Docker-related improvements are not that relevant, since they won't provide perfromance boost comparable with just running Ollama locally.

PR #946, opened 2025-10-09
Add params for google and anthropic and refactor LLMParams

devcrocod:
[KG-453](https://youtrack.jetbrains.com/issue/KG-453) Add specific params for non-OpenAI compatible clients

- add Anthropic Params
- add Google Params
- refactor LLMParams

## Motivation and Context
To be able to configure LLM requests more precisely and improvement of the API

## Breaking Changes
Yes,
two properties have been removed from `LLMParams` that were used exclusively by `GoogleLLMClient`.
These parameters were not suitable for all clients.
Instead, such parameters are now passed through specialized classes like `OpenAIParams`, `GoogleParams`, and others.

---

#### Type of the changes
- [x] New feature (non-breaking change which adds functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Tests improvement
- [x] Refactoring

#### Checklist
- [x] The pull request has a description of the proposed change
- [x] I read the [Contributing Guidelines](https://github.com/JetBrains/koog/blob/main/CONTRIBUTING.md) before opening the pull request
- [x] The pull request uses **`develop`** as the base branch
- [x] Tests for the changes have been added
- [x] All new and existing tests passed

##### Additional steps for pull requests adding a new feature
- [x] An issue describing the proposed change exists
- [x] The pull request includes a link to the issue
- [ ] The change was discussed and approved in the issue
- [x] Docs have been added / updated

PR #991, opened 2025-10-17
KG-452 Add documentation about LLMParams

marinkovicmarko:
Add documentation for LLM parameters (LLMParams + provider-specific parameters).

---

#### Type of the changes
- [x] Documentation update
- [ ] Tests improvement
- [ ] Refactoring

#### Checklist

### DIFF
--- a/docs/docs/llm-parameters.md
+++ b/docs/docs/llm-parameters.md
@@ -1,0 +1,418 @@
+# LLM parameters
+
+This page provides details about LLM parameters in the Koog agentic framework. LLM parameters let you control and customize the behavior of language models.
+
+## Overview
+
+LLM parameters are configuration options that let you fine-tune how language models generate responses. These parameters control aspects like response randomness, length, format, and tool usage. By adjusting the parameters, you optimize model behavior for different use cases, from creative content generation to deterministic structured outputs.
+
+In Koog, the `LLMParams` class incorporates LLM parameters and provides a consistent interface for configuring language model behavior. You can use LLM parameters in the following ways:
+
+- When creating a prompt:
+
+    <!--- INCLUDE
+    import ai.koog.prompt.prompt
+    import ai.koog.prompt.params.LLMParams
+    -->
+    ```kotlin
+    val prompt = prompt(
+        "dev-assistant",
+        LLMParams(
+            temperature = 0.7,
+            maxTokens = 500
+        )
+    ) {
+        // Add a system message to set the context
+        system("You are a helpful assistant.")
+
+        // Add a user message
+        user("Tell me about Kotlin")
+    }
+    ```
+    <!--- KNIT example-llm-parameters-01.kt -->
+
+    For more information about prompt creation, see [Prompt API](prompt-api.md).
+
+- When creating a subgraph:
+
+    <!--- INCLUDE
+    import ai.koog.agents.core.dsl.builder.strategy
+    import ai.koog.agents.ext.tool.SayToUser
+    import ai.koog.prompt.executor.clients.openai.OpenAIModels
+    import ai.koog.agents.ext.agent.subgraphWithTask
+    val searchTool = SayToUser
+    val calculatorTool = SayToUser
+    val weatherTool = SayToUser
+    val strategy = strategy<String, String>("strategy_name") {
+    -->
+    <!--- SUFFIX
+    }
+    -->
+    ```kotlin
+    val processQuery by subgraphWithTask<String, String>(
+        tools = listOf(searchTool, calculatorTool, weatherTool),
+        llmModel = OpenAIModels.Chat.GPT4o,
+        llmparams = LLMParams(
+            temperature = 0.7,
+            maxTokens = 500,
+        )
+    ) { userQuery ->
+        """
+        You are a helpful assistant that can answer questions about various topics.
+        Please help with the following query:
+        $userQuery
+        """
+    }
+    ```
+    <!--- KNIT example-llm-parameters-02.kt -->
+
+    For more information on how to create and implement your own subgraphs, see [Custom subgraphs](custom-subgraphs.md).
+
+- When updating a prompt in an LLM write session:
+
+    <!--- INCLUDE
+    import ai.koog.agents.core.dsl.builder.strategy
+    val strategy = strategy<Unit, Unit>("strategy-name") {
+    val node by node<Unit, Unit> {
+    -->
+    <!--- SUFFIX
+       }
+    }
+    -->
+    ```kotlin
+    llm.writeSession {
+        changeLLMParams(
+            LLMParams(
+                temperature = 0.7,
+                maxTokens = 500
+            )
+        )
+    }
+    ```
+    <!--- KNIT example-llm-parameters-03.kt -->
+
+    For more information about sessions, see [LLM sessions and manual history management](sessions.md).
+
+## LLM parameter reference
+
+The following table provides a reference of LLM parameters included in the `LLMParams` class and supported by all LLM providers that are available in Koog out of the box. 
+For a list of parameters that are specific to some providers, see [Provider-specific parameters](#provider-specific-parameters).
+
+| Parameter              | Type                           | Description                                                                                                                                                                                                                                                                      |
+|------------------------|--------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `temperature`          | Double                         | Controls randomness in the output. Higher values, such as 0.7–1.0, produce more diverse and creative responses, while lower values produce more deterministic and focused responses. Takes a value in the range of 0.0–2.0.                                                      |
+| `maxTokens`            | Integer                        | Maximum number of tokens to generate in the response. Useful for controlling response length.                                                                                                                                                                                    |
+| `numberOfChoices`      | Integer                        | Number of alternative responses to generate. Must be greater than 0.                                                                                                                                                                                                             |
+| `speculation`          | String                         | A speculative configuration string that influences model behavior, designed to enhance result speed and accuracy. Supported only by certain models, but may greatly improve speed and accuracy.                                                                                  |
+| `includeThoughts`      | Boolean                        | If `true`, requests the model to add reasoning blocks to the response. Responses may include detailed reasoning steps. When `false` or `null`, responses are typically shorter and faster.                                                                                       |
+| `thinkingBudget`       | Integer                        | The top limit for the use of reasoning tokens. Can be used to limit the amount of tokens used for reasoning when `includeThoughts` is set to `true`.                                                                                                                             |
+| `schema`               | Schema                         | Defines the structure for the model's response format, enabling structured outputs like JSON. For more information, see [Schema](#schema).                                                                                                                                       |
+| `toolChoice`           | ToolChoice                     | Controls tool calling behavior of the language model. For more information, see [Tool choice](#tool-choice).                                                                                                                                                                     |
+| `user`                 | String                         | Identifier for the user making the request, which can be used for tracking purposes.                                                                                                                                                                                             |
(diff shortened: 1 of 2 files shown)


docs/docs/llm-parameters.md
```diff
@@ -0,0 +1,418 @@
...
+
+- When creating a prompt:
+
+    <!--- INCLUDE
+    import ai.koog.prompt.prompt
+    import ai.koog.prompt.params.LLMParams
+    -->
+    ```kotlin
+    val prompt = prompt(
+        "dev-assistant",
```
azhebel: parameter names?

docs/docs/llm-parameters.md
```diff
@@ -0,0 +1,418 @@
...
+        )
+    }
+    ```
+    <!--- KNIT example-llm-parameters-03.kt -->
+
+    For more information about sessions, see [LLM sessions and manual history management](sessions.md).
+
+## LLM parameter reference
+
+The following table provides a reference of LLM parameters included in the `LLMParams` class and supported by all LLM providers that are available in Koog out of the box.
```
azhebel, line 98: Maybe the table should mention default values, wdyt?

docs/docs/llm-parameters.md
```diff
@@ -0,0 +1,418 @@
...
+        )
+    )
+    ```
+    <!--- KNIT example-llm-parameters-05.kt -->
+
+## Tool choice
+
+The `ToolChoice` class controls how the language model uses tools. It provides the following options:
+
+1. **Named** (`LLMParams.ToolChoice.Named`): the language model calls the specified tool.
```
azhebel: I don't think there is a need to name the subclass and then type it out from the top-level class in parentheses. Just `LLMParams.ToolChoice.Named` as the list item name is fine. Also, this is not a sequence, so should be unordered. Maybe even use a definition list? IMO it's what def lists are for. But not everyone likes how they look compared to lists with bullet points, and it requires adding the `def_list` extension: https://squidfunk.github.io/mkdocs-material/reference/lists/

docs/docs/llm-parameters.md
```diff
@@ -0,0 +1,418 @@
...
+3. **Auto** (`LLMParams.ToolChoice.Auto`): the language model automatically decides whether to call tools and which tool to call.
+4. **Required** (`LLMParams.ToolChoice.Required`): the language model calls at least one tool.
+
+
+<!--- INCLUDE
+import ai.koog.prompt.params.LLMParams
+-->
+```kotlin
+// Use a specific tool
+val specificToolParams = LLMParams(
```
azhebel, line 222: Having both the list of possible tool choices and these examples seems redundant. I'd leave just the list without examples in this case, because they are very simple and pretty much the same. Maybe add only one example for the named tool choice, because you need to add a name of the tool there. If you think the examples are useful, then maybe drop the list and have only examples with descriptions in comment?

docs/docs/llm-parameters.md
```diff
@@ -0,0 +1,418 @@
...
+<!--- KNIT example-llm-parameters-06.kt -->
+
+## Provider-specific parameters
+
+Koog supports provider-specific parameters for some LLM providers. These parameters extend the base `LLMParams` class
+and add provider-specific functionality. The following classes include parameters that are specific per provider:
+
+- DeepSeek: `DeepSeekParams`
+- OpenRouter: `OpenRouterParams`
+- The supported parameters for OpenAI may differ between the underlying OpenAI APIs:
```
azhebel: Adding this nesting level seems unnecessary. And this may also be a case for a def list or maybe just a flat bulleted list. Something like this:

`DeepSeekParams`
: Parameters specific to DeepSeek models.

`OpenRouterParams`
: Parameters specific to OpenRouter models.

`OpenAIChatParams`
: Parameters specific to the OpenAI Chat Completions API.

`OpenAIResponsesParams`
: Parameters specific to the OpenAI Responses API.

docs/docs/llm-parameters.md
```diff
@@ -0,0 +1,418 @@
...
+    - OpenAI Chat: `OpenAIChatParams`. Parameters specific to the OpenAI Chat Completions API.
+    - OpenAI Responses: `OpenAIResponsesParams`. Parameters specific to the OpenAI Responses API.
+
+Here is the complete reference of provider-specific parameters in Koog:
```
azhebel, line 238: Also maybe would be nice to have defaults mentioned?

docs/docs/llm-parameters.md
```diff
@@ -0,0 +1,418 @@
...
+| `presencePenalty`   | OpenAI Chat, DeepSeek, OpenRouter                   | Double                 | Prevents the model from reusing tokens that have already been included in the output. Higher values encourage the introduction of new tokens and topics. Takes a value in the range of -2.0 to 2.0.                                                                                                                                                                                                                                |
+| `stop`              | OpenAI Chat, DeepSeek, OpenRouter                   | List&lt;String&gt;     | Stop sequences. The model stops generating content when it encounters any of the sequences in the list. A list of strings containing maximum 4 items.                                                                                                                                                                                                                                                                              |
```
azhebel: This seems like the least obvious one in terms of syntax. What are "items" in a string? Like words? An example would be helpful here.

docs/docs/llm-parameters.md
```diff
@@ -0,0 +1,418 @@
...
+| `parallelToolCalls` | OpenAI Chat, OpenAI Responses                       | Boolean                | If `true`, multiple tool calls can run in parallel.                                                                                                                                                                                                                                                                                                                                                                                |
```
azhebel, line 248: This comment is more from the user's perspective, not related to content but to the API itself. Maybe I don't get something, but my understanding was that by default an LLM request node can return a tool call, because a tool also returns one specific result. If you want to expect parallel tool calls, you need to use a different node, because parallel tool calls will return a list (or a map?) of tool results. So I don't understand what we control here if this is controlled by the type of node. But I didn't look at the API myself, this was just my first thought when I read this)

docs/docs/llm-parameters.md
```diff
@@ -0,0 +1,418 @@
...
+| `topA`              | OpenRouter                                          | Double                 | Dynamically adjusts the sampling window based on model confidence. If the model is confident (there are dominant high-probability next tokens), it keeps the sampling window limited to a few top tokens. If the confidence is low (there are many tokens with similar probabilities), keeps more tokens in the sampling window. Takes a value in the range of 0.0–0.1 (inclusive). Higher value means greater dynamic adaptation. |
+| `transforms`        | OpenRouter                                          | List&lt;String&gt;     | List of context transforms. Defines how context is transformed when it exceeds the model's token limit. The default transformation is `middle-out` which truncates from the middle of the prompt. Use empty list for no transformations.                                                                                                                                                                                           |
```
azhebel: An example here would also be helpful or a link to relevant OpenRouter docs?

docs/docs/llm-parameters.md
```diff
@@ -0,0 +1,418 @@
...
+    additionalProperties = additionalPropertiesOf(
+        "top_p" to 0.95,
+        "frequency_penalty" to 0.5,
+        "presence_penalty" to 0.5
+    )
+)
+```
+<!--- KNIT example-llm-parameters-12.kt -->
+
+### Setting and overriding default parameters
```
azhebel: Is this also how I'd combine generic LLM params with provider-specific ones?

docs/docs/llm-parameters.md
```diff
@@ -0,0 +1,418 @@
...
+        // Add a system message to set the context
+        system("You are a helpful assistant.")
+
+        // Add a user message
+        user("Tell me about Kotlin")
+    }
+    ```
+    <!--- KNIT example-llm-parameters-01.kt -->
+
+    For more information about prompt creation, see [Prompt API](prompt-api.md).
```
devcrocod, line 34: It seems to me there shouldn’t be any indentation here

docs/docs/llm-parameters.md
devcrocod: This is a good idea, but it might be difficult to maintain in the future, since each provider can change its default values. It might be better to refer to their documentation instead.

docs/docs/llm-parameters.md
```diff
@@ -0,0 +1,418 @@
...
+For a list of parameters that are specific to some providers, see [Provider-specific parameters](#provider-specific-parameters).
+
+| Parameter              | Type                           | Description                                                                                                                                                                                                                                                                      |
+|------------------------|--------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `temperature`          | Double                         | Controls randomness in the output. Higher values, such as 0.7–1.0, produce more diverse and creative responses, while lower values produce more deterministic and focused responses. Takes a value in the range of 0.0–2.0.                                                      |
```
devcrocod: The default values differ between providers: some use the range [0, 1], some use [0, 2]

docs/docs/llm-parameters.md
marinkovicmarko: This indentation makes both the code sample and this sentence a part of the bullet point content (intentionally). If there's no indentation, it would put the content outside of the context of the specific bullet point.

docs/docs/llm-parameters.md
marinkovicmarko: I also think it might be a challenge to implement as the defaults may differ per provider, especially with more providers. Below the table I'm adding a list of links to provider docs with either request body specs or separate parameter reference pages (where available). If you think another approach might work better or there are more appropriate links or resources, I'd be happy to hear.

docs/docs/llm-parameters.md
marinkovicmarko: Agreed, removing (descriptive) names and converting to a bulleted list. Let's discuss the deflist plugin.

docs/docs/llm-parameters.md
marinkovicmarko: I'm leaning towards leaving the list and just the sample for `Named` with an appropriate intro sentence.

docs/docs/llm-parameters.md
marinkovicmarko: This is how you would create a general set of parameters that you'd likely use the most and a new similar set from it by just partially overriding the existing one and not having to define everything from scratch. Might be a specific use case, I'll remove "default" to prevent confusion, but we can also remove the section altogether.

docs/docs/llm-parameters.md
marinkovicmarko: My best guess is custom node implementations, but I'll let @devcrocod answer this

docs/docs/llm-parameters.md
azhebel: I get the default + overrides example for generic parameters. My question is whether it's possible to create an object that has some generic parameters and also provider-specific ones, so that you can use the same LLMParams object with different models. Intuitively, I feel like this can be a useful case. But maybe not?

Something like:
```
val params = LLMParams(<generic_params>)
    .default(DeepSeekParams(<deepseek_params>))
    .default(OpenRouterParams(<openrouter_params>))
```

Will this work?

devcrocod: Nope
We’re planning to move the common parameters for enabling reasoning into LLMParams ([link](https://github.com/JetBrains/koog/pull/946#issuecomment-3402270985)) and then design an API for the functionality you described

docs/docs/llm-parameters.md
devcrocod: By default, the `parallelToolCalls` parameter in openai is set to `true`
It controls how the model handles tool calls, whether it returns them one by one or as a list
And our clients/providers are public classes and can be used outside of agent strategies or in custom nodes
Ideally, we should fully implement the providers API

azhebel: @devcrocod so then this parameter will not affect how our nodes handle tool calls? Is it worth explaining this, wdyt?

azhebel (approved):

PR #1004, opened 2025-10-20
Replace string content and attachments list in messages with a unified content parts list to make the API more flexible and preserve text/attachment parts order

tiginamaria:
## Motivation and Context

In order to make order in text and attachments I have to introduce object `ContentPart` (`ContentPart.Text` + `ContentPart.Attachment`) and `parts: List<ContentPart>` attribute in `Message` interface

Also the prompt api has changed, so now to create a message with list of parts you don't need to add `attachment` block
```
val prompt = Prompt.build("test") {
    user {
        +"Hello, how are you?"
        +"Here is my photo"
        image("https://example.com/photo1.jpg")
        +"I'm good!"
        +"And here is mine"
        image("https://example.com/photo2.jpg")
    }
}
```
```
listOf(
    ContentPart.Text("Hello, how are you?\n"Here is my photo"),
    ContentPart.Image(AttachmentContent.URL("https://example.com/photo1.jpg"), "jpg", "image/jpg", "photo1.jpg"),
    ContentPart.Text("I'm good!\nAnd here is mine"),
    ContentPart.Image(AttachmentContent.URL("https://example.com/photo2.jpg"), "jpg", "image/jpg", "photo2.jpg"),
)
```

## Breaking Changes

---

#### Type of the changes
- [ ] Documentation update
- [ ] Tests improvement
- [x] Refactoring

#### Checklist


prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/dsl/ContentPartsBuilder.kt
```diff
@@ -96,15 +118,15 @@ public class AttachmentBuilder {
...
      *
      * @param path Path to local image file
      * @throws IllegalArgumentException if the path is not valid, the file does not exist, or is not a regular file.
      */
     public fun image(path: Path) {
         val fileData = path.fileData()
         image(
-            Attachment.Image(
+            ContentPart.Image(
                 content = AttachmentContent.Binary.Bytes(path.readByteArray()),
```
kpavlov, line 129: nit: This is expensive, because it reads the whole image into memory. It would be great to have `AttachmentContent.Binary.Path` and read bytes only for serialization

agents/agents-features/agents-features-a2a-core/src/commonMain/kotlin/ai/koog/agents/a2a/core/MessageConverters.kt
```diff
@@ -96,3 +99,64 @@ public fun Message.toA2AMessage(
...
+
+/**
+ * Converts Koog [ContentPart] to A2A [Part].
+ */
+public fun ContentPart.toA2APart(): Part = when (this) {
+    is ContentPart.Text -> TextPart(this.text)
+
+    is ContentPart.Attachment -> {
+        val file = when (val content = this.content) {
+            // Plain text files are not supported, convert them to binary files.
```
kpavlov, line 111: Hmm... There is a [TextPart](https://agent2agent.info/docs/concepts/part/#different-types-of-parts) in the spec

tiginamaria: Text files, text parts are supported

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/dsl/ContentPartsBuilder.kt
tiginamaria: Indeed, maybe next PR

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/builder/AIAgentGraphStrategyBuilder.kt
```diff
@@ -9,7 +9,7 @@ import kotlin.reflect.typeOf
 
 /**
  * A builder class responsible for constructing an instance of `AIAgentStrategy`.
```
EugeneTheDev: nit: Let's from now on replace backticks with square brackets when refering to types? It will create a proper reference in generated API docs, making navigation slightly easier

docs/docs/prompt-api.md
```diff
@@ -61,21 +61,15 @@ val prompt = prompt("multimodal_input") {
 
     user {
         +"Describe these images"
-
-        attachments {
-            image("https://example.com/test.png")
-            image(Path("/User/koog/image.png"))
-        }
+        
+        image("https://example.com/test.png")
```
kpavlov, line 65: is there a way to distinguish between URL and base64 string?

prompt/prompt-model/src/commonTest/kotlin/ai/koog/prompt/dsl/PromptBuilderTest.kt
```diff
@@ -169,6 +169,42 @@ class PromptBuilderTest {
...
+                    image(
+                        ContentPart.Image(
+                            content = AttachmentContent.URL("https://example.com/test.png"),
+                            format = "png",
+                            mimeType = "image/png",
+                            fileName = "test.png"
+                        )
+                    )
+                }
+            }
```
kpavlov, line 187: neat, but might be sweeter :candy:

agents/agents-features/agents-features-trace/src/jvmTest/kotlin/ai/koog/agents/features/tracing/mock/TestAgent.kt
```diff
@@ -31,15 +30,12 @@ internal val testClock: Clock = object : Clock {
  * the message content.
  *
  * @param content The text content of the user message.
- * @param attachmentsBlock A lambda function used to configure the media attachments
- *                          for the message, using the `AttachmentBuilder` DSL.
  * @return A `Message.User` object containing the message content, metadata,
  *         and any associated media attachments.
  */
-fun userMessage(content: String, attachmentsBlock: AttachmentBuilder.() -> Unit = {}): Message.User = Message.User(
+fun userMessage(content: String): Message.User = Message.User(
```
kpavlov, line 36: nit: it might make sense to add also `fun userMessage(block: *()->String)` to keep code DSLish

kpavlov (approved):
Thank you, @tiginamaria 
It makes sense to merge

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/message/Message.kt
```diff
@@ -17,9 +17,17 @@ import kotlin.jvm.JvmOverloads
 @Serializable
 public sealed interface Message {
     /**
-     * The content of the message.
+     * The content of the message aggregated from all [ContentPart.Text] parts joined to [String] by line separator.
```
EugeneTheDev: nit: Let's clarify a bit more in the KDoc: "The textual content...". And also "line separator" doesn't give much information, let's update to "separated by newlines" or something similar

prompt/prompt-model/src/commonTest/kotlin/ai/koog/prompt/dsl/PromptBuilderTest.kt
tiginamaria: You can just write `user("https://example.com/test.png")`, here I'm just testing that `ContentPart.Image` api also works

prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/bedrock/BedrockLLMClient.kt
```diff
@@ -451,7 +454,10 @@ public class BedrockLLMClient(
                                     }
                                 }
                             )
-                        )
+                        }
+                        else -> {
```
EugeneTheDev: Seems like this behavior is (and was) inconsistent with other clients. Here we skip unsupported parts silently, in other LLM clients (e.g. [AbstractOpenAILLMClient](https://github.com/JetBrains/koog/blob/d2d4d97b79897acc0d7e84b00628736e2e715874/prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client-base/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openai/base/AbstractOpenAILLMClient.kt#L356)) we explicitly throw an exception. I suggest being more explicit and throwing an exception here too

agents/agents-features/agents-features-trace/src/jvmTest/kotlin/ai/koog/agents/features/tracing/mock/TestAgent.kt
tiginamaria: I guess it's just a helper method for tests not public api, so I guess it good to have if there is a case for it in tests, I'll have a look, thanks

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/text/TextContentBuilder.kt
```diff
@@ -24,3 +24,22 @@ public open class TextContentBuilder : TextContentBuilderBase<String>() {
...
+ * ```kotlin
+ * TextContentBuilder().apply {
+ *     text { "Some text before markdown." }
+ *     text { "Some other text." }
+ * }
+ * ```
+ *
+ * @param init The text content builder
+ */
+public inline fun TextContentBuilderBase<*>.text(init: TextContentBuilder.() -> Unit) {
```
EugeneTheDev, line 43: Why is this function needed? If you're inside the `TextContentBuilderBase`, you already have all DSL functions to append the text. Also, code sample in KDoc is not entirely correct, you forgot unary plus operator to make it actually append the text:
```kotlin
text { +"some text" }
```

docs/docs/prompt-api.md
tiginamaria: So we now have builder for string url only, but maybe we can also add with `kotlin.io.encoding.Base64`. Otherwise, there is no way to distinguish the right function. Or you can always add `ContentPart.Image` and put inside right content type

examples/simple-examples/src/main/kotlin/ai/koog/agents/example/a2a/advancedjoke/JokeWriterAgentExecutor.kt
```diff
@@ -225,7 +225,7 @@ private fun jokeWriterStrategy() = strategy<A2AMessage, Unit>("joke-writer") {
         llm.writeSession {
             updatePrompt {
                 user {
-                    +text {
+                    text {
```
EugeneTheDev: I already commented on this `text` function in the comment above. Looks like it's not needed here, as we are already in a text builder when we start `user {}` block. So it should be possible to safely remove it, reducing the nesting too

EugeneTheDev (approved):
Generally looks good to me, nicely implemented. I have some minor comments, please take a look

docs/docs/prompt-api.md
EugeneTheDev: We don't provide `image` DSL function for Base64 images yet. So, there's only one `image(String)` function that clearly states in its KDoc that it excepts a URL. Maybe we can add convenience function for Base64 too, but then we would have to use different names for these functions. I'm not sure if we need a convenience function for Base64 currently, though. So I suggest leaving it as it is for now

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/builder/AIAgentGraphStrategyBuilder.kt
tiginamaria: Totally agree

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/text/TextContentBuilder.kt
tiginamaria: It's just to be consistent with `markdown` and `xml` dsl, as if we do not have it, users need to write `+text{ +"some text" }`, but `markdown { +"some text" }` without unary +, to append the text to the builder. Or am I getting something wrong?

examples/simple-examples/src/main/kotlin/ai/koog/agents/example/a2a/advancedjoke/JokeWriterAgentExecutor.kt
tiginamaria: Removed text block

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/text/TextContentBuilder.kt
EugeneTheDev: But why would you need to write `text {}` builder if you're already inside one? You can just append the strings directly. `xml` and `markdown` exist because they provide more dedicated methods for each textual format, so it's sort of like "and now a snippet of text with extended formatting"

EugeneTheDev: So the intended usage would be:
```kotlin
user {
    +"Plain line"

    markdown {
        h1("Header 1")
        +"more text"
    }
}
```

tiginamaria: True, I don't know, but I just was thinking that is I need to replace `markdown` with `text` I need to change a lot, but I guess it's not very common thing. Should I drop it then or leave as it is now?

tiginamaria: I'll leave it as I really think that it's not super extra, and I'd like to have same interface as for markdown, because if user will just use `text { }` block without it, the string won't be added anywhere, but the code will be green and not gray

EugeneTheDev: OK, makes sense. Let's leave it for the sake of consistency, but avoid using it unnecessarily in message builders (I see you changed it, so that's good)

PR #1014, opened 2025-10-22
KG-480. Return JSON element for input and output in Agent events

sdubov:
[KG-480](https://youtrack.jetbrains.com/issue/KG-480).

- Agent node execution events return String object for input and output data. It might be more convenient to pass JsonElement through events to be able to use more structured data and render it on a client side;
- Pass node input object to the NodeExecutionFailedEvent event;
- Add serialization handling for ReceivedToolResultType;
- Update Debugger and Tracing logic and tests.

## Motivation and Context
Add a structured JsonElement data for input/output properties for a node agent events. This can help to process this data on a client side.

## Breaking Changes
Yes (change input/output parameter types in agent events)

---

#### Type of the changes
- [ ] Documentation update
- [ ] Tests improvement
- [ ] Refactoring

#### Checklist

### DIFF
--- a/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/utils/SerializationUtil.kt
+++ b/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/utils/SerializationUtil.kt
@@ -1,0 +1,54 @@
+package ai.koog.agents.core.utils
+
+import ai.koog.agents.core.annotation.InternalAgentsApi
+import io.github.oshai.kotlinlogging.KotlinLogging
+import kotlinx.serialization.SerializationException
+import kotlinx.serialization.json.Json
+import kotlinx.serialization.json.JsonElement
+import kotlinx.serialization.serializerOrNull
+import kotlin.reflect.KType
+
+/**
+ * Utility object for handling serialization of input data to JSON using Kotlin Serialization.
+ */
+@InternalAgentsApi
+public object SerializationUtil {
+
+    private val json = Json {
+        prettyPrint = true
+    }
+
+    private val logger = KotlinLogging.logger { }
+
+    /**
+     * Attempts to serialize the given input data using the provided data type.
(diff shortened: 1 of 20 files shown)


agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/utils/SerializationUtil.kt
```diff
@@ -0,0 +1,54 @@
...
+    /**
+     * Serializes the given input data into a JSON element using the specified data type.
+     * If no serializer is found for the data type, the function returns null.
+     *
+     * @param data The object to be serialized.
+     * @param dataType The type of the object used to find the appropriate serializer.
+     * @return A JsonElement representing the serialized data, or null if serialization fails.
+     */
+    @InternalAgentsApi
+    public fun serializeDataToJsonElement(data: Any?, dataType: KType): JsonElement? {
```
EugeneTheDev, line 45: It is used only in neighboring `trySerializeDataToJsonElement`. Let's make it private or, even better, inline the implementation, since it looks quite short

EugeneTheDev (approved):
Looks good, only a small suggestion

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/utils/SerializationUtil.kt
sdubov: You right. This is a general naming for API depending on the behaviour that you want from this invocation. Where `try` prefix in method name implies that this is a safe method that does not throw, while the one without `try` prefix does not handle common errors and can throw and you can handle this as you want on an invocations side. Both API was added here intentionally. Please let me know if you have some objections or if I miss anything from your suggestion.

EugeneTheDev: If this is the intention, I don't have any objections

PR #1016, opened 2025-10-23
Docs: Don't pass previous conversation in functional agent logic

SebastianAigner:
Koog keeps track of the ongoing conversation with the LLM, so giving the draft back to the LLM is redundant and implies that it has to be done manually, so removing it from the sample.


kpavlov (approved):
Make sense, thank you

PR #1032, opened 2025-10-26
KG-505. Move Debugger feature to the agents-core

sdubov:
[KG-505](https://youtrack.jetbrains.com/issue/KG-505/Move-Debugger-feature-to-agents-core). Move Debugger feature to the agents-core

- Move Debugger feature from agents-features module into agents-core module to get access to the feature from the core agents flow and be able to install a feature through an external agent configuration.
- Add a list of system features (Debugger) that can be installed through agent configuration parameters;
- Add agent logic to read a system property and set up Debugger feature with a provided config.

## Motivation and Context
Debugger feature is essential for a debugging capabilities. It would be necessary to have it in agents-core to enable the functionality upon request.

This is also allows to install debugger feature upon request by a plugins, e.g. "AI Debugger" that can set the Debugger feature via VM options instead of requesting this from a user.

## Breaking Changes
Yes

#### Type of the changes
- [ ] Documentation update
- [ ] Tests improvement
- [x] Refactoring

#### Checklist


agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/pipeline/AIAgentPipeline.kt
```diff
@@ -114,6 +118,19 @@
      */
     protected val registeredFeatures: MutableMap<AIAgentStorageKey<*>, RegisteredFeature> = mutableMapOf()
 
+    /**
+     * A set containing the keys of all registered features.
+     */
+    public val features: Set<AIAgentStorageKey<*>>
```
EugeneTheDev, line 125: This is basically a duplicate of a `registeredFeatures`, so it is doubtful that a separate property is needed. And it is used in tests only, so shouldn't it be removed?

EugeneTheDev (dismissed):
While I understand the reason why this change was made - we want to always ensure that `Debugger` is installed - I have some concerns regarding the implementation. Specifically, moving feature implementation to the core, as it kinda obscures the border between core and non-core. 

I have an alternative proposal. What if we go another way, and move some stuff out of the core, to make it lighter and leave truly "core" functionality in there. Namely, I'm suggesting to move `GraphAIAgent` and `FunctionalAIAgent` with `AgentService` to a separate module(s). It would give us a number of benefits and would work as follows:
1. Keeping the core lean without bringing specific implementation details. Only truly "core" functionality is left in core, and concrete implementations leave in dedicated modules.
2. You get to keep debugger with all its logic in a separate module, as before, making its development more isolated and easier. Faster compilation and no unnecesary clashes and conflicts with changes in core.
3. To still ensure that debugger is always installed in the agents we provide, we just add an explicit dependency in module with concrete agent implementations that we moved out of the core before on the debugger feature, and install it explicitly, like any other feature. We can safely do this now, since it's a separate module and there won't be circular dependencies.  

IMHO such solution would look cleaner and ensure better structure. Please let me know what you think, and let's maybe discuss it in more details.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/pipeline/AIAgentPipeline.kt
sdubov: True, this was added for the test purpose, but there is a reason for that. `registeredFeatures` is protected and store a feature and feature config. If I make this public, the config will be exposed to public as well. It means that you can modify the config outside of the `install(MyFeature) { ... }` lambda which I did not want to break the current design. 

It might be that we want to expose this, and, I can imagine, that this can be helpful in some cases. For example,  when you want to spread the control for the feature config. I would propose to discuss this topic separately.

Ololoshechkin (approved):
Let's just mark Debugger as InternalAgentApi with OptIn, or -- create a new OptIn annotation with a warning that it's experimental API and might be changed.

Maybe call it `ExperimentalKoogAPI` :)

Ololoshechkin (approved):

PR #1040, opened 2025-10-28
KG-507. Support multi-responses from LLM in the subgraphWithTask API

sdubov:
[KG-507](https://youtrack.jetbrains.com/issue/KG-507/Support-multi-response-from-LLM-in-subgraphWithTask-API). Support multi-responses from LLM in the subgraphWithTask API

- Unify the logic for subgraphWithTask API to be able to support multi-responses from LLM;
- Add an optional parameter 'runMode' to control the task execution flow and run tools flow;
- Deprecate old methods;
- Add tests.

## Motivation and Context
Unify the subgraphWithTask API

## Breaking Changes
No

---

#### Type of the changes
- [ ] Documentation update
- [x] Tests improvement
- [ ] Refactoring

#### Checklist

### DIFF
--- a/agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt
+++ b/agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt
@@ -1,5 +1,6 @@
 package ai.koog.agents.ext.agent
 
+import ai.koog.agents.core.agent.ToolCalls
 import ai.koog.agents.core.agent.context.AIAgentContext
 import ai.koog.agents.core.agent.context.AIAgentGraphContextBase
 import ai.koog.agents.core.agent.context.DetachedPromptExecutorAPI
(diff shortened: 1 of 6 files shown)


agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt
```diff
@@ -141,15 +144,17 @@ public object SubgraphWithTaskUtils {
  * @param toolSelectionStrategy The strategy used to select tools for the subgraph operations.
  * @param llmModel Optional language model to be used within the subgraph. Defaults to null.
  * @param llmParams Optional parameters for configuring the language model behavior. Defaults to null.
+ * @param runMode The mode in which tools are executed. Defaults to sequential execution.
  * @param defineTask A suspending lambda function that defines the task for the subgraph, taking the input as a parameter.
  * @return A delegate that represents the created subgraph, allowing input and output operations.
  */
-@OptIn(InternalAgentToolsApi::class)
+@OptIn(InternalAgentToolsApi::class, InternalAgentsApi::class)
```
devcrocod, line 166: This is not related to this pr. Why do we need to split it into multiple optIn instances instead of using a single `InternalKoogApi`?

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt
```diff
@@ -342,11 +367,46 @@ public inline fun <reified Input : Any> AIAgentSubgraphBuilderBase<*, *>.subgrap
  * @param finishTool A descriptor for the tool that determines the condition to finalize the subgraph's operation.
  * @param defineTask A suspending lambda that defines the main task of the subgraph, producing a task description based on the input.
  */
-@OptIn(InternalAgentToolsApi::class)
+@Deprecated(
+    message = "Use setupSubgraphWithTask API that receive a runMode parameter instead.",
```
devcrocod, line 386: we’ll remove this method, and only `setupSubgraphWithTaskMultiple` will remain, correct?

Are we planning a separate deprecation cycle for `setupSubgraphWithTaskMultiple`, replacing it with `setupSubgraphWithTask` that takes a `runMode` parameter to simplify the API?

devcrocod (approved):
lgtm

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt
sdubov: Thank you for catching this. I missed to rename the API here to `setupSubgraphWithTask` that takes a `runMode` parameter. Actually, we can even simple drop the old one (add a required `runMode` parameter without deprecation) since it is an internal API. I left it deprecated since it was the public method.

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt
sdubov: I think it has a bit different purpose. But I would agree that we do not need two. I think it makes sense to keep one in future. cc @EugeneTheDev.

PR #1041, opened 2025-10-28
Get rid of concurrent unsafe global `RegisteredBasicJsonSchemaGenerator` and introduce client-specific generator implementations

Ololoshechkin:
See KG-431

---

#### Type of the changes
- [ ] Documentation update
- [ ] Tests improvement
- [x] Refactoring

#### Checklist


EugeneTheDev (changes requested):
Unfortunately, it wouldn't work the way you implemented it. You introduced circular dependencies, which is inidicated by the compilation failure. Now you have:
```
prompt-executor-model
|--prompt-executor-clients
|   |--prompt-structure
|   |   |--prompt-executor-model <- loop
```

If I remember it correctly, this is similar to the problem I faced when implementing structured output originally, that's partially why I decided to use maps. I agree these are suboptimal and not safe, but it seems we have to approach this more carefully. Maybe we can adjust dependencies in a way that would make such a solution possible.

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/utils/TestUtils.kt
```diff
@@ -304,7 +304,7 @@ object TestUtils {
         @OptIn(InternalStructuredOutputApi::class)
         fun getStructure(model: LLModel) = JsonStructuredData.createJsonStructure<WeatherReport>(
             json = Json,
-            schemaGenerator = RegisteredStandardJsonSchemaGenerators[model.provider] ?: StandardJsonSchemaGenerator,
+            schemaGenerator = getExecutor(model).clientFor(model)?.standardJsonSchemaGeneratorFor(model) ?: StandardJsonSchemaGenerator,
```
anna239, line 307: here you have to pass `model` each time, maybe you could say that LLMClient knows its model already?

prompt/prompt-executor/prompt-executor-clients/build.gradle.kts
```diff
@@ -13,7 +14,9 @@ kotlin {
             dependencies {
                 api(project(":prompt:prompt-model"))
                 api(project(":agents:agents-tools"))
-                api(project(":prompt:prompt-executor:prompt-executor-model"))
+                api(project(":prompt:prompt-markdown"))
```
anna239, line 17: Looks strange that you need dependency to markdown for accessing the llm

prompt/prompt-executor/prompt-executor-llms/src/commonMain/kotlin/ai/koog/prompt/executor/llms/MultiLLMPromptExecutor.kt
```diff
@@ -220,6 +221,8 @@ public open class MultiLLMPromptExecutor(
         return client.moderate(prompt, model)
     }
 
+    override fun clientFor(model: LLModel): LLMClient? = llmClients[model.provider]
```
anna239, line 224: executeStreaming should also use clientFor, and also I believe the logic with fallback should also be placed here

anna239 (changes requested):

prompt/prompt-executor/prompt-executor-clients/build.gradle.kts
Ololoshechkin: It was here from the beginning :)

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/utils/TestUtils.kt
Ololoshechkin: I thought about that, but:
- Some LLM clients (ex: grazie) work with multiple providers
- Some other LLM clients work with one provider but proxy to different underlying providers based on the model (ex: Bedrock). It's actually current issue that Anthropic via Bedrock has "Bedrock" mentioned as provider -- we probably should change it to be like in grazie in the future. But I don't want to change it in this minor PR due to backwards compatibility.
- And the rest (Google, OpenAI, ...) are just single provider clients. But at the same time there's no guarantee they won't be providing more custom schema options for some models in the future.

So, the only way to support the general case even now is to provide per-model (or in the future -- when we refactor Bedrock model providers --- per-provider, where provider is still model.provider) schema getter

prompt/prompt-executor/prompt-executor-clients/build.gradle.kts
anna239: You mean it was added through prompt-executor-model ? And why is it api dependency?

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/utils/TestUtils.kt
anna239: Can't you say that a client always connects to one particular model and return a specific client from the promptExecutor? Otherwise I don's see why you need a separation between LLMClient and PromptExecutor

Ololoshechkin:
Closed in favour of clashing @tiginamaria 's PR

Issue #986, opened 2025-10-16
Unable to resume from latest checkpoint when using react strategy and crashing on a tool execution (java.util.NoSuchElementException: reasoning_step)


codependent:
The current implementation of the reActStrategy throws an exception when resuming from the latest checkpoint in the case of failure on a tool execution.

In order to reproduce:

1. Start a reactStrategy agent with persistence enabled
2. Force a failure on a tool execution
3. Resume the agent.
4. The latest checkpoint is restored, it'll pick up on "nodeId: nodeExecuteTool"
5. The tool executes correctly and then moves on to `nodeCallLLMReason`
6. The exception is thrown:

```
java.util.NoSuchElementException: Key ai.koog.agents.core.agent.entity.AIAgentStorageKey@65419120(name=reasoning_step) not found in storage
	at ai.koog.agents.core.agent.entity.AIAgentStorage.getValue(AIAgentStorage.kt:77) ~[agents-core-jvm-0.5.0.jar:0.5.0]
```

The problem lies here:

```
    val nodeCallLLMReason by node<ReceivedToolResult, Unit> { result ->
        val reasoningStep = storage.getValue(reasoningStepKey)
        llm.writeSession {
```

It is trying to access the reasoningStepKey from the agent context storage but that field is only set on the first node of the workflow:

```
    val reasoningStepKey = createStorageKey<Int>("reasoning_step")
    val nodeSetup by node<String, String> {
        storage.set(reasoningStepKey, 0)
        it
```

Since we are skipping that node while restoring it doesn't find the key and crashes.

**Expected Behavior**

Resuming from a checkpoint should restore the agent context storage (e.g., reasoning_step) so downstream nodes can access it without crashing, even if the initial setup node isn’t re-executed.

aozherelyeva:
Hi @codependent, thank you for the catch!

I also reported it to team's youtrack: [KG-489](https://youtrack.jetbrains.com/issue/KG-489) Unable to resume from latest checkpoint when using react strategy and crashing on a tool execution (java.util.NoSuchElementException: reasoning_step), feel free to follow the issue.

cc @Rizzen

PR #1049, opened 2025-10-30
Make ReceivedToolResult @Serializable

Rizzen:
## Breaking Changes

---

#### Type of the changes
- [ ] Documentation update
- [ ] Tests improvement
- [ ] Refactoring

#### Checklist

### DIFF
--- a/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/environment/GenericAgentEnvironment.kt
+++ b/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/environment/GenericAgentEnvironment.kt
@@ -17,7 +17,7 @@ import io.github.oshai.kotlinlogging.KLogger
 import kotlinx.coroutines.async
 import kotlinx.coroutines.awaitAll
 import kotlinx.coroutines.supervisorScope
-import kotlinx.serialization.json.JsonObject
+import kotlinx.serialization.json.JsonElement
 
 internal class GenericAgentEnvironment(
     private val agentId: String,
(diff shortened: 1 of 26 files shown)


agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/environment/GenericAgentEnvironment.kt
```diff
@@ -76,7 +77,7 @@ internal class GenericAgentEnvironment(
         toolName: String,
         agentId: String,
         message: String,
-        result: Any?
+        result: JsonObject?
```
EugeneTheDev: Is tool result always an object though? E.g. I can return just a `String` or an `Int` from the tool, it won't be an object, right?

agents/agents-tools/src/commonMain/kotlin/ai/koog/agents/core/tools/Tool.kt
```diff
@@ -122,6 +138,30 @@ public abstract class Tool<TArgs, TResult> {
      */
     public fun encodeArgs(args: TArgs): JsonObject = ToolJson.encodeToJsonElement(actualArgsSerializer, args).jsonObject
 
+    /**
+     * Encodes the given result into a JSON representation using the configured result serializer.
+     *
+     * @param result The result object of type TResult to be encoded.
+     * @return A JsonObject representing the encoded result.
+     */
+    public fun encodeResult(result: TResult): JsonObject =
```
EugeneTheDev: Why do you wrap primitive tool results here to return `JsonObject`, just like with tool arguments? For tool arguments it's important since arguments must always be a valid JSON object for an LLM to call the tool successfully. But there are no such limitations for tool results, we can return primitive types just like that, right? Shouldn't we return `JsonElement` here to avoid unnecessary wrapping?

EugeneTheDev (approved):
Looks very useful, thank you. I have only a single question regarding `JsonObject` vs `JsonElement` for tool result, please see the comment.

aozherelyeva:
Fixes #986

PR #1064, opened 2025-11-01
Pin Ollama version in CI checks to avoid accidental bugs and regressions from Ollama side

EugeneTheDev:
It seems that the latest at the time of writing version of Ollama introduced some performance degradations, making our Ollama tests fail on CI with timeouts. I pinned exact version in CI config to try to fix it and to avoid similar problems in the future.

### DIFF
--- a/.github/workflows/ollama-tests.yml
+++ b/.github/workflows/ollama-tests.yml
@@ -12,6 +12,9 @@ on:
   push:
     branches: [ "main", "develop" ]
 
+env:
+  OLLAMA_VERSION: "0.12.6"
+
 jobs:
   integration-tests:
     name: ${{ matrix.job-name }}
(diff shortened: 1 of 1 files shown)


kpavlov (approved):

.github/workflows/ollama-tests.yml
```diff
@@ -100,7 +100,7 @@ jobs:
       - name: Install and start Ollama
         run: |
           echo "=== Installing Ollama ==="
-          curl -fsSL https://ollama.com/install.sh | sh
+          curl -fsSL https://ollama.com/install.sh | OLLAMA_VERSION=0.12.6 sh
```
kpavlov: An [environment variable can be set per job or workflow](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-variables#defining-environment-variables-for-a-single-workflow)

kpavlov (approved):

aozherelyeva:
Thank you!

PR #1078, opened 2025-11-06
[agents] Make systemPrompt and temperature optional, set default temperature to `null` in AIAgent factory functions

### LINKED ISSUE #716: Do not require a system prompt
If I do not set a `systemPrompt` in my `AIAgent` I get:
```
Error executing node (name: nodeCallLLM): Error from Anthropic API: 400 Bad Request: {"type":"error","error":{"type":"invalid_request_error","message":"system: text content blocks must be non-empty"},"request_id":""}
```

A `systemPrompt` should not be required.

EugeneTheDev:
* Make these parameters optional to make factory methods more flexible. System prompt and temperature are not always required when calling LLMs.
* Set default temperature to `null` instead of `1.0` to default behavior for each model.

Fix #716

### DIFF
--- a/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
+++ b/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
@@ -262,8 +262,8 @@ public interface AIAgent<Input, Output> : Closeable {
          * @param strategy The strategy that defines the agent's workflow, defaulting to the [singleRunStrategy].
          * @param toolRegistry The set of tools available for the agent, defaulting to an empty registry.
          * @param id Unique identifier for the agent. Random UUID will be generated if set to null.
-         * @param systemPrompt The system-level prompt used as context for the agent, defaulting to an empty string.
-         * @param temperature The randomness or creativity of the model's responses, with valid values ranging typically from 0.0 to 1.0. Defaults to 1.0.
+         * @param systemPrompt Optional system prompt for the agent.
+         * @param temperature Optional model temperature, with valid values ranging typically from 0.0 to 1.0. Defaults to 0.0.
          * @param numberOfChoices The number of response choices to be generated, defaulting to 1.
          * @param maxIterations The maximum number of iterations the agent is allowed to perform, defaulting to 50.
          * @param installFeatures A function to configure additional features into the agent during initialization. Defaults to an empty configuration.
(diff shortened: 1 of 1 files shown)


Ololoshechkin (approved):

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
```diff
@@ -328,8 +328,8 @@ public interface AIAgent<Input, Output> : Closeable {
             toolRegistry: ToolRegistry = ToolRegistry.EMPTY,
             id: String? = null,
             clock: Clock = Clock.System,
-            systemPrompt: String = "",
-            temperature: Double = 1.0,
+            systemPrompt: String? = null,
+            temperature: Double? = 0.0,
```
tiginamaria: I'm a bit frustrated, what to set here, as, for example, for OpenAI temperature is in 0..2 range (and [1.0](https://platform.openai.com/docs/api-reference/responses/create#responses_create-temperature) is default so seems super reasonable then). But for Anthropic it's 0..1 so I can agree that it's good to set it lower, but not 0, and the default value is [1.0](https://docs.claude.com/en/api/messages#body-temperature)

tiginamaria (approved):
Thank you! I'd suggest not to tweak the temperature (as 1.0 seems default to me), but other than that, all is good!

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
EugeneTheDev: My idea is to be as "rigid" and "determenistic" by default as possible. Otherwise the agent might behave more "randomly" with default settings, which might frustrate new users. And in my understanding setting it to zero would just give us the most determenistic behavior possible

tiginamaria: Ok, could also be, just wanted to highlight that it's not a default value for models, and it confuses me

EugeneTheDev: Hmm, actually a good point. I guess setting it to null should give us default behavior for each model. Wdyt?

tiginamaria: Good option! Let's set null to have a default behavior for models and mention it in doc string

PR #1080, opened 2025-11-07
Integration tests refactoring

tiginamaria:
## Motivation and Context

I need to make a huge change in clients (migrate them to KoogHttp) for this [KG-527](https://youtrack.jetbrains.com/issue/KG-527/Support-models-endpoint-for-clients)
So I need to prepare integration tests to make it smooth.

Split TestUtils into parts (there were a mixed logic and >1000 lines of code): 
1. Move tools and structured output examples to separate dirs
2. Add a separate `APIKeys` util

Improve test cases: 
1. Add checks that tool was called
2. Added some corner cases for `anyOf` and `null` args

Next steps (Separate PR):
1. Rename `executor` tests to `client`, add separate tests for executor (maybe unit is enough)
2. (Maybe) split tests by clients to make them more interpretable

## Breaking Changes

---

#### Type of the changes
- [ ] Documentation update
- [x] Tests improvement
- [x] Refactoring

#### Checklist


integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/executor/ExecutorIntegrationTestBase.kt
```diff
@@ -176,271 +159,126 @@ abstract class ExecutorIntegrationTestBase {
...
+        withRetry(times = 3, testName = "integration_testToolWithNestedListParams[${model.id}]") {
+            val response = executor.execute(prompt, model, listOf(LotteryTool.descriptor))
             assertTrue(response.isNotEmpty(), "Response should not be empty")
+            assertResponseContainsToolCall(response, LotteryTool.name)
         }
     }
 
     open fun integration_testToolsWithNullParams(model: LLModel) = runTest(timeout = 300.seconds) {
         Models.assumeAvailable(model.provider)
+        assumeTrue(model.provider != LLMProvider.Anthropic, "Anthropic does not support anyOf")
```
zarechneva-a, line 266: By excluding this check, we miss an important test – the null param check for Anthropic models. Let's remove anyOf param from the SimplePriceCalculatorTool or create a separate tool for this check?

aozherelyeva (changes requested):
Thanks! Just one thing with the null param check, and we're good.

aozherelyeva:
I ran the heavy tests – the Single- and Multiple executor integration tests fail with the following error:
```
[SingleLLMPromptExecutorIntegrationTest.[13] LLModel(provider=MistralAI, id=mistral-medium-latest, capabilities=[Temperature, Completion, Tools, ToolChoice, Basic, Standard, Image, Document, MultipleChoices], contextLength=128000, maxOutputTokens=null)[jvm]: integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/executor/SingleLLMPromptExecutorIntegrationTest.kt#L167](https://github.com/JetBrains/koog/commit/a97bd09ce761fedea5890f927a2163e97d3a8c78#annotation_41386986980)
kotlinx.serialization.json.internal.JsonDecodingException: Serializer for subclass 'reference' is not found in the polymorphic scope of 'OpenAIContentPart'.
Check if class with serial name 'reference' exists and serializer is registered in a corresponding SerializersModule.
To be registered automatically, class 'reference' has to be '@serializable', and the base class 'OpenAIContentPart' has to be sealed and '@serializable'.
JSON input: {"type":"reference","reference_ids":["calculator"]}
```
Could you please take a look? Thank you!

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/executor/ExecutorIntegrationTestBase.kt
tiginamaria: `Null` parameter is useless without `AnyOf`, that's my I've changed the test, otherwise, you just have an argument which always null, not sure, if it has a use case. But if the test was done for a purpose, I can rewrite it back

tiginamaria:
The error you reported is our fault, and we need to suppress it until we fix this: [KG-535](https://youtrack.jetbrains.com/issue/KG-535) Error while parsing OpenAI-like content when the content is json array but not parts

I'll suppress it with the issue in comment

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/executor/ExecutorIntegrationTestBase.kt
zarechneva-a: I did it to cover all possible combos, but the combination with `AnyOf` makes sense to me. Let's leave it as is then and see if there're usecases with `null` parameter only :)

aozherelyeva (approved):

PR #1083, opened 2025-11-09
[agents] Fix primitive tool descriptor generation to use correct wrapper key. Improve primitive tool type handling in related places

EugeneTheDev:


aozherelyeva (approved):
Thanks!

PR #1084, opened 2025-11-09
(KG-481) [prompt] Fix StructureFixingParser to do the right number of retires, add tests

EugeneTheDev:
Fixes [KG-481](https://youtrack.jetbrains.com/issue/KG-481) by correcting fixing retries behavior. Also added tests for this fixing parser


sdubov (approved):
:+1:

PR #1093, opened 2025-11-10
KG-256 merge response message parts in ExecutorIntegrationTestBase.kt…

mltheuser:
YouTrack link: https://youtrack.jetbrains.com/issue/KG-256

## Motivation and Context
This PR fixes a failing test case in `SingleLLMPromptExecutorIntegrationTest` and `MultipleLLMPromptExecutorIntegrationTest` when running against `GoogleModels.Gemini2_5Pro`. The Gemini API can return a response containing multiple parts, which our client currently interprets as multiple Assistant Messages. The test's use of the `.single()` extension function on this list caused an `IllegalArgumentException`, preventing validation.

This change introduces a workaround directly within the test. Instead of attempting to force the list to a single element, the test now:
1.  Receives the `List<Message>` from the executor.
2.  Calls a new `response.toSingleMessage()` extension function (as proposed in the issue discussion) to merge the multiple messages into a single `Message.Assistant`.
3.  Performs the existing validation checks on this newly created single message.

This approach allows the test for Gemini 2.5 Pro to pass without altering the core behavior of the `GoogleLLMClient`. Following this fix, `Gemini2_5Pro` has replaced `Gemini2_5Flash` in the list of models for this test.

## Breaking Changes
None. This change is confined to a test class and does not affect the public API.

---

#### Type of the changes
- [ ] Documentation update
- [x] Tests improvement
- [ ] Refactoring

#### Checklist
- [x] Tests for the changes have been added (The change is a fix to an existing, failing test)

### DIFF
--- a/integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/executor/ExecutorIntegrationTestBase.kt
+++ b/integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/executor/ExecutorIntegrationTestBase.kt
@@ -44,6 +44,7 @@ import ai.koog.prompt.markdown.markdown
 import ai.koog.prompt.message.AttachmentContent
 import ai.koog.prompt.message.ContentPart
 import ai.koog.prompt.message.Message
+import ai.koog.prompt.message.ResponseMetaInfo
 import ai.koog.prompt.params.LLMParams
 import ai.koog.prompt.params.LLMParams.ToolChoice
 import ai.koog.prompt.streaming.StreamFrame
@@ -362,7 +363,7 @@ abstract class ExecutorIntegrationTestBase {
 
             withRetry {
                 try {
-                    with(getExecutor(model).execute(prompt, model).single()) {
+                    with(getExecutor(model).execute(prompt, model).toSingleMessage()) {
                         when (scenario) {
                             MarkdownTestScenario.MALFORMED_SYNTAX,
                             MarkdownTestScenario.MATH_NOTATION,
@@ -920,5 +921,15 @@ abstract class ExecutorIntegrationTestBase {
         withClue("Models list should not be empty") {
             getLLMClientForProvider(provider).models().shouldNotBeEmpty()
         }
+    }
+
+    private fun List<Message>.toSingleMessage(): Message.Assistant {
+        if (this.isEmpty()) {
+            return Message.Assistant(parts = emptyList(), metaInfo = ResponseMetaInfo.Empty)
+        }
+
+        val allParts = this.flatMap { it.parts }
+
+        return Message.Assistant(parts = allParts, metaInfo = ResponseMetaInfo.Empty)
     }
 }
--- a/integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/executor/ExecutorIntegrationTestBase.kt
+++ b/integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/executor/ExecutorIntegrationTestBase.kt
@@ -363,12 +363,13 @@ abstract class ExecutorIntegrationTestBase {
 
             withRetry {
                 try {
-                    with(getExecutor(model).execute(prompt, model).toSingleMessage()) {
+                    with(
+                        getExecutor(model).execute(prompt, model)
+                            .filterIsInstance<Message.Assistant>()
+                            .toSingleMessage()
+                    ) {
                         when (scenario) {
-                            MarkdownTestScenario.MALFORMED_SYNTAX,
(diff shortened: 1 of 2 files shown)


mltheuser:
If you would prefer a more general solution by merging parts inside of `processGoogleCandidate` in `GoogleLLMCLient.kt` let me know.

aozherelyeva (approved):
LGTM, thank you!

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/executor/ExecutorIntegrationTestBase.kt
```diff
@@ -921,4 +922,14 @@ abstract class ExecutorIntegrationTestBase {
             getLLMClientForProvider(provider).models().shouldNotBeEmpty()
         }
     }
+
+    private fun List<Message>.toSingleMessage(): Message.Assistant {
+        if (this.isEmpty()) {
```
skarpovdev: Consider removing the explicit `this` reference - in Kotlin, you can just use `isEmpty()` instead of `this.isEmpty()` for cleaner, more idiomatic code.

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/executor/ExecutorIntegrationTestBase.kt
```diff
@@ -921,4 +922,14 @@ abstract class ExecutorIntegrationTestBase {
...
+            return Message.Assistant(parts = emptyList(), metaInfo = ResponseMetaInfo.Empty)
+        }
+
+        val allParts = this.flatMap { it.parts }
```
skarpovdev: The empty check pattern (lines 927-929) could be simplified using Kotlin's `ifEmpty()` extension. This would make the code more concise:
```kotlin
val allParts = ifEmpty { emptyList() }.flatMap { it.parts }
```
This handles both the empty and non-empty cases more idiomatically.

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/executor/ExecutorIntegrationTestBase.kt
```diff
@@ -921,4 +922,14 @@ abstract class ExecutorIntegrationTestBase {
             getLLMClientForProvider(provider).models().shouldNotBeEmpty()
         }
     }
+
+    private fun List<Message>.toSingleMessage(): Message.Assistant {
```
skarpovdev: since `flatMap` already returns an empty list when called on empty list, this can be a one-liner:
`private fun List<Message>.toSingleMessage(): Message.Assistant = Message.Assistant(parts = flatMap { it.parts }, metaInfo = ResponseMetaInfo.Empty)`

skarpovdev (changes requested):

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/executor/ExecutorIntegrationTestBase.kt
mltheuser: Good catch. Resolved.

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/executor/ExecutorIntegrationTestBase.kt
```diff
@@ -921,4 +922,7 @@ abstract class ExecutorIntegrationTestBase {
             getLLMClientForProvider(provider).models().shouldNotBeEmpty()
         }
     }
+
+    private fun List<Message>.toSingleMessage(): Message.Assistant =
```
antoniibelyshev: If one of the messages is a tool call message, then I'm not sure if the result will be reasonable. I would suggest adding `filterIsInstance<Message.Assistance>`. It will help to avoid errors if someone wants to reuse this function

skarpovdev (approved):

antoniibelyshev (changes requested):
Looks great! I just left a suggestion to make the method more robust for the future uses

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/executor/ExecutorIntegrationTestBase.kt
mltheuser: Good idea. Changed the receiver type as well so prior filtering is required.

antoniibelyshev (approved):
Thank you! LGTM

PR #1104, opened 2025-11-11
KG-178. Set id and parent id for every event in the agent execution flow

sdubov:
[KG-178](https://youtrack.jetbrains.com/issue/KG-178).

- Add coroutine context elements for describing invocation hierarchy for an agent. Create elements for agent, strategy, node, and subgraph layers;
- Updated all context objects to pass id and parent id fields;
- Update all remote events with id and parent id;
- Update tests.

## Motivation and Context
Add `id` and `parentId` parameters for each agent event

## Breaking Changes
No. Add backward compatibility for events.

---

#### Type of the changes
- [ ] Documentation update
- [ ] Tests improvement
- [x] Refactoring

#### Checklist


agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/context/element/SubgraphInfoContextElement.kt
```diff
@@ -0,0 +1,31 @@
...
+    /**
+     * A companion object that serves as the key for the `SubgraphInfoContextElement` in a `CoroutineContext`.
+     * This key enables the retrieval of a `SubgraphInfoContextElement` instance stored within a coroutine's context.
+     *
+     * Implements `CoroutineContext.Key` to provide a type-safe mechanism for accessing the associated context element.
+     */
+    public companion object Key : CoroutineContext.Key<SubgraphInfoContextElement>
+
+    override val key: CoroutineContext.Key<*> = Key
+}
```
Rizzen, line 31: Well, I have huge concerns about that approach in general.

Basically, now we have two separate contexts in our agent - one explicit and huge `AIAgentContext` and recently we introduced another one - in coroutine context. And both of these contexts are growing and duplicating each other. 

I strongly suggest to redesign this, move away from coroutine context and unify agent context.

cc @Ololoshechkin @EugeneTheDev

Rizzen (changes requested):

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/context/element/SubgraphInfoContextElement.kt
Rizzen: We may not to address it in this PR, but I would discuss this refactoring sooner than later

sdubov: @Rizzen, I believe, there are different concepts between `AIAgentContext` instance and coroutine context. 
- `AIAgentContext` is a global context for an agent execution. It is usually shared between all entities that we have while running an agent. True, there are cases when we copy the context for parallel nodes execution, but instances are not "fully isolated" (as we copy some properties by reference).
- `Coroutine Context` has a different nature. It is used to store an isolated context elements. The scope is defined and localised per an operation (execute node/execute subgraph). It is useful, when you run nodes in parallel and you need to isolate information about current node and receive it in child events (llm calls, tool calls). it might be quite tricky to support in the `AIAgentContext`, because we need to implement the whole management logic ourselves and use some unique ids for each scope. Plus we need to implement this taking into account the concurrent nature of the agent.

I share your concerns about duplicating logic from one side, but I think that spawning `AIAgentContext` instances might be not the best approach here. I'm definitely open for discussion here and looking forward to find the clear solution.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/context/element/AIAgentContextElementBase.kt
```diff
@@ -0,0 +1,31 @@
...
+    /**
+     * A unique identifier associated with a context element.
+     *
+     * This identifier is used to identify the group of events within the coroutine
+     * context and facilitate tracking, retrieval, or associating additional metadata
+     * specific to the element during AI agent operations.
+     */
+    public val id: String
+
+    /**
```
EugeneTheDev, line 22: Same here

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/context/element/AIAgentContextElementBase.kt
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * Represents the base interface for context elements used in the execution environment
+ * of an AI agent. This interface extends `CoroutineContext.Element`, making it suitable
+ * for integration with coroutine contexts to provide execution metadata or contextual
+ * information for AI agent operations.
+ */
+public interface AIAgentContextElementBase : CoroutineContext.Element {
+
+    /**
```
EugeneTheDev, line 13: "slop-py" KDoc (sorry for the AI pun). The second, longer, paragraph is not needed, IMHO, it doesn't add any value. A single sentence "A unique identifier..." is clear enough

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/model/events/agentEvents.kt
```diff
@@ -10,16 +10,34 @@ import kotlinx.serialization.Serializable
...
     val runId: String,
     override val timestamp: Long = Clock.System.now().toEpochMilliseconds(),
-) : DefinedFeatureEvent()
+) : DefinedFeatureEvent() {
+
+    /**
+     * @deprecated. Creates an instance of [AgentStartingEvent].
+     * Note! Do not relay on [id] and [parentId] parameters with this constructor.
+     */
+    @Deprecated(
```
EugeneTheDev, line 32: Here and in other similar places: I don't think it's actively used by anyone, let's just delete it to avoid unnecessary deprecations? Or is it important for Debugger functionality?

agents/agents-features/agents-features-opentelemetry/src/jvmTest/kotlin/ai/koog/agents/features/opentelemetry/OpenTelemetrySpanAsserts.kt
```diff
@@ -7,179 +7,178 @@ import kotlin.test.assertEquals
 import kotlin.test.assertNotNull
 import kotlin.test.assertTrue
 
-internal object OpenTelemetrySpanAsserts {
+private val logger = KotlinLogging.logger { }
 
-    private val logger = KotlinLogging.logger { }
+internal fun assertMapsEqual(expected: Map<*, *>, actual: Map<*, *>, message: String = "") {
```
EugeneTheDev, line 12: I feel like Kotest might help here, to avoid implementing it by hand. Can you check?

prompt/prompt-executor/prompt-executor-model/src/commonMain/kotlin/ai/koog/prompt/executor/model/PromptExecutor.kt
```diff
@@ -42,15 +42,15 @@ public interface PromptExecutor : AutoCloseable {
      * @param tools A list of `ToolDescriptor` objects that define the tools available for the execution.
      * @return A flow emitting `StreamFrame` objects that represent the streaming output of the language model.
      */
-    public fun executeStreaming(
+    public suspend fun executeStreaming(
```
EugeneTheDev, line 45: Please be careful here. The reason it didn't have `suspend` is because it already returns `Flow`, which itself is a lazy stream. So a common pattern when working with coroutines is to avoid making functions that return flows suspendable. If you need suspendable operations, you can perform them in the flow builder

EugeneTheDev (changes requested):
I don't have any major comments. A couple of things:
1. Please check the comments I've left, especially about making `executeStreaming` suspendable 
2. Some KDocs for newly introduced entities are a bit too verbose. Please consider making these more concise. Also, I would suggest using square brackets instead of backticks when referring to symbols in KDocs. I know AI Assistant loves using backticks when generating KDocs, but when you use square brackets, it would actually generate a proper reference to the symbol mentioned. This improves navigation in the docs.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/context/element/SubgraphInfoContextElement.kt
EugeneTheDev: I think it's not that bad, actually. These components serve two separate concerns:
1. Agent context is an "operational" part. It is what the agent needs to actually function. It contains not only data, but also a lot of logic, can be mutable, etc.
2. These entities represent an "observational" part. They are immutable pieces of data representing a flow of events during the agent execution.  It is to be expected that they are somewhat duplicated in terms of data, that's kinda their purpose - to tell us what's going on inside the agent. However, this duplication is minimal, we are not duplicating the biggest part of the agent context - the logic itself. 

So I would say I don't see anything outright "bad". But these are my two cents. If you have other concerns I didn't cover above, please share

sdubov:
The PR was a draft implementation. It was split on separate PRs and merged into the `develop` branch with a slightly changed approaches. Closing this draft PR for now.

PR #1119, opened 2025-11-14
Optimize dependencies

kpavlov:
# Optimize dependencies and update project setup

## Motivation and Context

Reduce the library’s transitive footprint by making JDBC drivers and the connection pool `compileOnly`, so applications are no longer forced to pull in specific database drivers they may not use. It also stops the Spring Boot starter from implicitly choosing a Ktor client engine, requiring consumers to select the engine that best fits their environment and avoiding version and engine conflicts. The updated docs clarify these explicit dependency requirements, and the logging configuration helps keep test logs readable while still surfacing useful debug information.

- Switched optional JDBC drivers to `compileOnly` for reduced transitive dependencies.
- Added `logback.xml` to configure logging levels and prevent excessive logs.
- Exclude Ktor client engine from spring-boot transitive dependencies and updated documentation

## Breaking Changes

1. JDBC drivers dependencies are no longer included and should be explicitly configured by users.
2. Ktor client engine dependency should be configured by user when using Koog Spring Boot starter

---

#### Type of the changes
- [ ] Documentation update
- [x] Dependency update
- [ ] Tests improvement
- [ ] Refactoring

#### Checklist

### DIFF
--- a/agents/agents-features/agents-features-sql/build.gradle.kts
+++ b/agents/agents-features/agents-features-sql/build.gradle.kts
@@ -35,10 +35,10 @@ kotlin {
                 api(libs.exposed.jdbc)
                 api(libs.exposed.json)
                 api(libs.exposed.kotlin.datetime)
-                api(libs.postgresql)
-                api(libs.mysql)
-                api(libs.h2)
-                api(libs.sqlite)
+                compileOnly(libs.h2)
+                compileOnly(libs.mysql)
+                compileOnly(libs.postgresql)
+                compileOnly(libs.sqlite)
                 implementation(libs.hikaricp)
             }
         }
@@ -52,6 +52,11 @@ kotlin {
                 implementation(libs.testcontainers)
                 implementation(libs.testcontainers.postgresql)
                 implementation(libs.testcontainers.mysql)
+
+                runtimeOnly(libs.postgresql)
+                runtimeOnly(libs.mysql)
+                runtimeOnly(libs.h2)
+                runtimeOnly(libs.sqlite)
             }
         }
     }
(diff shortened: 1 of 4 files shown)


agents/agents-features/agents-features-sql/build.gradle.kts
```diff
@@ -35,10 +35,10 @@ kotlin {
                 api(libs.exposed.jdbc)
                 api(libs.exposed.json)
                 api(libs.exposed.kotlin.datetime)
-                api(libs.postgresql)
-                api(libs.mysql)
-                api(libs.h2)
-                api(libs.sqlite)
+                compileOnly(libs.h2)
```
EugeneTheDev, line 38: Shouldn't drivers be `runtimeOnly`? Because we don't use components from these dependencies directly, but these components are loaded at runtime based on the data source type. That's why AFAIK these are usually made `runtimeOnly`

EugeneTheDev (changes requested):

agents/agents-features/agents-features-sql/build.gradle.kts
kpavlov: @EugeneTheDev, no, when `runtimeOnly` is used, the dependency is present in the generated pom.xml (for publication):
```xml
<dependency>
  <groupId>com.h2database</groupId>
  <artifactId>h2</artifactId>
  <version>2.4.240</version>
  <scope>runtime</scope>
</dependency>
```

EugeneTheDev: Ok, so we're doing this to avoid adding these dependencies forcibly to the end users, aiming for "provided" behavior instead. Makes sense, thanks

EugeneTheDev (approved):

PR #906, opened 2025-10-01
[KTOR-8881] Ktor/Koog configuration in application.yaml gives error

nomisRev:
## Motivation and Context

Fixes [KTOR-8881](https://youtrack.jetbrains.com/issue/KTOR-8881) Ktor/Koog configuration in application.yaml gives error

## Breaking Changes

Breaks supporting top-level `timeout` block.

---

#### Type of the changes
- [ ] New feature (non-breaking change which adds functionality)
- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Tests improvement
- [ ] Refactoring

#### Checklist
- [x] The pull request has a description of the proposed change
- [x] I read the [Contributing Guidelines](https://github.com/JetBrains/koog/blob/main/CONTRIBUTING.md) before opening the pull request
- [x] The pull request uses **`develop`** as the base branch
- [x] Tests for the changes have been added
- [x] All new and existing tests passed

##### Additional steps for pull requests adding a new feature
- [ ] An issue describing the proposed change exists
- [ ] The pull request includes a link to the issue
- [ ] The change was discussed and approved in the issue
- [ ] Docs have been added / updated

Issue #807, opened 2025-09-16
Ktor/Koog configuration in application.yaml gives error


strindberg:
This issue was reported to the Ktor Youtrack (https://youtrack.jetbrains.com/issue/KTOR-8881), but I was asked to create it here instead.

Put the following configuration in an application running Ktor with the Koog plugin (and with ktor-server-config-yaml in dependencies):

```ktor:
application:
        modules:
            - se.example.ai.ApplicationKt.main

koog:
    openai:
        apikey: $OPEN_AI_API_KEY
        baseUrl: https://api.openai.com
```

Expected result:
The application starts normally.

Actual result:
The application throws the following errors:
"ERROR Application - Failed to read Koog configuration from application config
io.ktor.server.config.ApplicationConfigurationException: Path timeout not found."

TuAbE:
If you run into the following error:| ERROR | main | Application : Failed to read Koog configuration from application config
io.ktor.server.config.ApplicationConfigurationException: Path timeout not found.

You probably need to add a top-level timeout block in your YAML file and list each parameter (namely connectTimeout, readTimeout, and writeTimeout). You can leave their values empty; the built-in defaults will be used.

Here is the complete format:
``` yaml
koog:
  openai:
    apikey: xxx
    baseUrl: xxx
timeout:
  connectTimeout:
  readTimeout:
  writeTimeout:
```

This feels like a bug, since defaults are already provided and the YAML section should not be required.
The issue is likely in the:
 private fun KoogAgentsConfig.TimeoutConfiguration.configure(config: ApplicationConfig) function; 
The config parameter should probably be nullable.
If this is indeed a bug, could you please let me fix it? :)

@nomisRev 
Translated by a large language model; the original text is as follows.
如果你遇到了下面的问题：
| ERROR | main | Application : Failed to read Koog configuration from application config
io.ktor.server.config.ApplicationConfigurationException: Path timeout not found.
那可能是你需要再yaml的最顶层写一个timeout块，并且写上各个参数（具体为：  connectTimeout， readTimeout，writeTimeout，他们的值可以为空，默认加载代码中已经指定的默认值）。
以下为完整格式：
``` yaml
koog:
  openai:
    apikey: xxx
    baseUrl: xxx
timeout:
  connectTimeout:
  readTimeout:
  writeTimeout:
```
感觉这个问题可能是个bug？毕竟已经有默认值，且如果不填写不会使用yaml里面的配置信息。
[...]

PR #1122, opened 2025-11-14
Fix: [KTOR-8881] Fix the issue with reading the configuration file.

TuAbE:
[KTOR-8881] Determine whether the timeout is configured in application.yaml; if it is, read the configuration instead of exiting directly.(#807 OR #906)

When a timeout is defined for any LLM provider in application.yaml, use that value instead of the built-in default.

## Motivation and Context

Previously the application silently ignored per-provider timeout values that were declared in application.yaml; it always fell back to the hard-coded default.

Whenever a timeout entry exists under an LLM provider in application.yaml, that value is now honoured, so operators can tune timeouts without touching code or rebuilding the artifact.

## Breaking Changes
No breaking changes.

If timeout is absent from application.yaml the service behaves exactly as before (uses the built-in default).

Only users who already declared timeout but assumed it was inactive will notice the difference—those entries will now take effect, which is the intended behaviour.

---

#### Type of the changes
- [ ] Documentation update
- [ ] Tests improvement
- [ ] Refactoring

#### Checklist


nomisRev (approved):

aozherelyeva:
Hi @TuAbE, could you please rebase the branch onto the freshest develop? I think we're good to go right after it!

aozherelyeva:
Ah, I see, the fix is already in develop. Closing the PR then.

PR #1123, opened 2025-11-14
KG-504 Change flushTextBuilder to flush whenever textBuilder is not empty

### LINKED ISSUE #1071: Text ending with newline character is ignored in prompt
My text part was missing from my prompts. After playing around I found out that it happens when the string in the text part ends with a newline character "\n"

Here is a screenshot of evaluation showing the text missing in the prompt object.

Image

Here is the snippet in the evaluation window:
```kotlin
prompt("newlineTest") {

    user {
        text("this ends with newline\n")
    }
}
```

Removing the newline character fixes the issue:

Image

mltheuser:
Related to: [KG-504](https://youtrack.jetbrains.com/issue/KG-504)

## Motivation and Context
The `ContentPartsBuilder` would fail to create a `ContentPart` for text that ended with `\n` or was followed by a DSL call like `br()` or `newline()`. This was because the internal `flushTextBuilder` method checked `caret.offset != 0` to detect pending text. A trailing newline can reset the offset to 0, causing the check to fail even when the `textBuilder` is not empty.

This change proposes to check `textBuilder.isNotEmpty()` instead, ensuring all content is properly flushed.

## Breaking Changes

None. This is a non-breaking bug fix.

---

#### Type of the changes
- [ ] Documentation update
- [ ] Tests improvement
- [ ] Refactoring

#### Checklist

### DIFF
--- a/prompt/prompt-model/src/commonTest/kotlin/ai/koog/prompt/dsl/PromptBuilderTest.kt
+++ b/prompt/prompt-model/src/commonTest/kotlin/ai/koog/prompt/dsl/PromptBuilderTest.kt
@@ -918,4 +918,15 @@ class PromptBuilderTest {
         assertEquals(1, prompt.messages[1].parts.size, "Should have only text part")
         assertEquals(expectedText, prompt.messages[1].parts[0], "Should have same text")
     }
+
+    @Test
+    fun testUserMessageWithTrailingNewline() {
+        val prompt = Prompt.build("test") {
+            user {
+                +"Text\n"
+            }
+        }
+
+        assertEquals(1, prompt.messages[0].parts.size, "Prompt should have one message with one part")
+    }
 }
--- a/prompt/prompt-model/src/commonTest/kotlin/ai/koog/prompt/dsl/PromptBuilderTest.kt
+++ b/prompt/prompt-model/src/commonTest/kotlin/ai/koog/prompt/dsl/PromptBuilderTest.kt
@@ -5,6 +5,7 @@ import ai.koog.prompt.message.AttachmentContent
 import ai.koog.prompt.message.ContentPart
 import ai.koog.prompt.message.Message
 import ai.koog.prompt.text.text
+import io.kotest.matchers.equals.shouldBeEqual
 import kotlin.test.Ignore
 import kotlin.test.Test
 import kotlin.test.assertEquals
--- a/prompt/prompt-model/src/commonTest/kotlin/ai/koog/prompt/dsl/PromptBuilderTest.kt
+++ b/prompt/prompt-model/src/commonTest/kotlin/ai/koog/prompt/dsl/PromptBuilderTest.kt
@@ -927,6 +927,6 @@ class PromptBuilderTest {
             }
         }
 
-        assertEquals(1, prompt.messages[0].parts.size, "Prompt should have one message with one part")
+        prompt.messages[0].parts shouldBeEqual listOf(ContentPart.Text("Text\n"))
     }
 }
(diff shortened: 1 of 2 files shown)


prompt/prompt-model/src/commonTest/kotlin/ai/koog/prompt/dsl/PromptBuilderTest.kt
```diff
@@ -918,4 +918,15 @@ class PromptBuilderTest {
...
+
+    @Test
+    fun testUserMessageWithTrailingNewline() {
+        val prompt = Prompt.build("test") {
+            user {
+                +"Text\n"
+            }
+        }
+
+        assertEquals(1, prompt.messages[0].parts.size, "Prompt should have one message with one part")
```
kpavlov:
```suggestion
        prompt.messages[0].parts shouldBeEqual listOf(ContentPart.Text("Text\n"))
```
A stricter check for prompt message contents

kpavlov (approved):
Thank you, @mltheuser 
The test could be further improved, but thank you for the fix!

prompt/prompt-model/src/commonTest/kotlin/ai/koog/prompt/dsl/PromptBuilderTest.kt
mltheuser: Good change, thanks!

prompt/prompt-model/src/commonTest/kotlin/ai/koog/prompt/dsl/PromptBuilderTest.kt
```diff
@@ -918,4 +919,15 @@ class PromptBuilderTest {
         assertEquals(1, prompt.messages[1].parts.size, "Should have only text part")
         assertEquals(expectedText, prompt.messages[1].parts[0], "Should have same text")
     }
+
+    @Test
+    fun testUserMessageWithTrailingNewline() {
```
EugeneTheDev, line 924: Let's add the test from the issue? Because as far as I understand the problem was specifically with trailing newline functions, not with the newlines themselves.
```kt
user {
    markdown {
        +"Text"
        br()
    }
}
```

EugeneTheDev (changes requested):
Thank you, I only have a small comment

prompt/prompt-model/src/commonTest/kotlin/ai/koog/prompt/dsl/PromptBuilderTest.kt
mltheuser: Hi Eugene, thanks for the feedback. I see your point.

But adding a test specific to br() might be a bit redundant.

I made a conscious choice to test with a literal \n because my investigation showed the bug was not in the br() or newline() functions themselves, but in the underlying builder logic that flushes the text. Any content ending with a newline character would trigger it.

Since the current fix solves the problem for br(), newline(), and literal \ns, I believe the existing test is the most direct and fundamental way to verify the fix and prevent regressions. 

That said I am also fine with adding it or making a parametrised test that covers all the versions: [newline(), br()]. That might be useful to proof that more then one "\n" at the end is supported and those (although very minimal) functions work as expected.

Let me know what you think.

For reference:
```kotlin
/**
 * Appends a newline character to the underlying text builder.
 *
 * This method is primarily used to ensure proper formatting
 * and separation of text content in the [TextContentBuilder] class.
 */
public fun newline() {
    append("\n")
}

/**
 * Adds two consecutive newline characters to the text content.
 *
 * This function is typically used to create a double line break, ensuring
 * greater spacing between sections of the textual content.
 */
public fun br() {
    newline()
    newline()
}
```

EugeneTheDev: Thanks for elaborating, it's fine as it is then

EugeneTheDev (approved):

PR #1143, opened 2025-11-17
KG-185 Validate name and description in ToolDescriptor

mltheuser:
Related to: [KG-185](https://youtrack.jetbrains.com/issue/KG-185)

## Motivation and Context
This change addresses an issue where `ToolDescriptor` could be instantiated with an empty `description`. 

Following the discussion in the original issue, this PR introduces validation directly into the `ToolDescriptor` constructor to enforce the "fail-fast" principle. Now, any attempt to create a tool with a blank name or description will immediately throw an `IllegalArgumentException`, providing clear and early feedback to the developer.

This fix also includes updated KDoc to reflect the new constructor behavior and adds a new, focused test class (`ToolDescriptorValidationTest.kt`) to verify the validation logic.

## Breaking Changes
Yes, this is a breaking change.

Code that previously created `ToolDescriptor` instances with an empty or blank `description` will now throw an `IllegalArgumentException` upon object creation. While the likelihood of developers intentionally providing empty descriptions is low, any such existing code will now fail at runtime. 

---

#### Type of the changes
- [ ] Documentation update
- [ ] Tests improvement
- [ ] Refactoring

#### Checklist

### DIFF
--- a/agents/agents-tools/src/commonTest/kotlin/ai/koog/agents/core/tools/SerialToToolDescriptionTest.kt
+++ b/agents/agents-tools/src/commonTest/kotlin/ai/koog/agents/core/tools/SerialToToolDescriptionTest.kt
@@ -34,12 +34,15 @@ class SerialToToolDescriptionTest {
     )
 
     @Serializable
+    @LLMDescription("Color value")
     enum class Color { RED, GREEN, BLUE }
 
     @Serializable
+    @LLMDescription("Singleton")
     object Singleton
 
     @Serializable
+    @LLMDescription("Holder")
     data class FreeFormHolder(
         // contextual => free-form property mapping
         @Contextual val meta: Any? = null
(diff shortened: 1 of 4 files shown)


mltheuser:
Screenshot 2025-11-18 at 08 20 09

The workflow shows that `jvmTest` and more specific `DebuggerConfigTest` failed. Problem is I can not recreate this failure locally (have run that test and the full jvmTest task at least 10 times now). `DebuggerConfigTest` also does not define any tools which leads me to believe that this test class fails for reasons unrelated to the changes in this PR. Might be that the test class is flaky when running on the slow pipeline agents.

Could someone try to kick off the workflow one more time (I can't do that without making changes)?

agents/agents-tools/src/commonTest/kotlin/ai/koog/agents/core/tools/SerialToToolDescriptionTest.kt
```diff
@@ -34,12 +34,15 @@ class SerialToToolDescriptionTest {
...
 
     @Serializable
+    @LLMDescription("Singleton")
     object Singleton
 
     @Serializable
+    @LLMDescription("Holder")
     data class FreeFormHolder(
         // contextual => free-form property mapping
         @Contextual val meta: Any? = null
```
Rizzen, line 48: I strongly believe we should not require @LLMDescription for DTOs

Rizzen (changes requested):

agents/agents-tools/src/commonTest/kotlin/ai/koog/agents/core/tools/SerialToToolDescriptionTest.kt
Rizzen: Or maybe I understood it wrong, but I'm not sure why you added it here, could you please clarify?

mltheuser: Hi @Rizzen. Let me clarify why I added those `@LLMDescription` annotations.

Because of my change, creating a `ToolDescriptor` with an empty string for description (which is the default for asToolDescriptor()) now throws an `IllegalArgumentException`, breaking the tests for `Color`, `Singleton`, etc.

To fix the failing tests, I had two options:

1.  Modify every call to `asToolDescriptor()` in the test file to pass a hardcoded description, like `asToolDescriptor("color", toolDescription = "a color")`.
2.  Add the `@LLMDescription` annotation to the test DTOs.

I chose option #2 because it required fewer line changes and felt like a more surgical fix *specifically for the test setup*.

**To be clear, I completely agree with you: we should not require `@LLMDescription` for all DTOs.** The changes in this test file are not meant to establish a new convention. They are simply a pragmatic way to make the existing tests compatible with the new, stricter validation introduced in `ToolDescriptor`.

Note: I could have also made description have no default and be non nullable in asToolDescriptor() but because of the fallback logic where it tries to get the description from `@LLMDescription` I felt it was out of scope for this particular issue.

I hope this clears things up! I'm definitely open to other ideas on how to best adjust these tests if you feel there's a cleaner approach.

mltheuser: Hi @Rizzen, I wanted to bring this PR back to your attention. It's been a week and I would like to get this merged.

Rizzen: @mltheuser Thank you for follow up! We need at least 2 approves, so I'll add more people to review :) We'll take a look at your changes and I hope merge in following couple of days.

tiginamaria (commented):
Thank you for the contribution! I agree with the Google fast fail problem, but I don't think we need to require description for all the providers. For example for OpenAI the description in optional (https://platform.openai.com/docs/api-reference/responses/create#responses_create-tools-custom_tool-description). And the possible reason for it is tokens count, as if the name of the tool already says a lot, you don't want to add extra tokens for the description.

My proposal here could be adding the check into the GoogleLLMClient or keep it as is. WDYT?

mltheuser:
@tiginamaria What you are saying makes a lot of sense. I didn't really think about this feature critically until now, just saw the task and solved it. You are probably right, we should leave everything as is. As you said, there are valid scenarios where you don't want to be forced to supply a description because the provider allows it and you want to save tokens. Even if the provider does not allow it I am convinced most of them will parse and validate the tools before running the actual LLM and therefor reject the request before incurring cost. When they reject it the user should get a useful failure message now that ([KG-552](https://youtrack.jetbrains.com/issue/KG-552) LLM clients: better error reporting) has been merged.

I think it would be best for Koog to close this PR and issue [KG-185](https://youtrack.jetbrains.com/issue/KG-185).

PS: While you are already reading this, could I ask you to also check back on [KG-314](https://youtrack.jetbrains.com/issue/KG-314)? I asked for some guidance on where to put the integration tests for that feature 2 weeks ago and have been sitting on the solution since.

mltheuser:
Before you close it, we might want to keep the name validation though. This is also done provider side currently. WDYT?

PR #1151, opened 2025-11-19
Improved documentation on LLM providers

innateteniuk:
* Added a table that provides LLM providers with their capabilities and recommended use cases to help users choose a provider based on their needs.
* Added information that explains two ways to work with LLM providers in Koog (LLM clients and prompt executors).

## Motivation and Context

The LLM providers page should help users choose an appropriate provider based on their needs.
Detailed documentation about LLM clients and prompt executors stays in the [Prompts](https://docs.koog.ai/prompt-api/) documentation.

## Breaking Changes

---

#### Type of the changes
- [x] Documentation update
- [ ] Tests improvement
- [ ] Refactoring

#### Checklist


docs/docs/llm-providers.md
```diff
@@ -1,17 +1,43 @@
...
-- OpenAI
-- Anthropic
-- Google
-- DeepSeek
-- OpenRouter
-- Bedrock
-- Mistral
-- Ollama
+
+| <div style="width:115px">LLM provider</div> | Key capabilities                                                                                                                                                                                      | Choose for                                   |
```
kpavlov: Nit: consider representing it as a feature matrix, ([example](https://docs.langchain4j.dev/integrations/language-models/))

kpavlov (approved):
Thank you, @innateteniuk. It looks much nicer now

docs/docs/llm-providers.md
```diff
@@ -1,17 +1,43 @@
...
+  Each client implements the `LLMClient` interface, handling authentication, request formatting, and response parsing for the provider.
+  For details, see [Running prompts with LLM clients](prompt-api.md#running-prompts-with-llm-clients).
+
+- **Prompt executors** that manage the LLM client lifecycle and provide a unified interface across providers.
+  They handle failures, retries, and switching between providers.
+  You can either create your own executor or use a pre-defined prompt executor for a specific provider.
+  For details, see [Running prompts with prompt executors](prompt-api.md#running-prompts-with-prompt-executors).
+
+!!! tip
+    Both LLM clients and prompt executors support streaming, multiple choices, and content moderation.
```
devcrocod: That’s not correct.
As far as I remember, some providers don’t support multiple choices, for example, deepseek
and only openai and mistral support moderation (need double-check)

docs/docs/llm-providers.md
```diff
@@ -1,17 +1,43 @@
...
+| OpenRouter                                  | One integration with an access to multiple models from multiple providers. Capabilities differ by model and include tools, vision, structured output (JSON Schema), and other depending on the model. | Flexibility, provider comparison, unified API |
+| Bedrock                                     | AWS-native access to multiple models. Capabilities include tools with tool choice, image input, document input, embeddings, and content moderation, all model-specific.                               | AWS-native integrations, multi-provider access |
+| Mistral                                     | Structured output (JSON Schema), tools with tool choice, multiple choices, vision (model-specific), embeddings, content moderation.                                                                   | European hosting                             |
+| Ollama (local models)                       | Local execution, tools, structured output (JSON Schema), content moderation, vision (model-specific).                                                                                                         | Privacy, local development                   |
```
devcrocod: Could you please add Qwen (DashScope)?
https://api.koog.ai/prompt/prompt-executor/prompt-executor-clients/prompt-executor-dashscope-client/index.html

docs/docs/llm-providers.md
devcrocod: In the example you provided, the table doesn’t fully fit within the page layout, so you have to scroll. That’s something to keep in mind

Otherwise, any visual indicator like ✅ or ❌  is read faster and more easily

docs/docs/llm-providers.md
innateteniuk: Just to clarify, AFAIC, Qwen is a familiy of LLMs rather than a provider. 
Would it make sense to add it as  **DashScope (Qwen models)**?
Please correct me if I'm wrong.

docs/docs/llm-providers.md
innateteniuk: Then, probably this variant will work:

> Both LLM clients and prompt executors provide the functionality supported by the corresponding provider.
> For example, if a provider supports response streaming, multiple choices generation, or content moderation, both the LLM client and prompt executor for that provider will enable these features.
> Availability of specific functionalities depends on the LLM provider's capabilities.

This one now ephasizes that both LLM clients and prompt executors offer the same functionality, but it depends on  the specific provider. WDYT?

docs/docs/llm-providers.md
devcrocod: Oh, I wish I knew for sure myself 🙂

But yes, you’re right. Qwen is the name of the model family. At the same time, the API link and the previous naming follow the `dashscope`

And according to the [documentation](https://www.alibabacloud.com/help/en/model-studio/qwen-api-reference), everything is processed through Alibaba Cloud.

devcrocod: Our client is called `dashscope`

innateteniuk: > Would it make sense to add it as DashScope (Qwen models)?

Or maybe **Alibaba (Dashscope)**, since the provider is defined as `LLMProvider.Alibaba ` in the source code.

docs/docs/llm-providers.md
innateteniuk: We can try to represent it as 2 tables:
- The 1st table lists all the providers and the reasons for coosing each of them.
- The 2nd table is a feature matrix hidden in a collapsible block. It won’t fit the page layout, but since it is hidden by default, it shouldn’t be a big problem, IMHO.

WDYT?

Ololoshechkin (approved):

docs/docs/llm-providers.md
```diff
@@ -1,17 +1,46 @@
...
+
+## Working with providers
+
+Koog provides two ways to work with LLM providers:
+
+- **LLM clients**: that provide direct control over specific providers.
+  Each client implements the `LLMClient` interface, handling authentication, request formatting, and response parsing for the provider.
+  For details, see [Running prompts with LLM clients](prompt-api.md#running-prompts-with-llm-clients).
+
+- **Prompt executors** that manage the LLM client lifecycle and provide a unified interface across providers.
```
Ololoshechkin: and wrap multiple LLM clients and provide fallback LLM client

Ololoshechkin (approved):

docs/docs/llm-providers.md
azhebel: As a suggestion, use the first column for both the LLM provider name and the reason for choosing it (i.e. biggest value) and the other columns for features. But I am actually in favor of splitting, except make the first one a list, followed by a proper feature matrix table. And also I'd make names links to the corresponding LLM provider websites.

docs/docs/llm-providers.md
```diff
@@ -1,17 +1,46 @@
...
+| Ollama (local models)                       | Local execution, tools, structured output (JSON Schema), content moderation, vision (model-specific).                                                                                                 | Privacy, local development                                              |
 
 !!! note
-    For more information on using LLMs from these providers to run prompts and create AI agents, refer to [Prompt API](prompt-api.md#choosing-between-llm-clients-and-prompt-executors).
+    The list above shows each provider's capabilities.
+    For the exact capabilities of a specific model within a provider, refer to [Model capabilities](model-capabilities.md).
+
+## Working with providers
+
+Koog provides two ways to work with LLM providers:
```
azhebel: "provides ways to work with providers" and "provide direct control over specific providers"

How about this?

> Koog enables you to work with LLM providers on two separate levels:
> 
> - Use an **LLM client** to interact directly with one LLM provider...
> - Use a **prompt executor** for a higher-level abstraction that simplifies common use cases, handles LLM client lifecycle management, and can switch between multiple LLM clients...

It doesn't have to be this exact wording, but I'd avoid "provide ... providers"

docs/docs/llm-providers.md
```diff
@@ -1,17 +1,46 @@
...
+  Each client implements the `LLMClient` interface, handling authentication, request formatting, and response parsing for the provider.
+  For details, see [Running prompts with LLM clients](prompt-api.md#running-prompts-with-llm-clients).
+
+- **Prompt executors** that manage the LLM client lifecycle and provide a unified interface across providers.
+  They handle failures, retries, and switching between providers.
+  You can either create your own executor or use a pre-defined prompt executor for a specific provider.
+  For details, see [Running prompts with prompt executors](prompt-api.md#running-prompts-with-prompt-executors).
+
+!!! tip
+    Both LLM clients and prompt executors provide the functionality supported by the corresponding provider.
```
azhebel: "provide ... provider" again

And I am not sure I understand what this tip is about. A prompt executor includes and interacts with an LLM client. So isn't it obvious that they both will support whatever Koog supports from the corresponding LLM provider? BTW I think this may be more important: if I am not mistaken, Koog may not support everything that some provider's LLM can do.

azhebel (approved):

docs/docs/llm-providers.md
devcrocod: Yes, this sounds better and doesn’t create a false impression that a feature will be available if it isn’t provided by the provider itself

docs/docs/llm-providers.md
```diff
@@ -1,17 +1,64 @@
...
+| [Anthropic](https://www.anthropic.com/)                                                                                                                     | Long contexts and prompt caching.                                                                                              |
+| [Google](https://ai.google.dev/)                                                                                                                            | Multimodal processing (audio, video), large contexts.                                                                          |
+| [DeepSeek](https://www.deepseek.com/)                                                                                                                       | Cost-effective reasoning and coding.                                                                                           |
+| [OpenRouter](https://openrouter.ai/)                                                                                                                        | One integration with an access to multiple models from multiple providersfor flexibility, provider comparison, and unified API. |
```
innateteniuk:
```suggestion
| [OpenRouter](https://openrouter.ai/)                                                                                                                        | One integration with an access to multiple models from multiple providers for flexibility, provider comparison, and unified API. |
```

PR #1163, opened 2025-11-21
Fix SerializationUtils: Catch IllegalArgumentException as serialization exception

kpavlov:
# Catch IllegalArgumentException in SerializationUtils

- Update `SerializationUtils` to handle `IllegalArgumentException` as serialization exception, and add corresponding unit test
- Add mockK dependency to test-utils

## Motivation and Context

When, for some reason, `OpenTelemetry.Feature.getNodeData` can't serialize an object, it throws an exception and breaks strategy execution. It turned out that SerializationUtils.encodeDataToStringOrNull does not follow a contract: it should return null on serialization failures. `IllegalArgumentException` should be caught instead, as recommended in the `SerializationException` KDoc. 

<details>
<summary>Exception in OpenTelemetry serialization</summary>
<pre>
 "message": "Captured type parameter T from generic non-reified function. Such functionality cannot be supported because T is erased, either specify serializer explicitly or make calling function inline with reified T.",
        "stackTrace": "java.lang.IllegalArgumentException: Captured type parameter T from generic non-reified function. Such functionality cannot be supported because T is erased, either specify serializer explicitly or make calling function inline with reified T.\n\tat kotlinx.serialization.internal.Platform_commonKt.kclass(Platform.common.kt:106)\n\tat kotlinx.serialization.SerializersKt__SerializersKt.serializerByKTypeImpl$SerializersKt__SerializersKt(Serializers.kt:216)\n\tat kotlinx.serialization.SerializersKt__SerializersKt.serializer(Serializers.kt:152)\n\tat kotlinx.serialization.SerializersKt.serializer(Unknown Source)\n\tat kotlinx.serialization.SerializersKt__SerializersKt.serializersForParameters(Serializers.kt:295)\n\tat kotlinx.serialization.SerializersKt.serializersForParameters(Unknown Source)\n\tat kotlinx.serialization.SerializersKt__SerializersKt.serializerByKTypeImpl$SerializersKt__SerializersKt(Serializers.kt:253)\n\tat kotlinx.serialization.SerializersKt__SerializersKt.serializer(Serializers.kt:152)\n\tat kotlinx.serialization.SerializersKt.serializer(Unknown Source)\n\tat ai.koog.agents.core.utils.SerializationUtils.encodeDataToString(SerializationUtils.kt:73)\n\tat ai.koog.agents.core.utils.SerializationUtils.encodeDataToStringOrNull(SerializationUtils.kt:55)\n\tat ai.koog.agents.core.utils.SerializationUtils.encodeDataToStringOrNull$default(SerializationUtils.kt:52)\n\tat ai.koog.agents.core.utils.SerializationUtils.encodeDataToStringOrDefault(SerializationUtils.kt:40)\n\tat ai.koog.agents.core.utils.SerializationUtils.encodeDataToStringOrDefault$default(SerializationUtils.kt:33)\n\tat ai.koog.agents.features.opentelemetry.feature.OpenTelemetry$Feature.getNodeData(OpenTelemetry.kt:563)\n\tat ai.koog.agents.features.opentelemetry.feature.OpenTelemetry$Feature.access$getNodeData(OpenTelemetry.kt:43)\n\tat ai.koog.agents.features.opentelemetry.feature.OpenTelemetry$Feature$install$7.invokeSuspend(OpenTelemetry.kt:221)\n\tat ai.koog.agents.features.opentelemetry.feature.OpenTelemetry$Feature$install$7.invoke(OpenTelemetry.kt)\n\tat ai.koog.agents.features.opentelemetry.feature.OpenTelemetry$Feature$install$7.invoke(OpenTelemetry.kt)\n\tat ai.koog.agents.core.feature.pipeline.AIAgentGraphPipeline$interceptNodeExecutionCompleted$$inlined$createConditionalHandler$1.invokeSuspend(AIAgentPipeline.kt:1287)\n\tat ai.koog.agents.core.feature.pipeline
</pre>
</details>

See the branch https://github.com/kpavlov/koog-spring-boot-assistant/tree/upgrade-koog

## Breaking Changes

---

#### Type of the changes
- [ ] Documentation update
- [ ] Tests improvement
- [ ] Refactoring

#### Checklist


agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/utils/SerializationUtils.kt
```diff
@@ -53,7 +53,7 @@ public object SerializationUtils {
     public fun encodeDataToStringOrNull(data: Any?, dataType: KType, json: Json? = null): String? =
         try {
             encodeDataToString(data, dataType, json)
-        } catch (e: SerializationException) {
+        } catch (e: IllegalArgumentException) {
```
sdubov, line 59: @kpavlov, could you please clarify the original code that rise this error? Could you please add this to description? I think we need to investigate the root cause of the problem.

I would not expect that we handle base exception instead of serialisation errors here. By updating this exception to the base `IllegalArgumentException` class, you might miss the errors that are unrelated to serialisation. the method will safely return you default value and you will skip the original error. The `encodeToString` contract specify the exact exception of `SerializationException`. I think we need to follow the contract here.

sdubov (changes requested):

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/utils/SerializationUtils.kt
kpavlov: @sdubov, right now the contract of `encodeToSeringOrNull(...)` is broken: "Returns the serialized string if successful, or null if serialization fails."
The method is expected to succeed, if serialization fails by whatever reason.
The stack trace and the link to reproducer are attached to the description.

While I comprehend your rationale for silently capturing all issues, this method was intentionally designed that way. Violating this contract triggers a critical runtime error in version 0.5.3, rendering the Agent entirely unusable.

sdubov: > if serialization fails by whatever reason

This is not what I would expect here. I would like to catch the serialisation exceptions (everything that is produced by `encodeToString` call), but not "everything". I would not expect this method to return `null` in case of `CancellationException` for example.

Even in your fix, you catch the scope of `IllegalArgumentException` type which is not "whatever failure". Here I do not fully understand the reasons to extent the scope to this particular exception type.

@kpavlov, sorry, It might be I miss something, but what was the root cause that result to the `IllegalArgumentException` exception here?

kpavlov: @sdubov, I read the [SerializationException](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-serialization-exception/) api docs and it says:

> For general "catch-all" patterns around deserialization of potentially untrusted/invalid/corrupted data it is recommended to catch IllegalArgumentException type to avoid catching irrelevant to serialization errors such as OutOfMemoryError or domain-specific ones.

Also, the exception thrown on serializing is IllegalArgumentException (after 0.5.3)

cc: @Ololoshechkin

sdubov:
I performed some investigation in the root cause of this exception. There are several outcomes I made from the investigation:
1. The first issue here was related to the `ModeratedResults` serialisation error. The class is not annotated with @Serializable and failed when we try to send this as a node output through the `OpenTelemetry` feature (the logic makes an invocation of `SerialisationUtils` to get a string out from the output data). This issue is fixed by adding necessary parameters for default `json` object configuration and annotating the class as `@Serializable`.
2. The second issue is related to the Kotlin serialisation logic. In the example provided by @kpavlov, the node transfer the streaming `Flow<String>` type through nodes. This cause the failure since `Flow` type does not have a serialiser. The error was raised when reading the property: `json.serializersModule.serializer`. The `getSerializer()` method throws the `IllegalArgumentException` instead (please see the `getSerializer()` documentation for details). I've extended the original fix, made by @kpavlov, to other methods inside the `SerializationUtils` class and added tests to verify a particular types serialisation. This helps to test the serialisation logic against real cases scenarios.

agents/agents-core/src/commonTest/kotlin/ai/koog/agents/core/utils/SerializationUtilsTest.kt
```diff
@@ -262,6 +270,46 @@ class SerializationUtilsTest {
         )
     }
 
+    @Test
+    @JsName("encodeDataToStringForModeratedMessageType")
+    fun `encodeDataToString for ModeratedMessage type`() {
+        val message = Message.User(
```
kpavlov, line 276: Just a heads up, for isolation purposes, it’s usually a good idea to create an inner serializable class within the test. This test case, in particular, seems more like ModeratedMessageTest.

sdubov (approved):

PR #1166, opened 2025-11-22
Fix reasoning message handling in strategy

devcrocod:
related to #1153 

## Motivation and Context
- fix reasoning message in nodeLLMRequest
- fix conditions on multiple requests
- add onReasoningMessage and onMultipleReasoningMessage

## Breaking Changes
None

---

#### Type of the changes
- [ ] Documentation update
- [ ] Tests improvement
- [x] Refactoring

#### Checklist

#### Additional Context
To add tests need to modify and refactor mock executor

### DIFF
--- a/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/session/AIAgentLLMSession.kt
+++ b/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/session/AIAgentLLMSession.kt
@@ -145,14 +145,10 @@ public sealed class AIAgentLLMSession(
      * to ensure compatibility with the underlying LLM client's behavior. It then executes the request
      * and retrieves the response from the LLM.
      *
-     * When `skipReasoningMessage` is true, filters out reasoning messages and returns the first
-     * non-reasoning response.
-     *
-     * @param skipReasoningMessage If true, skips reasoning messages and returns first non-reasoning response (default: true).
      * @return The response message from the language model after executing the request, represented
      *         as a [Message.Response] instance.
      */
-    public open suspend fun requestLLMWithoutTools(skipReasoningMessage: Boolean = true): Message.Response {
+    public open suspend fun requestLLMWithoutTools(): Message.Response {
         validateSession()
         /*
             Not all LLM providers support tool list when tool choice is set to "none", so we are rewriting all tool messages to regular messages,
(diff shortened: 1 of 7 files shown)


BLannoo (changes requested):
If I understand it correctly the goal is to fix the issue described in https://github.com/JetBrains/koog/pull/1153

And while this does look a lot cleaner solution and I would be interested in finding one more in this direction, it currently does not seem to fix the core issue:

Which is that https://github.com/JetBrains/koog/blob/a5c36d580cae4ce04dbb4f95959c393110101c74/examples/code-agent/step-01-minimal-agent/src/main/kotlin/Main.kt#L29 fails when not providing ToolCalls.SEQUENTIAL as a parameter.

The error is still:
AIAgentStuckInTheNodeException: AI Agent has run into a problem: When executing agent graph, stuck in node nodeSendToolResult because output Reasoning(

Maybe merging in develop could help, because some other issue related to it was fixed recently, so the version in develop now works with: ToolCalls.SEQUENTIAL

But I am quite sure we would like it to work without that parameter too.

devcrocod:
@BLannoo 
> And while this does look a lot cleaner solution and I would be interested in finding one more in this direction, it currently does not seem to fix the core issue:

I’ve fixed the issue, the exception was actually occurring in another node

aozherelyeva:
Could you please cover changes with some basic unit-tests? Apart from this, the code looks good to me!

aozherelyeva (approved):

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/session/AIAgentLLMSession.kt
```diff
@@ -143,7 +158,7 @@ public sealed class AIAgentLLMSession(
             .withUpdatedParams { toolChoice = null }
             .let { preparePrompt(it, emptyList()) }
 
-        return executeSingle(promptWithDisabledTools, emptyList())
+        return executeMultiple(promptWithDisabledTools, emptyList()).first { it !is Message.Reasoning }
```
EugeneTheDev, line 161: nit: seems like you can simplify this method by reusing `requestLLMMultipleWithoutTools` and only applying `first`

EugeneTheDev (approved):

tiginamaria (approved):
Thank you! Looks good!

BLannoo (approved):
Validated the codeagent step01 and step02 to work after the fix (including the SINGLE_RUN_SEQUENTIAL version)

PR #1181, opened 2025-11-25
Support /models request for Google client

tiginamaria:
## Breaking Changes

---

#### Type of the changes
- [ ] Documentation update
- [ ] Tests improvement
- [ ] Refactoring

#### Checklist

### DIFF
--- a/prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClient.kt
+++ b/prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClient.kt
@@ -734,8 +734,8 @@ public open class GoogleLLMClient(
         val models = mutableListOf<String>()
 
         while ((response == null) || response.nextPageToken != null) {
-            val parameters = response?.let {
-                mapOf("pageToken" to it.nextPageToken.toString())
+            val parameters = response?.nextPageToken?.let {
+                mapOf("pageToken" to it)
             } ?: emptyMap()
             try {
                 response = httpClient.get(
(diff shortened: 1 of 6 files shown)


prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClient.kt
```diff
@@ -722,6 +723,37 @@ public open class GoogleLLMClient(
         throw UnsupportedOperationException("Moderation is not supported by Google API.")
     }
 
+    /**
+     * Retrieves a list of available language models supported by the Google LLM client.
+     * https://ai.google.dev/api/models#method:-models.list
+     *
+     * @return A list of strings, each representing a model identifier available for use.
+     */
+    public override suspend fun models(): List<String> {
```
devcrocod, line 732: It seems to me that this method is not very meaningful in its current form. 
Right now, we only return a list of model names, but it would probably make more sense to also return their configurations. wdyt?

devcrocod (approved):

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClient.kt
```diff
@@ -722,6 +723,37 @@ public open class GoogleLLMClient(
...
+        var response: GoogleModelsResponse? = null
+        val models = mutableListOf<String>()
+
+        while ((response == null) || response.nextPageToken != null) {
+            val parameters = response?.let {
+                mapOf("pageToken" to it.nextPageToken.toString())
```
devcrocod: I don’t understand why we’re calling toString() here when nextPageToken is already a String

devcrocod: It may happen that we end up returning "null" when response is not null but nextPageToken is null

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClient.kt
tiginamaria: I agree, let's add more info there and have a common for all the providers object for it. I'll do it in separate PR!

As for now, the requested use case was to provide en ability to check if the user has access to the specific model or not

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClient.kt
tiginamaria: Yeh, will replace it

PR #1187, opened 2025-11-26
Add LLM parameters for Google, Anthropic, Mistral, and Alibaba (DashScope)

marinkovicmarko:
Add LLM parameters for recently introduced providers: Google, Anthropic, Mistral, and Alibaba (DashScope).

---

#### Type of the changes
- [x] Documentation update
- [ ] Tests improvement
- [ ] Refactoring

#### Checklist


docs/docs/llm-parameters.md
```diff
@@ -113,12 +113,17 @@ For a list of default values for each parameter, see the corresponding LLM provi
 
 - [OpenAI Chat](https://platform.openai.com/docs/api-reference/chat/create)
 - [OpenAI Responses](https://platform.openai.com/docs/api-reference/responses/create)
+- [Google](https://ai.google.dev/api/generate-content#generationconfig)
+- [Anthropic](https://platform.claude.com/docs/en/api/messages/create)
+- [Mistral](https://docs.mistral.ai/api/#operation/chatCompletions)
 - [DeepSeek](https://api-docs.deepseek.com/api/create-chat-completion#request)
-- [OpenRouter](https://openrouter.ai/docs/api-reference/parameters)
+- [OpenRouter](https://openrouter.ai/docs/api/reference/parameters)
+- Alibaba ([DashScope](https://www.alibabacloud.com/help/en/model-studio/qwen-api-reference?spm=a2c63.p38356.help-menu-2400256.d_2_1_0.f57d482eBkrliN))
```
devcrocod: remove please scm parameter from url
`https://www.alibabacloud.com/help/en/model-studio/qwen-api-reference`

docs/docs/llm-parameters.md
```diff
@@ -230,42 +235,57 @@ val specificToolParams = LLMParams(
...
+| Parameter           | Providers                                                                                            | Type                                        | Description                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                    |
```
devcrocod: Maybe it would make sense to split the information into separate tabs for each provider?
I understand that a single table is simpler, since it allows you to group shared parameters and avoid duplication.
However, parameters like `reasoning`, `thinkingConfig`, and `thinking`, while similar, are specific to individual providers. There are also provider-specific parameters. All of this makes the table large and harder to read

marinkovicmarko: This suggestion makes sense, but with some content reuse between tabs to not have to duplicate the content as that would be too error-prone and challenging to maintain. I have a working solution, will push once it's ready.

docs/docs/llm-parameters.md
```diff
@@ -10,88 +10,95 @@ In Koog, the `LLMParams` class incorporates LLM parameters and provides a consis
 
 - When creating a prompt:
```
azhebel, line 11: WDYT about converting these list items to tabs as well?

azhebel (approved):

docs/docs/llm-parameters.md
marinkovicmarko: I think it is a good idea and a better option in terms of presentation than what we have now. However, for tabs to work, I'd have to indent the tab content in the source, and indentation will exclude the code from Knit checks 😞

Issue #1165, opened 2025-11-22
`singleRunStrategy(ToolCalls.PARALLEL)` and `singleRunStrategy(ToolCalls.SEQUENTIAL)` are broken


Stream29:
## To Reproduce

Add dependency.

```kotlin
implementation("ai.koog:koog-agents:0.5.3")
```

Run code.

```kotlin
import ai.koog.agents.core.agent.AIAgent
import ai.koog.agents.core.agent.ToolCalls
import ai.koog.agents.core.agent.singleRunStrategy
import ai.koog.agents.core.tools.ToolDescriptor
import ai.koog.agents.core.tools.ToolRegistry
import ai.koog.agents.ext.tool.SayToUser
import ai.koog.prompt.dsl.ModerationResult
import ai.koog.prompt.dsl.Prompt
import ai.koog.prompt.executor.clients.openai.OpenAIModels
import ai.koog.prompt.executor.model.PromptExecutor
import ai.koog.prompt.llm.LLModel
import ai.koog.prompt.message.ContentPart
import ai.koog.prompt.message.Message
import ai.koog.prompt.message.ResponseMetaInfo
import ai.koog.prompt.streaming.StreamFrame
import kotlinx.coroutines.flow.Flow
import kotlinx.datetime.Clock

suspend fun main() {
    val mockedExecutor = object : PromptExecutor {
        override suspend fun execute(
            prompt: Prompt,
            model: LLModel,
            tools: List<ToolDescriptor>
        ): List<Message.Response> = listOf(
            Message.Tool.Call(
                id = null,
                tool = "say_to_user",
                parts = listOf(ContentPart.Text("""{"message": "Hello"}""")),
                metaInfo = ResponseMetaInfo(timestamp = Clock.System.now()),
            )
        )

        override fun executeStreaming(
            prompt: Prompt,
            model: LLModel,
            tools: List<ToolDescriptor>
        ): Flow<StreamFrame> = TODO("Not yet implemented")


        override suspend fun moderate(
            prompt: Prompt,
            model: LLModel
        ): ModerationResult = TODO("Not yet implemented")

        override fun close() {}
    }
    val agent = AIAgent(
        promptExecutor = mockedExecutor,
        systemPrompt = "",
        temperature = 0.0,
        llmModel = OpenAIModels.Chat.GPT4o,
        strategy = singleRunStrategy(ToolCalls.SEQUENTIAL),
        toolRegistry = ToolRegistry {
            tool(SayToUser)
        },
[...]

PR #1192, opened 2025-11-26
Fix empty list condition check

devcrocod:
Adds `isNotEmpty` condition to ensure lists are non-empty before call condition

## Motivation and Context
fix `onMultipleToolResults` and `onMultipleAssistantMessages`

## Breaking Changes
None

---

#### Type of the changes
- [ ] Documentation update
- [ ] Tests improvement
- [ ] Refactoring

#### Checklist


tiginamaria (approved):
Thank you for the fix!

tiginamaria:
It would also be nice to cover it with tests, but as it is asap quick fix, it's ok to do it after

devcrocod:
> It would also be nice to cover it with tests, but as it is asap quick fix, it's ok to do it after

true
I started writing tests and realized that our mock executor doesn’t support our response types at all. At the moment, it only handles strings and tool calls. Since we now have quite a few different response types, plus new concepts like content parts and structured outputs, it seems we need a significant overhaul of our mock. This can’t be fixed by simply adding a couple of new methods, and need separate pr for this

Stream29:
This PR also fixes: #1165 .

PR #1194, opened 2025-11-26
Preserve `LLMParams` fields after `Prompt.withUpdatedParams`

yakivy:
## Motivation and Context
Preserves all `LLMParams` fields that were not updated during `Prompt.withUpdatedParams`, including subtype of LLMParams instance. The biggest impact of this change is on `PromptExecutor.executeStructured`, where it erases all the properties that were not updated by the method for Native structured completion requests.

## Breaking Changes
No breaking changes.

## Solution
I used `.copy` method from `LLMParams` that is properly overwritten by its children instead of `LLMParamsUpdateContext.toParams`

---

#### Type of the changes
- [ ] Documentation update
- [ ] Tests improvement
- [x] Refactoring

#### Checklist
- [x] All new and existing tests passed (checked jvmTest for `prompt.*` package, agent tests are red on my machine even before the PR)

### DIFF
--- a/prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/dsl/Prompt.kt
+++ b/prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/dsl/Prompt.kt
@@ -176,6 +176,20 @@ public data class Prompt @JvmOverloads constructor(
             toolChoice = toolChoice,
             user = user
         )
+
+        /**
+         * Updates the given [LLMParams] instance with the current context's configuration.
+         *
+         * @param params The original instance of [LLMParams] to which the updates are applied.
+         * @return A new instance of [LLMParams] with updated values.
+         */
+        public fun applyToParams(params: LLMParams): LLMParams = params.copy(
+            temperature = temperature,
+            speculation = speculation,
+            schema = schema,
+            toolChoice = toolChoice,
+            user = user,
+        )
     }
 
     /**
@@ -189,5 +203,5 @@ public data class Prompt @JvmOverloads constructor(
      * @return A new `Prompt` instance with the updated parameters.
      */
     public fun withUpdatedParams(update: LLMParamsUpdateContext.() -> Unit): Prompt =
-        copy(params = LLMParamsUpdateContext(params).apply { update() }.toParams())
+        copy(params = LLMParamsUpdateContext(params).apply { update() }.applyToParams(params))
 }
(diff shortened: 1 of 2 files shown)


yakivy:
@Ololoshechkin @devcrocod could you take a look please?

yakivy:
@kpavlov hey, it looks like failed test is a bit flaky, should we restart or do you want me to dig deeper into the problem?

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/dsl/Prompt.kt
```diff
@@ -176,6 +176,20 @@ public data class Prompt @JvmOverloads constructor(
...
             user = user
         )
+
+        /**
+         * Updates the given [LLMParams] instance with the current context's configuration.
+         *
+         * @param params The original instance of [LLMParams] to which the updates are applied.
+         * @return A new instance of [LLMParams] with updated values.
+         */
+        public fun applyToParams(params: LLMParams): LLMParams = params.copy(
```
EugeneTheDev: Let's make this private, this is implementation detail

EugeneTheDev (changes requested):
Hi, sorry for the wait, thank you for the fix. 

A general note: this whole approach with `LLMParamsUpdateContext` wouldn't work with custom `LLMParams` child classes that some LLM clients provide (`OpenAIResponsesParams`, `AnthropicParams`, etc.). These can provide certain LLM provider specific params, and these will also be lost when `withUpdatedParams` is called, even with your fix. That's because it creates a new instance of the base `LLMParams`, not that specific child class. This is a more fundamental problem, because `LLMParamsUpdateContext` comes from the times when we didn't have custom LLMParams classes. So we will need to think of a proper way to make it respect custom params. But for now, let's merge your fix to at least preserve core LLM params. Only a small comment, though, please take a look

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/dsl/Prompt.kt
yakivy: I'm not sure I understand the proposal, as I will not be able to use it in `withUpdatedParams` if it's private

EugeneTheDev: Ah, my bad, missed it. Let's mark it `internal`

yakivy: Sure, no worries, done. Should I squash?

EugeneTheDev: Not necessary, we merge usually with "Squash & merge", so GitHub does it automatically

EugeneTheDev (approved):

PR #1197, opened 2025-11-28
[examples] use Markdown renderer for agent response text in demo app

joreilly:
Uses https://github.com/mikepenz/multiplatform-markdown-renderer for displaying agent response text

Screenshot 2025-11-28 at 10 59 04

cc @mikepenz :)


EugeneTheDev (approved):
Thank you, it definitely looks nicer now

PR #1198, opened 2025-11-28
KG-545 Fix `requestLLMOnlyCallingTools` ignoring tool calls after reasoning message

mltheuser:
## Motivation and Context
Related to [KG-545](https://youtrack.jetbrains.com/issue/KG-545).

Currently, `requestLLMOnlyCallingTools` relies on `executeSingle`, which returns the first message received from the LLM. When using models that output Chain of Thought or "Thinking" blocks (e.g., Nova, Claude) prior to calling a tool, the response sequence is often `[Message.Assistant(Thinking), Message.Tool.Call]`.

As a result:
1. The method returns the "Thinking" text message instead of the expected Tool Call.
2. The actual Tool Call is discarded and never executed or saved to history.

**Changes:**
*   Introduced `requestLLMMultipleOnlyCallingTools()` in `AIAgentLLMSession` to allow retrieving the full list of messages while enforcing `ToolChoice.Required`.
*   Updated `requestLLMOnlyCallingTools` to use this new method. It now persists **all** messages (preserving the reasoning context in the session history) but filters the return value to ensure the caller receives the `Message.Tool.Call`.

## Breaking Changes
None. This is a behavioral fix to ensure the method contract (returning a tool call) is honored when the LLM is "chatty" or provides reasoning.

---

#### Type of the changes
- [ ] Documentation update
- [ ] Tests improvement
- [ ] Refactoring

#### Checklist

### DIFF
--- a/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/session/AIAgentLLMSession.kt
+++ b/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/session/AIAgentLLMSession.kt
@@ -146,17 +146,42 @@ public sealed class AIAgentLLMSession(
     /**
      * Sends a request to the language model that enforces the usage of tools and retrieves the response.
      *
-     * This method updates the session's prompt configuration to mark tool usage as required before
-     * executing the request. Additionally, it ensures the session is active before proceeding.
+     * This method:
+     * 1. Validates that the session is active.
+     * 2. Updates the prompt configuration to mark tool usage as required (`ToolChoice.Required`).
+     * 3. Retrieves all generated messages (including potential Chain of Thought/Reasoning blocks).
+     * 4. Filters the result to return the first [Message.Tool.Call].
      *
-     * @return The response from the language model after executing the request with enforced tool usage.
+     * If no tool call is found (e.g., the model refused or asked a question), it falls back to the first message.
+     *
+     * @return The tool call response (or first response) from the language model.
      */
     public open suspend fun requestLLMOnlyCallingTools(): Message.Response {
+        validateSession()
+        // We use the multiple-response method to ensure we capture all context (e.g. thinking)
+        // even though we only return the specific tool call.
+        val responses = requestLLMMultipleOnlyCallingTools()
+        return responses.firstOrNull { it is Message.Tool.Call } ?: responses.first()
+    }
+
+    /**
+     * Sends a request to the language model that enforces the usage of tools and retrieves all responses.
+     *
+     * This is useful when the LLM returns multiple messages, such as a "thinking" block followed by tool calls,
+     * or multiple parallel tool calls.
+     *
+     * This method:
+     * 1. Validates that the session is active.
+     * 2. Updates the prompt configuration to mark tool usage as required (`ToolChoice.Required`).
+     *
+     * @return A list of responses from the language model.
+     */
+    public open suspend fun requestLLMMultipleOnlyCallingTools(): List<Message.Response> {
         validateSession()
         val promptWithOnlyCallingTools = prompt.withUpdatedParams {
             toolChoice = LLMParams.ToolChoice.Required
         }
-        return executeSingle(promptWithOnlyCallingTools, tools)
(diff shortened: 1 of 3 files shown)


agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/session/AIAgentLLMWriteSession.kt
```diff
@@ -392,13 +392,19 @@ public class AIAgentLLMWriteSession internal constructor(
...
-     * @return The response received from the Language Learning Model (LLM).
+     * Crucially, this method appends **all** received messages to the prompt history to preserve context.
+     *
+     * @return A list of responses received from the Language Model (LLM).
      */
-    override suspend fun requestLLMOnlyCallingTools(): Message.Response {
-        return super.requestLLMOnlyCallingTools().also { response -> appendPrompt { message(response) } }
+    override suspend fun requestLLMMultipleOnlyCallingTools(): List<Message.Response> {
+        return super.requestLLMMultipleOnlyCallingTools().also { responses ->
+            appendPrompt {
```
EugeneTheDev: nit: You can simplify it 
```kotlin
appendPrompt {
    messages(responses)
}

```

EugeneTheDev (approved):
Looks good, thank you. Please check a small comment

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/session/AIAgentLLMSession.kt
```diff
@@ -164,17 +164,42 @@ public sealed class AIAgentLLMSession(
...
-     * This method updates the session's prompt configuration to mark tool usage as required before
-     * executing the request. Additionally, it ensures the session is active before proceeding.
+     * This method:
+     * 1. Validates that the session is active.
+     * 2. Updates the prompt configuration to mark tool usage as required (`ToolChoice.Required`).
+     * 3. Retrieves all generated messages (including potential Chain of Thought/Reasoning blocks).
+     * 4. Filters the result to return the first [Message.Tool.Call].
      *
-     * @return The response from the language model after executing the request with enforced tool usage.
+     * If no tool call is found (e.g., the model refused or asked a question), it falls back to the first message.
```
kpavlov: Let’s also make sure this use case gets tested.

However, I’m not sure we should have a fallback; perhaps we should just fail.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/session/AIAgentLLMSession.kt
```diff
@@ -164,17 +164,42 @@ public sealed class AIAgentLLMSession(
...
+     *
+     * @return A list of responses from the language model.
+     */
+    public open suspend fun requestLLMMultipleOnlyCallingTools(): List<Message.Response> {
         validateSession()
         val promptWithOnlyCallingTools = prompt.withUpdatedParams {
             toolChoice = LLMParams.ToolChoice.Required
         }
-        return executeSingle(promptWithOnlyCallingTools, tools)
+        return executeMultiple(promptWithOnlyCallingTools, tools)
```
kpavlov, line 185: 👍🏻

kpavlov (commented):
Thank you, @mltheuser
I’m on board with the overall change.

Just one thing: should we handle the case where no tool call is found by failing? I’d lean towards that, but I’m not set in stone.

I’m happy to merge it, as we have tests in place to ensure we’re covering all edge cases.

`requestLLMOnlyCallingTools()`:
1. "If no tool call is found (e.g., the model refused or asked a question), it falls back to the first message."
2. What would happen when the model returns multiple tools

mltheuser:
Thanks for the review!

@EugeneTheDev I've applied the simplification using `messages()` as suggested — much cleaner.

@kpavlov Good call on the edge cases. I've updated `requestLLMOnlyCallingTools` to fail explicitly if no tool call is returned, rather than falling back to the text message. I also added new tests to cover that scenario and the multiple tool calls case (it now correctly returns the first one while preserving all context).

PR #1220, opened 2025-12-03
Allow agents to be reusable with context

### LINKED ISSUE #218: Support Chat-style agents
[Disclaimer: This is an opinionated proposal that only reflects my views.]

#### Problem
To support Chat-style agents, you need:
1. Either allow calling the same agent multiple times and keep it's state between runs
2. Either use user-input-gathering nodes and allow to have cyclic graph definitions

I think Koog should allow both.

#### Proposal

- Rearrange AgentConfig and Agent to keep it's state to support case 1.
- Make maxAgentInteractions a part of the agent state, and make it temporarily overridable in a sub-graph, to support case 2.
- Lift the path-check between nodeStart and nodeFinish?

#### Notes

Currently, here is what I am doing to support this:

```kotlin
class ChatAgent(
    private val initialPrompt: Prompt,
    private val model: LLModel,
    val maxAgentIterations: Int = 50,
    val promptExecutor: PromptExecutor,
    private val strategy: AIAgentStrategy,
    private val initialToolRegistry: ToolRegistry = ToolRegistry.Companion.EMPTY,
    private val installFeatures: AIAgent.FeatureContext.() -> Unit = {}
) {
    var toolRegistry: ToolRegistry = initialToolRegistry
    var prompt: Prompt = initialPrompt

    fun resetPrompt() {
        prompt = initialPrompt
    }

    fun resetToolRegistry() {
        toolRegistry = initialToolRegistry
    }

    fun updatePrompt(newPrompt: Prompt) {
        prompt = newPrompt
    }

    suspend fun runAndGetResult(input: String): String? {
        val agent = buildAgent()
        val result = agent.runAndGetResult(input)
        agent.agentConfig.prompt
        return result
    }

    private fun buildAgent(): AIAgent {
        val agentConfig = AIAgentConfig(
            prompt = prompt,
            model = model,
            maxAgentIterations = maxAgentIterations,
        )

        return AIAgent(
            promptExecutor = promptExecutor,
            strategy = strategy,
            agentConfig = agentConfig,
            toolRegistry = toolRegistry,
            installFeatures = {
                install(PromptCollector) {
                    collectPrompt = ::updatePrompt
                }
                installFeatures()
            },
        )
    }
}

private class PromptCollector {
    companion object Feature : AIAgentFeature<PromptCollectorConfig, PromptCollector> {
        override val key: AIAgentStorageKey<PromptCollector> =
            AIAgentStorageKey("agents-features-prompt-collector")

        override fun createInitialConfig(): PromptCollectorConfig = PromptCollectorConfig()

        override fun install(config: PromptCollectorConfig, pipeline: AIAgentPipeline) {
            val featureImpl = PromptCollector()

            pipeline.interceptAfterNode(this, featureImpl) { node, context, input, output ->
                config.collectPrompt(context.llm.readSession { prompt })
            }
        }
    }
}

private class PromptCollectorConfig : FeatureConfig() {
    var collectPrompt: (Prompt) -> Unit = {}
}
```

This has several drawbacks (non-exhaustive):
- recreating the agent everytime
- forced to use `interceptAfterNode` because all handlers do not give you the `AgentContext`

### LINKED ISSUE #1217: Reusing AIAgent sessions
Calling AIAgent.run() twice sequentially on the same instance fails the second time with:

```
Agent was already started. Please use AIAgentService.createAgentAndRun(agentInput) to run an agent multiple times.
```

`AIAgentService.createAgentAndRun(agentInput)` doesn't make sense because you basically lose the context, which ruins the point of using an agent in the first place.

The `AIAgent` [initializer](https://github.com/JetBrains/koog/blob/f247a0fcc822ab8da89bd266a48063ed324e16c2/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt#L202) implements StatefulSingleUseAIAgent pattern.

Neither documentation nor the codebase provides examples of an agent configured for multiple runs. I've also noticed that this behavior of single use isn't mentioned either.

```kt
val agent = AIAgent(
    promptExecutor = simpleGoogleAIExecutor("apiKey"),
    llmModel = GoogleModels.Gemini2_5Flash,
)

println(agent.run("Hello"))
println(agent.run("World")) // Fails here
```

Need guidance on how to configure AIAgent for reuse for scenarios like conversational context.

### LINKED ISSUE #1268: Using Agent as normal chat
I have a simple use case which is a normal chat bot and in the same time is able to call tools.
I tried to create an agent but i noticed that it only allows calling `run` once which is weird.
 I expect multiple messages between the user and the llm and in between the agent might decide to call tools. This is a very common (I think even the most used) use case for an agent but it's not supported.

PranavPurwar:
## Motivation and Context
Currently, the default agent types are designed to be used only once; they throws an `IllegalStateException` if you try to run it a second time. This prevents the creation of  conversational, or manual feedback based agents.

This change introduces a flag (`enforceSingleRun`) that allows the agent to be **reused** for multiple interactions. When reused, the agent keeps its context history and state across runs.

This enables the development of **conversational agents** that can remember what happened in the previous step. 

The change also updates the core run logic to ensure the agent's initial setup is only done once, guaranteeing the reuse of the context object for subsequent calls while correctly tracking the session.

Closes #1217 

Example usage:
```kt
val agent = AIAgent(
    promptExecutor = simpleGoogleAIExecutor(API_KEY),
    agentConfig = AIAgentConfig(
        prompt = Prompt.build("Simple Prompt") {
        }, model = GoogleModels.Gemini2_5Flash,
        enforceSingleRun = false
    )
)

println(agent.run("What is Koog?")) // Koog is a...
println(agent.run("What was my previous message?")) // "What is Koog?"
```

## Breaking Changes
No. The default behavior of `StatefulAIAgent` remains **single-use** (by assuming `enforceSingleRun = true` if not specified), replicating current behavior and throwing the `IllegalStateException` on the second run. People wishing to enable reuse must explicitly set the new parameter to `false`.

---

#### Type of the changes
- [x] Documentation update (Documentation will be updated in a follow-up PR, but the capability is included here)
- [x] Tests improvement
- [ ] Refactoring

#### Checklist

- [x] Docs have been added / updated (Marked for update, likely in a separate PR)

### DIFF
--- a/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
+++ b/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
@@ -266,6 +266,7 @@ public interface AIAgent<Input, Output> : Closeable {
          * @param temperature Optional model temperature, with valid values ranging typically from 0.0 to 1.0.
          * @param numberOfChoices The number of response choices to be generated, defaulting to 1.
          * @param maxIterations The maximum number of iterations the agent is allowed to perform, defaulting to 50.
+         * @param enforceSingleRun A flag indicating whether to enforce a single run of the agent. Defaults to true.
          * @param installFeatures A function to configure additional features into the agent during initialization. Defaults to an empty configuration.
          * @return An instance of [AIAgent] configured with the provided parameters.
          */
@@ -280,6 +281,7 @@ public interface AIAgent<Input, Output> : Closeable {
             temperature: Double? = null,
             numberOfChoices: Int = 1,
             maxIterations: Int = 50,
+            enforceSingleRun: Boolean = true,
             installFeatures: FeatureContext.() -> Unit = {}
         ): AIAgent<String, String> = AIAgent(
             id = id,
@@ -297,6 +299,7 @@ public interface AIAgent<Input, Output> : Closeable {
                 },
                 model = llmModel,
                 maxAgentIterations = maxIterations,
+                enforceSingleRun = enforceSingleRun
             ),
             toolRegistry = toolRegistry,
             installFeatures = installFeatures
@@ -317,6 +320,7 @@ public interface AIAgent<Input, Output> : Closeable {
          * @param temperature Optional model temperature, with valid values ranging typically from 0.0 to 1.0.
          * @param numberOfChoices The number of choices the model should generate per invocation. Defaults to `1`.
          * @param maxIterations The maximum number of iterations the agent can perform. Defaults to `50`.
+         * @param enforceSingleRun A boolean flag indicating whether to enforce a single run of the agent. Defaults to `true`.
          * @param installFeatures An extension function on `FeatureContext` to install custom features for the agent. Defaults to an empty lambda.
          * @return A configured [AIAgent] instance that can process inputs and generate outputs using the specified strategy and model.
          */
@@ -332,6 +336,7 @@ public interface AIAgent<Input, Output> : Closeable {
             temperature: Double? = null,
             numberOfChoices: Int = 1,
             maxIterations: Int = 50,
+            enforceSingleRun: Boolean = true,
             noinline installFeatures: FeatureContext.() -> Unit = {},
         ): GraphAIAgent<Input, Output> {
             return GraphAIAgent(
@@ -352,6 +357,7 @@ public interface AIAgent<Input, Output> : Closeable {
                     },
                     model = llmModel,
                     maxAgentIterations = maxIterations,
+                    enforceSingleRun = enforceSingleRun
                 ),
                 toolRegistry = toolRegistry,
                 clock = clock,
@@ -374,6 +380,7 @@ public interface AIAgent<Input, Output> : Closeable {
          * @param temperature Optional model temperature, with valid values ranging typically from 0.0 to 1.0.
          * @param numberOfChoices The number of response choices to generate when querying the language model. Default is 1.
          * @param maxIterations The maximum number of iterations the agent is allowed to perform during execution. Default is 50.
+         * @param enforceSingleRun Flag indicating whether to enforce a single run of the agent. Default is true.
          * @param installFeatures A lambda to configure and install features in the agent's context.
          * @return An AI agent instance configured with the provided parameters and ready to execute the specified strategy.
          */
@@ -387,6 +394,7 @@ public interface AIAgent<Input, Output> : Closeable {
             temperature: Double? = null,
             numberOfChoices: Int = 1,
             maxIterations: Int = 50,
+            enforceSingleRun: Boolean = true,
             installFeatures: FunctionalAIAgent.FeatureContext.() -> Unit = {},
         ): FunctionalAIAgent<Input, Output> = FunctionalAIAgent(
             promptExecutor = promptExecutor,
@@ -402,6 +410,7 @@ public interface AIAgent<Input, Output> : Closeable {
                 },
                 model = llmModel,
                 maxAgentIterations = maxIterations,
+                enforceSingleRun = enforceSingleRun
             ),
             installFeatures = installFeatures,
             toolRegistry = toolRegistry,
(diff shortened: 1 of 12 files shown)


Rizzen:
Hello! Thank you for a contribution!
In current paradigm we prefer to avoid of reusing the agents and running singular agent in parallel. Currently to run same agent few times we have `AIAgentService` interface with `FunctionalAIAgentService` and `GraphAIAgentService` implementations.

PranavPurwar:
> Hello! Thank you for a contribution! In current paradigm we prefer to avoid of reusing the agents and running singular agent in parallel. Currently to run same agent few times we have `AIAgentService` interface with `FunctionalAIAgentService` and `GraphAIAgentService` implementations.

Hello! Thanks for the feedback!

I understand the current preference for using `AIAgentService`, which is the best approach for running multiple, independent agents in parallel.

The `AIAgentService` creates a fresh context for every call, whereas stateful/conversational agent would require history preservation across sequential calls. For example, a coding assistant that needs to remember previous reviews, file edits, or a planning agent with user feedback.

This change enables that context reuse with ease. Since the old, single run behavior remains the default, it's non-breaking and allows building such stateful agents out of the box easily.

How can we align this stateful capability with the architectural paradigm?

Rizzen:
>The AIAgentService creates a fresh context for every call, whereas stateful/conversational agent would require history preservation across sequential calls.

I understand this scenario, but by default we don't want agent to be implicitly stateful. For these scenarios you could use Persistence feature or pass context to agent input between runs yourself.

salah-8:
> I understand this scenario, but by default we don't want agent to be implicitly stateful. For these scenarios you could use Persistence feature or pass context to agent input between runs yourself.

@Rizzen why not add the option to make it stateful and keep the default as it is? 
I think this change will be useful and simple way to use agent as chatbot with multiple messages without unnecessary complexity with the other workarounds.

PranavPurwar:
> > The AIAgentService creates a fresh context for every call, whereas stateful/conversational agent would require history preservation across sequential calls.
> 
> I understand this scenario, but by default we don't want agent to be implicitly stateful. For these scenarios you could use Persistence feature or pass context to agent input between runs yourself.

I think this will not be an issue because this PR keeps the default functionality as is. It will only be "stateful", when toggled by the flag. Otherwise the current behavior remains as is.

aozherelyeva:
Hi @PranavPurwar, would you please rebase on the latest develop and resolve the conflicts? Thank you!

Chrylo:
Hi I hope it is ok to write the comment here.

@PranavPurwar we are also running into the use case of sequential agent calls like with chat bots e.g.:
1) User makes a request
2) Our service processes it
3) Our service wants to verify the result but needs more information from the user
4) repeat until verify returns OK

If I understand correctly this PR will help as long as the agent is still running. I am wondering if it actually solves the root cause for supporting "interruptions" in the graph with asynchronous input. Imagine a user continuing his chat a week later. You don't want x Agents running all the time? 

From the comments I understand two different "workarounds" of handling this: 
1) Having some kind of endless node graph -> let it crash whenever user input is required  -> restore the agent via persistence and somehow inject new user input.
2) Let the agent end whenever user input is required -> always output the whole llm session in the final result -> save it in the database -> start a new agent -> write the old llm session into the new agent.

So I am kinda agree with @Rizzen answers because this flag may only "fix" a small use case but does not help with the actual more complex feature request. Basically what Alberto is saying here: #1217 

What is the best practice for this because I agree that requiring external input during the graph execution is a common use case. Would be awesome if an agent automatically saves his history into a database and restores it automatically whenever the same agent ID is used. Maybe an approach like this could be cleaner?

PranavPurwar:
Hi @Chrylo,

The current implementation in Koog makes things like history compression fairly unintuitive, as single-run agents make it impractical to manage context properly. Additionally, because the graph execution is continuous, there is no native way to "pause" the graph and wait for user data. This usually forces you to write verbose, unintuitive hacks just to replicate a simple interruption.

This PR aims to simplify those common scenarios—like when a user needs to provide clarifications, more data, or when results are sent back sequentially. While complex workflows might eventually need a more robust system, this provides a practical path for the majority of cases where you don't want to over-engineer the solution.

Regarding the concern about agents "running" indefinitely: they are only active during a `run` call. They don’t consume tokens while waiting; the conversation history is just stored as metadata. For very long delays, such as a user returning after a week, Persistence, is still the intended tool to use here. You can utilize this alongside Persistence if you wish. With Persistence, you can save the state to disk and retrieve it whenever the user returns. This feature focuses on making the interaction logic more flexible, while persistence handles the long-term storage.

Let me know if you have any further enquiries!

aozherelyeva:
@PranavPurwar, thanks for the PR. If the enhancement is still actual for you, would you please create a new PR and mention me in it? (so that I could set assignees). Thanks.

Issue #1210, opened 2025-12-01
Amazon Bedrock Models list is incomplete


chris-hatton:
[bedrock-all-region-models.sh](https://github.com/user-attachments/files/23845647/bedrock-all-region-models.sh)

The current list of Bedrock models in Koog, is a subset of what's available.

While not all models are available in all AWS regions; I believe Koog should provide the complete set.

Running:
```
aws --region us-east-1 bedrock list-foundation-models > models.json
```
...then processing the JSON to a Markdown table yielded the following; which is more than implemented in [BedrockModels.kt](https://github.com/JetBrains/koog/blob/develop/prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/bedrock/BedrockModels.kt):

### AWS Bedrock Models from all regions

| Bedrock ID                     | Model Name               | Provider      | Modalities       | Regions                          |
|-------------------------------|--------------------------|---------------|------------------|----------------------------------|
| `ai21.jamba-1-5-mini-v1:0      ` | Jamba 1.5 Mini           | AI21 Labs     | TEXT             | `us-east-1                     ` |
| `amazon.nova-2-multimodal-embeddings-v1:0` | Amazon Nova Multimodal Embeddings | Amazon        | IMAGE AUDIO VIDEO TEXT | `us-east-1                     ` |
| `amazon.nova-canvas-v1:0       ` | Nova Canvas              | Amazon        | TEXT IMAGE VIDEO SPEECH | `ap-south-1`, `ca-central-1`, `ap-northeast-2`, `eu-central-1`, `eu-west-1`, `ap-northeast-3`, `ap-southeast-1`, `us-west-2`, `ap-northeast-1`, `ap-southeast-2` |
| `amazon.nova-lite-v1:0         ` | Nova Lite                | Amazon        | TEXT IMAGE VIDEO SPEECH | `ap-south-1`, `eu-west-2`, `eu-west-3`, `ca-central-1`, `ap-northeast-2`, `eu-north-1`, `eu-central-1`, `eu-west-1`, `ap-northeast-3`, `ap-southeast-1`, `us-west-2`, `ap-northeast-1`, `ap-southeast-2` |
| `amazon.nova-lite-v1:0:24k     ` | Nova Lite                | Amazon        | IMAGE VIDEO TEXT | `us-east-1                     ` |
[...]

PR #1224, opened 2025-12-04
Add fallback model support and warning mechanism for unsupported Bedrock models

aozherelyeva:
## Motivation and Context

[KG-595](https://youtrack.jetbrains.com/issue/KG-595) Relax getBedrockModelFamily error to a warning when using custom models

When the user adds a Bedrock model from a family that's not added to the list we have, it's good to have a fallback mechanism.
E.g. the user knows that the LLM API of a new model works exactly like the Anthropic one. In this case, they'll get a warning and can continue using the fallback Anthropic API until their own serializer is not added.

Source: https://github.com/JetBrains/koog/issues/1210#issuecomment-3601786108

## Breaking Changes

---

#### Type of the changes
- [ ] Documentation update
- [ ] Tests improvement
- [x] Refactoring

#### Checklist

### DIFF
--- a/koog-ktor/src/jvmMain/kotlin/ai/koog/ktor/BedrockConfig.kt
+++ b/koog-ktor/src/jvmMain/kotlin/ai/koog/ktor/BedrockConfig.kt
@@ -49,7 +49,7 @@ public fun KoogAgentsConfig.bedrock(
     }
     addLLMClient(
         LLMProvider.Bedrock,
-        BedrockLLMClient(client, moderationGuardrailsSettings, clock)
+        BedrockLLMClient(client, moderationGuardrailsSettings, null, clock)
     )
 }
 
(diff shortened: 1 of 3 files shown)


prompt/prompt-executor/prompt-executor-clients/src/commonMain/kotlin/ai/koog/prompt/executor/clients/Warnings.kt
```diff
@@ -0,0 +1,33 @@
+package ai.koog.prompt.executor.clients
+
+import io.github.oshai.kotlinlogging.KotlinLogging
+
+private val logger = KotlinLogging.logger {}
+
+/**
+ * Warning for Koog LLM clients
+ */
+public object LLMClientWarning {
```
EugeneTheDev: Is such a complex approach really needed? IMHO a regular `logger.warn` would suffice

EugeneTheDev (changes requested):
Generally looks good, a minor comment

EugeneTheDev (approved):

PR #1226, opened 2025-12-04
KG-508 Update Tool API to fix name and descriptor discrepancy and make it more robust

EugeneTheDev:
**This PR introduces breaking changes in Tool API**

Currently `argsSerializer`, `resultSerializer`, `name`, `description` and `descriptor` are meant to be overriden to provide your own values. Since we can also generate `descriptor` automatically now, it introduces a discrepancy between property values and the actual state of things. It's described in [KG-508](https://youtrack.jetbrains.com/issue/KG-508). So current version makes for unstable and somewhat confusing API.

In this PR, I've moved these configurable tool properties to the constructors instead. This avoids confusion, since now the ways in which you can configure the tool are visible more explicitly - either provide your own `descriptor`, or opt for an automatic generation and only provide `name` and `description`. Updated API also looks a bit more concise and fluent, IMHO, since you have to write less `override` now. Compare these:
```kt
// Before
object DummyTool : SimpleTool<Unit>() {
    override val argsSerializer = Unit.serializer()

    override val name: String = "dummy"
    override val description: String = "Dummy tool for testing"

    override suspend fun doExecute(args: Unit): String = "Dummy result"
}

// After
object DummyTool : SimpleTool<Unit>(
    argsSerializer = Unit.serializer(),
    name = "dummy",
    description = "Dummy tool for testing"
) {
    override suspend fun execute(args: Unit): String = "Dummy result"
}
```

Also, another breaking change, as you have already seen from the example above, is `doExecute` removal. It is not needed anymore, since tools support primitive types now, so you can just use a regular `execute`

### DIFF
--- a/agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt
+++ b/agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt
@@ -29,7 +29,6 @@ import ai.koog.prompt.markdown.markdown
 import ai.koog.prompt.message.Message
 import ai.koog.prompt.params.LLMParams
 import ai.koog.prompt.processor.ResponseProcessor
-import kotlinx.serialization.KSerializer
 import kotlinx.serialization.serializer
 
 /**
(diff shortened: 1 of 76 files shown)


agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt
```diff
@@ -138,11 +106,12 @@ public object SubgraphWithTaskUtils {
...
-internal inline fun <reified Output> identityTool(): Tool<Output, Output> = object : Tool<Output, Output>() {
-    override val argsSerializer: KSerializer<Output> = serializer()
-    override val resultSerializer: KSerializer<Output> = serializer()
-    override val name: String = SubgraphWithTaskUtils.FINALIZE_SUBGRAPH_TOOL_NAME
-    override val description: String = SubgraphWithTaskUtils.FINALIZE_SUBGRAPH_TOOL_DESCRIPTION
+internal inline fun <reified Output> identityTool(): Tool<Output, Output> = object : Tool<Output, Output>(
+    argsSerializer = serializer(),
+    resultSerializer = serializer(),
+    name = SubgraphWithTaskUtils.FINALIZE_SUBGRAPH_TOOL_NAME,
+    description = SubgraphWithTaskUtils.FINALIZE_SUBGRAPH_TOOL_DESCRIPTION
```
kpavlov, line 115: Is it a copy from finalize tool?

kpavlov (approved):
Hey @EugeneTheDev, could you please add a PR description and update the identity tool’s name and description before we merge?

agents/agents-ext/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt
EugeneTheDev: It was like this before. This `identityTool` is internal and is used actually only as a finalize tool to provide the final result, hence the finalize tool name. Sounds like we can leave it as it is. Or do you want me to make name and description configurable (although there's no use for it currently). wdyt?

EugeneTheDev:
Updated description

PR #1228, opened 2025-12-05
Add multiplatform `SystemConfigReader` interface with JVM-specific implementation

kpavlov:
# Add multiplatform system configuration and secrets readers with JVM-specific implementation

Introduce `SystemConfigReader` and `SystemSecretsReader` interfaces to standardize access to configuration values and secrets across different platforms. This change utilizes the `expect`/`actual` mechanism to define common contracts with specific platform implementations.

*   Implement fully functional readers for the JVM target: the configuration reader resolves values from environment variables and system properties (supports name normalization), while the secrets reader strictly uses environment variables for security.
*   Add stub implementations for Android, Apple, JS, and WasmJS platforms that currently throw a `NotImplementedError` when accessed.
*   Define the common interfaces and factory functions to allow retrieval of configuration strings and secrets.

## Motivation and Context
Should be used in Debugger Config and LLM Clients

See also #1231 

## Breaking Changes
No

#### Type of the changes
- [ ] Documentation update
- [ ] Tests improvement
- [ ] Refactoring

#### Checklist

### DIFF
--- a/utils/src/androidMain/kotlin/ai/koog/utils/system/SystemConfigReader.android.kt
+++ b/utils/src/androidMain/kotlin/ai/koog/utils/system/SystemConfigReader.android.kt
@@ -1,0 +1,5 @@
+package ai.koog.utils.system
+
+public actual fun systemConfigReader(): SystemConfigReader {
+    throw NotImplementedError("SystemConfigReader is not yet supported on Android platform")
+}
--- a/utils/src/appleMain/kotlin/ai/koog/utils/system/SystemConfigReader.apple.kt
+++ b/utils/src/appleMain/kotlin/ai/koog/utils/system/SystemConfigReader.apple.kt
@@ -1,0 +1,5 @@
+package ai.koog.utils.system
+
+public actual fun systemConfigReader(): SystemConfigReader {
+    throw NotImplementedError("SystemConfigReader is not yet supported on Apple platforms")
+}
--- a/utils/src/jvmMain/kotlin/ai/koog/utils/system/SystemConfigReader.jvm.kt
+++ b/utils/src/jvmMain/kotlin/ai/koog/utils/system/SystemConfigReader.jvm.kt
@@ -1,0 +1,3 @@
+package ai.koog.utils.system
+
+public actual fun systemConfigReader(): SystemConfigReader = JvmSystemConfigReader
--- a/utils/src/wasmJsMain/kotlin/ai/koog/utils/system/SystemConfigReader.wasmJs.kt
+++ b/utils/src/wasmJsMain/kotlin/ai/koog/utils/system/SystemConfigReader.wasmJs.kt
@@ -1,0 +1,5 @@
+package ai.koog.utils.system
+
+public actual fun systemConfigReader(): SystemConfigReader {
+    throw NotImplementedError("SystemConfigReader not yet supported on WasmJS platform")
+}
--- a/utils/src/androidMain/kotlin/ai/koog/utils/system/SystemSecretsReader.android.kt
+++ b/utils/src/androidMain/kotlin/ai/koog/utils/system/SystemSecretsReader.android.kt
@@ -1,0 +1,5 @@
+package ai.koog.utils.system
+
+public actual fun systemSecretsReader(): SystemSecretsReader {
+    throw NotImplementedError("SystemSecretsReader is not yet supported on Android platform")
+}
--- a/utils/src/appleMain/kotlin/ai/koog/utils/system/SystemSecretsReader.apple.kt
+++ b/utils/src/appleMain/kotlin/ai/koog/utils/system/SystemSecretsReader.apple.kt
@@ -1,0 +1,5 @@
+package ai.koog.utils.system
+
+public actual fun systemSecretsReader(): SystemSecretsReader {
+    throw NotImplementedError("SystemSecretsReader is not yet supported on Apple platforms")
+}
--- a/utils/src/commonMain/kotlin/ai/koog/utils/system/SystemConfigReader.kt
+++ b/utils/src/commonMain/kotlin/ai/koog/utils/system/SystemConfigReader.kt
@@ -1,0 +1,30 @@
+package ai.koog.utils.system
+
+/**
+ * An interface for reading system configuration properties.
+ *
+ * This interface provides a method to retrieve system configuration values by their names.
+ * Implementations can vary based on the operating system or platform being used.
+ */
+public interface SystemConfigReader {
+
+    /**
+     * Retrieves the value of a system configuration property by its name.
+     *
+     * @param name The name of the system configuration property to retrieve.
+     * @return The value of the configuration property if it exists, or null if not found.
+     */
+    public fun getConfigVariable(name: String): String?
+}
+
+/**
+ * Provides an implementation of the [SystemConfigReader] interface for the current platform.
+ *
+ * This function returns an instance of [SystemConfigReader] that allows access
+ * to system-level configuration properties based on platform-specific implementations.
+ * On unsupported platforms, this function will throw a [NotImplementedError].
+ *
+ * @return An instance of [SystemConfigReader] for accessing system configuration properties.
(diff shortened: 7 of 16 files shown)


utils/src/androidMain/kotlin/ai/koog/utils/system/SystemConfigReader.android.kt
```diff
@@ -0,0 +1,9 @@
+package ai.koog.utils.system
+
+internal object AndroidSystemConfigReader : SystemConfigReader {
```
EugeneTheDev: Why is it inernal? Shouldn't we make this public, since `systemConfigReader` is public too?

utils/src/androidMain/kotlin/ai/koog/utils/system/SystemConfigReader.android.kt
```diff
@@ -0,0 +1,9 @@
...
+    override fun get(name: String): String? {
+        TODO("Not yet implemented")
```
EugeneTheDev: Please throw `NotImplementedException` with a proper message instead of `TODO`, `TODO` is more for work-in-process kind of things

utils/src/appleMain/kotlin/ai/koog/utils/system/SystemConfigReader.apple.kt
```diff
@@ -0,0 +1,9 @@
+package ai.koog.utils.system
+
+internal object AppleSystemConfigReader : SystemConfigReader {
```
EugeneTheDev: Same about `internal` and `TODO`

utils/src/jvmMain/kotlin/ai/koog/utils/system/SystemConfigReader.jvm.kt
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * The property name provided is automatically converted to a format compatible with
+ * system environment variable conventions by transforming the name to lowercase and
+ * replacing underscores with dots. It attempts to fetch the value from both the original
+ * name and the transformed format.
+ *
+ * Implements:
+ * - SystemConfigReader: Provides the ability to retrieve configuration properties.
+ */
+internal object JvmSystemConfigReader : SystemConfigReader {
```
EugeneTheDev: `public`

utils/src/jvmMain/kotlin/ai/koog/utils/system/SystemConfigReader.jvm.kt
```diff
@@ -0,0 +1,46 @@
...
+     *
+     * @param name The name of the property to retrieve. It may be found directly in the environment
+     *             variables or after transforming it to a system property format (lowercase with
+     *             underscores replaced by dots).
+     * @return The corresponding value if it exists in the environment variables or cache,
+     *         or null if no value is found.
+     */
+    override fun get(name: String): String? {
+        return cache.computeIfAbsent(name) {
+            System.getenv(name) ?: System.getProperty(name.lowercase().replace('_', '.'))
```
EugeneTheDev: Custom logic with lowercase and replace doesn't look straightforward and can lead to confusion. Let's use `name` as it is everywhere

utils/src/wasmJsMain/kotlin/ai/koog/utils/system/SystemConfigReader.wasmJs.kt
```diff
@@ -0,0 +1,10 @@
+package ai.koog.utils.system
+
+internal object WasmJsSystemConfigReader : SystemConfigReader {
```
EugeneTheDev: Same about `internal` and `TODO`

utils/src/jvmMain/kotlin/ai/koog/utils/system/SystemConfigReader.jvm.kt
```diff
@@ -0,0 +1,46 @@
...
+ * system environment variable conventions by transforming the name to lowercase and
+ * replacing underscores with dots. It attempts to fetch the value from both the original
+ * name and the transformed format.
+ *
+ * Implements:
+ * - SystemConfigReader: Provides the ability to retrieve configuration properties.
+ */
+internal object JvmSystemConfigReader : SystemConfigReader {
+
+    /**
```
EugeneTheDev: Empty doc comment, do we need it here?

EugeneTheDev (changes requested):

utils/src/androidMain/kotlin/ai/koog/utils/system/SystemConfigReader.android.kt
kpavlov: No, the interface and getter function are public. The implementation should not be public

utils/src/androidMain/kotlin/ai/koog/utils/system/SystemConfigReader.android.kt
kpavlov: TODO throws NotImplementedException

EugeneTheDev: But it gives you a warning, because it's not intended for production use. It's better semantically to throw a proper exception

utils/src/jvmMain/kotlin/ai/koog/utils/system/SystemConfigReader.jvm.kt
kpavlov: I’m not sure about that. It’s a pretty [standard way/convention](https://docs.spring.io/spring-boot/how-to/properties-and-configuration.html#howto.properties-and-configuration.external-properties-location) to handle things by using `CONFIG_PROP` as an environment variable and `config.prop` as a system property. Debugger is using the same pattern as well

EugeneTheDev: It's fine that it's an established patter, the main concern here is that your input gets modified unexpectedly. So e.g. if I pass org.my_propery, it would be transformed to org.my.property and the function will return null. And it can be confusing if I know that the property is there

kpavlov: @EugeneTheDev, how should we handle this? Should we only normalize the name if it’s in UPPER_CASE_WITH_UNDERSCORES?
Is the current setup working well enough?

I suggest keeping this implementation, and users can implement their own config reader

utils/src/androidMain/kotlin/ai/koog/utils/system/SystemConfigReader.android.kt
kpavlov: I removed dummy readers at all

utils/src/jvmMain/kotlin/ai/koog/utils/system/SystemConfigReader.jvm.kt
kpavlov: okay, because it might be multiple implementations

utils/src/jvmMain/kotlin/ai/koog/utils/system/SystemConfigReader.jvm.kt
EugeneTheDev: Do you think it should be reader's concern? IMHO we should keep reader as simple as possible - only read what is asked. And if the application allows several configuration options, like in your Spring example with env `MY_PROPERTY` and system `my.property`, it should be application's concern, and I would expect the usage site to implement something as follows:
```kt
val myProperty = reader.getConfigVariable("MY_PROPERTY")
    ?: reader.getConfigVariable("my.property")
    ?: throw NoSuchElementException("Can't find myProperty")
    
```

kpavlov: This wouldn't work as convention over configuration then.
This behavior is only applicable to JVM. For other platforms there should not be such convention.
@Ololoshechkin, what do you think?

kpavlov:
Merged as a part of #1231

PR #1253, opened 2025-12-10
Introduce initial ACP integration

tiginamaria:
## Motivation and Context

ACP feature to create ACP-compatible agents in Koog
- [x] Raw access to the protocol from the feature in order not to complicate it until ACP becomes stable
- [x] Example
- [x] README (Model.md)
- [x] Documentation

## Breaking Changes

---

#### Type of the changes
- [ ] Documentation update
- [ ] Tests improvement
- [ ] Refactoring

#### Checklist


agents/agents-features/agents-features-acp/src/jvmMain/kotlin/ai/koog/agents/features/acp/MessageConverters.kt
```diff
@@ -0,0 +1,205 @@
...
+import com.agentclientprotocol.common.Event.SessionUpdateEvent
+import com.agentclientprotocol.model.ContentBlock
+import com.agentclientprotocol.model.EmbeddedResourceResource
+import com.agentclientprotocol.model.SessionUpdate
+import com.agentclientprotocol.model.SessionUpdate.AgentMessageChunk
+import com.agentclientprotocol.model.ToolCallId
+import com.agentclientprotocol.model.ToolCallStatus
+import kotlinx.datetime.Clock
+
+public const val UNKNOWN_FORMAT: String = "unknown"
```
EugeneTheDev, line 18: Since these consts are public, let's add a brief doc comment for them

agents/agents-features/agents-features-acp/src/jvmMain/kotlin/ai/koog/agents/features/acp/MessageConverters.kt
```diff
@@ -0,0 +1,205 @@
...
+        is ContentBlock.Text -> {
+            ContentPart.Text(text)
+        }
+    }
+}
+
+/**
+ * Converts a [Message.Response] to a list of ACP [SessionUpdateEvent].
+ */
+public fun Message.Response.toAcpEvents(): List<SessionUpdateEvent> {
```
EugeneTheDev: nit: to make it a bit easier to read, let's move this function right after the `toKoogMessage`. The idea is, more general functions go first, in this case converting from ACP to Koog and vice versa, and then secondary auxilary functions go after

agents/agents-features/agents-features-acp/src/jvmMain/kotlin/ai/koog/agents/features/acp/MessageConverters.kt
```diff
@@ -0,0 +1,205 @@
...
+    )
+}
+
+/**
+ * Converts a single [ContentBlock] of ACP prompt to a Koog [ContentPart].
+ */
+public fun ContentBlock.toKoogContentPart(): ContentPart {
+    return when (this) {
+        // https://agentclientprotocol.com/protocol/content#audio-content
+        is ContentBlock.Audio -> {
```
EugeneTheDev, line 160: Here and for `Image` you can put `uri` from ACP's entities into our `fileName` field. Technically it's not really correct, but at least we will preserve this information somehow, rather than just leaving it as an unnamed file

agents/agents-features/agents-features-acp/src/jvmMain/kotlin/ai/koog/agents/features/acp/MessageConverters.kt
```diff
@@ -0,0 +1,205 @@
...
+import kotlinx.datetime.Clock
+
+public const val UNKNOWN_FORMAT: String = "unknown"
+public const val UNKNOWN_MIME_TYPE: String = "unknown/unknown"
+public const val UNKNOWN_URI: String = "unknown"
+public const val UNKNOWN_FILE_NAME: String = "unknown"
+public const val UNKNOWN_TOOL_CALL_ID: String = "unknown"
+public const val UNKNOWN_TOOL_DESCRIPTION: String = "unknown"
+
+private fun parseFormat(mimeType: String?): String {
```
EugeneTheDev: This is not quite right, actually, You're using this to get the file extension, but you can't always get it from the mime type. That is, the second part of mime type != always file extension. I see the limitation though, we can't determine it reliably in all the cases for ACP, since there's no such information in some sources entities. I would suggest leaving the comment here for clarity, acknowledging the problem and explaining the limitation. So that future us could potentially find a better solution or remove it altogether

agents/agents-features/agents-features-acp/src/jvmMain/kotlin/ai/koog/agents/features/acp/MessageConverters.kt
```diff
@@ -0,0 +1,205 @@
...
+                content = AttachmentContent.Binary.Base64(data),
+                format = parseFormat(mimeType),
+                mimeType = mimeType
+            )
+        }
+
+        // https://agentclientprotocol.com/protocol/content#embedded-resource
+        is ContentBlock.Resource -> {
+            when (val resource = this.resource) {
+                is EmbeddedResourceResource.BlobResourceContents -> {
```
EugeneTheDev, line 181: Same thing about `uri` here, we can put it into `fileName` too

agents/agents-features/agents-features-acp/src/jvmMain/kotlin/ai/koog/agents/features/acp/MessageConverters.kt
```diff
@@ -0,0 +1,205 @@
...
+                    ContentPart.File(
+                        content = AttachmentContent.Binary.Base64(resource.blob),
+                        format = parseFormat(resource.mimeType),
+                        mimeType = resource.mimeType ?: UNKNOWN_MIME_TYPE
+                    )
+                }
+
+                is EmbeddedResourceResource.TextResourceContents -> {
+                    ContentPart.File(
```
EugeneTheDev, line 191: And here as well, `uri` into `fileName`

agents/agents-features/agents-features-acp/src/jvmMain/kotlin/ai/koog/agents/features/acp/MessageConverters.kt
```diff
@@ -0,0 +1,205 @@
...
+                    )
+                )
+            }
+
+            is Message.Tool.Call -> {
+                add(
+                    SessionUpdateEvent(
+                        update = SessionUpdate.ToolCall(
+                            toolCallId = ToolCallId(response.id ?: UNKNOWN_TOOL_CALL_ID),
+                            // TODO: Support tool description in the event
```
EugeneTheDev: I see that you added it already in the events. Can we use it here now?

agents/agents-features/agents-features-acp/src/jvmMain/kotlin/ai/koog/agents/features/acp/MessageConverters.kt
```diff
@@ -0,0 +1,205 @@
...
+                mimeType = this.mimeType,
+            )
+        }
+
+        is ContentPart.File ->
+            when (val content = this.content) {
+                is AttachmentContent.Binary.Base64 -> ContentBlock.Resource(
+                    resource = EmbeddedResourceResource.BlobResourceContents(
+                        blob = content.base64,
+                        // TODO: add uri to the file
```
EugeneTheDev: Similar to how I suggested to use `uri` as a `fileName`, let's do the reverse here: let's set `fileName` as `uri` everywhere. Again, technically not correct, but we're preserving at least some infromation for now

agents/agents-features/agents-features-acp/src/jvmMain/kotlin/ai/koog/agents/features/acp/MessageConverters.kt
```diff
@@ -0,0 +1,205 @@
...
+ */
+public fun ContentPart.toAcpContentBlock(): ContentBlock {
+    return when (this) {
+        is ContentPart.Text -> {
+            ContentBlock.Text(this.text)
+        }
+
+        is ContentPart.Audio -> {
+            ContentBlock.Audio(
+                data = this.content.toString(),
```
EugeneTheDev: Careful here, we can't just use `toString`. ACP treats `Audio` as a binary content, but in our case it can be either an embedded binary content or a URL to a content. So there has to be an additional condition. If it's a `Binary` content, treat it as a `ContentBlock.Audio`. Otherwise, it's a `ContentBlock.ResourceLink`.

agents/agents-features/agents-features-acp/src/jvmMain/kotlin/ai/koog/agents/features/acp/MessageConverters.kt
```diff
@@ -0,0 +1,205 @@
...
+                    ContentBlock.ResourceLink(
+                        name = this.fileName ?: UNKNOWN_FILE_NAME,
+                        uri = content.url
+                    )
+                }
+            }
+
+        is ContentPart.Image -> {
+            ContentBlock.Image(
+                data = this.content.toString(),
```
EugeneTheDev: Same as with `ContentPart.Audio`. Also, nit, let's move it up next to the `ContentPart.Image` handling, to make it easier to read again, since these are kinda similar and you already put them together in the other conversion extension function.

agents/agents-features/agents-features-acp/src/jvmMain/kotlin/ai/koog/agents/features/acp/MessageConverters.kt
```diff
@@ -0,0 +1,205 @@
...
+                mimeType = this.mimeType,
+            )
+        }
+
+        is ContentPart.Video -> {
+            throw AcpException("Video content is not supported yet in Acp content blocks.")
```
EugeneTheDev: Since this is the only place you use this custom exception type for now, I suggest removing it and using regular `IllegalArgumentException`

koog-agents/build.gradle.kts
```diff
@@ -40,6 +40,7 @@ val excluded = setOf(
 
 val included = setOf(
     ":agents:agents-core",
+    ":agents:agents-features:agents-features-acp",
```
EugeneTheDev: Are we including it by default? E.g. A2A is not included because it brings other dependencies, mainly A2A SDK. Here it would bring ACP SDK as a dependency too, so I suggest not including it into the meta `koog-agents` dependency, and specifying in the docs that users are supposed to add this feature dependency explicitly

examples/simple-examples/src/main/kotlin/ai/koog/agents/example/acp/KoogAgentSupport.kt
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    override suspend fun cancel() {
+        logger.info { "Canceling ACP agent" }
+        agentJob?.cancelAndJoin()
+    }
+
+    private fun koogStrategy() = strategy<Unit, Unit>("acp-agent") {
+        val nodePlanPrompt by node<Unit, String>("plan") {
+            "You have a task! Crate a plan for it."
```
EugeneTheDev: typo: "Crate" -> "Create"

EugeneTheDev (changes requested):
Overall nicely done, thank you. I left some comments, mainly regarding message converters.

docs/docs/agent-client-protocol.md
```diff
@@ -0,0 +1,433 @@
+# Agent Client Protocol
+
+Agent Client Protocol (ACP) is a standardized protocol that enables client applications to communicate with AI agents through a consistent, bidirectional interface.
+
+ACP provides a structured way for agents to interact with clients, supporting real-time event streaming, tool call notifications, and session lifecycle management.
+
+The Koog framework provides integration with ACP, enabling you to build ACP-compliant agents that can communicate with standardized client applications.
```
EugeneTheDev, line 7: Here and below, "The Koog framework" sounds kinda heavy. Can we just say "Koog"?

docs/docs/agent-client-protocol.md
```diff
@@ -0,0 +1,433 @@
...
+
+To learn more about the protocol, see the [Agent Client Protocol](https://agentclientprotocol.com) documentation.
+
+## What is ACP?
```
EugeneTheDev: This section sounds like a repetition of what was just said above. Let's remove it?

docs/docs/agent-client-protocol.md
```diff
@@ -0,0 +1,433 @@
...
+val acpContent: List<ContentBlock> = TODO()
+-->
+```kotlin
+// Convert ACP content blocks to Koog message
+val koogMessage = acpContent.toKoogMessage(clock)
+
+// Append to existing prompt
+fun Prompt.appendPrompt(content: List<ContentBlock>): Prompt {
+    return withMessages { messages ->
+        messages + listOf(content.toKoogMessage(clock))
```
EugeneTheDev, line 225: nit: You don't have to wrap the right part in `listOf`. Kotlin overloads `plus` operator for collections, so you can just write `+ element` and this statement will return a new collection with this element appended

EugeneTheDev (changes requested):
Sorry, a few more comments that I missed initially

EugeneTheDev:
Also, let's update PR title a bit to make it more descriptive: "Introduce initial ACP integration" or something similar

agents/agents-features/agents-features-acp/src/jvmMain/kotlin/ai/koog/agents/features/acp/MessageConverters.kt
tiginamaria: Found a way to do it, fixed

koog-agents/build.gradle.kts
tiginamaria: Yeah, my bad, removed it

agents/agents-features/agents-features-acp/src/jvmMain/kotlin/ai/koog/agents/features/acp/MessageConverters.kt
```diff
@@ -0,0 +1,283 @@
...
+            when (val content = this.content) {
+                is AttachmentContent.Binary.Base64,
+                is AttachmentContent.Binary.Bytes -> {
+                    ContentBlock.Audio(
+                        data = content.asBase64(),
+                        mimeType = this.mimeType,
+                    )
+                }
+
+                is AttachmentContent.PlainText -> {
```
EugeneTheDev: Media attachments (Audio, Image, Video) can't have `PlainText` content type, there's a check for it in the constructor. You can throw in this case, just to be sure, but it shouldn't be happening really

EugeneTheDev (approved):
Thanks, a small comment, but other than that good to go

PR #1281, opened 2025-12-17
Fix integration tests

aozherelyeva:
## Motivation and Context

Fixing some recent integration tests failures. Also, used AI to fix some of the non-testing code parts, and although I can describe why it's done, a thorough review would be appreciated.

## Breaking Changes

---

#### Type of the changes
- [ ] Documentation update
- [x] Tests improvement
- [ ] Refactoring

#### Checklist


integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/capabilities/ModelCapabilitiesIntegrationTest.kt
```diff
@@ -204,6 +205,12 @@ class ModelCapabilitiesIntegrationTest {
                 }
 
                 LLMCapability.Document -> {
+                    // TODO KG-620 GPT-5.1-Codex fails to process the text input file
+                    if (model == OpenAIModels.Chat.GPT5_1Codex) {
+                        assumeTrue(false, "Skipping document capability test for ${model.id}, see KG-620")
+                        return@runTest
+                    }
```
skarpovdev: assumeTrue is capable of skipping the test, no need for if statement


```suggestion
                    assumeTrue(
                        model != OpenAIModels.Chat.GPT5_1Codex,
                        "Skipping document capability test for ${model.id}, see KG-620"
                    )
```

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/capabilities/ModelCapabilitiesIntegrationTest.kt
```diff
@@ -204,6 +205,12 @@ class ModelCapabilitiesIntegrationTest {
                 }
 
                 LLMCapability.Document -> {
+                    // TODO KG-620 GPT-5.1-Codex fails to process the text input file
```
skarpovdev: I'd rather add a comment in [KG-620](https://youtrack.jetbrains.com/issue/KG-620) issue to remove this block than have a TODO. Having a TODO doesn't guarantee the block will be removed because one has to remember to do it once the issue is fixed and also fives unnecessary warnings when committing changes in this file.

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/executor/ExecutorIntegrationTestBase.kt
```diff
@@ -140,17 +140,48 @@ abstract class ExecutorIntegrationTestBase {
...
+                    maxTokens = 512
+                )
+            }
+
+            else -> LLMParams(maxTokens = 256)
+        }
+    }
+
+    private fun createNoReasoningParams(model: LLModel): LLMParams {
+        return when (model.provider) {
```
skarpovdev: nit: It can be an expression body instead of using return.

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/executor/ExecutorIntegrationTestBase.kt
```diff
@@ -140,17 +140,48 @@ abstract class ExecutorIntegrationTestBase {
...
+            is LLMProvider.Google -> {
+                val thinkingConfig = if (model.id == GoogleModels.Gemini3_Pro_Preview.id) {
+                    GoogleThinkingConfig(
+                        includeThoughts = false,
+                    )
+                } else {
+                    GoogleThinkingConfig(
+                        includeThoughts = false,
+                    )
+                }
```
skarpovdev: I don't get it. Why is `GoogleThinkingConfig` the same in both cases?

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClient.kt
```diff
@@ -192,7 +192,7 @@ public open class GoogleLLMClient(
                         outputTokensCount = it.candidatesTokenCount,
                     )
                 }
-                response.candidates.firstOrNull()?.let { candidate ->
+                response.candidates?.firstOrNull()?.let { candidate ->
```
skarpovdev: If `candidates` is now nullable, I would check if it's null first and throw if it is:
```
val candidates = response.candidates ?: throw LLMClientException(clientName, "Null candidates in Google API response")
```
While `firstOrNull()` is fine in this case, there is also `isNotEmpty()` and `isEmpty()` later on on same variable, which is inconsistent, unclear why, and doesn't prevent silent failures.

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/agent/AIAgentIntegrationTest.kt
```diff
@@ -1007,7 +1007,7 @@ class AIAgentIntegrationTest : AIAgentTestBase() {
 
                 with(state) {
                     withClue("${CalculatorToolNoArgs.descriptor.name} tool should be called for model $model") {
-                        actualToolCalls shouldBe listOf(CalculatorToolNoArgs.descriptor.name)
+                        actualToolCalls.shouldContain(CalculatorToolNoArgs.descriptor.name)
```
skarpovdev, line 1010: Why? Is the test flaky when we check for the exact list?
Can the list have a different size or order or what is the issue?

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/executor/ExecutorIntegrationTestBase.kt
```diff
@@ -140,17 +140,48 @@ abstract class ExecutorIntegrationTestBase {
...
                     GoogleThinkingConfig(
                         includeThoughts = true,
-                        thinkingLevel = GoogleThinkingLevel.LOW // with HIGH thoughts often exceed maxTokens causing test failures
+                        thinkingLevel = GoogleThinkingLevel.HIGH
                     )
                 } else {
                     GoogleThinkingConfig(
                         includeThoughts = true,
-                        thinkingBudget = 256
+                        thinkingBudget = 512
```
skarpovdev: The reason is not documented, but I assume it's the same as in the comment below?
We can extract these into global vars btw.

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/utils/Models.kt
```diff
@@ -103,7 +103,8 @@ object Models {
     @JvmStatic
     fun reasoningCapableModels(): Stream<LLModel> {
         return Stream.of(
-            OpenAIModels.Chat.GPT5_2,
+            // Replaced 5.2 with 5.1-Codex because of the unstable 5.2 behaviour, see KG-625
```
skarpovdev, line 106: IMHO this comment can be removed. It's better to describe this in PR. It won't be useful for anyone looking at the code later on.

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/capabilities/ModelCapabilitiesIntegrationTest.kt
skarpovdev: Same for other todos.
If you want to keep the mention of the YT issue, it can be just a comment.

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/executor/ExecutorIntegrationTestBase.kt
zarechneva-a: Yeah, I forgot to remove the old condition that I used to check Gemini 3 behaviour

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/executor/ExecutorIntegrationTestBase.kt
```diff
@@ -161,15 +192,16 @@ abstract class ExecutorIntegrationTestBase {
...
             system("You are a helpful assistant.")
             user("What is the capital of France?")
         }
 
         withRetry(times = 3, testName = "integration_testExecute[${model.id}]") {
             getExecutor(model).execute(prompt, model) shouldNotBeNull {
                 shouldNotBeEmpty()
-                with(shouldForAny { it is Message.Assistant }.first()) {
+                shouldForAny { it is Message.Assistant }
+                with(filterIsInstance<Message.Assistant>().first()) {
```
skarpovdev: minor: with `shouldForAny` and then `filterIsInstance` you iterate twice. Instead, you can create a variable  with something like the following chain:

filterIsInstance<Message.Assistant>().firstOrNull().shouldNotBeNull()

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/agent/AIAgentIntegrationTest.kt
zarechneva-a: Yup, some models call NothingTool because of the parameters’ directive, but then want to call the Calculator tool because it's what's logically demanded by the request. Therefore, we don't mind having some parallel tool calls, but we have to make sure that the explicitly called tools are in the list of the called ones.

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/executor/ExecutorIntegrationTestBase.kt
```diff
@@ -1052,7 +1086,12 @@ abstract class ExecutorIntegrationTestBase {
             getLLMClient(model).execute(prompt, model) shouldNotBeNull {
                 shouldNotBeEmpty()
                 withClue("No reasoning messages found") { shouldForAny { it is Message.Reasoning } }
-                assertResponseContainsReasoning(this)
+                // Some Google models aren't providing meta info
+                if (model.provider == LLMProvider.Google) {
+                    assertResponseContainsReasoning(this, false)
+                } else {
+                    assertResponseContainsReasoning(this)
+                }
```
skarpovdev: nit: can be a one-liner `assertResponseContainsReasoning(this, model.provider != LLMProvider.Google)`

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/utils/Models.kt
zarechneva-a: I can't agree here: I usually use the latest LLM profiles in the code, hence I would later question why the latest GPT profile is not used here, and this comment will be helpful. I suggest removing the comment when the new GPT (5.3?) is released and its behaviour is more stable.

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/utils/TestUtils.kt
```diff
@@ -68,8 +68,9 @@ object TestUtils {
     fun assertResponseContainsReasoningWithEncryption(response: List<Message>) {
         with(response) {
             shouldNotBeEmpty()
-            shouldForAny { it is Message.Reasoning }
-            with(first { it is Message.Reasoning } as Message.Reasoning) {
+            val reasoningMessage = firstOrNull { it is Message.Reasoning } as? Message.Reasoning
```
skarpovdev: Let's simplify. At the moment it's "find anything, then cast", but `filterIsInstance` will be more idiomatic:
`val reasoningMessage = response.filterIsInstance<Message.Reasoning>().firstOrNull().shouldNotBeNull()`

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClient.kt
```diff
@@ -425,8 +422,11 @@ public open class GoogleLLMClient(
 
         val functionCallingConfig = when (val toolChoice = googleParams.toolChoice) {
             LLMParams.ToolChoice.Auto -> GoogleFunctionCallingConfig(GoogleFunctionCallingMode.AUTO)
+
```
skarpovdev: Are all these empty lines here and below for aesthetics?
I'd keep the changes unrelated to your necessary edits to a minimum and revert these, unless done by a formatter.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClient.kt
```diff
@@ -691,7 +701,7 @@ public open class GoogleLLMClient(
      * @return A list of choices, where each choice is a list of response messages
      */
     private fun processGoogleResponse(response: GoogleResponse): List<List<Message.Response>> {
-        if (response.candidates.isEmpty()) {
+        if (response.candidates?.isEmpty() ?: true) {
```
skarpovdev:
```suggestion
        if (response.candidates.isNullOrEmpty()) {
```

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClient.kt
zarechneva-a: I dived into docs and realized that actually this field shouldn't be nullable. Let me fix it quickly.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/extension/AIAgentNodes.kt
```diff
@@ -450,6 +450,19 @@ public fun AIAgentSubgraphBuilderBase<*, *>.nodeExecuteMultipleToolsAndSendResul
         }
 
         llm.writeSession {
+            // Ensure all originating tool-call messages exist in the prompt before adding results.
```
EugeneTheDev: This sounds like a core behavior fix that should affect how we work with `Prompt` in general. I think a better place for it would be in the [PromptBuilder.ToolMessageBuilder.result](https://github.com/JetBrains/koog/blob/e7c88315a9f24aa35102ef6f3ffbb9090177ed88/prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/dsl/PromptBuilder.kt#L264) itself. As you can see, we already do prompt transformation here, so this one would likely look similar. I suggest implementing this transformation as a first step, before the already existing one.<br>

Also, it's a bit unclear how do problematic cases look like, so elaborating a bit and maybe even providing a few examples in the comments would be very helpful for understanding, why this fix is needed.<br>


very nit: I would say block comments (/**/) would look slightly better and easier to read/edit here, since there are multiple lines of text

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/extension/AIAgentNodes.kt
kpavlov: nit: would be great to avoid the C&P

kpavlov (approved):
Looks good, just minor comment

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClient.kt
zarechneva-a: I formatted the file with klint formatter and the lines were added. Do they really make such a difference? I mean, they increase readability as well.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClient.kt
zarechneva-a: Thanks! I believe this is already committed

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClient.kt
skarpovdev: Weird, ktlint doesn't do it for me and neither does IJ's formatter. I disagree on readability though.

skarpovdev: Not a big deal in this case, but it's generally a good idea to focus the changes on what the PR is about.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/extension/AIAgentNodes.kt
zarechneva-a: Thanks! Done

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/extension/AIAgentNodes.kt
```diff
@@ -471,8 +484,12 @@ public fun AIAgentSubgraphBuilderBase<*, *>.nodeLLMSendMultipleToolResults(
 ): AIAgentNodeDelegate<List<ReceivedToolResult>, List<Message.Response>> =
     node(name) { results ->
         llm.writeSession {
+            // Ensure corresponding tool-call messages are present before adding results.
```
EugeneTheDev: Similar to the above, should be fixed by implementing the logic inside the `call`. Would also avoid logic duplication

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClient.kt
EugeneTheDev: Good point. I would say in this case we can keep them, since these formatting changes are pretty minor, but in the future that's a good idea to keep what you said in mind.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClient.kt
```diff
@@ -364,28 +365,24 @@ public open class GoogleLLMClient(
...
-                GoogleFunctionDeclaration(
-                    name = tool.name,
-                    description = tool.description,
-                    parameters = buildJsonObject {
-                        put("type", "object")
-                        put("properties", JsonObject(properties))
-                        putJsonArray("required") {
-                            addAll(tool.requiredParameters.map { JsonPrimitive(it.name) })
-                        }
+        val googleTools = tools.map { tool ->
```
EugeneTheDev: If I understand it correctly, the changes in this part are only formatting related. And the formatting actually got a bit worse, less readable. Can you please revert it?

prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openai/OpenAILLMClient.kt
```diff
@@ -777,11 +795,14 @@ public open class OpenAILLMClient(
                         metaInfo = metaInfo
                     )
 
-                    is Item.OutputMessage -> Message.Assistant(
-                        content = output.text(),
-                        finishReason = output.status?.name,
-                        metaInfo = metaInfo
-                    )
+                    is Item.OutputMessage -> {
+                        val text = output.text().ifBlank { response.outputText.orEmpty() }
```
EugeneTheDev, line 806: Can you please clarify, why this `ifBlank` was added? AFAIK, according to KDocs, `outputText` is just a convenience field that contains only concatenated text items parts. That's what `text()` already does, so this addition seems redundant to me.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openai/OpenAILLMClient.kt
```diff
@@ -366,7 +366,25 @@ public open class OpenAILLMClient(
...
+            is OpenAIChatParams -> super.executeMultipleChoices(prompt, model, tools)
+
+            is OpenAIResponsesParams -> {
+                // Responses API does not currently expose a native "n" parameter,
+                // so we issue multiple independent responses and aggregate them.
+                // This path is required for models like gpt-5.1-codex that only
+                // support the Responses endpoint and return 404 on Chat Completions.
+                val choices = (params.numberOfChoices ?: 1).coerceAtLeast(1)
+                buildList {
+                    repeat(choices) {
```
EugeneTheDev: You're running requests sequentially here, which introduces a bottleneck. I suggest doing it concurrently instead:
```kotlin
                coroutineScope {
                    List(choices) {
                        async {
                            val response = getResponseWithResponsesAPI(prompt, params, model, tools)
                            processResponsesAPIResponse(response)
                        }
                    }.awaitAll()
                }

```

EugeneTheDev (changes requested):
Thank you for your fixes, I have a few minor comments regarding the fixes in the framework

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClient.kt
zarechneva-a: Eeeeeh, shared my weird formatter behaviour in the DM 🌚

zarechneva-a: UPD: succeeded!

prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openai/OpenAILLMClient.kt
zarechneva-a: > That's what text() already does

Didn't know, thanks! will remove

zarechneva-a: UPD: I removed it and some tests failed. So I remembered why it was added in the first place. It's a fallback because the Responses API provides the final text in two places. So in this case, if `output.text()` is blank, we also check `response.outputText` before passing the empty string as an OutputMessage

EugeneTheDev: Oh, that's why, thank you for clarifying

EugeneTheDev (approved):
Looks good now, thank you

PR #1296, opened 2025-12-22
feat(embeddings): KG-104 Control Embedding Dimensions & Batch Support

### LINKED ISSUE #340: Support dimensions for embedding models
- add dimensions property to embedding request for OpenAI ([link](https://platform.openai.com/docs/api-reference/embeddings/create#embeddings-create-dimensions))
- check if ollama supports `dimensions`

mltheuser:
# feat: Control Embedding Dimensions & Batch Support

## Context

**Tickets:** KG-104, KG-538

Hey team, this PR aims to tackle two requested improvements to koog's embedding capabilities: **Dimension Control** and **Batch Processing**.

## Review wanted
I will stop working on this PR for a bit to give the team time to review the approach and provide feedback. The current state has the generic interface changes I intend to make and an example implementation of these patterns for the `Google` provider. Looking forward to your thoughts!

## Design Philosophy

The implementation heavily follows Koog's established patterns—specifically mimicking the `LLMParams` architecture used for completions. This ensures the API feels familiar and keeps our provider-specific configurations consistent.

### Capability-based Model Declaration
I’ve added a specific capability so we can validate which models support controlling the dimension size at runtime and let callers know what's supported.

```kotlin
public val GeminiEmbedding001: LLModel = LLModel(
    provider = LLMProvider.Google,
    id = "gemini-embedding-001",
    // Explicitly declaring dimension support
    capabilities = listOf(LLMCapability.Embed, LLMCapability.Embedding.Dimensions), 
    // ...
)
```

### Extensible `EmbeddingParams`
Just like `LLMParams`, we now have `EmbeddingParams`. It starts with a base class for common needs (dimensions) and allows providers to extend it for niche features (like Google’s `taskType`).

```kotlin
// Base class
public open class EmbeddingParams(
    public val dimensions: Int? = null,
)

// Provider-specific extension
public class GoogleEmbeddingParams(
    dimensions: Int? = null,
    public val taskType: GoogleEmbeddingTaskType? = null,
    public val title: String? = null,
) : EmbeddingParams(dimensions)
```

### Smart, Backward-Compatible Interface
I’ve updated `LLMEmbeddingProvider` to support batching without breaking existing implementations.

I included a default "polyfill" for `embedBatch`. If a provider doesn't have a native batch endpoint yet, the interface automatically handles parallelizing the requests using coroutines (possible to run into rate limit exceeded though). This lets us roll out the feature interface immediately while updating individual providers incrementally.

```kotlin
interface LLMEmbeddingProvider {
    suspend fun embed(
        text: String,
        model: LLModel,
        params: EmbeddingParams = EmbeddingParams() 
    ): List<Double>
    
    // Default implementation handles concurrency for us
    suspend fun embedBatch(
        texts: List<String>,
        model: LLModel,
        params: EmbeddingParams = EmbeddingParams()
    ): List<List<Double>> = coroutineScope {
        texts.map { async { embed(it, model, params) } }.awaitAll()
    }
}
```

### Polymorphic Parameter Conversion
To keep things type-safe inside the providers, I used the established pattern for converting base params to provider-specific params:

```kotlin
internal fun EmbeddingParams.toGoogleEmbeddingParams(): GoogleEmbeddingParams =
    (this as? GoogleEmbeddingParams) ?: GoogleEmbeddingParams(dimensions = dimensions)
```

## Testing (Draft)
*   Added unit tests for the new parameter mapping.
*   Verified that `embedBatch` defaults correctly to parallel execution for providers without overrides.
*   Tested manually against Google Gemini to ensure dimensions are actually being truncated/resized by the API.


mltheuser:
@eugenekarpenko , @kpavlov , @sdubov would be nice if one of you could do an initial review for this.

embeddings/embeddings-llm/build.gradle.kts
```diff
@@ -25,7 +25,8 @@ kotlin {
         commonTest {
             dependencies {
                 implementation(kotlin("test"))
-                implementation(project(":prompt:prompt-executor:prompt-executor-clients:prompt-executor-openai-client"))
+                // TODO: Re-enable after OpenAI migration
+                // implementation(project(":prompt:prompt-executor:prompt-executor-clients:prompt-executor-openai-client"))
```
sdubov: Would it be possible to add a placeholder for OpenAI for now and skip dimensions (or handle it somehow when explicitly specify dimension capability with OpenAI models) instead of disabling the embedding for OpenAI clients?

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/utils/Models.kt
```diff
@@ -61,14 +61,24 @@ object Models {
 
     @JvmStatic
     fun embeddingModels(): Stream<LLModel> {
+        // TODO: Re-enable after provider migration complete
         return Stream.of(
-            BedrockModels.Embeddings.AmazonTitanEmbedText,
-            OpenAIModels.Embeddings.TextEmbedding3Large,
-            MistralAIModels.Embeddings.MistralEmbed,
+            // BedrockModels.Embeddings.AmazonTitanEmbedText,
```
sdubov: Same comment here. It seems we might loose some base embedding tests if we just exclude these models for now. WDYT?

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleEmbeddingParams.kt
```diff
@@ -0,0 +1,95 @@
...
+ * Specifies the intended use case to help the model produce better embeddings.
+ *
+ * **Polymorphic Usage**: Users can call `embed()` with either:
+ * - Generic `EmbeddingParams(dimensions = 256)` - works with any provider
+ * - Specific `GoogleEmbeddingParams(dimensions = 256, taskType = RETRIEVAL_QUERY)` - Google-specific features
+ *
+ * The conversion function [toGoogleEmbeddingParams] handles both cases transparently.
+ */
+@Serializable
+public enum class GoogleEmbeddingTaskType(public val apiValue: String) {
```
sdubov: Could you please extract this class into a separate file? It seems a separate entity for me that would be easier to teach when it is located separately from parameters. WDYT?

Also, could you please add documentation for `apiValue` parameter? It would be clear to understand and read the code.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClient.kt
```diff
@@ -762,18 +763,35 @@ public open class GoogleLLMClient(
...
+        
+        val googleParams = params.toGoogleEmbeddingParams()
 
         val request = GoogleEmbeddingRequest(
             model = "models/${model.id}",
             content = GoogleContent(
                 parts = listOf(GooglePart.Text(text))
-            )
+            ),
+            outputDimensionality = googleParams.dimensions,
```
sdubov: What do you think about adding parameters as a single parameter to the request instance instead of a separate parameters. Request can process parameters internally in that case. Current request instance does not implement any API that define parameters.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openai/OpenAILLMClient.kt
```diff
@@ -398,10 +398,15 @@ public open class OpenAILLMClient(
...
      * @param model The model to use for embedding. Must have the Embed capability.
+     * @param params Embedding parameters (dimensions support TODO in OpenAI migration)
      * @return A list of floating-point values representing the embedding.
      * @throws IllegalArgumentException if the model does not have the Embed capability.
      */
-    override suspend fun embed(text: String, model: LLModel): List<Double> {
+    override suspend fun embed(
+        text: String,
+        model: LLModel,
+        params: ai.koog.prompt.params.EmbeddingParams
```
sdubov: (Here and in other places above) Could you please use the import for importing `EmbeddingParams` class symbols instead of a FQN here. It help to make code shorter and easy to ready:
```
import ai.koog.prompt.params.EmbeddingParams
...
    override suspend fun embed(
        text: String,
        model: LLModel,
        params: EmbeddingParams
...
```

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/params/EmbeddingParams.kt
```diff
@@ -0,0 +1,43 @@
...
+ *   Only applicable to models that support variable dimensions
+ *   (models with [ai.koog.prompt.llm.LLMCapability.Embedding.Dimensions] capability).
+ *   If null, uses model's default dimension.
+ */
+@Serializable
+public open class EmbeddingParams(
+    public val dimensions: Int? = null,
+) {
+    init {
+        dimensions?.let {
```
sdubov: This is a strange requirement. The class parameters allows to pass null for dimensions, but you prevent null by validation in constructor. I think it is a bit redundant. Please let me know if I miss something.

sdubov (changes requested):
@mltheuser thank you for the contribution! It looks very cool extension for the embedding models capability. I looked through the code and added some comments about different parts of a code. I would also like to provide general proposals with improvements if you do not mind:
1. Would it be possible to split the updates related to "dimensions" and "batch" to two different PRs? Currently, we use the "squash and merge" policy for GitHub when merging PRs. Since there are different changes, it would be better to have them as separate commits.
2. My general feeling is that we need to support embedding capability for models independently from "dimensions" parameter. So, if the model has embedding capability, it still should be available. I would propose to have `EmbeddedParameters` common for all modes with embeddings and add dimensions for Google Parameters only for now. This will allow you to keep a list of models that supports embeddings the same and extend them when a models is updated with dimensions.

mltheuser:
Thanks a lot @sdubov! I will work through your comments soon.

embeddings/embeddings-llm/src/commonTest/kotlin/ai/koog/embeddings/local/LLMEmbedderTest.kt
```diff
@@ -75,15 +79,89 @@ class LLMEmbedderTest {
...
+            Vector(listOf(0.3, 0.4)),
+            Vector(listOf(0.5, 0.6))
+        )
+        
+        // Mock individual embeddings
+        texts.forEachIndexed { i, text -> mockClient.mockEmbedding(text, vectors[i]) }
+        
+        val results = mockClient.embedBatch(texts, model, EmbeddingParams())
+        
+        assertEquals(3, results.size)
```
EugeneTheDev: Looks like it can be simplified to a single assert:
```kotlin
assertEquals(
    listOf(
        listOf(...),
        ...,
    ),
    results,
)
```

prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/bedrock/BedrockLLMClient.kt
```diff
@@ -354,7 +354,11 @@ public class BedrockLLMClient(
         }
     }
 
-    override suspend fun embed(text: String, model: LLModel): List<Double> {
+    override suspend fun embed(
+        text: String,
+        model: LLModel,
+        params: ai.koog.prompt.params.EmbeddingParams
```
EugeneTheDev: Here and in other clients that don't support `EmbeddingParams` yet (I reckon it's all clients except Google), let's also add an additional check and throw if `param.dimensions` is not null. This will provide more clear feedback to the users that this feature is not supported yet

EugeneTheDev (changes requested):
Generally looks good, but I have a few questions. And why did you disable embedding tests for some providers (unit and integration)? We should not break existing functionality, so let's keep these tests to check that

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/params/EmbeddingParams.kt
EugeneTheDev: I think you misread it. This validation will only be triggered when the `dimensions` param is not null because of the `dimensions?.let {}`

prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/bedrock/BedrockLLMClient.kt
mltheuser: Given the feedback I decided to make dimensions a provider specific field of EmbeddingParams. You are right that this reduces coupling and avoids confusion at the cost of a little more code (given that all our providers support custom dimensions for at least on of their models).

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleEmbeddingParams.kt
mltheuser: Done. Also cleaned up the enum and it's description.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClient.kt
mltheuser: I like that idea. I have create a factory method in GoogleEmbeddingRequest for this purpose.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openai/OpenAILLMClient.kt
mltheuser: Ah thanks, I missed that one! Fixed.

embeddings/embeddings-llm/src/commonTest/kotlin/ai/koog/embeddings/local/LLMEmbedderTest.kt
mltheuser: As suggested by sdubov I will remove everything from this PR that is not related to embedding dimensions. But will keep this in mind for the follow up task.

embeddings/embeddings-llm/build.gradle.kts
mltheuser: By making dimensions a provider specific parameter there is no need to migrate all in one go anymore. Re-enabled.

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/utils/Models.kt
mltheuser: Re-enabled. By the way this was never meant to be merged like this I was intending to migrate all providers in one go and just commented them out to get early feedback on a working version.

mltheuser:
Thanks for the feedback @sdubov and @EugeneTheDev!

I've refactored the PR based on your suggestions:

I've removed the batch embedding feature entirely from this PR. It now focuses solely on dimension control, which should make review much easier. I'll follow up with batch embedding in a separate PR.

Exactly as @sdubov suggested! The dimensions parameter is now provider specific, while the base EmbeddingParams stays simple and common to all providers. This way we can add new features incrementally without any breaking changes.

About the disabled tests: Sorry for the confusion! Those were temporarily commented out just to get the code compiling while I was iterating - this was meant for early feedback, not a merge-ready state. I would have migrated all providers before requesting an actual merge. I've now restored them properly.

My plan is: get this foundational dimension control merged (currently just Google), then incrementally add the other providers (OpenAI, Bedrock, etc.) in follow-up PRs, and finally tackle batch embedding as the last piece.

Let me know if this approach works better!

mltheuser:
Is one of you able to edit the PR title in github? It seems I can't since the PR is owned by the koog repo. Would be great if you could prepend "feat" or similar.

mltheuser:
@sdubov @EugeneTheDev please check back when you find the time.

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/utils/Models.kt
```diff
@@ -70,6 +70,15 @@ object Models {
         )
     }
 
+    /**
+     * Returns embedding models that support variable output dimensions via the dimensions parameter.
+     * Only includes models with [LLMCapability.Embedding.Dimensions].
+     */
+    @JvmStatic
+    fun dimensionCapableEmbeddingModels(): Stream<LLModel> {
+        return embeddingModels().filter { it.capabilities.contains(LLMCapability.Embedding.Dimensions) }
```
sdubov: There is a helper method `LLModel#supports()` that make a safe check for model capabilities. I think you can re-use it here.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClient.kt
```diff
@@ -826,4 +825,16 @@ public open class GoogleLLMClient @JvmOverloads constructor(
             )
         }
     }
+
+    private fun validateEmbeddingRequest(model: LLModel, params: GoogleEmbeddingParams) {
+        require(model.capabilities.contains(LLMCapability.Embed)) {
```
sdubov: You should be able to replace `model.capabilities.contains(LLMCapability.Embed)` with the code `model.supports(LLMCapability.Embed)` for consistency.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleEmbeddingParams.kt
```diff
@@ -0,0 +1,97 @@
...
+/**
+ * Google-specific embedding parameters.
+ *
+ * @property dimensions Desired output embedding dimensions (mapped to `outputDimensionality`).
+ *   Only applicable to models with [LLMCapability.Embedding.Dimensions].
+ *   If null, uses model's default dimension.
+ * @property taskType Specifies the intended use case for the embeddings.
+ * @property title Document title (only valid with taskType=RETRIEVAL_DOCUMENT).
+ */
+public class GoogleEmbeddingParams(
```
sdubov, line 45: It seem that that this class need to be serialzable as well.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleEmbeddingParams.kt
```diff
@@ -0,0 +1,97 @@
...
+    public val dimensions: Int? = null,
+    public val taskType: GoogleEmbeddingTaskType? = null,
```
sdubov, line 47: `GoogleEmbeddingTaskType` need to have a `@Serializable` annotation as well.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleEmbeddingParams.kt
```diff
@@ -0,0 +1,97 @@
...
+    public val title: String? = null,
+) : EmbeddingParams() {
+
+    init {
+        dimensions?.let {
+            require(it > 0) { "dimensions must be > 0, but was $it" }
+        }
+        title?.let {
```
sdubov, line 55: This check still looks strange for me. You check the title to require other parameter to be a defined value. Why would anyone call a constructor with `taskType` any, but `GoogleEmbeddingTaskType.RETRIEVAL_DOCUMENT` with this check? Do we expect to receive and deserialize these parameters from LLM?

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/params/EmbeddingParams.kt
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * Base parameters for embedding generation.
+ *
+ * This is an `open class` to allow provider-specific subclasses to add additional
+ * parameters while maintaining polymorphism. This mirrors the [LLMParams] pattern.
+ *
+ * @see LLMParams for the equivalent pattern used in completion/chat models.
+ */
+@Serializable
+public open class EmbeddingParams {
```
sdubov: Sorry, @mltheuser, could you please clarify, why is it implemented as an open class instead of an interface? I see several issue with current approach:
1. Current implementation of the `hashCode` and the `equals` methods breaks the comparison logic as every instance have the same hash code.
2. Why would I need to make an instance of a base `EmbeddingParams` type?

sdubov (changes requested):
Looks good for me in general. The major comment is about the `EmbeddingParams` type. @mltheuser, could you please address to these question? As well as some other failed jobs including conventional PRs job (validates the commit messages structure).

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/params/EmbeddingParams.kt
mltheuser: You are completely right. The main motivation behind this was to keep in line with with the LLMParams class I used for reference. Then again, now after removing the dimensions field from this class it becomes quite apparent that this is not the right patter for the job. I will refactor it to be an interface like you suggested.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleEmbeddingParams.kt
mltheuser: `GoogleEmbeddingParams` itself is never serialized or deserialized. Its fields are manually extracted and placed into `GoogleEmbeddingRequest` (which is @Serializable). See `GoogleEmbeddingRequest.from()`.


`GoogleEmbeddingRequest` is a wire format model (what gets sent to the Google API), while `GoogleEmbeddingParams` is a user-facing model (what developers pass into the SDK) that might contain field names that need to be supported across multiple providers for a more unified api. These two concerns should be decoupled — the naming, types, and structure serve different audiences.

Marking it @Serializable would be misleading since direct serialization would produce incorrect output (wrong field names, wrong structure). The class intentionally isn't serializable.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClient.kt
mltheuser: Good catch! That works, thanks.

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/utils/Models.kt
mltheuser: Done.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleEmbeddingParams.kt
mltheuser: The parameter are given by the user not an LLM. The init block says: "if title is provided, then taskType must be RETRIEVAL_DOCUMENT". This is a constraint from the Google Embedding API. We could also remove the check and let the http request fail but I consider an early sanity check the better approach. I have also played around with some alternative ways to write it but have always come back to the field?.let { doSomeCheck } pattern. We could also replace it with if (field != null) { doSomeCheck } if you prefer it? The checks on init {} should stay though.

mltheuser:
@sdubov , @EugeneTheDev addressed your comments.

aozherelyeva:
Hi @mltheuser, sorry for the delay. I will now close this one as outdated. 
Please rebase to the `develop` and open a new PR in case the change is still actual. Sorry for the long processing 😞

PR #1317, opened 2026-01-04
KG-596 Fix thought signature on tool call with gemini 3.0

mltheuser:
## Motivation and Context

Related to: [KG-596](https://youtrack.jetbrains.com/projects/KG/issues/KG-596)

Gemini 3.0 models enforce stricter validation of [thoughtSignature](file:///Users/ku76uh/Developer/jetbrains/fork/koog/prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/jvmTest/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClientTest.kt#401-429) for function calls. When the model returns parallel tool calls, only the *first* call in a turn receives a [thoughtSignature](file:///Users/ku76uh/Developer/jetbrains/fork/koog/prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/jvmTest/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClientTest.kt#401-429). On subsequent turns, the API expects this signature to be echoed back exactly. Without proper handling, multi-turn agentic conversations with parallel tools fail with cryptic API errors.

**The Problem:** The [GoogleLLMClient](file:///Users/ku76uh/Developer/jetbrains/fork/koog/prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClient.kt#101-844) wasn't preserving [thoughtSignature](file:///Users/ku76uh/Developer/jetbrains/fork/koog/prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/jvmTest/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClientTest.kt#401-429) across turns, and wasn't correctly re-grouping parallel tool calls/results when constructing requests—leading to malformed conversation structures that newer Gemini models reject.

## How It's Solved

1. **Preserve [thoughtSignature](file:///Users/ku76uh/Developer/jetbrains/fork/koog/prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/jvmTest/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClientTest.kt#401-429):** When processing model responses, we now extract the [thoughtSignature](file:///Users/ku76uh/Developer/jetbrains/fork/koog/prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/jvmTest/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClientTest.kt#401-429) from `GooglePart.FunctionCall` and store it in `Message.Tool.Call.metaInfo.metadata`. When building subsequent requests, we restore it.

2. **Correctly batch parallel tool calls/results:** The [createGoogleRequest](file:///Users/ku76uh/Developer/jetbrains/fork/koog/prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClient.kt#280-477) function now uses a buffering strategy to re-group interleaved messages. The key insight: if a tool call has a [thoughtSignature](file:///Users/ku76uh/Developer/jetbrains/fork/koog/prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/jvmTest/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClientTest.kt#401-429), it starts a new turn; if it doesn't, it's a parallel call in the same turn. This lets us batch calls into a single [model](file:///Users/ku76uh/Developer/jetbrains/fork/koog/prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClient.kt#775-805) role [GoogleContent](file:///Users/ku76uh/Developer/jetbrains/fork/koog/prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/models/GoogleGenerateContent.kt#48-53) and results into a single [user](file:///Users/ku76uh/Developer/jetbrains/fork/koog/prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/jvmTest/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClientTest.kt#59-74) role [GoogleContent](file:///Users/ku76uh/Developer/jetbrains/fork/koog/prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/models/GoogleGenerateContent.kt#48-53), as the API requires.

3. **Clean, idiomatic implementation:** The buffering logic uses a [when](file:///Users/ku76uh/Developer/jetbrains/fork/koog/prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/jvmTest/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClientTest.kt#59-74) expression to make the three states explicit (new turn, starting fresh, parallel call), keeping the code readable and maintainable.

## Breaking Changes

None. This is a backward-compatible fix that enables correct behavior with Gemini 3.0+ while remaining compatible with earlier models.

---

#### Type of the changes
- [ ] Documentation update
- [ ] Tests improvement
- [ ] Refactoring

#### Checklist

### DIFF
--- a/integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/utils/Models.kt
+++ b/integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/utils/Models.kt
@@ -35,6 +35,7 @@ object Models {
     @JvmStatic
     fun googleModels(): Stream<LLModel> {
         return Stream.of(
+            GoogleModels.Gemini3_Pro_Preview,
             GoogleModels.Gemini2_5Pro,
             GoogleModels.Gemini2_5Flash,
         )
--- a/prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClient.kt
+++ b/prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClient.kt
@@ -65,6 +65,8 @@ import kotlinx.serialization.json.JsonPrimitive
 import kotlinx.serialization.json.add
 import kotlinx.serialization.json.buildJsonArray
 import kotlinx.serialization.json.buildJsonObject
+import kotlinx.serialization.json.contentOrNull
+import kotlinx.serialization.json.jsonPrimitive
 import kotlinx.serialization.json.put
 import kotlinx.serialization.json.putJsonArray
 import kotlin.uuid.ExperimentalUuidApi
@@ -287,12 +289,25 @@ public open class GoogleLLMClient(
         val systemMessageParts = mutableListOf<GooglePart.Text>()
         val contents = mutableListOf<GoogleContent>()
         val pendingCalls = mutableListOf<GooglePart.FunctionCall>()
+        val pendingResults = mutableListOf<GooglePart.FunctionResponse>()
 
         fun flushCalls() {
             if (pendingCalls.isNotEmpty()) {
                 contents += GoogleContent(role = "model", parts = pendingCalls.toList())
                 pendingCalls.clear()
             }
+        }
+
+        fun flushResults() {
+            if (pendingResults.isNotEmpty()) {
+                contents += GoogleContent(role = "user", parts = pendingResults.toList())
+                pendingResults.clear()
+            }
+        }
+
+        fun flushAll() {
+            flushCalls()
+            flushResults()
         }
 
         for (message in prompt.messages) {
@@ -302,13 +317,13 @@ public open class GoogleLLMClient(
                 }
 
                 is Message.User -> {
-                    flushCalls()
+                    flushAll()
                     // User messages become 'user' role content
                     contents.add(message.toGoogleContent(model))
                 }
 
                 is Message.Assistant -> {
-                    flushCalls()
+                    flushAll()
                     contents.add(
                         GoogleContent(
                             role = "model",
@@ -318,7 +333,7 @@ public open class GoogleLLMClient(
                 }
 
                 is Message.Reasoning -> {
-                    flushCalls()
+                    flushAll()
                     contents.add(
                         GoogleContent(
                             role = "model",
@@ -334,35 +349,49 @@ public open class GoogleLLMClient(
                 }
 
                 is Message.Tool.Result -> {
-                    flushCalls()
-                    contents.add(
-                        GoogleContent(
-                            role = "user",
-                            parts = listOf(
-                                GooglePart.FunctionResponse(
-                                    functionResponse = GoogleData.FunctionResponse(
-                                        id = message.id,
-                                        name = message.tool,
-                                        response = buildJsonObject { put("result", message.content) }
-                                    )
-                                )
+                    // Just buffer results. We only flush when we know the current tool turn is complete.
+                    pendingResults.add(
+                        GooglePart.FunctionResponse(
+                            functionResponse = GoogleData.FunctionResponse(
+                                id = message.id,
+                                name = message.tool,
+                                response = buildJsonObject { put("result", message.content) }
                             )
                         )
                     )
                 }
 
                 is Message.Tool.Call -> {
+                    val signature = message.metaInfo.metadata?.get("thoughtSignature")?.jsonPrimitive?.contentOrNull
+
+                    // Determine whether to flush based on current state and signature presence.
+                    // - signature != null on a subsequent call indicates a new turn (sequential calls)
+                    // - signature == null on a subsequent call indicates parallel calls in the same turn
+                    when {
+                        pendingCalls.isNotEmpty() && signature != null -> {
+                            // New turn: flush the previous turn's calls and results
+                            flushCalls()
+                            flushResults()
+                        }
+                        pendingCalls.isEmpty() -> {
+                            // Starting fresh: ensure no stale results from a previous turn
+                            flushResults()
+                        }
+                        // else: parallel call in same turn, just buffer it
+                    }
+
                     pendingCalls += GooglePart.FunctionCall(
                         functionCall = GoogleData.FunctionCall(
                             id = message.id,
                             name = message.tool,
                             args = json.decodeFromString(message.content)
-                        )
+                        ),
+                        thoughtSignature = signature
                     )
                 }
             }
         }
-        flushCalls()
+        flushAll()
 
         val googleTools = tools
             .map { tool ->
@@ -630,14 +659,29 @@ public open class GoogleLLMClient(
                         }
                     }
 
-                    is GooglePart.FunctionCall -> add(
-                        Message.Tool.Call(
-                            id = Uuid.random().toString(),
-                            tool = part.functionCall.name,
-                            content = part.functionCall.args.toString(),
-                            metaInfo = metaInfo
+                    is GooglePart.FunctionCall -> {
+                        // Store thoughtSignature in metadata to preserve it for subsequent turns
+                        val signature = part.thoughtSignature
+
+                        val callMetaInfo = if (signature != null) {
+                            metaInfo.copy(
+                                metadata = buildJsonObject {
+                                    metaInfo.metadata?.forEach { (key, value) -> put(key, value) }
+                                    put("thoughtSignature", signature)
+                                }
+                            )
+                        } else {
+                            metaInfo
+                        }
+                        add(
+                            Message.Tool.Call(
+                                id = Uuid.random().toString(),
+                                tool = part.functionCall.name,
+                                content = part.functionCall.args.toString(),
+                                metaInfo = callMetaInfo
+                            )
                         )
-                    )
+                    }
 
                     is GooglePart.InlineData -> {
                         val inlineData = part.inlineData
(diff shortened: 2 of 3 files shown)


integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/utils/Models.kt
```diff
@@ -35,6 +35,7 @@ object Models {
     @JvmStatic
     fun googleModels(): Stream<LLModel> {
         return Stream.of(
+            GoogleModels.Gemini3_Pro_Preview,
```
mltheuser, line 38: You will want to include Gemini3 for the sequential and parallel tool calling tests at least.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClient.kt
```diff
@@ -630,14 +659,29 @@ public open class GoogleLLMClient(
...
-                            metaInfo = metaInfo
+                    is GooglePart.FunctionCall -> {
+                        // Store thoughtSignature in metadata to preserve it for subsequent turns
+                        val signature = part.thoughtSignature
+
+                        val callMetaInfo = if (signature != null) {
+                            metaInfo.copy(
+                                metadata = buildJsonObject {
+                                    metaInfo.metadata?.forEach { (key, value) -> put(key, value) }
+                                    put("thoughtSignature", signature)
```
mltheuser: Alternatively we could add an encrypted field to Message.Tool.Call to mirror the pattern for Message.Reasoning. But since though signature preservation is not a thing yet for most providers storing it as metadata seemed more idiomatic to me.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/jvmTest/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClientTest.kt
```diff
@@ -384,4 +397,65 @@ class GoogleLLMClientTest {
         filePart.mimeType shouldBe "application/pdf"
         (filePart.content as AttachmentContent.Binary.Bytes).asBytes() shouldBe fileData
     }
+
+    @Test
+    fun `should preserve thoughtSignature for function calls`() = runTest {
```
mltheuser: This test is nice to have but needs some more setup to work then the other tests in the class. You might prefer removing it to reduce maintenance effort for this test class going forward. The integration tests are also sufficient for catching this type of problem.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClient.kt
```diff
@@ -334,35 +349,49 @@ public open class GoogleLLMClient(
...
+                                id = message.id,
+                                name = message.tool,
+                                response = buildJsonObject { put("result", message.content) }
                             )
                         )
                     )
                 }
 
                 is Message.Tool.Call -> {
+                    val signature = message.metaInfo.metadata?.get("thoughtSignature")?.jsonPrimitive?.contentOrNull
```
kpavlov: 1. Extract the "thoughtSignature" key to a constant

2. Make metadata handling for thoughtSignature defensive (avoid .jsonPrimitive on non-primitives).

```suggestion
                    val signature = message.metaInfo.metadata?.get("thoughtSignature")?.let {
    it as? JsonPrimitive
}?.contentOrNull
```

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/jvmTest/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClientTest.kt
kpavlov: The test is good and useful. Let's also add/update this unit test to verify that multiple `Tool.Result` messages in sequence produce a single user-role content with multiple parts.

Also add tests for:

1. parallel calls (two FunctionCall parts, only first has signature),
2. malformed non-primitive thoughtSignature in metadata to ensure parsing doesn't throw (e.g., metadata contains JSON array/object under thoughtSignature).

kpavlov (approved):
Thank you, @mltheuser, very useful fix. Brings value
I added a couple of comments on test scenarios, and making it more defensive.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClient.kt
```diff
@@ -334,35 +349,49 @@ public open class GoogleLLMClient(
...
+
+                    // Determine whether to flush based on current state and signature presence.
+                    // - signature != null on a subsequent call indicates a new turn (sequential calls)
+                    // - signature == null on a subsequent call indicates parallel calls in the same turn
+                    when {
```
antoniibelyshev: I think, this `when` could be simplified into
`if (signature != null) {
    flushAll()
}`

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClient.kt
```diff
@@ -630,14 +659,29 @@ public open class GoogleLLMClient(
...
-                        Message.Tool.Call(
-                            id = Uuid.random().toString(),
-                            tool = part.functionCall.name,
-                            content = part.functionCall.args.toString(),
-                            metaInfo = metaInfo
+                    is GooglePart.FunctionCall -> {
+                        // Store thoughtSignature in metadata to preserve it for subsequent turns
+                        val signature = part.thoughtSignature
+
+                        val callMetaInfo = if (signature != null) {
```
antoniibelyshev: The thinking is already added as a separate part inside the
```if (part.thoughtSignature != null && part.thought == false)```
statement, and 
I think, we can avoid this by not storing the `thoughtSignature` in the metainfo. To solve the initial problem, we could use the `Reasoning` messages as indicators of the start of a new message from the API, and append those reasoning messages to the following message when creating an API request.

antoniibelyshev (changes requested):
Thank you! Looks almost ready. I saw a couple issues which might need attention.
1. The reasoning inside a function call seems to be duplicated: it is added as a separate `Reasoning` message, and then also as part of the function call.
2. Probably, it's better to not introduce a new field in the `metaInfo`, and just use the `Resoning` messages as delimiters indicating where a new api response starts.
Not sure about both of the points though. I'm open for a discussion here.

Rizzen (approved):

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClient.kt
mltheuser: Great suggestion! I've refactored the implementation to do exactly this.`Message.Reasoning` now acts as the carrier for the thoughtSignature. `processGoogleCandidate` emits it right before the Tool.Call, and `createGoogleRequest` listens for it to trigger the necessary step flushing and signature attachment. Only thing that makes this a bit annoying to work with is the thought summary case but it is manageable. Thanks for pointing this out!

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClient.kt
mltheuser: We now use reasoning messages to indicate a new step which simplifies the flush logic.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClient.kt
mltheuser: As suggested by @antoniibelyshev we store the though signature in reasoning messages now which means we avoid this metadata handling all together.

mltheuser:
Thank you everyone for the in-depth review. I have switched to the pattern where we store thought signatures in reasoning messages as suggested by @antoniibelyshev and I have extended the tests as suggested by @kpavlov (your point about verifying the structural logic (especially for parallel calls) was also spot on)

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/jvmTest/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClientTest.kt
mltheuser: Extended the tests. Since we moved away from the metadata, point 2. is not applicable anymore.

antoniibelyshev (approved):
Thank you! LGTM

jay-anbaric:
Is there somewhere official that the public can track what will be release? We are keen to see this fix being part of the next release as this is stopping us from using gemini 3 models with tool calls.

PR #1323, opened 2026-01-06
Heuristic required tool choice

antoniibelyshev:
## Motivation and Context

Models without the tool choice capability (e.g. most of the ollama models) cannot rely on the `toolChoice` in the model api. This makes some workflows inconsistent, in particular, `subgraphWithTask` does not work with such models since it relies on the `toolChoice` parameter. This pr introduces a workaround for this issue using `LLMBasedToolCallFixProcessor`, which now will use prompts to force LLM to generate a tool when the `toolChoice` is set to `Required`.

## Breaking Changes

---

#### Type of the changes
- [ ] Documentation update
- [ ] Tests improvement
- [x] Refactoring

#### Checklist

### DIFF
--- a/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/session/AIAgentLLMSession.kt
+++ b/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/session/AIAgentLLMSession.kt
@@ -107,7 +107,7 @@ public open class AIAgentLLMSession(
         val promptWithOnlyCallingTools = preparePromptWithToolChoice(LLMParams.ToolChoice.Required)
         val responses = executeMultiple(promptWithOnlyCallingTools, tools)
         return responses.firstOrNull { it is Message.Tool.Call }
-            ?: error("requestLLMOnlyCallingTools expected at least one Tool.Call but received: ${responses.map { it::class.simpleName }}")
+            ?: responses.first { it !is Message.Reasoning }
     }
 
     public open override suspend fun requestLLMMultipleOnlyCallingTools(): List<Message.Response> {
--- a/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/session/AIAgentLLMSession.kt
+++ b/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/session/AIAgentLLMSession.kt
@@ -95,31 +95,33 @@ public open class AIAgentLLMSession(
         return executeMultiple(promptWithDisabledTools, emptyList()).first { it !is Message.Reasoning }
     }
 
-    private fun preparePromptWithToolChoice(toolChoice: LLMParams.ToolChoice) =
-        prompt.withUpdatedParams {
-            this.toolChoice = toolChoice
-        }
-
     public open override suspend fun requestLLMOnlyCallingTools(): Message.Response {
         validateSession()
-        // We use the multiple-response method to ensure we capture all context (e.g. thinking)
-        // even though we only return the specific tool call.
-        val promptWithOnlyCallingTools = preparePromptWithToolChoice(LLMParams.ToolChoice.Required)
+        val promptWithOnlyCallingTools = prompt.withUpdatedParams {
+            toolChoice = LLMParams.ToolChoice.Required
+        }
         val responses = executeMultiple(promptWithOnlyCallingTools, tools)
+
+        // some models might fail to produce a tool call
+        // it's better to not fail here and allow the user to handle that
         return responses.firstOrNull { it is Message.Tool.Call }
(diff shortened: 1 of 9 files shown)


agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/session/AIAgentLLMSession.kt
```diff
@@ -94,29 +95,31 @@ public open class AIAgentLLMSession(
         return executeMultiple(promptWithDisabledTools, emptyList()).first { it !is Message.Reasoning }
     }
 
+    private fun preparePromptWithToolChoice(toolChoice: LLMParams.ToolChoice) =
```
EugeneTheDev: IMHO this block of code looks small and clear enough to just inline it, as it was before, without dedicated method

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/session/AIAgentLLMSession.kt
```diff
@@ -94,29 +95,31 @@ public open class AIAgentLLMSession(
...
     public open override suspend fun requestLLMOnlyCallingTools(): Message.Response {
         validateSession()
         // We use the multiple-response method to ensure we capture all context (e.g. thinking)
         // even though we only return the specific tool call.
-        val responses = requestLLMMultipleOnlyCallingTools()
+        val promptWithOnlyCallingTools = preparePromptWithToolChoice(LLMParams.ToolChoice.Required)
+        val responses = executeMultiple(promptWithOnlyCallingTools, tools)
         return responses.firstOrNull { it is Message.Tool.Call }
-            ?: error("requestLLMOnlyCallingTools expected at least one Tool.Call but received: ${responses.map { it::class.simpleName }}")
+            ?: responses.first { it !is Message.Reasoning }
```
EugeneTheDev: I understand the idea here: for weak models, if they failed to produce a tool call, you want to return the first assistant message. Can you make a few adjustments:
1. Doesn't this condition essentially mean `first { it is Message.Assistant }`? Can you rewrite it like this to make the intention clearer?
2. Add explanatory comment, otherwise it might look a bit confusing, why do we want assistant message in tools-only request

agents/agents-core/src/commonTest/kotlin/ai/koog/agents/core/agent/session/AIAgentLLMWriteSessionTest.kt
```diff
@@ -392,6 +395,8 @@ class AIAgentLLMWriteSessionTest {
     }
 
     @Test
+    // This behavior is not supported yet.
+    @Ignore
```
EugeneTheDev, line 402: Can you explain, why did you disable this test, why did it break?

EugeneTheDev (changes requested):

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/session/AIAgentLLMSession.kt
antoniibelyshev: True. Initially I wanted to add more functionality here, but then decided not to do that :slightly_smiling_face: I will remove it

agents/agents-core/src/commonTest/kotlin/ai/koog/agents/core/agent/session/AIAgentLLMWriteSessionTest.kt
antoniibelyshev: The test assumes that when the `llm.readSession { requestLLMOnlyCallingTools() }` is called, all reasoning responses are added to the prompt to improve later LLM calls. However, this behavior was incorrectly implemented: in the previous implementation, *all* responses from the LLM were added to the prompt, while only one tool call was returned from the request, leaving unhandled tool calls in the message history. I updated the implementation, so that only the returned message is saved to the prompt, but it left reasoning messages out of the prompt.

The issue with the reasoning messages is related not only to the `requestLLMOnlyCallingTools`: many requestLLM methods returning a single message actually drop the reasoning messages. So I decided to leave that out of the scope of this pr. Moreover, I saw that we are planning to deprecate the single-message responses (at least, we have such task in our planning table), and decided not to create a ticket for fixing the issue with the reasoning messages.

EugeneTheDev: Got it. Can you elaborate in the comment a bit, then, to leave these details there as well. And let's add TODO to remove this test in the future, if single message responses will be dropped?

EugeneTheDev (approved):

PR #1366, opened 2026-01-16
feat(agents): Add java-style chained builders for agent-planner components

antoniibelyshev:
Added java-style chained builders for planners:
    - `PlannerAIAgentBuilder` for creating planner agents
    - Builders for creating `GOAP` components

Added `JavaAIAgentPlanner` for synchronous implementations of the `AIAgentPlanner` in Java

closes [KG-642](https://youtrack.jetbrains.com/issue/KG-642)

### DIFF
--- a/agents/agents-planner/src/commonMain/kotlin/ai/koog/agents/planner/AIAgentPlanner.kt
+++ b/agents/agents-planner/src/commonMain/kotlin/ai/koog/agents/planner/AIAgentPlanner.kt
@@ -5,6 +5,7 @@ import ai.koog.agents.core.agent.context.AIAgentFunctionalContext
 import ai.koog.agents.core.agent.exception.AIAgentMaxNumberOfIterationsReachedException
 import io.github.oshai.kotlinlogging.KotlinLogging
 import kotlin.reflect.KType
+import kotlin.reflect.typeOf
 
 /**
  * An abstract base planner component, which can be used to implement different types of AI agent planner execution flows.
@@ -16,11 +17,19 @@ import kotlin.reflect.KType
  * 2. Execute a step in the plan: [executeStep]
  * 3. Repeat steps 1 and 2 until the plan is considered completed. Then the final [State] is returned.
  *
- * @property stateType [KType] of the [State].
+ * @param stateType [KType] of the [State].
  */
 public abstract class AIAgentPlanner<State, Plan>(
-    public val stateType: KType,
+    // FIXME: require the type explicitly when we decide, what to do with it in Java API
+    stateType: KType? = null,
 ) {
+    /**
+     * [KType] of the [State]
+     */
+    public val stateType: KType = stateType ?: typeOf<Any?>().also {
+        logger.warn { "State type is not specified, some agent features might behave unexpectedly." }
+    }
+
     private companion object {
         private val logger = KotlinLogging.logger { }
     }
--- a/agents/agents-planner/src/commonMain/kotlin/ai/koog/agents/planner/PlannerAIAgent.kt
+++ b/agents/agents-planner/src/commonMain/kotlin/ai/koog/agents/planner/PlannerAIAgent.kt
@@ -19,6 +19,7 @@ import ai.koog.agents.core.tools.ToolRegistry
 import ai.koog.prompt.executor.model.PromptExecutor
 import io.github.oshai.kotlinlogging.KotlinLogging
 import kotlinx.datetime.Clock
+import kotlin.jvm.JvmStatic
 
 /**
  * Represents an instance of planner agent using [AIAgentPlannerStrategy].
@@ -47,8 +48,18 @@ public class PlannerAIAgent<State, Plan>(
     logger = logger,
     id = id,
 ) {
-    private companion object {
+    /**
+     * Companion object providing the static `builder()` method.
+     */
+    public companion object {
         private val logger = KotlinLogging.logger {}
+
+        /**
+         * Creates a new instance of [PlannerAIAgentBuilder] for configuring and building a planner AI agent.
+         */
+        @JvmStatic
+        public fun <State, Plan> builder(strategy: AIAgentPlannerStrategy<State, Plan>): PlannerAIAgentBuilder<State, Plan> =
+            PlannerAIAgentBuilder(strategy)
     }
 
     override val pipeline: AIAgentFunctionalPipeline = AIAgentFunctionalPipeline(agentConfig, clock)
--- a/agents/agents-planner/src/commonMain/kotlin/ai/koog/agents/planner/goap/ActionBuilderApi.kt
+++ b/agents/agents-planner/src/commonMain/kotlin/ai/koog/agents/planner/goap/ActionBuilderApi.kt
@@ -1,0 +1,41 @@
+package ai.koog.agents.planner.goap
+
+/**
+ * API for building [Action] instances.
+ */
+public interface ActionBuilderApi<State> {
+    /**
+     * Sets the name of the action.
+     */
+    public fun name(name: String): ActionBuilderApi<State>
+
+    /**
+     * Sets the description of the action.
+     */
+    public fun description(description: String?): ActionBuilderApi<State>
+
+    /**
+     * Sets the precondition for the action.
+     */
+    public fun precondition(precondition: Condition<State>): ActionBuilderApi<State>
+
+    /**
+     * Sets the belief for the action.
+     */
+    public fun belief(belief: Belief<State>): ActionBuilderApi<State>
+
+    /**
+     * Sets the cost function for the action.
+     */
+    public fun cost(cost: Cost<State>): ActionBuilderApi<State>
+
+    /**
+     * Sets the execute function for the action.
+     */
+    public fun executeAsync(execute: Execute<State>): ActionBuilderApi<State>
+
+    /**
+     * Builds the [Action].
+     */
+    public fun build(): Action<State>
+}
--- a/agents/agents-planner/src/commonMain/kotlin/ai/koog/agents/planner/goap/ActionBuilderApi.kt
+++ b/agents/agents-planner/src/commonMain/kotlin/ai/koog/agents/planner/goap/ActionBuilderApi.kt
@@ -32,7 +32,7 @@ public interface ActionBuilderApi<State> {
     /**
      * Sets the execute function for the action.
      */
-    public fun executeAsync(execute: Execute<State>): ActionBuilderApi<State>
+    public fun execute(execute: Execute<State>): ActionBuilderApi<State>
 
     /**
      * Builds the [Action].
--- a/agents/agents-planner/src/commonMain/kotlin/ai/koog/agents/planner/goap/Entities.kt
+++ b/agents/agents-planner/src/commonMain/kotlin/ai/koog/agents/planner/goap/Entities.kt
@@ -1,18 +1,37 @@
 package ai.koog.agents.planner.goap
 
 import ai.koog.agents.core.agent.context.AIAgentFunctionalContext
+import kotlin.jvm.JvmStatic
+
+public typealias Condition<State> = (State) -> Boolean
+public typealias Belief<State> = (State) -> State
+public typealias Cost<State> = (State) -> Double
+public typealias Execute<State> = suspend (AIAgentFunctionalContext, State) -> State
+public typealias GoalValue = (Double) -> Double
 
 /**
  * Represents an action that can be performed by the agent.
  */
 public class Action<State>(
     public val name: String,
-    public val description: String? = null,
-    public val precondition: (State) -> Boolean,
-    public val belief: (State) -> State,
-    public val cost: (State) -> Double,
-    public val execute: suspend (AIAgentFunctionalContext, State) -> State
-)
+    public val description: String?,
+    public val precondition: Condition<State>,
+    public val belief: Belief<State>,
+    public val cost: Cost<State>,
+    public val execute: Execute<State>
+) {
+
+    /**
+     * Companion object for builder api.
+     */
+    public companion object {
+        /**
+         * Returns a new [ActionBuilder].
+         */
+        @JvmStatic
+        public fun <State> builder(): ActionBuilder<State> = ActionBuilder()
+    }
+}
 
 /**
  * Represents a goal that the agent wants to achieve.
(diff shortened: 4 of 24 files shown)


agents/agents-planner/src/androidMain/kotlin/ai/koog/agents/planner/AIAgentPlannerBuilder.kt
```diff
@@ -0,0 +1,22 @@
+@file:Suppress("MissingKDocForPublicAPI", "EXPECT_ACTUAL_CLASSIFIERS_ARE_IN_BETA_WARNING")
+
+package ai.koog.agents.planner
+
+import kotlin.reflect.KType
+
+public actual class AIAgentPlannerBuilder<State, Plan> actual constructor(
+    stateType: KType?
+) : AIAgentPlannerBuilderApi<State, Plan> {
+    private val delegate = AIAgentPlannerBuilderImpl<State, Plan>(stateType)
```
Ololoshechkin: Why not just `by` delegation?
```
public actual class AIAgentPlannerBuilder<State, Plan> actual constructor(
    stateType: KType?
) : AIAgentPlannerBuilderApi<State, Plan> by AIAgentPlannerBuilderImpl<State, Plan>(stateType)
```
And without overrides inside.

agents/agents-core/src/jvmMain/kotlin/ai/koog/agents/core/utils/CoroutineUtils.kt
```diff
@@ -54,3 +56,9 @@ public suspend fun <T> AIAgentConfig.submitToMainDispatcher(block: () -> T): T {
 
     return result.await()
 }
+
+@OptIn(InternalAgentsApi::class)
+public suspend operator fun <C : AIAgentContext, I, O> BiFunction<C, I, O>.invoke(context: C, input: I): O =
```
EugeneTheDev: What is the purpose of this function? If it's part of the internal api, shouldn't there be an internal api annotation? Also, making it an operator instead of a regular function makes it harder to understand the purpose and the effect of such a function. So I suggest making it a regular extension function with descriptive name, like these above

agents/agents-planner/src/androidMain/kotlin/ai/koog/agents/planner/goap/Builders.kt
```diff
@@ -0,0 +1,36 @@
+@file:Suppress("MissingKDocForPublicAPI", "EXPECT_ACTUAL_CLASSIFIERS_ARE_IN_BETA_WARNING")
+
+package ai.koog.agents.planner.goap
+
+public actual class ActionBuilder<State> actual constructor() : ActionBuilderApi<State> {
+    private val delegate = ActionBuilderImpl<State>()
```
EugeneTheDev: I think you can reduce the amount of duplicated code in all of your builders. Take a look at how `AIAgentContext` and its implementations are implemented. You can move `delegate` to the constructor, and then use `ActionBuilderApi<State> by delegate` in the builder class declaration, thus avoiding the need to repeat all these methods in the class body explicitly by hand. BTW `apply` doesn't have any effect in the method bodies.

agents/agents-planner/src/commonMain/kotlin/ai/koog/agents/planner/goap/GOAPPlannerBuilder.kt
```diff
@@ -26,14 +26,21 @@ public class GOAPPlannerBuilder<State>(
...
     ) {
         actions.add(Action(name, description, precondition, belief, cost, execute))
     }
 
+    /**
+     * Defines an action available to the GOAP agent.
+     *
+     * Returns the builder instance for chained calls.
+     */
+    public fun action(action: Action<State>): GOAPPlannerBuilder<State> = apply { actions.add(action) }
```
EugeneTheDev: Please mark such "for Java" methods with `@JavaAPI`

agents/agents-planner/src/commonMain/kotlin/ai/koog/agents/planner/PlannerAIAgent.kt
```diff
@@ -47,8 +48,18 @@ public class PlannerAIAgent<State, Plan>(
...
+    /**
+     * Companion object providing the static `builder()` method.
+     */
+    public companion object {
         private val logger = KotlinLogging.logger {}
+
+        /**
+         * Creates a new instance of [PlannerAIAgentBuilder] for configuring and building a planner AI agent.
+         */
+        @JvmStatic
```
EugeneTheDev, line 60: Let's add `JavaAPI`

agents/agents-planner/src/commonMain/kotlin/ai/koog/agents/planner/goap/Entities.kt
```diff
@@ -1,18 +1,36 @@
 package ai.koog.agents.planner.goap
 
 import ai.koog.agents.core.agent.context.AIAgentFunctionalContext
+import kotlin.jvm.JvmStatic
+
+public typealias Condition<State> = suspend AIAgentFunctionalContext.(State) -> Boolean
```
EugeneTheDev: Extension lambas are not Java-friendly, unfortunately. Previously it was implemented with context as an explicit first argument. Let's update the type as it was before, with context as argument, not as a receiver

agents/agents-planner/src/commonTest/kotlin/ai/koog/agents/planner/GOAPPlannerAgentTest.kt
```diff
@@ -79,9 +79,9 @@ class GOAPPlannerAgentTest {
         val initialState = SimpleState()
         val finalState = agent.run(initialState)
 
-        assertTrue(finalState.hasKey, "Agent should have obtained the key")
-        assertTrue(finalState.doorUnlocked, "Agent should have unlocked the door")
-        assertTrue(finalState.treasureFound, "Agent should have found the treasure")
+        finalState.hasKey.shouldBeTrue()
```
EugeneTheDev: You're missing helpful string descriptions here now. Let's maybe leave it as it was, or wrap it in `withClue` with the same explanation (although the second option would look more cumbersome, IMHO)

agents/agents-planner/src/commonMain/kotlin/ai/koog/agents/planner/AIAgentPlanner.kt
```diff
@@ -16,13 +18,28 @@ import kotlin.reflect.KType
  * 2. Execute a step in the plan: [executeStep]
  * 3. Repeat steps 1 and 2 until the plan is considered completed. Then the final [State] is returned.
  *
- * @property stateType [KType] of the [State].
+ * @param stateType [KType] of the [State].
  */
 public abstract class AIAgentPlanner<State, Plan>(
-    public val stateType: KType,
+    stateType: KType? = null,
```
EugeneTheDev, line 24: This might break our features, if the `stateType` won't match the actual type of the state. I understand the problem, there's no `KType` in Java. And since this is needed only for serialization, I plan to take a look at how we can remove KTypes completely. But for now, let's at least put a `FIXME` comment here with warning, that this won't work with some of our features that require serialization

agents/agents-planner/src/commonMain/kotlin/ai/koog/agents/planner/AIAgentPlannerBuilderApi.kt
```diff
@@ -0,0 +1,45 @@
...
+import ai.koog.agents.core.agent.context.AIAgentFunctionalContext
+
+public typealias BuildPlan<State, Plan> = suspend AIAgentFunctionalContext.(State, Plan?) -> Plan
+public typealias ExecuteStep<State, Plan> = suspend AIAgentFunctionalContext.(State, Plan) -> State
+public typealias IsPlanCompleted<State, Plan> = suspend AIAgentFunctionalContext.(State, Plan) -> Boolean
+
+/**
+ * API for building [AIAgentPlanner] instances.
+ */
+public interface AIAgentPlannerBuilderApi<State, Plan> {
```
EugeneTheDev: I don't think we need builders for `AIAgentPlanner`. The idea was that if you need your own planner, you implement an interface. Since a planner can be more complicated in terms of logic, it's more clear and convenient to use such an approach with interface implementation instead of builders with lambdas. So I suggest removing all these builders.

EugeneTheDev (changes requested):
Let's also mark all builder `Api` classes with `JavaAPI`, to make it clear that these are mostly designed to be used with the Java style

agents/agents-planner/src/commonMain/kotlin/ai/koog/agents/planner/goap/Entities.kt
antoniibelyshev: I though that this function would not be available for the Java API anyway, since it's a suspend function, so I might as well make the context a receiver, not a parameter, since that would be more convenient for Kotlin users. Am I wrong here?

agents/agents-planner/src/commonMain/kotlin/ai/koog/agents/planner/AIAgentPlannerBuilderApi.kt
antoniibelyshev: Is it possible to implement `suspend fun` in Java? That was my concern, I thought it's not possible. If it is, the builder is indeed not needed

agents/agents-planner/src/commonMain/kotlin/ai/koog/agents/planner/goap/Entities.kt
EugeneTheDev: `suspend` functions are still accessible, they just have this additional "continuation" parameter on the Java side. But I remeber we discussed some time ago with Vadim that extension lambdas somehow turn out pretty inconvenient on the Java side. And even without Java, in pure Kotlin, extension lambdas are not always convenient in terms of API discoverability. E.g. I find myself often typing `this.` to see avaliable options in autocomplete (and this is not only with Koog, but with Kotlin functions that use such a pattern in general). So it's better to make it more discoverable and convenient with explicit argument instead of implicit receiver here, IMHO. And as a bonus point we get to be a tad more Java friendly.

agents/agents-planner/src/commonMain/kotlin/ai/koog/agents/planner/AIAgentPlannerBuilderApi.kt
EugeneTheDev: Ah yes, good point. I faced the same problem when trying to make class tools more Java-friendly. It is possible to implement such functions in Java, however, there's a bug in Kotlin plugin in IntelliJ currently that falsely reports such code as red. So it compiles and runs, just the devex is not that nice. And since it's related to what I'm already thinking about with tools, let's keep it `suspend` for now, again, adding `FIXME` clarifying comment here. And then we'll tackle this common problem with overriding suspendable functions later

agents/agents-planner/src/androidMain/kotlin/ai/koog/agents/planner/goap/Builders.kt
antoniibelyshev: I think, that's not possible with builders. The problem is that `ActionBuilderImpl` methods return the `ActionBuilderImpl` isntance, which is incompatible with `ActionBuilder` return types, which are instances of `ActionBuilder`. And actually you need `apply` here to return `this`, and not the `delegate` instance.

agents/agents-planner/src/commonMain/kotlin/ai/koog/agents/planner/AIAgentPlannerBuilderApi.kt
antoniibelyshev: Ok, it's possible, but is it convenient? I do not think we should remove `suspend`, I think we should keep the builder :slightly_smiling_face:

EugeneTheDev: Using builder for complex logic is less convenient, IMHO. Especially in Java, where lambdas are not first-class citizens. We should not remove suspend either. I would suggest leaving the interface as it was before. I don't think creating custom planners would fall under the "basic usage" anyways, since we provide GOAP and simple planner already. And in the meantime we can think what we can do about this "red code" when overriding suspendable methods in Java, since again, the problem is not unique and I will have to think about it too for tool classes.

agents/agents-planner/src/jvmMain/java/ai/koog/agents/planner/JavaAIAgentPlanner.java
```diff
@@ -0,0 +1,61 @@
...
+    );
+
+    abstract protected Boolean isPlanCompleted(
+        AIAgentFunctionalContext context,
+        State state,
+        Plan plan
+    );
+
+    @Override
+    protected Plan buildPlan(
```
EugeneTheDev: nit: it's better to put `final` to such overrides, so that they won't be overriden by accident by the users

agents/agents-planner/src/commonMain/kotlin/ai/koog/agents/planner/goap/ActionBuilderApi.kt
```diff
@@ -0,0 +1,41 @@
...
+
+    /**
+     * Sets the cost function for the action.
+     */
+    public fun cost(cost: Cost<State>): ActionBuilderApi<State>
+
+    /**
+     * Sets the execute function for the action.
+     */
+    public fun executeAsync(execute: Execute<State>): ActionBuilderApi<State>
```
EugeneTheDev: You're naming it `executeAsync` because for the JVM target you also have non-suspendable `execute`. But let's do this the other way around? Let's call this one `execute`, and then for Java-specific API we can have `executeSync` that accepts non-suspendable labmda

agents/agents-planner/src/jvmMain/java/ai/koog/agents/planner/JavaAIAgentPlanner.java
```diff
@@ -0,0 +1,61 @@
+package ai.koog.agents.planner;
+
+import ai.koog.agents.core.agent.context.AIAgentFunctionalContext;
+import kotlin.coroutines.Continuation;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public abstract class JavaAIAgentPlanner<State, Plan> extends AIAgentPlanner<State, Plan> {
```
EugeneTheDev, line 10: Let's annotate it with `@JavaAPI`

EugeneTheDev (approved):
A few minor comments, looks good now

EugeneTheDev:
Also, can you please make the PR title a bit more descriptive

PR #1381, opened 2026-01-19
feat(agents): support OpenTelemetry metrics

Fant1k34:
## Motivation and Context

PR introduces metrics support based on OpenTelemetry specification. Such metrics allow their listeners observe the behaviour of agent in run-time. All changes are done within OpenTelemetry feature

[Link to issue](https://youtrack.jetbrains.com/issue/KG-136/Add-Open-Telemetry-metrics-support-in-Koog)

Added configuration:
- `addMeterExporter(MetricExporter)` -- function within OpenTelemetry feature that accept MetricExporter and sent library metrics via such exporter

Example of OpenTelemetry feature configuration:
```
install(OpenTelemetry) {
    setServiceInfo(
        "calculator",
        "0.0.1"
    )
    addResourceAttributes(
        mapOf(AttributeKey.stringKey("service.instance.id") to "run-1")
    )
    addMeterExporter(
        OtlpGrpcMetricExporter.builder()
            .setEndpoint("http://localhost:17011")
            .setTimeout(1, TimeUnit.SECONDS)
            .build()
        )
    }
```

Added metrics:
- gen_ai.client.token.usage -- see [here](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-metrics/#metric-gen_aiclienttokenusage)
- gen_ai.client.operation.duration -- see [here](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-metrics/#metric-gen_aiclientoperationduration)
- gen_ai.client.tool.count -- custom metric allows observe tool calls and their statuses 

Visualisation (via OpenTelemetry plugin):
telegram-cloud-photo-size-2-5438313832905380635-y
image

## Breaking Changes

---

#### Type of the changes
- [ ] Documentation update
- [ ] Tests improvement
- [ ] Refactoring
- [ ] CI/CD changes
- [ ] Dependencies update

#### Checklist


docs/docs/opentelemetry-support.md
```diff
@@ -350,30 +395,95 @@ The following event types are supported in line with OpenTelemetry's [Semantic c
...
+
+### Available metrics
+
+- gen_ai.client.token.usage
+    - Instrument: Counter (LongCounter)
+    - Unit: `token`
+    - Description: Total token count per operation and token type
+    - When emitted: after an LLM call finishes; recorded separately for input and output tokens
+    - Key attributes:
+        - `gen_ai.operation.name` (required)
```
egorklimov: (?) chat; generate_content; text_completion

docs/docs/opentelemetry-support.md
```diff
@@ -350,30 +395,95 @@ The following event types are supported in line with OpenTelemetry's [Semantic c
...
+      conventions, [gen_ai.client.token.usage](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-metrics/#metric-gen_aiclienttokenusage)
+
+- gen_ai.client.operation.duration
+    - Instrument: Histogram (DoubleHistogram)
+    - Unit: seconds (`s`)
+    - Description: Operation duration distribution; in Koog it is recorded for tool calls
+    - Recommended explicit bucket boundaries (seconds): 0.01, 0.02, 0.04, 0.08, 0.16, 0.32, 0.64, 1.28, 2.56, 5.12,
+      10.24, 20.48, 40.96, 81.92
+    - Key attributes:
+        - `gen_ai.operation.name` (required)
```
egorklimov: + execute_tool

docs/docs/opentelemetry-support.md
```diff
@@ -350,30 +395,95 @@ The following event types are supported in line with OpenTelemetry's [Semantic c
...
+      conventions, [gen_ai.client.operation.duration](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-metrics/#metric-gen_aiclientoperationduration)
+
+- gen_ai.client.tool.count
+    - Instrument: Counter (LongCounter)
+    - Unit: `tool call`
+    - Description: Custom metric that counts tool calls and their statuses
+    - When emitted: on tool completion, failure, or validation failure
+    - Key attributes:
+        - `gen_ai.operation.name` (required) — `EXECUTE_TOOL`
+        - `gen_ai.tool.name` (recommended)
```
egorklimov: We should be cautious with dynamic values since they can affect the cardinality and performance of the system, as well as the monitoring backend. [A general recommendation](https://prometheus.io/docs/practices/instrumentation/#do-not-overuse-labels) is that cardinality is not critical if it remains below 100. The current metric cardinality is approximately $3 + 3*n$, where $n$ is the number of tools.

To address this, let's add a method to filter tool names for the attribute, enabling us to limit the cardinality. Additionally, let's document this approach.

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/attribute/GenAIAttributes.kt
```diff
@@ -358,6 +383,19 @@ internal object SpanAttributes {
                 override val key: String = super.key.concatKey("result")
                 override val value: HiddenString = HiddenString(result.toString())
             }
+
+            // gen_ai.tool.status
```
sdubov: Why not to report this as a custom attribute? or a special koog attribute. I think it would be cleaner to easier to separate attribute if we keep `gen_ai` attributes consistent with ones defined in the Open Telemetry documentation. WDYT?

examples/simple-examples/src/main/kotlin/ai/koog/agents/example/calculator/Calculator.kt
```diff
@@ -53,9 +60,33 @@ suspend fun main() {
                     println("Result: ${eventContext.result}")
                 }
             }
+
+            install(OpenTelemetry) {
```
sdubov: This seems for me unrelated to the Open Telemetry metrics updates. Is it a leftover from testing?

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/attribute/GenAIAttributes.kt
```diff
@@ -48,8 +48,10 @@ import kotlinx.serialization.json.putJsonArray
  * - gen_ai.tool.call.id (recommended)
  * - gen_ai.tool.description (recommended)
  * - gen_ai.tool.name (recommended)
+ * - gen_ai.token.type (required)
+ * - gen_ai.tool.status (custom)
```
sdubov: Please see the comment about the custom attribute below.

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetry.kt
```diff
@@ -546,6 +564,45 @@ public class OpenTelemetry {
                     span = inferenceSpan,
                     path = patchedExecutionInfo
                 )
+
+                eventContext.responses.lastOrNull()?.let { message ->
+                    message.metaInfo.inputTokensCount?.let { inputTokensCount ->
+                        tokensCounter.add(
+                            inputTokensCount.toLong(),
+                            Attributes.builder()
```
sdubov: Would not it be safer to and easier to use existing wrapper and converter from Koog Attributes to SDK attributes using PII protection and auto type conversion here. Something like this:
```
...
listOf(
  GenAIAttributes.Operation.Name(GenAIAttributes.Operation.OperationNameType.TEXT_COMPLETION),
  GenAIAttributes.Provider.Name(provider),
  GenAIAttributes.Token.Type(GenAIAttributes.Token.TokenType.INPUT),
  GenAIAttributes.Response.Model(eventContext.model)
).toSdkAttributes(config.isVerbose)
...
```

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetry.kt
```diff
@@ -546,6 +564,45 @@ public class OpenTelemetry {
                     span = inferenceSpan,
                     path = patchedExecutionInfo
                 )
+
+                eventContext.responses.lastOrNull()?.let { message ->
+                    message.metaInfo.inputTokensCount?.let { inputTokensCount ->
+                        tokensCounter.add(
```
sdubov: What do you think about extracting this logic a separate extension method (similar to builder methods to start and end spans) that will receive necessary parameters from `eventContext` and create attributes. It will make the feature code a bit cleaner. For example:
```
fun LongCounter.addInputTokensCounter(
    config: OpenTelemetryConfig, 
    inputTokensCount: Long,
    provider: LLMProvider,
    model: LLModel
) {
    this.add(
        inputTokensCount.toLong(),
        listOf<Attribute>(
            GenAIAttributes.Operation.Name(GenAIAttributes.Operation.OperationNameType.TEXT_COMPLETION),
            GenAIAttributes.Provider.Name(provider),
            GenAIAttributes.Token.Type(GenAIAttributes.Token.TokenType.INPUT),
            GenAIAttributes.Response.Model(eventContext.model)
        ).toSdkAttributes(config.isVerbose)
}

fun LongCounter.addInputTokensCounter(...) { ... }

```

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/metric/TokenCounter.kt
```diff
@@ -0,0 +1,24 @@
+package ai.koog.agents.features.opentelemetry.metric
```
sdubov: Would you expect other methods here? I would rename it to `tokenCounter.kt` to not confuse when new methods are added here as this file is to aggregate different token counter top-level methods.

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/metric/ToolCallCounter.kt
```diff
@@ -0,0 +1,25 @@
+package ai.koog.agents.features.opentelemetry.metric
```
sdubov: Same comment about the file name.

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/metric/OperationDurationHistogram.kt
```diff
@@ -0,0 +1,28 @@
+package ai.koog.agents.features.opentelemetry.metric
```
sdubov: Same comment about the file name.

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/metric/GenAIMetricNames.kt
```diff
@@ -0,0 +1,56 @@
...
+
+            object Count : Tool {
+                override val name: String
+                    get() = super.name.concatKey("count")
+
+                override val description: String
+                    get() = "Tool calls count"
+
+                override val unit: String
+                    get() = "tool call"
```
sdubov: unit is "tool call"?

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetry.kt
```diff
@@ -689,6 +796,30 @@ public class OpenTelemetry {
                     span = executeToolSpan,
                     path = patchedExecutionInfo
                 )
+
+                val failedToolCall = eventCallStorage.endEventCallAndReturn(eventContext.eventId)
+
+                failedToolCall?.let { toolCall ->
+                    toolCall.getDurationSec()?.let { sec ->
+                        toolCallDurationHistogram.record(
```
sdubov: Maybe it worth combining those two methods end + record. Let me propose an idea. Here, you want to create a new record with a specific metric. Maybe, instead of having a storage of events, you have a metrics collector class (similar to spans collector that we already have). This metrics collector also will be responsible for storing needed metrics and use this data to create a proper records. You can call the wrapper method record (or end) and it will create a record for a metric under the hood. So, you do not need to perform it with two separate steps: end + record. You will do it as a single step in this collector instead. Wdyt?

sdubov (changes requested):
I like the consistency for current metrics support implementation. Thank you. Please check my comments about some metrics implementation details.

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetryConfig.kt
```diff
@@ -250,8 +275,23 @@ public class OpenTelemetryConfig : FeatureConfig() {
...
 
+        val metricProvider = SdkMeterProvider.builder()
+            .setResource(resource)
+
+        val metricExporters = createMetricExporters()
+
+        metricExporters.forEach { exporter ->
+            val reader = PeriodicMetricReader
+                .builder(exporter)
+                .setInterval(Duration.ofSeconds(1))
```
sdubov: Does it makes sense to move this interval duration in the configuration as well?

Fant1k34: It does! Thanks, I added it

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/metric/GenAIMetricNames.kt
Fant1k34: Yeap, because this metric -- too calls count. So similarly to token count, metrics here is tool call

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetry.kt
```diff
@@ -480,6 +495,9 @@ public class OpenTelemetry {
                     span = inferenceSpan,
                     path = patchedExecutionInfo
                 )
+
+                // Store llm call
+                eventCallStorage.addEventCall(eventContext.eventId, eventContext.model.provider.display)
```
sdubov: I've also noticed that you use this `EventCallStorage` to get start-end time of a metric through `getDuration()` API. With current implementation, the start and end time should match the timing for a corresponding span. Maybe, you can re-use the duration from the span data instead. It would allow you to get rid of this storage class with necessity to manage it at all. Like this:
```
(inferenceSpan.span as? ReadableSpan)?.toSpanData()?.let { spanData ->
                    val duration =
                        (spanData.endEpochNanos - spanData.startEpochNanos).toDuration(DurationUnit.NANOSECONDS)
                    toolCallDurationHistogram.record(
                        duration.inWholeSeconds.toDouble(),
                       // attributes
                }
```

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetry.kt
Fant1k34: Thank you, I implemented this approach as you described:
* Introduced `MetricCollector` class which is responsible for storing necessary metrics, and recording them based on the MetricEvent that appears in OpenTelemetry class like that:
```
    metricCollector.recordEvent(
                    ToolCallEnded(
                        id = eventContext.eventId,
                        timestamp = System.currentTimeMillis(),
                        toolName = eventContext.toolName,
                        status = ToolCallStatus.FAILED
                    )
                )
```
* Under the hood, `MetricCollector` works with `MetricEventStorage` that is responsible for storing matching events via methods `startEvent` and `endEvent`. This approach allows to store paired events and compute some metrics based on difference between

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetry.kt
Fant1k34: Yeap, I agree we could get rid of storage class in this case

However, after implementing [this approach](https://github.com/JetBrains/koog/pull/1381#discussion_r2714527255) I have figured out, that OpenTelemetry class that has almost 900 lines is not much cluttered with metrics handling anymore, because of convenient `recordEvent` method

Additionally, Metric Collector separates logic of ai agent pipeline handling and metric recording. Moreover, it's not coupled with SpanCollector class and works separately (metrics for metric, traces for traces)

One more potential benefit is possibility to monitor states between 2 events. In future, we can save some value of starting event (for example, total spend tokens) and the value of ending event (total spend tokens). And we can record the difference between these states as metric (tokens spend tokens for this LLM Call) or something like that

If we reuse the logic from span collector:
*  we would need extra logic for OpenTelemetry class to get duration from span or add span as an argument to the MetricCollector, that leads to coupled, dependent classes. Ideally, metrics and spans should not be dependent, as they provide different way to observe system/agent behaviour
* we'll need follow strict matching "span" -> "metric" in the future, otherwise we could not record time property of operation

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/metric/TokenCounter.kt
Fant1k34: Makes sense, thanks! I have renamed it!

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetry.kt
Fant1k34: I guess, after implementing MetricCollector, such extracting could be excessive, as new MetricCollector class works with `add` or `record` function only once (except operationDurationHistogram -- twice, but different set of attributes). Wdyt?

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetry.kt
Fant1k34: Yes, it would!
Thank you, replaced Attributes.builder with `listOf` and `toSdkAttributes` functions, additionally added isVerbose attribute there not to propagate config futher

Fant1k34: Surprisingly, such way of attributes creation (via `listOf().toSdkAttributes(config.isVerbose)`) leads to wrong behaviour of metrics
It can be observed on the screenshots (the first screenshot -- how it works with `AttributesBuilder`, the second -- via `toSdkAttributes`)

image

image

For some reason, second option leads to fall of the metric, however it's cumulative one and it cannot become less that it was before

Thoroughly explore this issue by myself and with the help of chatGPT, we came to conclusion that the reason for such behaviour is custom wrapper under the `Attributes` class (`toSdkAttributes` function). For current implementation, it's not optimized and creates a new Attributes object each time. In terms of tracing feature that's totally fine, but not for metric one. Sdk should understand, what has changed from the previous timeframe, and since `toSdkAttributes` does create Attributes instance each time, it cannot properly link the change of metric. 

This behaviour can be observed via configuring LoggingMeterExporter. Each time it prints metrics' state and it has different hashcode and object reference in the system.

Since the problem appears not in plugin (that I use for visualization purpose), but in library (sdk cannot link attributes), I suggest revert changes to `AttributesBuilder.build()` back

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/attribute/GenAIAttributes.kt
Fant1k34: Thanks, I added it as a custom koog attribute. Additionally, I extracted metric `gen_ai.tool.call.count` to `koog.tool.count` as a custom metric name to be consistent with Open Telemetry documentation as well

examples/simple-examples/src/main/kotlin/ai/koog/agents/example/calculator/Calculator.kt
Fant1k34: Yeap, this is unrelated to Open Telemetry metrics feature. But I thought, that it could be great to have at least one example with configured metrics feature to check, if everything works as expected
So yes, it's a leftover from testing and I did that intentionally to have one example to test. Since it's in examples project in koog, I thought that it's fine to enable such feature here

If it's better to remove from Calculator example, I can do so!

docs/docs/opentelemetry-support.md
Fant1k34: Thank you! I have worked on this comment and came to the decision of [Views usage](https://opentelemetry.io/docs/languages/java/sdk/#views)
Therefore, I added `addMetricFilter` method to configure cardinality on agent-developer's side

docs/docs/opentelemetry-support.md
Fant1k34: fixed it to text_completion, thank you!

docs/docs/opentelemetry-support.md
Fant1k34: Mentioned it, thank you!

docs/docs/opentelemetry-support.md
egorklimov: Using Views is a cool idea, but they only let us drop attribute keys. They don’t support filtering by allowed attribute values, so we’d either lose the `tool.name` breakdown or make the metric much less useful 😢 

WDYT about adding a config allowlist for tools, e.g. I can set 
```kotlin
allowedTools = setOf("SayToUser", "plus")
```
Then metrics would keep `tool.name` for those tools and map all other values to something like `filtered`, which controls cardinality without losing overall totals and insights about selected tools

egorklimov: Don’t forget to mention in the docs that gen_ai.tool.name is a dynamic attribute, so a high number of tools can increase metric cardinality and impact performance. You can also note that users can use feature X to control this metric

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/metric/MetricEventStorage.kt
```diff
@@ -0,0 +1,58 @@
+package ai.koog.agents.features.opentelemetry.metric
+
+import io.github.oshai.kotlinlogging.KotlinLogging
+import java.util.concurrent.locks.ReentrantReadWriteLock
+import kotlin.concurrent.write
+
+internal class MetricEventStorage {
+    private val storage = mutableMapOf<String, MetricEvent>()
+    private val lock = ReentrantReadWriteLock()
```
egorklimov: (?) Maybe ConcurrentHashMap?

egorklimov (changes requested):

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/metric/MetricEventStorage.kt
Fant1k34: Thank you, it sounds like a great idea, done so!

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetry.kt
Fant1k34: Thank you! I fixed it that way we discussed

docs/docs/opentelemetry-support.md
Fant1k34: Thank you, fixed it! Implemented `addToolCallNamesMapping` configuration
```
addToolCallNamesMapping(
                    setOf("plus", "minus", "multiply", "divide"),
                    "unknown"
                )
```

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/metric/MetricObjects.kt
```diff
@@ -0,0 +1,8 @@
+package ai.koog.agents.features.opentelemetry.metric
+
+internal data class MetricFilter(val metricName: String, val attributesKeysToRetain: Set<String>)
+
+internal data class ToolCallMapper(
```
egorklimov: Let’s add a `map` method here since this is a mapper class.

Then you’ll be able to do something like:
```
val toolName = toolCallMapper?.map(metricEvent.toolName) ?: metricEvent.toolName
```

egorklimov (changes requested):

docs/docs/opentelemetry-support.md
egorklimov: I'd suggest something like
```kotlin
restrictToolNameCardinality(allowedNames: Set<String>, fallback: String = "filtered")
```
or
```kotlin
setToolNameAllowlist(allowed: Set<String>, otherValue: String = "filtered")
```

I read the current name as meaning that all tool names from the set will be mapped to "unknown", and I don’t understand the motivation behind the method

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/metric/MetricObjects.kt
Fant1k34: Thanks, I added ConfiguredToolCallMapper and NoOp implementation = DefaultToolCallMapper

docs/docs/opentelemetry-support.md
Fant1k34: Renamed to `restrictToolNameCardinality` and set default value to `fallback`

egorklimov (approved):

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/metric/KoogMetricName.kt
```diff
@@ -0,0 +1,6 @@
+package ai.koog.agents.features.opentelemetry.metric
+
+internal sealed interface KoogMetricName : MetricName {
```
sdubov: This instance keeps other fields (`description` and `unit`) as well as the `name`. Maybe, it worth to consider dropping the `Name` suffix?

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/metric/KoogMetricNames.kt
```diff
@@ -0,0 +1,19 @@
...
+    sealed interface Tool : KoogMetricName {
+        override val name: String
+            get() = super.name.concatKey("tool")
+
+        object Count : Tool {
+            override val name: String
+                get() = super.name.concatKey("count")
+
+            override val description: String
+                get() = "Tool calls count"
```
sdubov: The `description` and the `unit` can use the direct initialisation here since you do not need to compute it:
```
override val description: String = "Tool calls count"
override val unit: String = "tool call"
```

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/metric/MetricName.kt
```diff
@@ -0,0 +1,9 @@
+package ai.koog.agents.features.opentelemetry.metric
+
+internal sealed interface MetricName {
```
sdubov: I have a question about the name for this entity. This class store multiple fields, including `description` and `unit` as well as `name`. Maybe, use a generalised name for them, e.g. `Metric`? WDYT?

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetryConfig.kt
```diff
@@ -120,6 +141,52 @@ public class OpenTelemetryConfig : FeatureConfig() {
...
+    }
+
+    /**
+     * Restricts tool names in the attributes' metric and sets the fallback tool name when a tool is not allowed.
+     * Helps to manage cardinality of the metric.
+     *
+     * @param allowedToolNames A set of allowed tool names
+     * @param fallbackToolName The fallback / default tool name if not in the allowed set
+     */
+    public fun restrictToolNameCardinality(
```
sdubov: (TBD) @Fant1k34, as we discussed, the main concern from my side here is naming. It sounds a bit confusing for an end user from my perspective, because it is not quite clear what do we restrict here: metrics or spans or both.

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/metric/GenAIMetricName.kt
```diff
@@ -0,0 +1,6 @@
+package ai.koog.agents.features.opentelemetry.metric
+
+internal sealed interface GenAIMetricName : MetricName {
```
sdubov: @Fant1k34, the instance keeps other fields (description and unit) as well as the name. Maybe, it worth to consider dropping the Name suffix?

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/metric/GenAIMetricNames.kt
```diff
@@ -0,0 +1,40 @@
+package ai.koog.agents.features.opentelemetry.metric
+
+internal object GenAIMetricNames {
```
sdubov: @Fant1k34, the instance keeps other fields (description and unit) as well as the name. Maybe, it worth to consider dropping the Name suffix?

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/metric/GenAIMetricNames.kt
```diff
@@ -0,0 +1,40 @@
...
+        sealed interface Token : Client {
+            override val name: String
+                get() = super.name.concatKey("token")
+
+            object Usage : Token {
+                override val name: String
+                    get() = super.name.concatKey("usage")
+
+                override val description: String
+                    get() = "Total token count"
```
sdubov: The `description` and the `unit` fields can have a direct initialisation here since you do not need to compute it:

```
override val description: String = "Total token count"
override val unit: String = "token"
```

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/metric/GenAIMetricNames.kt
```diff
@@ -0,0 +1,40 @@
...
+
+        sealed interface Operation : Client {
+            override val name: String
+                get() = super.name.concatKey("operation")
+
+            object Duration : Operation {
+                override val name: String
+                    get() = super.name.concatKey("duration")
+
+                override val description: String
```
sdubov: The `description` and the `unit` fields can have a direct initialisation here since you do not need to compute it:

override val description: String = "Operation duration"
override val unit: String = "s"

sdubov (approved):
As we discussed, I have two concerns here:
1. `restricToolNameCardinality` config which I found a bit confusing. As I noticed, I would implement this as an extension for a metric adapter (similar to the span adapter).
2. Single responsibility for MetricCollector class. I would prefer it to be able to track any metric and help to create it if needed (keep the timestamp for a start event if needed for now).

Other than that, looks good for me. We can address some of those questions later in a further commits. For now, I will approve this one to not block the development.

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/metric/GenAIMetricNames.kt
Fant1k34: Thank you! It makes sence

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/metric/GenAIMetricNames.kt
Fant1k34: I agree, thanks!

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetry.kt
```diff
@@ -663,6 +732,26 @@ public class OpenTelemetry {
...
+                        toolName = eventContext.toolName,
+                        toolCallStatus = KoogAttributes.Koog.Tool.Call.StatusType.ERROR
+                    )
+                )
+
+                metricCollector.getMetricEvent(eventContext.eventId)?.let { storedMetricEvent ->
+                    metricCollector.recordHistogramMetricEvent(
+                        metricEvent = createExecuteToolDurationHistogramMetricEvent(
+                            id = eventContext.eventId,
+                            duration = Clock.System.now() - storedMetricEvent.timestamp,
```
egorklimov: I decided to check whether it’s correct to use `Clock.System.now()` and whether it could be affected by time zones or system time changes. I found out that it’s better to use `System.nanoTime()` or `TimeSource.Monotonic`, since these provide a monotonic clock for measuring elapsed time and are independent of system clock changes.

Image

egorklimov (changes requested):

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetry.kt
egorklimov: All Koog events already use Clock.System.now() for timestamps, which is correct for tracking event time, but it can cause issues when calculating durations. I think we can keep it as is in this PR, but maybe @sdubov  would like to consider refactoring it later.

egorklimov (approved):

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/attribute/attributes.kt
```diff
@@ -47,25 +27,32 @@ private fun Attribute.toSdkAttribute(verbose: Boolean): Pair<AttributeKey<*>, An
             val unwrappedValue = if (verbose) value.value else value.toString()
             Pair(AttributeKey.stringKey(key), unwrappedValue)
         }
+
```
sdubov: Unnecessary extra newlines.

examples/simple-examples/src/main/kotlin/ai/koog/agents/example/calculator/Calculator.kt
sdubov: Examples are usually dedicated to a specific portion of functionality in Koog. If you like to have an example that shows how to work with OpenTelemetry metrics, if you be better to add this into the open telemetry examples or create a separate example. Please see the `koog/examples/simple-examples/src/main/kotlin/ai/koog/agents/example/features/opentelemetry` directory.

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/attribute/attributes.kt
sdubov: Unnecessary new lines here.

examples/simple-examples/src/main/kotlin/ai/koog/agents/example/calculator/Calculator.kt
egorklimov: Reverted changes in the Calculator example

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/attribute/attributes.kt
egorklimov: Fixed

egorklimov: Fixed

docs/docs/opentelemetry-support.md
```diff
@@ -351,23 +416,95 @@ The following event types are supported in line with OpenTelemetry's [Semantic c
...
+        - `gen_ai.tool.name` (recommended, if `gen_ai.operation.name` is `EXECUTE_TOOL`)
+        - `gen_ai.tool.call.status` (custom, if `gen_ai.operation.name` is `EXECUTE_TOOL`) — `SUCCESS`, `ERROR`, or
+          `VALIDATION_FAILED`
+        - `gen_ai.provider.name` (recommended, if `gen_ai.operation.name` is `TEXT_COMPLETION`)
+        - `gen_ai.response.model` (recommended, if `gen_ai.operation.name` is `TEXT_COMPLETION`)
+        - `gen_ai.token.type` (recommended, if `gen_ai.operation.name` is `TEXT_COMPLETION`)
+    - Spec: see OpenTelemetry semantic
+      conventions, [gen_ai.client.operation.duration](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-metrics/#metric-gen_aiclientoperationduration)
+
+- gen_ai.client.tool.count
```
egorklimov: Currently we have `koog.tool.count` which is incorrect.

egorklimov (changes requested):

docs/docs/opentelemetry-support.md
egorklimov: We decided to use koog.gen_ai.client.tool.call.count

examples/simple-examples/src/main/kotlin/ai/koog/agents/example/features/opentelemetry/OpenTelemetryExample.kt
```diff
@@ -37,13 +41,35 @@ fun main() = runBlocking {
...
+            )
+
+            // Add a console logger to view metrics
+            addMetricExporter(
+                LoggingMetricExporter.create(),
+                0.5.seconds
+            )
+
+            // Send metrics to OpenTelemetry collector
+            addMetricExporter(
```
egorklimov: todo: remove

egorklimov (approved):

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/metric/MetricCollector.kt
```diff
@@ -0,0 +1,128 @@
...
+import io.github.oshai.kotlinlogging.KotlinLogging
+import io.opentelemetry.api.metrics.DoubleHistogram
+import io.opentelemetry.api.metrics.LongCounter
+import io.opentelemetry.api.metrics.Meter
+import java.util.concurrent.ConcurrentHashMap
+import kotlin.time.Clock
+
+internal class MetricCollector(private val meter: Meter, private val config: OpenTelemetryConfig) {
+
+    private val counters = ConcurrentHashMap<String, LongCounter>()
```
sdubov, line 14: @egorklimov, @Fant1k34 , I've just noticed that you do not drain `counters` and `histogram` maps as you do for `metricEvents` map. What do you think, do we need to clear maps on agent close as well?

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetry.kt
```diff
@@ -675,6 +733,26 @@ public class OpenTelemetry {
...
+                        toolName = eventContext.toolName,
+                        toolCallStatus = KoogAttributes.Koog.Tool.Call.StatusType.SUCCESS
+                    )
+                )
+
+                metricCollector.getMetricEvent(eventContext.eventId)?.let { storedMetricEvent ->
+                    metricCollector.recordHistogramMetricEvent(
+                        metricEvent = createExecuteToolDurationHistogramMetricEvent(
+                            id = eventContext.eventId,
+                            duration = Clock.System.now() - storedMetricEvent.timestamp,
```
sdubov: Does it worth to extract this (duration calculation for metric event) into a private helper function?

sdubov (approved):
@egorklimov , @Fant1k34 , amazing work! Thank you. LGTM. Please see couple minor comments from my side.

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/metric/MetricCollector.kt
egorklimov: Counters will still present in SDK, so there is no data leakage

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetry.kt
egorklimov: Added duration extension method, thanks!

PR #1384, opened 2026-01-20
[prompt] Converse API support in Bedrock LLM client

### LINKED ISSUE #1050: Use Bedrock Converse instead of invokeModel
Bedrock already has an API called `converse` that abstracts over different models: https://docs.aws.amazon.com/sdk-for-kotlin/api/latest/bedrockruntime/aws.sdk.kotlin.services.bedrockruntime/-bedrock-runtime-client/converse.html

Using this API would reduce a lot of the code that does per-model-family stuff in `BedrockLLMClient`. Using converse would also enable automatic support for all Bedrock models.

EugeneTheDev:
Adds [Converse API](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_Converse.html) support to Bedrock LLM client. Multimodality and streaming are supported.
Also adds new `apiMethod` parameter to Bedrock LLM client settings/constructor to control the API used: manual `InvokeModel` or higher-level `Converse`. Default is `InvokeModel` for now. There's also a new integration test suite `BedrockConverseApiIntegrationTest` to verify that it works as expected.

Fixes #1050


prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/bedrock/BedrockModels.kt
```diff
@@ -742,17 +742,20 @@ public object BedrockModels : LLModelDefinitions {
     }
 }
 
+// TODO remove, has no effect now
```
serge-p7v: Perhaps it's OK to remove it?

EugeneTheDev: I also think so. My only concern is that multimodality works only with this new API. `InvokeModel` is still text-only in Bedrock client. Do you think it would be OK to make the models multi-modal now and encourage users to switch to the new API?

serge-p7v:
The version of aws-sdk-kotlin = "1.5.16" was published in August 2025. I see two fixes in aws-sdk-kotlin that mention Converse since then (and some others in bedrockruntime). Perhaps it's safe to update the dependency to the latest version.

UPDATE: there is a PR about that, but it's failing - https://github.com/JetBrains/koog/pull/603

prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/bedrock/BedrockModels.kt
serge-p7v: I think yes. It seems that langchain4j and Spring AI have already (almost) switched from InvokeModel to Converse API.
But Spring AI also kept InvokeModel for embedding models:

> The Converse API does not support embedding operations, so these will remain in the current API and the embedding model functionality in the existing InvokeModel API will be maintained

https://docs.spring.io/spring-ai/reference/api/bedrock.html

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/executor/ExecutorIntegrationTestBase.kt
```diff
@@ -527,19 +512,23 @@ abstract class ExecutorIntegrationTestBase {
                         ImageTestScenario.LARGE_IMAGE_ANTHROPIC, ImageTestScenario.LARGE_IMAGE -> {
                             val message = e.message.shouldNotBeNull()
 
-                            message.shouldContain("Status code: 400")
-                            message.shouldContain("image exceeds")
+                            listOf(
```
antoniibelyshev, line 515: I think, it's better to make these messages consistent, rather than adding new options here

prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/bedrock/converse/BedrockConverseParams.kt
```diff
@@ -0,0 +1,166 @@
...
+        speculation = speculation,
+        schema = schema,
+        toolChoice = toolChoice,
+        user = user,
+        additionalProperties = additionalProperties,
+        topP = topP,
+        stopSequences = stopSequences,
+    )
+
+    override fun equals(other: Any?): Boolean = when {
```
antoniibelyshev, line 119: Do we need `equals` and `hashCode` here?

antoniibelyshev (approved):
LGTM

prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/bedrock/BedrockModels.kt
EugeneTheDev: Ok, makes sense, I'll remove this

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/executor/ExecutorIntegrationTestBase.kt
EugeneTheDev: Some of these messages come from the LLM providers themselves, so we can't control them. E.g. BedrockClient doesn't respond with `Status code: 400`, and we have to look at the error text itself that says "image exceeds"

prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/bedrock/converse/BedrockConverseParams.kt
EugeneTheDev: Yes, since these are not data classes, we don't get them for free. We use these as structure classes though, so we need to have proper comparison/hash code operations implemented, just like in other parameter classes

EugeneTheDev:
@serge-p7v I see the error with dependabot AWS SDK version update PR. There were new methods introduced, so tests have to be adjusted. Looks trivial, I will update it manually here

serge-p7v (approved):
LGTM

PR #1396, opened 2026-01-22
feat(prompt): Include examples in the prompt with StructuredRequest.Native to help LLMs better understand desired data format 

### LINKED ISSUE #1328: Structured Output should include examples in the prompt when Native mode is used
Looking at the design of the `executeStructured` [function](https://github.com/JetBrains/koog/blob/81fcc1c3a59203d6c6a13be621e91d64c0385404/prompt/prompt-structure/src/commonMain/kotlin/ai/koog/prompt/structure/PromptExecutorExtensions.kt#L167) in Koog, there's a flaw in the design that stops Structured Output from working properly:

- **Context:** When requesting structured output from an inference platform: 
  1. Some providers support structured output deep in the inference engine, based on a JSON Schema.
  2. Other providers that don't support this, rely on simply 'asking the LLM nicely' to format its output according to an example, then retrying if the result wasn't deserializable.

...Koog's `executeStructured` API abstracts over which of these **output constraint methods** is being used, and has implementation details to drive each method.  There is a bug here...

- **Problem:** The Koog framework wrongly assumes that inference platforms that support `Native` JSON Schema constraint do not require the example JSON in their context: They do still need it, and here's why:
  - While the JSON schema provides a guardrail, preventing the LLM from generating non compliant tokens, that does nothing to _prime the context_ toward generating the kind of JSON you want.
  - :bulb: The result is like sticking someone (the LLM) in a maze that blocks them from straying off the path (generating the wrong token), but with a blindfold on, so they just keep bumping into the walls frustrated. The person will eventually reach the exit (LLM will complete valid output) but only through coercion at a token-by-token level, leading to sub-standard output since the models latent layers will be less involved in correct generation.
  - The relevant code is here: See how example objects are only serialized and [presented to](https://github.com/JetBrains/koog/blob/81fcc1c3a59203d6c6a13be621e91d64c0385404/prompt/prompt-structure/src/commonMain/kotlin/ai/koog/prompt/structure/json/JsonStructure.kt#L87) the LLM in the `Manual` fallback case (2) where the inference engine doesn't natively support JSON Schema.

- **Solution:** Koog should still be formatting the 'example object' as a trailing part of the prompt AND setting the JSON schema, for best results.

## Verification of issue

I have tested this in the following ways:
- ✅ Verified that my inference provider (GPT-OSS on llama.cpp) is using `Native` structured output by placing a breakpoint on that branch
- ⚠️ Verified that the `JsonStructure.defaultDefinitionPrompt` is **not** being called (neither examples nor JSON schema being serialized and fed into the LLM token stream to prime the model to aim for correct output)
- 🙅 Observed that when running a simple example of 'populate this imaginary person model', the JSON syntax is always correct but **most of the time** the first field 'name' contains something like "_Please provide an example of the JSON structure_" instead of any name.
    - This is the LLM 'bouncing through the maze' of the initial JSON syntax then crying out for help because it doesn't actually know where it _should_ be going, only where it _can't_ go!
    - ✅ If I provide the JSON example _myself_ in the prompt, this problem goes away and generation is rock solid!
    - 🤔 However, always providing myself in the prompt is not a tidy solution since - let's repeat:
      `executeStructured` API is **transparent** to the **underlying output constraint method** being used and we wouldn't want to repeat the `Manual` method's example provision.

EugeneTheDev:
In certain cases, it's not enough to just constrain an LLM to a given data format, since it might "get lost". It helps to provide an LLM with examples regarding what type of data we want in the output, not only in which format. 

Fixes #1328

### DIFF
--- a/prompt/prompt-structure/src/commonMain/kotlin/ai/koog/prompt/structure/PromptExecutorExtensions.kt
+++ b/prompt/prompt-structure/src/commonMain/kotlin/ai/koog/prompt/structure/PromptExecutorExtensions.kt
@@ -81,7 +81,7 @@ public data class StructuredRequestConfig<T>(
                     // If examples are supplied, append them
                     if (mode.structure.examples.isNotEmpty()) {
                         user {
-                            StructuredOutputPrompts.examplesPrompt(this, mode.structure)
+                            mode.structure.examples(this)
                         }
                     }
                 }.withUpdatedParams { schema = mode.structure.schema }
--- a/prompt/prompt-structure/src/commonMain/kotlin/ai/koog/prompt/structure/PromptExecutorExtensions.kt
+++ b/prompt/prompt-structure/src/commonMain/kotlin/ai/koog/prompt/structure/PromptExecutorExtensions.kt
@@ -77,7 +77,14 @@ public data class StructuredRequestConfig<T>(
 
             // Rely on built-in model capabilities to provide structured response.
             is StructuredRequest.Native -> {
-                prompt.withUpdatedParams { schema = mode.structure.schema }
+                prompt(prompt) {
+                    // If examples are supplied, append them
+                    if (mode.structure.examples.isNotEmpty()) {
+                        user {
+                            StructuredOutputPrompts.examplesPrompt(this, mode.structure)
+                        }
+                    }
+                }.withUpdatedParams { schema = mode.structure.schema }
             }
         }
     }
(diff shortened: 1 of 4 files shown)


prompt/prompt-structure/src/commonMain/kotlin/ai/koog/prompt/structure/PromptExecutorExtensions.kt
antoniibelyshev: Don't we need to show examples for the manual mode too?

antoniibelyshev (approved):
Thank you, looks good!

I have a minor question on the design of the structured request prompts: can we allow customizing the examples prompt for the `Native` schema as we do for the `Manual` schema? AFAIU, `structure.definition` can be customized, but it is not used for the native schema.

antoniibelyshev (approved):
Thank you!

PR #1397, opened 2026-01-23
fix: upgrade kotlinx-datetime to 0.7.1-0.6.x-compat for compatibility

mafahir:
## Summary

Upgrades kotlinx-datetime from 0.6.2 to 0.7.1-0.6.x-compat to resolve runtime `ClassNotFoundException` when Koog is used alongside projects that depend on kotlinx-datetime 0.7.x.

## Problem

When a project uses both Koog 0.6.0 and kotlinx-datetime 0.7.x, Gradle resolves to 0.7.1 (the higher version). This causes a runtime error:

```
java.lang.NoClassDefFoundError: kotlinx/datetime/Clock$System
    at ai.koog.prompt.executor.clients.openai.OpenAILLMClient.<init>(OpenAILLMClient.kt:104)
```

This happens because:
- Koog 0.6.0 uses `kotlinx.datetime.Clock.System` which exists in kotlinx-datetime 0.6.2
- kotlinx-datetime 0.7.x changed `Clock` to be a typealias to `kotlin.time.Clock`, removing the actual class

## Solution

Use `0.7.1-0.6.x-compat` instead of `0.6.2`. This artifact:
- Preserves `kotlinx.datetime.Clock` and `kotlinx.datetime.Instant` as real classes
- Is compatible with kotlinx-datetime 0.7.x resolution (won't cause conflicts)
- Shows deprecation warnings encouraging migration to `kotlin.time.Clock`

This is the [recommended approach](https://github.com/Kotlin/kotlinx-datetime/releases/tag/v0.7.1) for libraries that need to support downstream projects using kotlinx-datetime 0.7.x.

## Testing

- Full `./gradlew compileKotlin` passes (176 tasks)
- Deprecation warnings are expected and intentional

## Related

- Closes: KG-544
- kotlinx-datetime release notes: https://github.com/Kotlin/kotlinx-datetime/releases/tag/v0.7.1


sproshev:
Hi! Please use the latest `develop` as the base branch

skarpovdev:
@mafahir Should we expect updates in this PR or is it ok to close it?

PR #1421, opened 2026-01-27
feat(agents): JBRes-7677: opentelementry support for MCP

vova-jb:
## Motivation and Context
opentelemetry spans for Mcp tools

## Breaking Changes
none

---

#### Type of the changes
- [* ] New feature (non-breaking change which adds functionality)

### DIFF
--- a/agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/attribute/McpAttributes.kt
+++ b/agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/attribute/McpAttributes.kt
@@ -1,0 +1,93 @@
+package ai.koog.agents.features.opentelemetry.attribute
+
+/**
+ * MCP (Model Context Protocol) specific attributes following OpenTelemetry semantic conventions.
+ * Based on:
+ * https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/mcp.md
+ * https://github.com/open-telemetry/semantic-conventions/blob/main/model/mcp/common.yaml
+ *
+ * These attributes are used for instrumenting MCP client operations including:
+ * - Tools operations
+ * - Session management
+ * - Protocol version information
+ * - TODO: Resources and prompts operations
+ */
+internal object McpAttributes {
+
+    sealed interface Mcp : Attribute {
+        override val key: String
+            get() = "mcp"
+
+        /**
+         * The name of the request or notification method.
+         * This is a REQUIRED attribute for all MCP operations.
+         *
+         * @see [ai.koog.agents.features.opentelemetry.integration.mcp.McpMethod]
+         */
+        sealed interface Method : Mcp {
+            override val key: String
+                get() = super.key.concatKey("method")
+
+            data class Name(private val name: String) : Method {
+                override val key: String = super.key.concatKey("name")
+                override val value: String = name
+            }
+        }
+
+        /**
+         * Identifies the MCP session.
+         * This is a RECOMMENDED attribute when the request is part of a session.
+         *
+         * Example: "191c4850af6c49e08843a3f6c80e5046"
+         */
+        sealed interface Session : Mcp {
+            override val key: String
+                get() = super.key.concatKey("session")
+
+            data class Id(private val id: String) : Session {
+                override val key: String = super.key.concatKey("id")
+                override val value: String = id
+            }
+        }
+
+        /**
+         * The version of the Model Context Protocol in use.
+         * This is a RECOMMENDED attribute.
+         *
+         * Example: "2025-06-18"
+         */
+        sealed interface Protocol : Mcp {
+            override val key: String
+                get() = super.key.concatKey("protocol")
+
+            data class Version(private val version: String) : Protocol {
+                override val key: String = super.key.concatKey("version")
+                override val value: String = version
+            }
+        }
+    }
+
+    /**
+     * Network transport and protocol attributes for MCP operations.
+     */
+    sealed interface Network : Attribute {
+        override val key: String
+            get() = "network"
+
+        /**
+         * The transport protocol used for the MCP session.
+         * This is RECOMMENDED.
+         *
+         * https://github.com/open-telemetry/semantic-conventions/blob/main/model/mcp/common.yaml
+         *
+         * Valid values:
+         * - "pipe" for stdio transport
+         * - "tcp" for HTTP transport
+         * - "quic" for HTTP/3 transport
+         */
+        data class Transport(private val transport: String) : Network {
+            override val key: String = "network.transport"
+            override val value: String = transport
+        }
+    }
+}
--- a/agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetry.kt
+++ b/agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetry.kt
@@ -1,16 +1,17 @@
 package ai.koog.agents.features.opentelemetry.feature
 
-import ai.koog.agents.core.agent.GraphAIAgent
 import ai.koog.agents.core.agent.entity.AIAgentStorageKey
 import ai.koog.agents.core.agent.execution.AgentExecutionInfo
 import ai.koog.agents.core.annotation.InternalAgentsApi
 import ai.koog.agents.core.feature.AIAgentFunctionalFeature
 import ai.koog.agents.core.feature.AIAgentGraphFeature
+import ai.koog.agents.core.feature.handler.tool.ToolCallEventContext
+import ai.koog.agents.core.feature.model.AIAgentError
 import ai.koog.agents.core.feature.pipeline.AIAgentFunctionalPipeline
 import ai.koog.agents.core.feature.pipeline.AIAgentGraphPipeline
 import ai.koog.agents.core.feature.pipeline.AIAgentPipeline
+import ai.koog.agents.core.tools.ToolRegistry
 import ai.koog.agents.core.utils.SerializationUtils
-import ai.koog.agents.features.opentelemetry.attribute.CommonAttributes
 import ai.koog.agents.features.opentelemetry.attribute.SpanAttributes
 import ai.koog.agents.features.opentelemetry.event.AssistantMessageEvent
 import ai.koog.agents.features.opentelemetry.event.ChoiceEvent
@@ -18,6 +19,8 @@ import ai.koog.agents.features.opentelemetry.event.ModerationResponseEvent
 import ai.koog.agents.features.opentelemetry.event.SystemMessageEvent
 import ai.koog.agents.features.opentelemetry.event.ToolMessageEvent
 import ai.koog.agents.features.opentelemetry.event.UserMessageEvent
+import ai.koog.agents.features.opentelemetry.integration.SpanAdapter
+import ai.koog.agents.features.opentelemetry.integration.mcp.McpMethod
 import ai.koog.agents.features.opentelemetry.span.GenAIAgentSpan
 import ai.koog.agents.features.opentelemetry.span.SpanCollector
 import ai.koog.agents.features.opentelemetry.span.SpanType
@@ -28,6 +31,7 @@ import ai.koog.agents.features.opentelemetry.span.endInvokeAgentSpan
 import ai.koog.agents.features.opentelemetry.span.endNodeExecuteSpan
 import ai.koog.agents.features.opentelemetry.span.endStrategySpan
 import ai.koog.agents.features.opentelemetry.span.endSubgraphExecuteSpan
+import ai.koog.agents.features.opentelemetry.span.enrichExecuteToolSpanWithMcpAttrs
 import ai.koog.agents.features.opentelemetry.span.startCreateAgentSpan
 import ai.koog.agents.features.opentelemetry.span.startExecuteToolSpan
 import ai.koog.agents.features.opentelemetry.span.startInferenceSpan
@@ -35,8 +39,11 @@ import ai.koog.agents.features.opentelemetry.span.startInvokeAgentSpan
 import ai.koog.agents.features.opentelemetry.span.startNodeExecuteSpan
 import ai.koog.agents.features.opentelemetry.span.startStrategySpan
 import ai.koog.agents.features.opentelemetry.span.startSubgraphExecuteSpan
+import ai.koog.agents.mcp.metadata.McpMetadataKeys
 import ai.koog.prompt.message.Message
 import io.github.oshai.kotlinlogging.KotlinLogging
+import kotlinx.serialization.json.JsonElement
+import kotlinx.serialization.json.JsonObject
 import kotlin.reflect.KType
 
 /**
@@ -264,8 +271,7 @@ public class OpenTelemetry {
                 logger.debug { "Execute OpenTelemetry before agent started handler" }
 
                 val messages = eventContext.agent.agentConfig.prompt.messages.toList()
-                val tools = (eventContext.agent as? GraphAIAgent<*, *>)?.toolRegistry?.tools?.map { it.descriptor }?.toList()
-                    ?: emptyList()
+                val tools = eventContext.context.llm.toolRegistry.tools.map { it.descriptor }.toList()
 
                 // Create CreateAgentSpan
                 val createAgentSpan = startCreateAgentSpan(
@@ -538,7 +544,14 @@ public class OpenTelemetry {
                             }
 
                             is Message.Tool.Call -> {
-                                add(ChoiceEvent(provider, message, arguments = message.contentJsonResult.getOrNull(), index = index))
+                                add(
+                                    ChoiceEvent(
+                                        provider,
+                                        message,
+                                        arguments = message.contentJsonResult.getOrNull(),
+                                        index = index
+                                    )
+                                )
                             }
                         }
                     }
@@ -583,15 +596,13 @@ public class OpenTelemetry {
 
             //region Tool Call
 
+            @OptIn(InternalAgentsApi::class)
             pipeline.interceptToolCallStarting(this) intercept@{ eventContext ->
                 logger.debug { "Execute OpenTelemetry tool call handler" }
 
-                val patchedExecutionInfo = eventContext.executionInfo
-                    .appendRunId(eventContext.runId)
-                    .appendId(eventContext.eventId)
-
+                val path = eventContext.executionInfo.appendRunId(eventContext.runId).appendId(eventContext.eventId)
                 val parentSpan = spanCollector.getParentSpanForEvent(
-                    executionInfo = patchedExecutionInfo,
+                    executionInfo = path,
                 ) ?: return@intercept
 
                 val executeToolSpan = startExecuteToolSpan(
@@ -603,122 +614,60 @@ public class OpenTelemetry {
                     toolDescription = eventContext.toolDescription,
                     toolCallId = eventContext.toolCallId
                 )
+                val mcpToolMetaData = eventContext.context.llm.toolRegistry.getMcpToolMeta(eventContext.toolName)
+                if (mcpToolMetaData != null) {
+                    executeToolSpan.enrichExecuteToolSpanWithMcpAttrs(
+                        toolName = eventContext.toolName,
+                        sessionId = mcpToolMetaData[McpMetadataKeys.McpSessionId],
+                        method = McpMethod.TOOLS_CALL,
+                        serverPort = mcpToolMetaData[McpMetadataKeys.ServerPort]?.toIntOrNull(),
+                        serverAddress = mcpToolMetaData[McpMetadataKeys.ServerUrl],
+                        mcpProtocolVersion = requireNotNull(mcpToolMetaData[McpMetadataKeys.McpProtocolVersion]) {
+                            "MCP protocol version is required for MCP tool calls."
+                        },
+                        mcpTransportType = requireNotNull(mcpToolMetaData[McpMetadataKeys.McpTransportType]) {
+                            "MCP transport type is required for MCP tool calls."
+                        },
+                    )
+                }
 
                 spanAdapter?.onBeforeSpanStarted(executeToolSpan)
-                spanCollector.collectSpan(executeToolSpan, patchedExecutionInfo)
+                spanCollector.collectSpan(executeToolSpan, path)
             }
 
             pipeline.interceptToolCallCompleted(this) intercept@{ eventContext ->
                 logger.debug { "Execute OpenTelemetry tool result handler" }
-
-                // Get the current ExecuteToolSpan
-                val patchedExecutionInfo = eventContext.executionInfo
-                    .appendRunId(eventContext.runId)
-                    .appendId(eventContext.eventId)
-
-                val executeToolSpan = spanCollector.getStartedSpan(
-                    executionInfo = patchedExecutionInfo,
-                    eventId = eventContext.eventId,
-                    spanType = SpanType.EXECUTE_TOOL
-                ) ?: return@intercept
-
-                eventContext.toolDescription?.let { toolDescription ->
-                    executeToolSpan.addAttribute(
-                        // gen_ai.tool.description
-                        attribute = SpanAttributes.Tool.Description(description = toolDescription)
-                    )
-                }
-
-                spanAdapter?.onBeforeSpanFinished(span = executeToolSpan)
-
-                val toolResult = eventContext.toolResult ?: kotlinx.serialization.json.JsonObject(emptyMap())
-                endExecuteToolSpan(
-                    span = executeToolSpan,
+                val toolResult = eventContext.toolResult ?: JsonObject(emptyMap())
+                endAndRemoveExecuteToolSpan(
                     toolResult = toolResult,
-                    verbose = config.isVerbose
-                )
-                spanCollector.removeSpan(
-                    span = executeToolSpan,
-                    path = patchedExecutionInfo
+                    config = config,
+                    spanAdapter = spanAdapter,
+                    spanCollector = spanCollector,
+                    eventContext = eventContext,
                 )
             }
 
             pipeline.interceptToolCallFailed(this) intercept@{ eventContext ->
                 logger.debug { "Execute OpenTelemetry tool call failure handler" }
-
-                // Get the current ExecuteToolSpan using executionInfo.path()
-                val patchedExecutionInfo = eventContext.executionInfo
-                    .appendRunId(eventContext.runId)
-                    .appendId(eventContext.eventId)
-
-                val executeToolSpan = spanCollector.getStartedSpan(
-                    executionInfo = patchedExecutionInfo,
-                    eventId = eventContext.eventId,
-                    spanType = SpanType.EXECUTE_TOOL
-                ) ?: return@intercept
-
-                eventContext.toolDescription?.let { toolDescription ->
-                    executeToolSpan.addAttribute(
-                        // gen_ai.tool.description
-                        attribute = SpanAttributes.Tool.Description(description = toolDescription)
-                    )
-                }
-
-                executeToolSpan.addAttribute(
-                    attribute = CommonAttributes.Error.Type(eventContext.message)
-                )
-
-                // End the ExecuteToolSpan span
-                spanAdapter?.onBeforeSpanFinished(executeToolSpan)
-                val toolResult = kotlinx.serialization.json.JsonObject(emptyMap())
-                endExecuteToolSpan(
-                    span = executeToolSpan,
-                    toolResult = toolResult,
+                endAndRemoveExecuteToolSpan(
+                    spanAdapter = spanAdapter,
+                    spanCollector = spanCollector,
+                    toolResult = JsonObject(emptyMap()),
+                    config = config,
+                    eventContext = eventContext,
                     error = eventContext.error,
-                    verbose = config.isVerbose
-                )
-                spanCollector.removeSpan(
-                    span = executeToolSpan,
-                    path = patchedExecutionInfo
                 )
             }
 
             pipeline.interceptToolValidationFailed(this) intercept@{ eventContext ->
                 logger.debug { "Execute OpenTelemetry tool validation error handler" }
-
-                // Get the current ExecuteToolSpan using executionInfo.path()
-                val patchedExecutionInfo = eventContext.executionInfo
-                    .appendRunId(eventContext.runId)
-                    .appendId(eventContext.eventId)
-
-                val executeToolSpan = spanCollector.getStartedSpan(
-                    executionInfo = patchedExecutionInfo,
-                    eventId = eventContext.eventId,
-                    spanType = SpanType.EXECUTE_TOOL
-                ) ?: return@intercept
-
-                eventContext.toolDescription?.let { toolDescription ->
-                    executeToolSpan.addAttribute(
-                        // gen_ai.tool.description
-                        attribute = SpanAttributes.Tool.Description(description = toolDescription)
-                    )
-                }
-
-                executeToolSpan.addAttribute(
-                    attribute = CommonAttributes.Error.Type(eventContext.message)
-                )
-
-                // End the ExecuteToolSpan span
-                spanAdapter?.onBeforeSpanFinished(executeToolSpan)
-                endExecuteToolSpan(
-                    span = executeToolSpan,
+                endAndRemoveExecuteToolSpan(
+                    spanAdapter = spanAdapter,
+                    spanCollector = spanCollector,
                     toolResult = null,
+                    config = config,
+                    eventContext = eventContext,
                     error = eventContext.error,
-                    verbose = config.isVerbose
-                )
-                spanCollector.removeSpan(
-                    span = executeToolSpan,
-                    path = patchedExecutionInfo
                 )
             }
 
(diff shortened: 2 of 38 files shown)


examples/simple-examples/src/main/kotlin/ai/koog/agents/example/features/langfuse/Langfuse.kt
```diff
@@ -18,15 +27,22 @@ import io.opentelemetry.exporter.otlp.http.trace.OtlpHttpSpanExporter
  *  1. Set up a Langfuse project and credentials as described [here](https://langfuse.com/docs/get-started#create-new-project-in-langfuse)
  *  2. Get Langfuse credentials as described [here](https://langfuse.com/faq/all/where-are-langfuse-api-keys)
  *  3. Set `LANGFUSE_HOST`, `LANGFUSE_PUBLIC_KEY`, and `LANGFUSE_SECRET_KEY` environment variables
+ *  4. Set `OPENAI_API_KEY` from [here](https://platform.openai.com/account/api-keys)
  *
  * @see <a href="https://langfuse.com/docs/opentelemetry/get-started#opentelemetry-endpoint">Langfuse OpenTelemetry Docs</a>
  */
 suspend fun main() {
-    simpleOpenAIExecutor(ApiKeyService.openAIApiKey).use { executor ->
+    val server = startSseMcpServer(CIO, port = 30001, host = "localhost", ToolRegistry {
```
vova-jb: Image

agents/agents-tools/src/commonMain/kotlin/ai/koog/agents/core/tools/Tool.kt
```diff
@@ -24,6 +26,7 @@ public abstract class Tool<TArgs, TResult>(
     public val argsSerializer: KSerializer<TArgs>,
     public val resultSerializer: KSerializer<TResult>,
     public val descriptor: ToolDescriptor,
+    private val meta: Map<KClass<*>, Any> = emptyMap(),
```
tiginamaria: For me it's fine like that (better then string or storage key), but need @sdubov oppinion here

agents/agents-tools/src/commonMain/kotlin/ai/koog/agents/core/tools/ToolRegistry.kt
```diff
@@ -187,7 +186,13 @@ public class ToolRegistry private constructor(tools: List<Tool<*, *>> = emptyLis
 
         /**
          * A constant representing an empty registry with no tools.
+         * TODO: ToolRegistry is mutable but stored as an immutable object.
```
tiginamaria: Maybe put deprecation with empty() replacement?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/handler/tool/McpToolSupport.kt
```diff
@@ -0,0 +1,67 @@
...
+ * @property serverPort The port number on which the server listens, if applicable.
+ * @property siteUrl The server's public website URL, if provided by the server.
+ * @property instructions Optional server-provided instructions or documentation for using its tools.
+ * @property mcpProtocolVersion The MCP protocol version supported by this server connection.
+ * @property mcpTransportType The transport protocol type used for this connection (e.g., pipe for stdio, tcp for HTTP).
+ * TODO: support [sessionId] when it would be available in the sdk client
+ * @property sessionId Unique identifier for the current MCP session, if session management is enabled.
+ */
+@InternalAgentsApi
+public class McpServerMeta(
```
tiginamaria: For @sdubov: this was added to the core intentionally, as we don't wnat Otel dependend on MCP

tiginamaria (commented):
Thank you! Looks good to me, but I need @sdubov's second independent opinion here

agents/agents-tools/src/commonMain/kotlin/ai/koog/agents/core/tools/Tool.kt
tiginamaria: Also not sure how it will work with java api

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
```diff
@@ -91,7 +91,7 @@ public expect abstract class AIAgent<Input, Output> constructor() : Closeable {
             promptExecutor: PromptExecutor,
             agentConfig: AIAgentConfig,
             strategy: AIAgentGraphStrategy<Input, Output>,
-            toolRegistry: ToolRegistry = ToolRegistry.EMPTY,
+            toolRegistry: ToolRegistry = ToolRegistry.empty(),
```
sdubov: @vova-jb, good point. I think it worth creating a separate ticket and separate update though

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/handler/tool/McpToolSupport.kt
sdubov: Folks, as I see the metadata is used not for features, but for tool registry. I have a proposal to discuss here: What do you think if we add a general metadata key-value property (not directly referring MCP parameters) for our internal tools in the core module. MCP adds special MCP-related attributes to this metadata. We add this metadata to tool events, OTel will read this metadata and add necessary attributes to spans converting all or filtered MCP data into spans. WDYT?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/handler/tool/McpToolSupport.kt
```diff
@@ -0,0 +1,67 @@
...
+    public val serverUrl: String?,
+    public val serverPort: Int?,
+    public val siteUrl: String?,
+    public val instructions: String?,
+    public val mcpProtocolVersion: String,
+    public val mcpTransportType: McpTransportType,
+    public val sessionId: String?,
```
sdubov: We have session Id inside the framework. Maybe name it `mcpSessionId` to not mess with Koog session id?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/handler/tool/McpToolSupport.kt
```diff
@@ -0,0 +1,67 @@
...
+ * @property mcpProtocolVersion The MCP protocol version supported by this server connection.
+ * @property mcpTransportType The transport protocol type used for this connection (e.g., pipe for stdio, tcp for HTTP).
+ * TODO: support [sessionId] when it would be available in the sdk client
+ * @property sessionId Unique identifier for the current MCP session, if session management is enabled.
+ */
+@InternalAgentsApi
+public class McpServerMeta(
+    public val serverUrl: String?,
+    public val serverPort: Int?,
+    public val siteUrl: String?,
```
sdubov: Sorry, could you please clarify the difference between serverUrl and siteUrl? I've read the properties KDoc, but I feel that I not exactly got it.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/handler/tool/McpToolSupport.kt
```diff
@@ -0,0 +1,67 @@
...
+ * @property value The canonical name of the transport protocol.
+ */
+public enum class McpTransportType(public val value: String) {
+    /**
+     * Pipe-based transport using standard input/output streams.
+     *
+     * This transport is typically used for local MCP servers running as separate
+     * processes that communicate via stdio. Suitable for development and local tools.
+     */
+    Pipe("pipe"),
```
sdubov: @vova-jb, is not it called "stdio" mcp transport type by MCP providers?

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/attribute/McpAttributes.kt
```diff
@@ -0,0 +1,180 @@
...
+                override val key: String = super.key.concatKey("version")
+                override val value: String = version
+            }
+        }
+    }
+
+    /**
+     * JSON-RPC protocol attributes for MCP operations.
+     */
+    sealed interface JsonRpc : Attribute {
```
sdubov: JSON-RPC seems to be non-MCP span attributes. @vova-jb, does it maker sense to move it into a separate `JsonRpcAttributes` class?

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/attribute/McpAttributes.kt
```diff
@@ -0,0 +1,180 @@
...
+        data class Transport(private val transport: String) : Network {
+            override val key: String = "network.transport"
+            override val value: String = transport
+        }
+    }
+
+    /**
+     * Server address and port attributes for MCP client operations.
+     */
+    sealed interface Server : Attribute {
```
sdubov: The attribute is already added to `CommonAttributes`. Let's drop this one. The same for `Address` and `Port`.

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetry.kt
```diff
@@ -72,8 +82,7 @@ public class OpenTelemetry {
                 logger.debug { "Execute OpenTelemetry before agent started handler" }
 
                 val messages = eventContext.agent.agentConfig.prompt.messages.toList()
-                val tools = (eventContext.agent as? GraphAIAgent<*, *>)?.toolRegistry?.tools?.map { it.descriptor }?.toList()
-                    ?: emptyList()
+                val tools = eventContext.context.llm.toolRegistry.tools.map { it.descriptor }.toList()
```
sdubov, line 274: 👍

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetry.kt
```diff
@@ -552,142 +561,53 @@ public class OpenTelemetry {
...
 
+            @OptIn(InternalAgentsApi::class)
             pipeline.interceptToolCallStarting(this) intercept@{ eventContext ->
                 logger.debug { "Execute OpenTelemetry tool call handler" }
 
-                val patchedExecutionInfo = eventContext.executionInfo
-                    .appendRunId(eventContext.runId)
-                    .appendId(eventContext.eventId)
-
+                val path = eventContext.withRunAndEventId()
```
sdubov: @vova-jb , sorry, but I might loose the logic between the naming and the extension target here. Could you please clarify why is this an extension for context object as we extend not the context, but an execution info instead? Also, I have a feeling that this extension name might become invalid as we would need to append any other property to the span identifier :) I would propose to make this an extension of ExecutionInfo, pass context (or separate properties) as parameter(s) and name it more general, e.g. `patchExecutionInfo` or even `createSpanIdentifier`.

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetry.kt
```diff
@@ -552,142 +561,53 @@ public class OpenTelemetry {
...
-                    id = eventContext.eventId,
-                    toolName = eventContext.toolName,
-                    toolArgs = eventContext.toolArgs,
-                    toolDescription = eventContext.toolDescription,
-                    toolCallId = eventContext.toolCallId
-                )
-
-                spanAdapter?.onBeforeSpanStarted(executeToolSpan)
-                spanCollector.collectSpan(executeToolSpan, patchedExecutionInfo)
+                val toolCallSpan = startCallToolSpan(tracer, parentSpan, eventContext)
```
sdubov: The name of the span according to Otel semantic convention is `ExecuteTool`, not the `CallTool`. I would prefer to keep the naming for methods naming consistent to keep it easier for reading and understanding.

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetry.kt
```diff
@@ -800,6 +724,109 @@ public class OpenTelemetry {
...
 
+        @OptIn(InternalAgentsApi::class)
+        private fun startCallToolSpan(
+            tracer: Tracer,
+            parentSpan: GenAIAgentSpan,
+            eventContext: ToolCallStartingContext
+        ): GenAIAgentSpan {
+            val mcpTool = eventContext.context.llm.toolRegistry.getMcpToolMeta(eventContext.toolName)
+            return if (mcpTool != null) {
+                startMcpClientSpan(
```
sdubov: @vova-jb, am I right that we create whether McpClientSpan or ExecuteToolSpan? Do we expect that one replace another here? Should not it be that MCP span is wrapped into Execute Tool Span? Does semantic convention define this?

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetry.kt
```diff
@@ -800,6 +724,109 @@ public class OpenTelemetry {
...
+                path = path
+            )
+        }
+
+        @OptIn(InternalAgentsApi::class)
+        private fun SpanCollector.getStartCallToolSpan(
+            patchedExecutionInfo: AgentExecutionInfo,
+            eventContext: ToolCallEventContext
+        ): GenAIAgentSpan? {
+            val mcpTool = eventContext.context.llm.toolRegistry.getMcpToolMeta(eventContext.toolName)
```
sdubov: @vova-jb, do you have tests to verify this? How would this work when you have MCP tool registered in the tool registry and you invoke a non-map tool?

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetry.kt
```diff
@@ -800,6 +724,109 @@ public class OpenTelemetry {
...
+                )
+            }
+            spanCollector.removeSpan(
+                span = callToolSpan,
+                path = path
+            )
+        }
+
+        @OptIn(InternalAgentsApi::class)
+        private fun SpanCollector.getStartCallToolSpan(
```
sdubov: Would it make sense to name it `getStarted...` instead of `getStart...`?

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetry.kt
```diff
@@ -800,6 +724,109 @@ public class OpenTelemetry {
...
+                startMcpClientSpan(
+                    tracer = tracer,
+                    parentSpan = parentSpan,
+                    id = eventContext.eventId,
+                    toolName = eventContext.toolName,
+                    toolArgs = eventContext.toolArgs,
+                    toolDescription = eventContext.toolDescription,
+                    toolCallId = eventContext.toolCallId,
+                    sessionId = mcpTool.sessionId,
+                    requestId = null,
```
sdubov: requestId = null?

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/span/mcpClientSpan.kt
```diff
@@ -0,0 +1,216 @@
...
+        builder.addAttribute(McpAttributes.Mcp.Session.Id(session))
+    }
+
+    // mcp.protocol.version (RECOMMENDED)
+    mcpProtocolVersion.let { version ->
+        builder.addAttribute(McpAttributes.Mcp.Protocol.Version(version))
+    }
+
+    // network.transport (RECOMMENDED)
+    mcpTransportType.let { transport ->
```
sdubov: `mcpTransportType` is not an optional parameter. You can unwrap it from `let`.

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/span/mcpClientSpan.kt
```diff
@@ -0,0 +1,216 @@
...
+        builder.addAttribute(McpAttributes.JsonRpc.Request.Id(reqId))
+    }
+
+    // mcp.session.id (RECOMMENDED)
+    sessionId?.let { session ->
+        builder.addAttribute(McpAttributes.Mcp.Session.Id(session))
+    }
+
+    // mcp.protocol.version (RECOMMENDED)
+    mcpProtocolVersion.let { version ->
```
sdubov: `mcpProtocolVersion` is not an optional parameter here, you can unwrap it from `let`.

agents/agents-features/agents-features-opentelemetry/src/jvmTest/kotlin/ai/koog/agents/features/opentelemetry/span/McpSpanTest.kt
```diff
@@ -0,0 +1,186 @@
...
+        val toolResult = buildJsonObject {
+            put("result", JsonPrimitive("success"))
+        }
+
+        endMcpClientSpan(
+            span = span,
+            toolCallResult = toolResult
+        )
+
+        val resultAttr = span.attributes.find { it.key == "gen_ai.tool.call.result" }
```
sdubov: You can also generate get the key from the instance instead of hardcode the string here. This would help if the structure is changed at some point.

agents/agents-features/agents-features-opentelemetry/src/jvmTest/kotlin/ai/koog/agents/features/opentelemetry/span/McpSpanTest.kt
```diff
@@ -0,0 +1,186 @@
...
+        assertNotNull(resultAttr)
```
sdubov: Would that make sense to verify the exact content of the attribute value as well?

agents/agents-features/agents-features-opentelemetry/src/jvmTest/kotlin/ai/koog/agents/features/opentelemetry/span/McpSpanTest.kt
```diff
@@ -0,0 +1,186 @@
...
+            method = McpMethod.TOOLS_CALL,
+            mcpProtocolVersion = "2025-06-18",
+            mcpTransportType = McpTransportType.Pipe
+        )
+
+        assertEquals(SpanKind.CLIENT, clientSpan.kind)
+    }
+
+    @Test
+    fun `MCP spans should support hierarchical structure`() {
```
sdubov: For me it looks like a more general test that is verified under SpanTests.

agents/agents-features/agents-features-opentelemetry/src/jvmTest/kotlin/ai/koog/agents/features/opentelemetry/span/McpSpanTest.kt
```diff
@@ -0,0 +1,186 @@
...
+            span = span,
+            toolCallResult = toolResult
+        )
+
+        val resultAttr = span.attributes.find { it.key == "gen_ai.tool.call.result" }
+        assertNotNull(resultAttr)
+    }
+
+    @Test
+    fun `MCP client have expected SpanKind`() {
```
sdubov: What do you think if we create a single test that verifies all common parameters for span instead of checking them one by one in separate tests.

agents/agents-mcp-server/src/jvmTest/kotlin/ai/koog/agents/mcp/server/KoogToolAsMcpToolTest.kt
```diff
@@ -26,11 +27,11 @@ import kotlin.test.assertNotNull
 import kotlin.test.assertTrue
 import kotlin.time.Duration.Companion.seconds
 
+@InternalAgentsApi
```
sdubov: Could you please clarify why did you mark a test class as `InternalAgentToolsApi`. This annotation is used for internal methods that we explicitly mark for a users to show that it is not a public API.

agents/agents-mcp/src/jvmTest/kotlin/ai/koog/agents/mcp/McpToolTest.kt
```diff
@@ -27,6 +31,7 @@ import kotlin.time.Duration.Companion.minutes
 import kotlin.time.Duration.Companion.seconds
 
 @Execution(ExecutionMode.SAME_THREAD)
+@InternalAgentsApi
```
sdubov: Same comment as above for tests.

agents/agents-tools/src/commonMain/kotlin/ai/koog/agents/core/tools/Tool.kt
sdubov: I have couple notes here: 
1. I would name it `metaInfo` to match the naming with `Message` class that we use.
2. I'm not sure how this approach should be extended if we would like to pass this meta info through network. I would assume that this should be serrializable class. Otherwise, this data is only for for internal usage.
3. How would that work with Java API?
4. What is the problem with key-value approach here?

examples/simple-examples/src/main/kotlin/ai/koog/agents/example/features/langfuse/Langfuse.kt
```diff
@@ -18,15 +27,22 @@ import io.opentelemetry.exporter.otlp.http.trace.OtlpHttpSpanExporter
  *  1. Set up a Langfuse project and credentials as described [here](https://langfuse.com/docs/get-started#create-new-project-in-langfuse)
  *  2. Get Langfuse credentials as described [here](https://langfuse.com/faq/all/where-are-langfuse-api-keys)
  *  3. Set `LANGFUSE_HOST`, `LANGFUSE_PUBLIC_KEY`, and `LANGFUSE_SECRET_KEY` environment variables
+ *  4. Set `OPENAI_API_KEY` from [here](https://platform.openai.com/account/api-keys)
```
sdubov, line 30: Changes in this file seem unrelated to the code from the PR. Maybe we can do it separately if needed?

examples/simple-examples/env.template.properties
```diff
@@ -4,11 +4,25 @@
...
+#
+# git clone https://github.com/langfuse/langfuse.git
+# cd langfuse
+# docker compose up
+#
+# then open http://localhost:3000
+#
+LANGFUSE_SECRET_KEY=sk-lf-...
+LANGFUSE_PUBLIC_KEY=pk-lf-...
+LANGFUSE_HOST=http://localhost:3000
```
sdubov: Could you please skip the value part as it is defined for other env vars to keep it consistent?

sdubov (commented):
@vova-jb, thank you for your implementation. I looked through general approach and through details. Please see my comments in review. I would like to discuss two general questions that was raised in my mind:
1. How do we compose / wrap spans for MCP vs regular tool call spans?
2. How do we present meta information for tools?

I would also add someone from the team to discuss this together if you do not mind.

examples/simple-examples/src/main/kotlin/ai/koog/agents/example/features/langfuse/Langfuse.kt
vova-jb: It is related to the 2nd commit. It makes sense to me to have an example in the same pr as the iml

agents/agents-tools/src/commonMain/kotlin/ai/koog/agents/core/tools/Tool.kt
tiginamaria: For key-value you mean string-string? It bas another option, but It's not type-safe and if I want MCP+ACP tool (for example) and it could cause key clashing. On the other hand it's easy to serialize and it's for sure java-friendly

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetry.kt
vova-jb: McpOpenTelemetryIntegrationTest

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/attribute/McpAttributes.kt
vova-jb: We don't have JSONJSON data available from mcp client for now, so I will remove it

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/handler/tool/McpToolSupport.kt
```diff
@@ -0,0 +1,60 @@
+package ai.koog.agents.core.feature.handler.tool
+
+import ai.koog.agents.core.annotation.InternalAgentsApi
+
+/**
+ * Metadata keys for tools provided by MCP servers.
+ */
+@InternalAgentsApi
+public object McpMetaDataKeys {
```
EugeneTheDev: You're adding MCP-specific entities to the `agents-core` module, while MCP tool support is implemented in its own module. This would make core even bigger and also violate separation of concerns, which makes it harder to search and reason about related components.

Also nit: `McpMetadataKeys` instead of `MetaData`

agents/agents-tools/src/commonMain/kotlin/ai/koog/agents/core/tools/Tool.kt
```diff
@@ -19,11 +19,13 @@ import kotlinx.serialization.json.jsonObject
  * @property argsSerializer A [KSerializer] responsible for encoding and decoding the arguments required for the tool execution.
  * @property resultSerializer A [KSerializer] responsible for encoding and decoding the result returned by the tool execution.
  * @property descriptor A [ToolDescriptor] representing the tool's schema, including its name, description, and parameters.
+ * @property metaData A map of arbitrary metadata associated with the tool.
  */
 public abstract class Tool<TArgs, TResult>(
     public val argsSerializer: KSerializer<TArgs>,
     public val resultSerializer: KSerializer<TResult>,
     public val descriptor: ToolDescriptor,
+    public val metaData: Map<String, String> = emptyMap(),
```
EugeneTheDev: It's not that bad, since we don't have `Any` here. However, this still looks like an abstraction with a specific purpose in mind (just for MCP telemetry support for now), rather than something that is actually needed for the general tool usage. I'm not strongly against it, but I think we need to be mindful of where do we add these dynamic "metadata" fields. For now, maybe we can keep it. A couple of nits though:    
1. Rename it to `metadata` instead of `metaData`, we already have such fields, sounds a bit better
2. Add `metadata` parameter to the secondary constructor too

EugeneTheDev (changes requested):
I have question regarding the separation of concerns. 

I see you're trying to avoid making `agents-features-opentelemetry` depend on `agents-mcp`. I assume that is to avoid unnecessary pulling MCP-related dependencies, such as MCP SDK, when it's not used by the user, which is the right concern. However, trying to achieve that, you're leaking these implementation details to the `agents-core`, spreading your components over multiple modules and making it harder to reason about. 

One possible solution is to create a small new module, something like `agents-mcp-opentelemetry`, and to move these `McpMetadataKeys` and other mcp-related entities from `agents-core` to this new module. And then `agents-features-opentelemetry` and `agents-mcp` can both have a dependency on this helper module, and we won't be pulling MCP SDK unnecessary too. This is a nice solution from the dependency management point of view, the downside is, however, that you will be creating a tiny module with just a couple of classes inside. But I think it's worth it, considering that it will provide us with better separation of concerns and proper dependency management. WDYT?

vova-jb:
@EugeneTheDev I made a bit more generalised module `agents-mcp-metadata`, It looked to me more appropriate and more extendable in the future

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetry.kt
```diff
@@ -603,122 +614,60 @@ public class OpenTelemetry {
...
                 )
+                val mcpToolMetaData = eventContext.context.llm.toolRegistry.getMcpToolMeta(eventContext.toolName)
+                if (mcpToolMetaData != null) {
+                    executeToolSpan.enrichExecuteToolSpanWithMcpAttrs(
+                        toolName = eventContext.toolName,
+                        sessionId = mcpToolMetaData[McpMetadataKeys.McpSessionId],
+                        method = McpMethod.TOOLS_CALL,
+                        serverPort = mcpToolMetaData[McpMetadataKeys.ServerPort]?.toIntOrNull(),
+                        serverAddress = mcpToolMetaData[McpMetadataKeys.ServerUrl],
+                        mcpProtocolVersion = requireNotNull(mcpToolMetaData[McpMetadataKeys.McpProtocolVersion]) {
```
sdubov: @vova-jb , how critical is this to be not null? I'm a bit concern that it might fail the agent execution if mcp did not provide you a protocol version. Would that be reasonable to assume that any metadata key can be not defined and just skip all missing (null) map entries from tool metadata?

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetry.kt
```diff
@@ -603,122 +614,60 @@ public class OpenTelemetry {
...
+                            "MCP protocol version is required for MCP tool calls."
+                        },
+                        mcpTransportType = requireNotNull(mcpToolMetaData[McpMetadataKeys.McpTransportType]) {
```
sdubov: Same comment here.

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/feature/OpenTelemetry.kt
```diff
@@ -603,122 +614,60 @@ public class OpenTelemetry {
                     toolDescription = eventContext.toolDescription,
                     toolCallId = eventContext.toolCallId
                 )
+                val mcpToolMetaData = eventContext.context.llm.toolRegistry.getMcpToolMeta(eventContext.toolName)
```
sdubov: I would keep the naming pattern consistent based on the [comment](https://github.com/JetBrains/koog/pull/1421#discussion_r2758480145) from @EugeneTheDev:
`mcpToolMetaData` -> `mcpToolMetadata`

agents/agents-mcp/src/commonMain/kotlin/ai/koog/agents/mcp/McpTool.kt
```diff
@@ -19,13 +20,16 @@ import kotlinx.serialization.json.jsonObject
  * 2. Calling the MCP tool through the MCP client
  * 3. Converting MCP tool results back to agent framework tool results
  */
+@InternalAgentsApi
 public class McpTool(
     private val mcpClient: Client,
     descriptor: ToolDescriptor,
+    metaData: Map<String, String>,
```
sdubov: I guess it also with renaming:
`metaData` -> `metadata`

agents/agents-mcp/src/commonMain/kotlin/ai/koog/agents/mcp/McpToolRegistryProvider.kt
```diff
@@ -56,19 +64,58 @@ public object McpToolRegistryProvider {
...
+    private fun buildToolRegistry(
+        sdkTools: List<Tool>,
+        mcpToolParser: McpToolDescriptorParser,
+        serverInfo: McpServerInfo,
+        mcpClient: Client
+    ): ToolRegistry = ToolRegistry {
+        sdkTools.forEach { sdkTool ->
+            try {
+                val toolDescriptor = mcpToolParser.parse(sdkTool)
+                val toolMetaData = mapOf(
```
sdubov, line 104: `toolMetaData` -> `toolMetadata`.

sdubov (approved):
I like current implementation a lot. Nice approach with extracting metadata into a separate package. Approve. I've added several comments in `OpenTelemetry` feature that I think worth looking at.
1. I would loosen the requirement here in the feature when getting entries from a map
2. Naming pattern

Other than that, LGTM. @vova-jb, thank you!

EugeneTheDev (approved):

PR #1424, opened 2026-01-27
Introduce a transforming interceptor to EventHandler

serge-p7v:
Add onLLMPromptTransforming handler that can transform Prompt before passing it to onLLMCallStarting.

## Motivation and Context
For some use cases like RAG or chat memory it might be useful to intercept every request to an LLM in order to augment it somehow (with relevant documents from a knowledge base for RAG or with previous messages for chat memory).

## Breaking Changes
None.

---

#### Type of the changes
- [ ] Documentation update
- [ ] Tests improvement
- [ ] Refactoring
- [ ] CI/CD changes
- [ ] Dependencies update

#### Checklist
- [X] All new and existing tests passed

### DIFF
--- a/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/ContextualPromptExecutor.kt
+++ b/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/ContextualPromptExecutor.kt
@@ -38,17 +38,36 @@ public class ContextualPromptExecutor(
         @OptIn(ExperimentalUuidApi::class)
         val eventId = Uuid.random().toString()
 
-        logger.debug { "Executing LLM call (event id: $eventId, prompt: $prompt, tools: [${tools.joinToString { it.name }}])" }
-        context.pipeline.onLLMCallStarting(eventId, context.executionInfo, context.runId, prompt, model, tools, context)
+        logger.debug { "Transforming prompt (event id: $eventId, prompt: $prompt, tools: [${tools.joinToString { it.name }}])" }
+        val transformedPrompt = context.pipeline.onLLMPromptTransforming(
+            eventId,
+            context.executionInfo,
+            context.runId,
+            prompt,
+            model,
+            tools,
+            context
+        )
 
-        val responses = executor.execute(prompt, model, tools)
+        logger.debug { "Executing LLM call (event id: $eventId, prompt: $transformedPrompt, tools: [${tools.joinToString { it.name }}])" }
+        context.pipeline.onLLMCallStarting(
+            eventId,
+            context.executionInfo,
+            context.runId,
+            transformedPrompt,
+            model,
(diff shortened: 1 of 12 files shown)


agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/ContextualPromptExecutor.kt
```diff
@@ -38,17 +38,36 @@ public class ContextualPromptExecutor(
         @OptIn(ExperimentalUuidApi::class)
         val eventId = Uuid.random().toString()
 
-        logger.debug { "Executing LLM call (event id: $eventId, prompt: $prompt, tools: [${tools.joinToString { it.name }}])" }
-        context.pipeline.onLLMCallStarting(eventId, context.executionInfo, context.runId, prompt, model, tools, context)
+        logger.debug { "Transforming prompt (event id: $eventId, prompt: $prompt, tools: [${tools.joinToString { it.name }}])" }
+        val transformedPrompt = context.pipeline.onLLMPromptTransforming(
```
serge-p7v, line 42: TODO: other methods here like "executeStreaming" or "moderate" should also call "onLLMPromptTransforming".

sdubov: @serge-p7v, could you please clarify the general question here. With this update, we have two events with same set of parameters one after another. It looks a bit redundant for me, if I do not miss anything, as you can transform prompt inside the `onLLMCallStarting` handler and `prompt` is a variable in the `AIAgentLLMWriteSession` that can be updated. Would that work as well? Why do we need a separate interceptor here?

serge-p7v: Currently, it is not possible to change the prompt inside of `onLLMCallStarting` (the prompt is immutable); for the intercepting approach we need to modify `onLLMCallStarting`. There might be two issues here:
* Prompt transformations through AIAgentLLMWriteSession would change the conversation state. In the PR they are applied to a current call to an LLM (they are kind of "transient").
* `onLLMCallStarting` would become a possibly mutating handler and it would be harder to reason about what it can do in a chain. In the PR `onLLMPromptTransforming` does only prompt transformation, so by chaining multiple `onLLMCallStarting` with multiple `onLLMPromptTransforming` we can be sure where the transformation is.

About the same set of parameters: good idea, thank you, most likely the transformer needs only the prompt and the context!

serge-p7v: @sdubov has suggested a great idea with extracting memory-specific handlers into a different feature. Converting the PR to draft.

PR #1376, opened 2026-01-19
Introduce a new feature for long-term memory

serge-p7v:
This PR has three layers:

1. A set of **strategies** with predefined graph **nodes** for common use-cases (like Retrieval-Augmented Generation or memory extraction).
2. A new **feature** that accepts a repository interface and has methods for interacting with it (store and search).
3. A **repository interface** and one implementation (for pgvector) for communicating with a vector database.

The approach with strategies and nodes is suitable for fixed pipelines (like naive RAG). For non-deterministic pipelines (like Agentic RAG) I added two tools that accept a repository interface to access a vector database. There is a wrapper ScopedMemoryRecordRepository that filters requests to the underlying repository by adding metadata filter.

The PR is focused on "long-term" memory, the storage&retrieval of chat messages will be added in a separate PR.

## Motivation and Context
The idea behind this PR is to provide a more flexible way to implement use-cases that require storage of chat messages or documents in external databases (use-cases like chat-bot with memory or RAG). Existing interfaces like RankedDocumentStorage and AgentMemoryProvider were designed for different use-cases.

## Breaking Changes
None.

---

#### Type of the changes
- [x] New feature (non-breaking change which adds functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Tests improvement
- [ ] Refactoring
- [ ] CI/CD changes
- [ ] Dependencies update

#### Checklist
- [x] The pull request has a description of the proposed change
- [x] I read the [Contributing Guidelines](https://github.com/JetBrains/koog/blob/main/CONTRIBUTING.md) before opening the pull request
- [x] The pull request uses **`develop`** as the base branch
- [x] Tests for the changes have been added
- [x] All new and existing tests passed

##### Additional steps for pull requests adding a new feature
- [ ] An issue describing the proposed change exists
- [ ] The pull request includes a link to the issue
- [ ] The change was discussed and approved in the issue
- [ ] Docs have been added / updated

PR #1437, opened 2026-01-30
Long term memory with interceptors

serge-p7v:
The PR introduces the following:
* a new feature with pipeline interceptors that can augment a user prompt before sending it to an LLM via provided retriever and optionally store assistant responses in a provided repository;
* `MemoryRecordRepository` abstraction for working with vector databases and three implementations: in-memory `EphemeralMemoryRecordRepository`, persistent `PGVectorMemoryRecordRepository` and a delegating `ScopedMemoryRecordRepository`;
* two tools for similarity and keyword search for Agentic RAG use cases.

## Motivation and Context
The purpose of this feature is to provide a convenient integration with vector databases and agentic memory providers. So for automatic Retrieval-Augmented Generation users would only need to install the feature with their implementation of `MemoryRecordRepository` interface and their `MemoryRecordRetriever` lambda function without changing their agent strategies. Existing interfaces like `RankedDocumentStorage` and `AgentMemoryProvider` were designed for different use-cases.

## Breaking Changes
None.

---

#### Type of the changes
- [ ] Documentation update
- [ ] Tests improvement
- [ ] Refactoring
- [ ] CI/CD changes
- [ ] Dependencies update

#### Checklist


serge-p7v:
The PR is already too long, that's why I decided to move the following things to the next PR:
* all the logic related to memory nodes and strategies from the first PR (https://github.com/JetBrains/koog/pull/1376) is not included, but some of it will return, e.g. store and search methods inside the feature and `nodeExtractAndSaveMemoryRecord` to extract typed records from chat and storing them;
* filter expression DSL instead of unsafe strings in `SearchRequest`;
* parameters for RAG like query transformation and results post-processing.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/ContextualPromptExecutor.kt
```diff
@@ -82,19 +91,32 @@ public class ContextualPromptExecutor(
...
-                context.pipeline.onLLMStreamingStarting(
-                    eventId,
-                    context.executionInfo,
-                    context.runId,
-                    prompt,
-                    model,
-                    tools,
-                    context
-                )
+        return kotlinx.coroutines.flow.flow {
```
EugeneTheDev, line 94: nit: import it instead

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/ContextualPromptExecutor.kt
```diff
@@ -134,7 +156,7 @@ public class ContextualPromptExecutor(
             }
     }
 
-    // TODO: Add Pipeline interceptors for this method
+    // TODO: Add Pipeline interceptors for this method. Without them features cannot modify prompts before calls to LLMs.
```
EugeneTheDev, line 159: Can't we just use `onLLMCallStarting` here?

agents/agents-features/agents-features-memory/src/commonMain/kotlin/ai/koog/agents/memory/feature/Memory2.kt
```diff
@@ -0,0 +1,292 @@
...
+    /**
+     * Configuration for the Memory2 feature.
+     *
+     * This class allows configuring:
+     * - The record repository
+     * - Whether to persist assistant messages as memory records
+     * - A retriever for retrieving relevant context (similarity, keyword, or hybrid)
+     * - How context should be inserted into prompts
+     */
+    public class Config : FeatureConfig() {
```
EugeneTheDev, line 64: I have some suggestions regarding configuration (and consequentially `Memory2` class) properties. Currently, it doesn't look consistent enough, the way writing/reading (augumenting) are configured. The main idea is that since these operations mirror each other, their configurations should be similar and mirror each other too in terms of pattern/naming. On top of that, ideally it should be clear enough, which property enables/disables automatic reading or writing just by looking at them. Currently, enabling/disabling part also looks a bit implicit and therefore confusing. 

I propose the following structure:
* `memoryRecordRepository` - looks good, just configures the repository to store memory, no changes needed.
* `retriever`, `contextInsertionMode`, `systemPromptTemplate`, `userContextTemplate` are all related to reading from memory and augumenting the prompt. `retriever` controls, whether this automatic augmentation is enabled or not. So if it is not set, the other three option have have no effect. Moreover, these three options are also related to each other, and e.g. if you don't use `SYSTEM_MESSAGE` context insertion mode, then `systemPromptTemplate` also has no effect. So the refined structure could be as follows:
```kotlin
// single property that controls all automatic augmentation behavior
// naming is not ideal, we can discuss a better name
// I feel like "augmentation" might also sound too comples and confusing
public var promptAugmentation: PromptAugmentation? = null

// Config class
class PromptAugmentation(
    val retriever: MemoryRecordRetriever,
    val contextInsertion: ContextInsertion,
)

// Config classes to control related context insertion configurations
sealed interface ContextInsertion {
    class SystemMessage(
        val systemPromptTemplate: String = ...,
    ) : ContextInsertion
    
    class UserMessageBeforeLast(
        val userContextTemplate: String = ...,
    ) : ContextInsertion
    
    class UserMessageAugmentLast(
        val userContextTemplate: String = ...,
    ) : ContextInsertion
}


```

* `persistAssistantReponses`. Let's make it mirror the option above. When you use similar approaches, namings and patterns, it makes it eaasier to reason about for someone who's not familiar with the code yet, i.e., for the users using this feature.
Also, why are we peristing only assistant messages by default? Can there be some useful information in user messages or tool calls too? Wdyt expanding the number of modes here too?

```kotlin
// mirroring the "read/augmentation" configuration
public var promptPersistence: PromptPersistence? = null

// Making it a separate class in case we want to add more configuration properties in the future
class PromptAugmentation(
    // control the types of messages to persist
   // it's a list to make it easier support different combinations of message types, in case we'll add more (e.g. Reasoning or Tool) in the future
    val persistMessages: List<PersistMessages>
)

enum class PersistMessages {
    User,
    Assistant,
 }
```

rag/vector-storage/src/commonMain/kotlin/ai/koog/rag/vector/database/MemoryRecordRepository.kt
```diff
@@ -0,0 +1,279 @@
...
+ * This interface is designed to:
+ * - Support reactive/coroutine-based operations
+ * - Return similarity scores with results
+ * - Support streaming via Flow
+ * - Provide comprehensive CRUD operations
+ * - Support direct vector search
+ * - Support multiple search types
+ *
+ */
+public interface MemoryRecordRepository {
```
EugeneTheDev, line 166: Regarding all these memory-related repositories in the `vector-rag` module:
1. There already exists `VectorStorage` and related abstractions. You're inventing your own abstraction that partially looks similar, but doesn't reuse existing abstractions.
2. You're putting memory-specific entities in the module that is not related to memory, which doesn't look clean in terms of separation of concerns principle

So my suggestions are:
1. Consider reusing existing vector storage abstractions. I think it's fine to modify these if it turns out they don't suite your needs because they're not flexible and generic enough. You can still have your own abstractions if they would bring some new functionality. If you do so, consider, which approach would be the best here: inheritance or delegation. I guess the latter would be cleaner, to avoid too cumbersome hierarchies.
2. Regarding the separation of concerns, you can either make these abstractions generic enough, not mentioning memory in any way, or move these to the memory feature module. And since we already have a number of RAG and vector storage abstractions, I suggest looking into the second option first. Maybe you can completely reuse (with minor modifications) existing abstractions, and then these custom classes won't be needed, or they would be small and specific enough to place them in the memory feature module

EugeneTheDev (changes requested):
Thanks for working on this feature, I have a number of suggestions

rag/vector-storage/src/commonMain/kotlin/ai/koog/rag/vector/database/ScopedMemoryRecordRepository.kt
```diff
@@ -0,0 +1,172 @@
...
+ *
+ * This class wraps an existing [MemoryRecordRepository] and:
+ * - For [add] and [update] operations: adds the [scopeValue] as a metadata field to all records
+ * - For [search] operations: adds a filter expression to only return records matching the [scopeValue]
+ *
+ * @property delegate The underlying [MemoryRecordRepository] to delegate operations to
+ * @property scopeKey The metadata field name used to store the memory scope ID, defaults to [DEFAULT_MEMORY_SCOPE_FIELD_NAME]
+ * @property scopeValue The scope ID to apply to all operations
+ */
+public class ScopedMemoryRecordRepository(
```
serge-p7v, line 18: After some thought this particular implementation in the PR looks redundant: it's relatively easy to create it in the calling code and after introducing FilterExpression DSL it has to be updated.
What is really missing is the ability to set a "namespace" in the interface methods. "Namespace" can be a table/collection/index name or something more granular for working with incomparable/incompatible records inside one database.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/ContextualPromptExecutor.kt
serge-p7v: Fixed in a new PR

agents/agents-features/agents-features-memory/src/commonMain/kotlin/ai/koog/agents/memory/feature/Memory2.kt
serge-p7v: Thank you! I think with shared memoryRecordRepository I violated interface segregation principle. I fixed that in the new PR https://github.com/JetBrains/koog/pull/1490

rag/vector-storage/src/commonMain/kotlin/ai/koog/rag/vector/database/MemoryRecordRepository.kt
serge-p7v: Thank you! I moved all RAG/vector related code from this PR into a new PR - https://github.com/JetBrains/koog/pull/1474

serge-p7v:
I reworked this PR and split it in two according to the suggestions above:
* the code related to RAG (or vector databases) is now in https://github.com/JetBrains/koog/pull/1474 ("low-level" abstractions)
* the code related to the new memory feature is now in https://github.com/JetBrains/koog/pull/1490 ("high-level" abstractions)
I'm closing this PR.

PR #1452, opened 2026-02-03
feat(prompt)!: make models list method in clients to return list of LLModel instances instead of String

tiginamaria:
Make the models list method in clients return a list of LLModel instances instead of a String
BREAKING:
* `LLModel.capabilities` and `LLModel.contextLenght ` became nullable
* `LLMClient` models not returns `List<LLModel>` instead of `List<String>`
* `LLModelDefinitions.list()` and `allModelsIn` now internal for test-use only


prompt/prompt-executor/prompt-executor-clients/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/LLModelDefinitionsExt.kt
```diff
@@ -15,11 +17,19 @@ import kotlin.reflect.full.memberProperties
  *   and its nested objects.
  */
 public fun allModelsIn(obj: Any): List<LLModel> {
```
EugeneTheDev, line 21: Are these workarounds still needed, if we have a proper explicit `models` list now? Same about the `list` funciton below

prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openai/OpenAILLMClient.kt
```diff
@@ -559,7 +559,11 @@ public open class OpenAILLMClient @JvmOverloads constructor(
             )
         }
 
-        return openAIResponse.data.map { it.id }
+        val models = OpenAIModels.models.associateBy { it.id }
+
+        return openAIResponse.data.mapNotNull {
```
EugeneTheDev: Please note that this approach will implicitly skip models that are unknown to us, which might be confusing for users. Throwing an exception in such case won't be optimal too. How about creating `LLModel` instances in place for unknown models? The main info needed for that is provider (which is always the same per client, OpenAI in this case) and model id (which API returns). I guess you can set `capabilities` and `contextLenght` to `null/-1`, since we don't have this data from the API. But having a proper model id would already allow calling it when using this client. 

The sane goes for other clients where the same approach is implemented

EugeneTheDev (changes requested):

prompt/prompt-executor/prompt-executor-clients/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/LLModelDefinitionsExt.kt
tiginamaria: This is required for testing, I can make it internal

prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openai/OpenAILLMClient.kt
tiginamaria: I can changle it, no problem. I also thought about it, but I desided that having null/-1 camabilities and contextLenght is more confusing that having (maybe) not all models. I mean that we have pretty whole list of models from main providers, do you think the case defining custom models is common?

prompt/prompt-executor/prompt-executor-clients/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/LLModelDefinitionsExt.kt
EugeneTheDev: Ah, I see. I guess you can't make it `internal` since it is used in other modules. But you can annotate these functions with `@VisibleForTesting` to make it clear that it's meant for tests only and signal improper usage.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openai/OpenAILLMClient.kt
tiginamaria: The idea was that you get the models from the client, choose the models and set to the agent context, but having empty capabilities, we need to do extra work to set them. 
Maybe we can add custom models as parameter?
```
override suspend fun models(customModels: List<LLModel>? = null): List<LLModel> {
    ...        
    val models = (OpenAIModels.models + customModels).associateBy { it.id }
    return openAIResponse.data.mapNotNull {
        models[it.id]
    }
}
```

prompt/prompt-executor/prompt-executor-clients/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/LLModelDefinitionsExt.kt
antoniibelyshev: Would it be cleaner to always define `models = list()`? I see that we require that in tests. This way, we do not need to manually add a new model to the `models`: it will appear there once it's defined inside the `LLModelsDefinition`

prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openai/OpenAILLMClient.kt
EugeneTheDev: The probem is that users might not always use the latest version of the framework, so they won't always have the most up-to-date list of models. And LLM providers also release new models quite often, so there's also a certain gap between model release and new Koog version with updated model list

tiginamaria: Now I'm thinking that returning Strings = ids were a good idea initially, wdyt? Maybe this change is senseless having your concern

antoniibelyshev: Can we change the capabilities logic so that we can set explicitly specify that a model *does not* have a capability? Then we would have three options for every capability, `has<name>Capability`, `doesNotHave<name>Capability`, and `null` (or an explicit `unsure` type) signaling that we do not know that. though, seems like a big change.

EugeneTheDev: Yes, that's sounds like a more major refactoring. Maybe another, not so major options, would be to make this property nullable. So we can differentiate between the cases when the capabilites are empty vs when we don't know these. Still a breaking change, but at least it's less breaking

prompt/prompt-executor/prompt-executor-clients/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/LLModelDefinitionsExt.kt
tiginamaria: list() is jvm-only, unfortunately

prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openai/OpenAILLMClient.kt
tiginamaria: I'm still a bit worried about users who get a list of models for a client and set them to the agent without capabilities (they could easily get, e.g., a tool call not supported, as no capabilities are set). So what if we use the current solution but add `addCustomModel` method to the `LLModelDefinitions` and append the model to the list of available models under this definition

antoniibelyshev: If they retrieve an existing model and it does not have a tool call capability, they will also get the same error

tiginamaria: But they expect it as they see a list of capabilities, but here they get the same model as in `LLModelDefinitions` but without

tiginamaria: So they need an additional step every time to substitute the LLModel they get from .models() with one from `LLModelDefinitions` or custom

antoniibelyshev: > the same model as in LLModelDefinitions but without

why? If we have the model, we can return the instance with the correct list of capabilities. If not, we return a model with no capabilities (as we do not know its capabilities).

Users can't really "see" what they received, since it's inside the strategy. For example, if they need a model with the tool calling capability, they'll need a programmatic filtering anyway, since existing models could lack the necessary capability.

Also, can we loosen the requirement for the capabilities in the executors from failing to to a warning when the model does not have a necessary capability? Then a model with no capabilities (when we did not know the capabilities) could still be used for tool calling.

tiginamaria: Ah, got you. Maybe in this way it's alright

tiginamaria: So what I propose is to as you said make capabilities nullable and return default `LLModel` with `null` capabilities and `-1` `contextLenght` but also add `addCustomModel` to `LLModelDefinitions` in case user wants to define new model and get not empty capabilities after calling `LLModel.models()`. Can you please take one more look at it

EugeneTheDev: Let's also make `contextLength` nullable? This should communicate the state "we don't know" more clearly than some `-1`, which can be perceived as confusing

EugeneTheDev (approved):
Please take a look at my comment regarding the nullable `contextLength`. Also, if your PR has breaking changes (as you indicated in the `BREAKING` section), please use `!` after the type/scope in the PR title. So it should be `feat(prompt)!: ...`

antoniibelyshev (approved):
Thank you, looks good!

PR #1457, opened 2026-02-05
docs: add API reference link parsing from FQNs

marinkovicmarko:
Adds a new functionality that allows users to create API reference links in the documentation using fully qualified names as link targets instead of absolute URLs to the specific API documentation page. 

The implementation is a custom Python script triggered by the mkdocs `on_page_markdown` hook. The script also includes link verification, so all API references that are resolved to URLs are verified against the current API reference documentation at [api.koog.ai](https://api.koog.ai).

Warnings and notes:
* Links will be resolved only if they point to an existing API documentation page at [api.koog.ai](https://api.koog.ai). Local Dokka-generated API references are not supported.
* The current version of the [Agent Client Protocol](https://docs.koog.ai/agent-client-protocol/) documentation page includes links to non-existing API reference pages, so the links were ignored during initial replacement.
* The current version of the [Class-based tools](https://docs.koog.ai/class-based-tools/) page includes links to API entities that no longer exist. Proper links to be added after the page has been updated.


docs/process-api-links.py
```diff
@@ -0,0 +1,326 @@
+import re
+from urllib import request, error
+from urllib.parse import urljoin
+
+yellow = '\033[33m'
+reset = '\033[0m'
+
+try:
```
EugeneTheDev: I don't think we should have such checks. We control the environment by using uv, so it should be fine. So please add Beautiful Soup dependency to `docs/pyproject.toml` and run `uv sync` from the `docs` directory to update `uv.lock`.

EugeneTheDev (changes requested):
Looks useful, thank you

docs/process-api-links.py
marinkovicmarko: Thanks for the comment! I removed the check, added `beautifulsoup4` to `pyproject.toml`, and ran the sync.

EugeneTheDev (approved):

PR #1467, opened 2026-02-09
ci: add explicit skip condition for tests job to report status for required checks

EugeneTheDev:
Since `tests` job in `checks.yaml` is set as required for all PRs targeting `develop`, an explicit `if` skip condition is needed here. Otherwise the job won't run when `compilation` is skipped, thus never reporting the status and leading to stuck PR status checks.


tiginamaria (approved):
Looks logical and good

PR #1474, opened 2026-02-10
feat(rag): add vector database abstraction with filter expression DSL and pgvector implementation

serge-p7v:
The purpose of this PR is to provide a convenient integration with vector databases. It will be used with upcoming feature that can make Retrieval-Augmented Generation easier to implement.

### DIFF
--- a/rag/vector-storage/src/commonMain/kotlin/ai/koog/rag/vector/database/FilterExpressionDsl.kt
+++ b/rag/vector-storage/src/commonMain/kotlin/ai/koog/rag/vector/database/FilterExpressionDsl.kt
@@ -1,0 +1,749 @@
+package ai.koog.rag.vector.database
+
+import kotlinx.serialization.KSerializer
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.descriptors.PrimitiveKind
+import kotlinx.serialization.descriptors.PrimitiveSerialDescriptor
+import kotlinx.serialization.descriptors.SerialDescriptor
+import kotlinx.serialization.encoding.Decoder
+import kotlinx.serialization.encoding.Encoder
+import kotlinx.serialization.json.*
+import kotlin.jvm.JvmInline
+
+/**
+ * Kotlin DSL for building filter expressions for vector databases.
+ *
+ * Example usage:
+ * ```kotlin
+ * val filter = filterExpression {
+ *     "category" eq "books"
+ *     and {
+ *         "price" lt 100
+ *         "inStock" eq true
+ *     }
+ * }
+ *
+ * // Or using infix functions:
+ * val filter2 = filterExpression {
+ *     ("category" eq "books") and ("price" lt 100)
+ * }
+ *
+ * // Complex expressions:
+ * val filter3 = filterExpression {
+ *     (("category" eq "books") or ("category" eq "electronics")) and ("price" lte 50)
+ * }
+ * ```
+ */
+
+// ==================== Core Types ====================
+
+/**
+ * Custom serializer for Any? types used in filter expressions.
+ * Supports: String, Number (Int, Long, Double), Boolean, null, List, Key, Value, Expression
+ */
+public object AnySerializer : KSerializer<Any?> {
+    override val descriptor: SerialDescriptor = PrimitiveSerialDescriptor("Any", PrimitiveKind.STRING)
+
+    override fun serialize(encoder: Encoder, value: Any?) {
+        val jsonEncoder = encoder as? JsonEncoder
+            ?: throw IllegalStateException("AnySerializer only works with JSON format")
+        val jsonElement = toJsonElement(value)
+        jsonEncoder.encodeJsonElement(jsonElement)
+    }
+
+    override fun deserialize(decoder: Decoder): Any? {
+        val jsonDecoder = decoder as? JsonDecoder
+            ?: throw IllegalStateException("AnySerializer only works with JSON format")
+        val element = jsonDecoder.decodeJsonElement()
+        return fromJsonElement(element)
+    }
+
+    private fun toJsonElement(value: Any?): JsonElement = when (value) {
+        null -> JsonNull
+        is String -> JsonPrimitive(value)
+        is Number -> JsonPrimitive(value)
+        is Boolean -> JsonPrimitive(value)
+        is Key -> buildJsonObject {
+            put("__type", "Key")
+            put("name", value.name)
+        }
+
+        is Value -> buildJsonObject {
+            put("__type", "Value")
+            put("value", toJsonElement(value.value))
+        }
+
+        is Expression -> buildJsonObject {
+            put("__type", "Expression")
+            put("type", value.type.name)
+            put("left", toJsonElement(value.left))
+            put("right", toJsonElement(value.right))
+        }
+
+        is List<*> -> JsonArray(value.map { toJsonElement(it) })
+        else -> JsonPrimitive(value.toString())
+    }
+
+    private fun fromJsonElement(element: JsonElement): Any? = when (element) {
+        is JsonNull -> null
+        is JsonPrimitive -> when {
+            element.isString -> element.content
+            element.booleanOrNull != null -> element.boolean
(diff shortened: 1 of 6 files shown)


serge-p7v:
The upcoming "LongTermMemory" feature is in the other PR - https://github.com/JetBrains/koog/pull/1490
The `VectorDatabase` from this PR can be used as an underlying storage for `RetrievalStorage` from the other PR.

rag/vector-storage/src/commonMain/kotlin/ai/koog/rag/vector/database/FilterExpressionDsl.kt
```diff
@@ -0,0 +1,749 @@
...
+ * ```
+ */
+
+// ==================== Core Types ====================
+
+/**
+ * Custom serializer for Any? types used in filter expressions.
+ * Supports: String, Number (Int, Long, Double), Boolean, null, List, Key, Value, Expression
+ */
+public object AnySerializer : KSerializer<Any?> {
```
EugeneTheDev, line 44: Let's avoid `Any?` serializers. I think it introduces unnecessary complexity and is generally less robust. If you need to support a set of serializable values, it might be better to do it like this:

```kotlin
@Serializable
sealed interface Value<T> {
    val value: T
    val type: String
    
    @Serializable
    data class String(override val value: String) : Value<String> {
        override val type = "string"
    }
    
    // other serializable values ...
}
```

This is a more standard approach, and it gives you more type-safety, and also no custom serializers required for simple cases like that

rag/vector-storage/src/commonMain/kotlin/ai/koog/rag/vector/database/VectorDatabase.kt
```diff
@@ -0,0 +1,298 @@
...
+
+    /**
+     * The main textual content to be embedded and searched
+     */
+    public val content: String
+
+    /**
+     * Flexible key-value metadata for filtering and custom fields
+     */
+    public val metadata: Map<String, JsonPrimitive>
```
EugeneTheDev, line 29: Can be `JsonObject`

rag/vector-storage/src/jvmTest/kotlin/ai/koog/rag/vector/database/PGVectorDatabase.kt
```diff
@@ -0,0 +1,614 @@
...
+ * @param username The database username
+ * @param password The database password
+ * @param tableName Name of the default table to store records (default: "memory_records").
+ *                  Can be overridden per operation using the `namespace` parameter.
+ * @param vectorDimension The dimension of embedding vectors (default: 1536 for OpenAI embeddings)
+ * @param embedder Optional embedder for generating embeddings when records don't have them.
+ *                 If null and a record without embedding is added, an exception will be thrown.
+ * @param json JSON serializer for metadata
+ */
+public class PGVectorDatabase(
```
EugeneTheDev, line 44: Why is it in `jvmTest`, shouldn't it be in `jvmMain`? 

Also, manually passing url, username, password etc. leads to more parameters in constructor and is not flexible enough (what if I want to specify custom connection properties?). I suggest using Exposed and accepting a `Database` and a `tableName`, take a look at how a similar postgres storage is implemented [here](https://github.com/JetBrains/koog/blob/daa0e78119298413933a6ec889fb61315c9b9e1a/agents/agents-features/agents-features-sql/src/jvmMain/kotlin/ai/koog/agents/features/sql/providers/PostgresPersistencyStorageProvider.kt#L16)

rag/vector-storage/src/jvmTest/kotlin/ai/koog/rag/vector/database/PGVectorDatabase.kt
```diff
@@ -0,0 +1,614 @@
...
+    private val jdbcUrl: String,
+    private val username: String,
+    private val password: String,
+    private val tableName: String = "memory_records",
+    private val vectorDimension: Int = 1536,
+    private val embedder: Embedder? = null,
+    private val json: Json = Json { ignoreUnknownKeys = true }
+) : VectorDatabase {
+
+    private fun getConnection(): Connection {
```
EugeneTheDev, line 54: This connection is opened but never closed, which might lead to memory leak. If you would follow my suggestion from the previous comment, then it won't be a problem, since you would do something like this:
```kotlin
newSuspendedTransaction(Dispatchers.IO, database) { block() }
```

rag/vector-storage/src/jvmTest/kotlin/ai/koog/rag/vector/database/PGVectorDatabase.kt
```diff
@@ -0,0 +1,614 @@
...
+     */
+    public suspend fun migrate(namespace: String) {
+        withContext(Dispatchers.IO) {
+            getConnection().use { conn ->
+                conn.createStatement().use { stmt ->
+                    stmt.executeUpdate("CREATE EXTENSION IF NOT EXISTS vector")
+
+                    stmt.executeUpdate(
+                        """
+                        CREATE TABLE IF NOT EXISTS $namespace (
```
EugeneTheDev, line 82: Maybe it's not that big of a deal in this particular case, but using string templates for constructing SQL statements might be vulnarable to SQLi, e.g., if an application passes `namespace` parameter from a user or from an LLM directly. I would suggest reviewing all such cases in this class and finding a safer approach, such as manual validation or built-in Exposed/database quoting support. If you're using Exposed, you can even switch to their ORM model for creating and working with tables, which should handle such cases automatically

EugeneTheDev (changes requested):

rag/vector-storage/src/jvmTest/kotlin/ai/koog/rag/vector/database/PGVectorDatabase.kt
serge-p7v: In the first (now closed) PR I added pgvector implementation with Exposed into jvmMain, but in the new PR I decided to add to the the jvmTest without Exposed because of the following:
* I didn't want to introduce an implementation that is not production ready and is not the most popular vector database (e.g. for RDBMS we can assume that most users have PostgreSQL or MySQL, but for vector databases we cannot assume that most users have pgvector). And current pgvector implementation can be useful for local testing and checking if `VectorDatabase` interface is adequate. 
* I didn't want to introduce new dependency (Exposed) to the RAG module, because it doesn't support vector databases such as Chroma, Pinecone, Weaviate or Milvus.

rag/vector-storage/src/commonMain/kotlin/ai/koog/rag/vector/database/VectorDatabase.kt
```diff
@@ -0,0 +1,298 @@
...
+
+/**
+ * Base sealed interface for search requests.
+ * Using a sealed interface eliminates invalid states and allows search-type-specific parameters.
+ *
+ * @property limit Maximum number of results to return (topK)
+ * @property similarityThreshold Minimum similarity score for results (0.0 to 1.0)
+ * @property filterExpression Metadata filter expression for pre-filtering (built using [filterExpression] DSL)
+ */
+public sealed interface SearchRequest {
```
serge-p7v, line 66: should not be sealed

rag/vector-storage/src/commonMain/kotlin/ai/koog/rag/vector/database/VectorDatabase.kt
```diff
@@ -0,0 +1,298 @@
...
+import kotlinx.serialization.json.JsonPrimitive
+
+/**
+ * Represents a record that can be stored in a vector database.
+ *
+ * @property id Unique identifier for the record (optional for new records, required for updates)
+ * @property content The main textual content to be embedded and searched
+ * @property metadata Flexible key-value metadata for filtering and custom fields
+ */
+public sealed interface Record {
```
serge-p7v, line 15: sealed or not?

serge-p7v:
Closing in favor of https://github.com/JetBrains/koog/pull/1563

PR #1475, opened 2026-02-10
refactor(koog-agents)!: update Kotlin to 2.3.10 and move Clock/Instant to kotlin.time  

devcrocod:
- Kotlin: 2.2.21 → 2.3.10
- kotlinx-serialization: 1.8.1 → 1.10.0
- kotlinx-datetime: 0.6.2 → 0.7.1
- Replace `kotlinx.datetime.Clock` -> `kotlin.time.Clock`, `kotlinx.datetime.Instant` -> `kotlin.time.Instant`

closes [KG-544](https://youtrack.jetbrains.com/issue/KG-544) Upgrade the dependency version of kotlinx-datetime from 0.6.2 to 0.7.1


kpavlov (commented):
Thank you, @devcrocod
Looks good overall

Copilot has found some leftovers, was it intentional?

devcrocod:
@kpavlov 

> Copilot has found some leftovers, was it intentional?

Yeap, I kept `Clock` and `Instant` from `kotlinx.datetime` in public api. Starting with version 0.7.0, kotlinx.datetime introduced typealiases for these types. Since kotlinx.datetime is dependent transitively, this should make the migration to kotlin.time easier for users, though it won’t eliminate the issue

In a minor release, we can switch the public api over to kotlin.time as well, and in some places even remove the kotlinx.datetime dependency entirely

kpavlov (approved):

PR #1479, opened 2026-02-11
feat(prompt)!: Support streaming reasoning messages

### LINKED ISSUE #1264: OpenAI Responses Reasoning Summary
When using the OpenAI Responses API with reasoning models (e.g. gpt-5 series), the [OpenAILLMClient](https://github.com/JetBrains/koog/blob/develop/prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openai/OpenAILLMClient.kt) does not stream reasoning summary text deltas to the caller. Additionally, encrypted reasoning content may not be properly accumulated when using stateless mode (`store=false`), preventing reasoning context from being passed to subsequent turns.

## **Affected Component**

`ai.koog.prompt.executor.clients.openai.OpenAILLMClient.executeResponsesStreaming()`

## **Current Behaviour**

Unless I am missing something, the `executeResponsesStreaming` method only handles three event types:

```kotlin
processStreamingChunk = {
    when (it) {
        is OpenAIStreamEvent.ResponseOutputItemDone -> { /* tool calls */ }
        is OpenAIStreamEvent.ResponseCompleted -> { /* end frame */ }
        is OpenAIStreamEvent.ResponseOutputTextDelta -> { /* text output */ }
        else -> null  // All other events discarded
    }
}
```

This means:

1. **`ResponseReasoningSummaryTextDelta` events are ignored** → Reasoning summaries (e.g., `"**Investigating error source**"`) are never streamed incrementally, only available in the final `ResponseCompleted` event.

2. **`ResponseOutputItemDone` for reasoning items is not captured** → When `item.type == "reasoning"`, the encrypted reasoning content (`item.encryptedContent`) is not extracted and included in `StreamFrame.End.messages`.

3. **`StreamFrame.End` is emitted without reasoning messages** → The `messages` field is always `null`, even though reasoning items should be included for context passing in subsequent turns (at least with Responses API in stateless mode).

### **1. Stream Reasoning Summaries Incrementally**

When `ReasoningConfig.summary` is set to `DETAILED` or `AUTO`, the API emits:

```json
{"type":"response.reasoning_summary_text.delta","delta":"**Investigating","...}
{"type":"response.reasoning_summary_text.delta","delta":" error","...}
{"type":"response.reasoning_summary_text.delta","delta":" source**","...}
```

These should be emitted as `StreamFrame.Append` or even better a different StreamFrame type (e.g. StreamFrame.Reasoning), so UIs can display reasoning progress in real-time.

### **2. Accumulate Encrypted Reasoning for Context Passing**

When using `store=false` (stateless mode) and `include=["reasoning.encrypted_content"]`, the encrypted reasoning must be captured from `response.output_item.done` events and passed back to subsequent API calls via `Item.Reasoning` in the input.

The client should populate `StreamFrame.End.messages` with:

```kotlin
StreamFrame.End(
    finishReason = null,
    metaInfo = metaInfo,
    messages = listOf(
        Message.Reasoning(
            id = "rs_abc123",
            encrypted = "gAAAAABpO1po77ABItkkgBK...",  // From item.encryptedContent
            content = "**Investigating error source**",  // From item.summary
            metaInfo = metaInfo
        )
    )
)
```

This allows frameworks like the Agents API to automatically include reasoning in the conversation history.

## **Proposed Fix**

In its simplest form, in `OpenAILLMClient.executeResponsesStreaming()`:

```kotlin
private fun executeResponsesStreaming(
    prompt: Prompt,
    model: LLModel,
    tools: List<ToolDescriptor>,
    params: OpenAIResponsesParams
): Flow<StreamFrame> {
    val reasoningBuffer = mutableListOf<Message.Reasoning>()
    
    return httpClient.sse(
        path = settings.responsesAPIPath,
        request = request,
        processStreamingChunk = {
            when (it) {
                // NEW: Stream reasoning summary incremental deltas
                is OpenAIStreamEvent.ResponseReasoningSummaryTextDelta -> {
                    StreamFrame.Append(it.delta)
                }
                
                is OpenAIStreamEvent.ResponseOutputItemDone -> {
                    when (val item = it.item) {
                        is Item.FunctionToolCall -> {
                            StreamFrame.ToolCall(item.callId, item.name, item.arguments)
                        }
                        // NEW: Accumulate reasoning for context passing
                        is Item.Reasoning -> {
                            reasoningBuffer.add(Message.Reasoning(
                                id = item.id,
                                encrypted = item.encryptedContent,
                                content = item.summary.joinToString("\n") { it.text },
                                metaInfo = ResponseMetaInfo.Empty
                            ))
                            null
                        }
                        else -> null
                    }
                }

                is OpenAIStreamEvent.ResponseCompleted -> {
                    StreamFrame.End(
                        finishReason = null,
                        metaInfo = createMetaInfo(it.response.usage),
                        messages = reasoningBuffer.takeIf { it.isNotEmpty() }  // NEW
                    )
                }

                is OpenAIStreamEvent.ResponseOutputTextDelta -> {
                    StreamFrame.Append(it.delta)
                }

                else -> null
            }
        }
    ).filterNotNull()
}
```

Although I would actually propose to create a `StreamFrame.Reasoning` which could be used by all reasoning models to stream reasoning separately from actual final output, or at the bare minimum have some delimiting marker. Otherwise it would not be possible to distinguish between the two in the UI.

## **Workaround**

I have implement a custom `OpenAILLMClient` subclass that manually handles `ResponseReasoningSummaryTextDelta` and `ResponseOutputItemDone` events for reasoning items that wraps and unwraps reasoning frames through the Append frame to get around this.

## **Additional Context**

- The Responses API models already exist in `ai.koog.prompt.executor.clients.openai.models` (e.g., `OpenAIStreamEvent.ResponseReasoningSummaryTextDelta`)
- The serialisation is working correctly
- Looks like only the streaming event handler needs updating

## **References**

- [OpenAI Responses API Documentation](https://platform.openai.com/docs/api-reference/responses/streaming)
- Event types: `response.reasoning_summary_text.delta`, `response.output_item.done` (when `item.type == "reasoning"`)

tiginamaria:
Reasoning messages have been previously ignored while streaming. This PR introduces the way to have them as separate objects of `StreamFrame` (including the summary and encrypted version)
`StreamFrame` now has 3 types of events: `DeltaFrame` (for partial messages) and `CompleteFrame`, (for complete messages, merged marts for partial, arrives after the message was completed), and `End` (after all messages are delivered)

BREAKING:
* The whole streaming API, including `StreamFrame` contains breaking changes

closes: #1264 KG-592

### DIFF
--- a/prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/message/Message.kt
+++ b/prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/message/Message.kt
@@ -168,6 +168,8 @@ public sealed interface Message {
      *
      * @property id An optional identifier for the reasoning process.
      * @property encrypted The encrypted content of the reasoning message.
+     * @property parts The parts of the reasoning message. Only the [ContentPart.Text] part is allowed.
+     * @property summary An optional summary of the reasoning process.
      * @property content The content of the message as a string.
      * @property role The [Role] of the message, indicating its source or function in the chat (e.g., assistant, user).
      *                Defaults to [Role.Assistant].
@@ -179,14 +181,15 @@ public sealed interface Message {
         public val id: String? = null,
         public val encrypted: String? = null,
         override val parts: List<ContentPart.Text>,
+        public val summary: List<ContentPart.Text>? = null,
         override val metaInfo: ResponseMetaInfo
     ) : Response {
 
         /**
          * Single content part reasoning message constructor
          */
-        public constructor(id: String? = null, encrypted: String? = null, content: String, metaInfo: ResponseMetaInfo) :
-            this(id, encrypted, listOf(ContentPart.Text(content)), metaInfo)
+        public constructor(id: String? = null, encrypted: String? = null, summary: String? = null, content: String, metaInfo: ResponseMetaInfo) :
+            this(id, encrypted, listOf(ContentPart.Text(content)), summary?.let { listOf(ContentPart.Text(it)) }, metaInfo)
 
         override val role: Role = Role.Reasoning
 
--- a/prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/message/Message.kt
+++ b/prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/message/Message.kt
@@ -169,7 +169,7 @@ public sealed interface Message {
      * @property id An optional identifier for the reasoning process.
      * @property encrypted The encrypted content of the reasoning message.
      * @property parts The parts of the reasoning message. Only the [ContentPart.Text] part is allowed.
-     * @property summary An optional summary of the reasoning process.
+     * @property summary An optional summary of the reasoning process. Only the [ContentPart.Text] part is allowed.
      * @property content The content of the message as a string.
      * @property role The [Role] of the message, indicating its source or function in the chat (e.g., assistant, user).
      *                Defaults to [Role.Assistant].
@@ -188,8 +188,19 @@ public sealed interface Message {
         /**
          * Single content part reasoning message constructor
          */
-        public constructor(id: String? = null, encrypted: String? = null, summary: String? = null, content: String, metaInfo: ResponseMetaInfo) :
-            this(id, encrypted, listOf(ContentPart.Text(content)), summary?.let { listOf(ContentPart.Text(it)) }, metaInfo)
+        public constructor(
+            id: String? = null,
+            encrypted: String? = null,
+            summary: String? = null,
+            content: String,
+            metaInfo: ResponseMetaInfo
+        ) : this(
+            id = id,
+            encrypted = encrypted,
+            parts = listOf(ContentPart.Text(content)),
+            summary = summary?.let { listOf(ContentPart.Text(it)) },
+            metaInfo = metaInfo
+        )
 
         override val role: Role = Role.Reasoning
 
--- a/prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/streaming/StreamFrame.kt
+++ b/prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/streaming/StreamFrame.kt
@@ -13,14 +13,75 @@ import kotlinx.serialization.json.jsonObject
 public sealed interface StreamFrame {
 
     /**
-     * Represents a frame of a streaming response from a LLM that appends some text.
+     * The interface representing a delta or partial frame of a streaming response from a LLM.
+     */
+    public sealed interface CompleteFrame : StreamFrame {
+        public val index: Int?
+    }
+
+    /**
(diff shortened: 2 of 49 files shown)


prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/message/Message.kt
```diff
@@ -179,14 +181,15 @@ public sealed interface Message {
         public val id: String? = null,
         public val encrypted: String? = null,
         override val parts: List<ContentPart.Text>,
+        public val summary: List<ContentPart.Text>? = null,
```
antoniibelyshev, line 184: summary sounds like it should be a single string, not list. Are there reasons to keep it as a list?

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/streaming/StreamFrame.kt
```diff
@@ -13,14 +13,75 @@ import kotlinx.serialization.json.jsonObject
 public sealed interface StreamFrame {
 
     /**
-     * Represents a frame of a streaming response from a LLM that appends some text.
+     * The interface representing a delta or partial frame of a streaming response from a LLM.
```
antoniibelyshev: kdoc seems to be swapped for delta and complete frame 🙂

antoniibelyshev (changes requested):
Thank you! The new api with `StreamFrame` looks good.
AFAIU, for every message in the response, the flow will contain all its deltas, and then its complete version (at least, that's what I saw when run executeStreaming with an openai client). So, every message basically appears in the flow twice, which is confusing. Can we stream only deltas? We can also have a helper method to make the flow more granular, e.g. `Flow<StreamFrame>.messages(): List<Message.Response>`

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/message/Message.kt
tiginamaria: The OpenAI format is a list, and the user in the issue asked to keep it as is. I think the idea is that you could have several reasoning and summaries for each

tiginamaria:
The Complete frames were added to indicate that a message is complete (otherwise, if you stream [Tool, Tool], the only way to know there are two tools in the stream is the changed `index`, so you need to manually check that the index has changed). And previously, only complete tool calls were inserted into the flow, which I think is unfair and changes the way the clients stream them

We can stream only deltas, though, but I really like the `Anthropic` and `OpenAI` approach here with delta and complete events.

antoniibelyshev (approved):
Thank you! I agree about the complete frames, it's nice to follow the pattern of OpenAI and Anthropic. The new streaming api looks great!

PR #1480, opened 2026-02-11
feat(agents)!: introduce AIAgentFunctionalContextBase, inherit functional and planner context from it

antoniibelyshev:
Introduced `AIAgentFunctionalContextBase`, moved to it the functionality from the `AIAgentFunctionalContext`.
Inherited `AIAgentFunctionalContext` from the base functional context.
Introduced `AIAgentPlannerContext`, inherited it from the base functional context.
Replaced functional context with planner context for planner agent and related components.

BREAKING:
Planner agent and its components now use the `AIAgentPlannerContext`, not the `AIAgentFunctionalContext`.

note: that's a preparation step for [KG-672](https://youtrack.jetbrains.com/issue/KG-672)


tiginamaria (approved):
Changes look good, thank you! More generally, I don't really get why Planning is a separate agent (which caused these changes) and not just a set of additional functions for functional strategy. But probably I don't know the details, so maybe it's just some thoughts for you to consider or not

Rizzen (approved):
Context changes looks good

antoniibelyshev:
@tiginamaria In simple words, a planning strategy is a loop where every iteration is dedicated to a specific step in a plan. So, we need a planner pipeline (and thus a planning context) with additional handlers, e.g. `onStepExecutionStarted` (I will add the handlers in a separate pr)

PR #1485, opened 2026-02-12
build: publish XCFramework binary for SPM

Amaneusz:
# Add Swift Package Manager (SPM) Support via XCFramework

## Summary
Adds infrastructure to build and distribute Koog as an XCFramework for iOS/macOS development via Swift Package Manager.

## Changes
1. **XCFramework Build Configuration** (`.github/workflows/build-xcframework.yml`)
   - New CI workflow that builds XCFramework on releases and manual trigger
   - Runs on macOS with Xcode support
   - Creates versioned zip archives with SHA-256 checksums
   - Uploads artifacts to GitHub releases

2. **iOS Target Support** (`ai.kotlin.multiplatform.gradle.kts`)
   - Configure iOS targets (arm64, simulator arm64, x64) to produce Framework binaries
   - Set up XCFramework assembly from individual frameworks
   - Use 'Koog' as framework name (follows Swift PascalCase convention)

3. **Module Export** (`koog-agents/build.gradle.kts`)
   - Export all included modules through Framework binaries
   - Makes Koog APIs accessible to Swift consumers

## Architecture
- **Single XCFramework**: `Koog.xcframework` contains all Koog modules
- **Distribution**: GitHub Releases (binary artifacts)
- **Consumption**: Separate SPM repository will reference these artifacts via Package.swift

## Testing
- [x] Manual: Verified XCFramework builds locally
- [x] Manual: Framework structure validated
- [x] Manual: Tested import in sample Xcode project 
- [ ] CI: Automated testing (future work)

## Next Steps (Follow-up Work)
1. Create separate SPM repository with Package.swift
2. Add modular framework distribution (separate frameworks per module)
3. Add automated Swift compilation tests

## Related Issues
Closes KG-682

image

### DIFF
--- a/.github/workflows/build-xcframework.yml
+++ b/.github/workflows/build-xcframework.yml
@@ -5,6 +5,11 @@ on:
     types: [published]
   workflow_dispatch: # Manual trigger
 
+env:
+  JAVA_VERSION: 17
+  JAVA_DISTRIBUTION: 'corretto'
+  JAVA_OPTS: "-Xms8g -Xmx8g -Dfile.encoding=UTF-8 -Djava.awt.headless=true -Dkotlin.daemon.jvm.options=-Xmx6g"
+
 jobs:
   build-xcframework:
     name: Build iOS XCFramework
@@ -31,14 +36,8 @@ jobs:
           echo "version=$VERSION" >> $GITHUB_OUTPUT
           echo "Using version: $VERSION"
 
-      - name: Set up JDK 17
-        uses: actions/setup-java@v5
-        with:
-          java-version: 17
-          distribution: corretto
-
       - name: Setup Gradle
-        uses: gradle/actions/setup-gradle@v4
+        uses: gradle/actions/setup-gradle@v5
         with:
           cache-read-only: false
 
@@ -46,8 +45,6 @@ jobs:
         run: xcodebuild -version
 
       - name: Build XCFramework
-        env:
-          JAVA_OPTS: "-Xms8g -Xmx8g -Dfile.encoding=UTF-8 -Djava.awt.headless=true -Dkotlin.daemon.jvm.options=-Xmx6g"
         run: |
           echo "Assembling XCFramework binary..."
           ./gradlew :koog-agents:assembleKoogXCFramework -Pkoog.build.xcframework=true --no-daemon --stacktrace
--- a/.github/workflows/build-xcframework.yml
+++ b/.github/workflows/build-xcframework.yml
@@ -1,0 +1,89 @@
+name: Build XCFramework
+
+on:
+  release:
+    types: [published]
+  workflow_dispatch: # Manual trigger
+  push:
+    branches:
+      - amanowicz/KG-682_spm_support
+
+jobs:
+  build-xcframework:
+    name: Build iOS XCFramework
+    runs-on: macos-latest  # Required for Xcode
+    timeout-minutes: 60
+    permissions:
+      contents: write  # Required for uploading release assets
+
+    steps:
+      - name: Configure Git
+        run: |
+          git config --global core.autocrlf input
+
+      - uses: actions/checkout@v5
+
+      - name: Extract version
+        id: version
+        run: |
+          if [ "${{ github.event_name }}" == "release" ]; then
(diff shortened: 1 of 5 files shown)


prompt/prompt-structure/src/commonMain/kotlin/ai/koog/prompt/structure/json/generator/JsonSchemaConsts.kt
```diff
@@ -46,6 +49,9 @@ public object JsonSchemaConsts {
         public const val BOOLEAN: String = "boolean"
         public const val ARRAY: String = "array"
         public const val OBJECT: String = "object"
+
+        @OptIn(ExperimentalObjCName::class)
+        @ObjCName("JSON_NULL")
         public const val NULL: String = "null"
```
Amaneusz, line 55: This change was necessary as `NULL` is reserved keyword in Objective C.

I've stumbled upon this when I was testing in Xcode whether the imported Koog module is working correctly - it was failing during the build because of this conflict.

kpavlov (approved):
Thank you, @Amaneusz
LGTM

Amaneusz:
Converting to draft: tests on macos have significantly slowed down due to changes in gradle files. Will mark as ready when fixed

kpavlov (approved):

.github/workflows/build-xcframework.yml
```diff
@@ -0,0 +1,86 @@
...
+        run: |
+          if [ "${{ github.event_name }}" == "release" ]; then
+            VERSION="${{ github.event.release.tag_name }}"
+          else
+            VERSION="${{ github.sha }}"
+          fi
+          echo "version=$VERSION" >> $GITHUB_OUTPUT
+          echo "Using version: $VERSION"
+
+      - name: Set up JDK 17
```
EugeneTheDev: AFAIK it's not necessary to setup Java separately, the following step that sets up Gradle handles this already. To make sure the correct JDK version, distro and options are used, you can set workflow envs as we do in other workflows, [example from checks.yml](https://github.com/JetBrains/koog/blob/7ec7af4c4d124ac40d3e89634774a4ba102cff2b/.github/workflows/checks.yml#L21)

.github/workflows/build-xcframework.yml
```diff
@@ -0,0 +1,86 @@
...
+        uses: actions/setup-java@v5
+        with:
+          java-version: 17
+          distribution: corretto
+
+      - name: Setup Gradle
+        uses: gradle/actions/setup-gradle@v4
```
EugeneTheDev: Our other workflows use `v5`, is there any specific reason you use `v4`?

.github/workflows/build-xcframework.yml
```diff
@@ -0,0 +1,86 @@
...
+        with:
+          cache-read-only: false
+
+      - name: Verify Xcode installation
+        run: xcodebuild -version
+
+      - name: Build XCFramework
+        env:
+          JAVA_OPTS: "-Xms8g -Xmx8g -Dfile.encoding=UTF-8 -Djava.awt.headless=true -Dkotlin.daemon.jvm.options=-Xmx6g"
```
EugeneTheDev: Can be moved to global `env`, see the comment above

convention-plugin-ai/src/main/kotlin/ai/koog/gradle/xcframework/XCFrameworkConfig.kt
```diff
@@ -0,0 +1,72 @@
...
+    /**
+     * Configures XCFramework for iOS targets if enabled.
+     * Creates framework binaries for all targets defined for the project.
+     *
+     * @param project The Gradle project
+     */
+    fun KotlinMultiplatformExtension.configureXCFrameworkIfRequested(project: Project) {
+        if (isEnabled(project)){
+            val xcf = project.XCFramework("Koog")
+            targets.withType(KotlinNativeTarget::class.java)
```
EugeneTheDev: nit: can you use more Kotlin-like inline function here? `targets.withType<KotlinNativeTarget>()`?

EugeneTheDev (approved):
Generally looks good, a few minor comments

PR #1487, opened 2026-02-12
test: check multiple similar event handlers

aozherelyeva:
Added an integration test to check the changes in scope of [KG-678](https://youtrack.jetbrains.com/issue/KG-678) Allow to register multiple handlers for agent Feature

BREAKING:
* none

DEPRECATED:
* none

### DIFF
--- a/integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/agent/AIAgentIntegrationTest.kt
+++ b/integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/agent/AIAgentIntegrationTest.kt
@@ -1326,7 +1326,6 @@ class AIAgentIntegrationTest : AIAgentTestBase() {
                 systemPrompt = "$systemPrompt JUST CALL THE TOOLS, NO QUESTIONS ASKED.",
                 strategy = singleRunStrategy(ToolCalls.SEQUENTIAL),
                 llmModel = model,
-                temperature = 1.0,
                 toolRegistry = toolRegistry,
                 maxIterations = 10
             ) {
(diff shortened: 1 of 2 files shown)


integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/agent/AIAgentIntegrationTest.kt
```diff
@@ -1306,6 +1307,53 @@ class AIAgentIntegrationTest : AIAgentTestBase() {
         }
     }
 
+    @ParameterizedTest
+    @MethodSource("allModels")
+    fun integration_MultipleHandlersForToolCallEvents(model: LLModel) = runTest(timeout = 180.seconds) {
```
sdubov, line 1312: I see this timeout pattern in number of tests:
```
= runTest(timeout = 180.seconds) { ... }
```
Maybe, it worth extracting this 3 minutes timeout to some shared variable that we can reuse?

aozherelyeva:
UPD: decided to leave the checks

sdubov (approved):
@aozherelyeva , thank you for writing the test. My general point here - let's consider whether we need to have this as integration tests. This is quite an internal logic that does not need integration from my perspective. The point to verify here is the following: if we registered two handlers on the same agent event - both handlers are executed (while it is not critical how to we get those agent events from a real LLM or from a mocked executor).

@aozherelyeva , I would leave this up to you to consider at that point.

aozherelyeva:
> This is quite an internal logic that does not need integration from my perspective

I tend to agree here :)

PR #1488, opened 2026-02-12
docs(agents): typo

ununhexium:
Typo


Rizzen:
Thanks for the contribution! 🙌

Could you please update the PR title to follow our conventional commit/PR naming format? I believe the problem is in scope - it should be `agents` instead of `agent`

Once that’s updated, we’ll be good to go. Thanks!

Rizzen (approved):

PR #1451, opened 2026-02-03
[WIP] Make writeSession re-entrant

serge-p7v:
This PR adds optional re-entrancy for `writeSession` in `AIAgentLLMContextImpl`: callers can pass `sessionId` and if there is already an active write session with the same `sessionId` it will be reused.

## Motivation and Context
Seemingly safe call to `writeSession` in the PR https://github.com/JetBrains/koog/pull/1437 (see `Memory2`) might cause deadlock. The problem in that PR is that calling a `writeSession` from an active `writeSession` causes a deadlock. 

## Breaking Changes
None

---

#### Type of the changes
- [ ] New feature (non-breaking change which adds functionality)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Tests improvement
- [x] Refactoring
- [ ] CI/CD changes
- [ ] Dependencies update

#### Checklist
- [x] The pull request has a description of the proposed change
- [x] I read the [Contributing Guidelines](https://github.com/JetBrains/koog/blob/main/CONTRIBUTING.md) before opening the pull request
- [x] The pull request uses **`develop`** as the base branch
- [x] Tests for the changes have been added
- [ ] All new and existing tests passed

##### Additional steps for pull requests adding a new feature
- [ ] An issue describing the proposed change exists
- [ ] The pull request includes a link to the issue
- [ ] The change was discussed and approved in the issue
- [ ] Docs have been added / updated

PR #1490, opened 2026-02-12
feat(agents): add LongTermMemory feature for memory retrieval and ingestion

serge-p7v:
This PR introduces a new feature with pipeline interceptors that can:
* augment prompts before sending them to an LLM with memory records from a provided storage;
* extract memory records from messages and ingest them into the provided storage.

### DIFF
--- a/agents/agents-features/agents-features-longterm-memory/Module.md
+++ b/agents/agents-features/agents-features-longterm-memory/Module.md
@@ -1,0 +1,27 @@
+# Module agents-features-longterm-memory
+
+Provides the `LongTermMemory` feature for AI agents, enabling persistent storage and retrieval of memory records (documents) across agent runs via vector databases or other storage backends. Supports Retrieval-Augmented Generation (RAG) and message ingestion as two independently configurable flows.
+
+### Overview
+
+The agents-features-longterm-memory module adds long-term memory capabilities to Koog AI agents:
+
+- **Retrieval (RAG)**: Searches a memory store for context relevant to the user's query and augments the LLM prompt before each call
+- **Ingestion**: Extracts and persists conversation messages into a memory store for future retrieval
+- **Flexible storage**: Plug any backend via `RetrievalStorage` / `IngestionStorage` interfaces; an in-memory `InMemoryRecordStorage` is included for testing
+- **Configurable timing**: Ingest per-LLM-call or on agent completion
+- **Prompt augmentation modes**: System prompt or user prompt or custom implementation
+
+### Key Components
+
+| Component                                                                                                                    | Description                                                            |
+|------------------------------------------------------------------------------------------------------------------------------|------------------------------------------------------------------------|
+| [`LongTermMemory`](src/commonMain/kotlin/ai/koog/agents/longtermmemory/feature/LongTermMemory.kt)                            | Agent feature with DSL config for retrieval & ingestion                |
+| [`RetrievalStorage`](src/commonMain/kotlin/ai/koog/agents/longtermmemory/retrieval/RetrievalStorage.kt)                      | Interface for searching memory records                                 |
+| [`IngestionStorage`](src/commonMain/kotlin/ai/koog/agents/longtermmemory/ingestion/IngestionStorage.kt)                      | Interface for adding memory records                                    |
+| [`SearchStrategy`](src/commonMain/kotlin/ai/koog/agents/longtermmemory/retrieval/SearchStrategy.kt)              | Converts user query into a `SearchRequest` (similarity or keyword)     |
+| [`MemoryRecordExtractor`](src/commonMain/kotlin/ai/koog/agents/longtermmemory/ingestion/extraction/MemoryRecordExtractor.kt) | Transforms messages into `MemoryRecord`s for storage                   |
+| [`PromptAugmenter`](src/commonMain/kotlin/ai/koog/agents/longtermmemory/retrieval/augmentation/PromptAugmenter.kt)           | Interface for augmenting prompts with relevant context                 |
+| [`SystemPromptAugmenter`](src/commonMain/kotlin/ai/koog/agents/longtermmemory/retrieval/augmentation/SystemPromptAugmenter.kt)               | Inserts retrieved context as a system message                          |
+| [`UserPromptAugmenter`](src/commonMain/kotlin/ai/koog/agents/longtermmemory/retrieval/augmentation/UserPromptAugmenter.kt)                   | Inserts retrieved context as a user message                            |
+| [`InMemoryRecordStorage`](src/commonMain/kotlin/ai/koog/agents/longtermmemory/storage/InMemoryRecordStorage.kt)              | In-memory storage implementing both retrieval and ingestion interfaces |
--- a/agents/agents-features/agents-features-longterm-memory/src/commonMain/kotlin/ai/koog/agents/longtermmemory/feature/LongTermMemory.kt
+++ b/agents/agents-features/agents-features-longterm-memory/src/commonMain/kotlin/ai/koog/agents/longtermmemory/feature/LongTermMemory.kt
@@ -1,0 +1,538 @@
+package ai.koog.agents.longtermmemory.feature
+
+import ai.koog.agents.core.agent.context.AIAgentContext
+import ai.koog.agents.core.agent.context.featureOrThrow
+import ai.koog.agents.core.agent.entity.AIAgentStorageKey
+import ai.koog.agents.core.agent.entity.createStorageKey
+import ai.koog.agents.core.annotation.ExperimentalAgentsApi
+import ai.koog.agents.core.feature.AIAgentFunctionalFeature
+import ai.koog.agents.core.feature.AIAgentGraphFeature
+import ai.koog.agents.core.feature.AIAgentPlannerFeature
+import ai.koog.agents.core.feature.config.FeatureConfig
+import ai.koog.agents.core.feature.pipeline.AIAgentFunctionalPipeline
+import ai.koog.agents.core.feature.pipeline.AIAgentGraphPipeline
+import ai.koog.agents.core.feature.pipeline.AIAgentPipeline
+import ai.koog.agents.core.feature.pipeline.AIAgentPlannerPipeline
+import ai.koog.agents.longtermmemory.ingestion.IngestionSettings
+import ai.koog.agents.longtermmemory.ingestion.IngestionStorage
+import ai.koog.agents.longtermmemory.ingestion.IngestionTiming
+import ai.koog.agents.longtermmemory.ingestion.extraction.MemoryRecordExtractor
+import ai.koog.agents.longtermmemory.retrieval.RetrievalSettings
+import ai.koog.agents.longtermmemory.retrieval.RetrievalStorage
+import ai.koog.agents.longtermmemory.retrieval.SearchStrategy
+import ai.koog.agents.longtermmemory.retrieval.augmentation.PromptAugmenter
+import ai.koog.agents.longtermmemory.retrieval.augmentation.SystemPromptAugmenter
+import ai.koog.prompt.dsl.Prompt
+import ai.koog.prompt.message.Message
+import ai.koog.prompt.streaming.StreamFrame
+import ai.koog.prompt.streaming.toMessageResponses
+import io.github.oshai.kotlinlogging.KotlinLogging
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
+import kotlin.coroutines.cancellation.CancellationException
+
+/**
+ * Memory feature that incorporates persistent storage of memory records (documents) in vector databases.
+ *
+ * This feature provides two main capabilities that can be configured independently:
+ * 1. **Retrieval (RAG)**: Retrieves relevant context from the memory record repository using a
+ *    configured search request builder and inserts it into the prompt before sending to the LLM.
+ *    Configured via [retrievalSettings]. When null, no retrieval/augmentation is performed.
+ * 2. **Ingestion**: Saves messages to the memory record repository for future retrieval.
+ *    Configured via [ingestionSettings]. When null, no messages are persisted.
+ *
+ * @see RetrievalSettings
+ * @see IngestionSettings
+ */
+@ExperimentalAgentsApi
+public class LongTermMemory(
+    private val retrievalSettings: RetrievalSettings? = null,
+    private val ingestionSettings: IngestionSettings? = null,
+) {
+    /**
+     * Buffer to accumulate streaming frames by runId.
+     * Frames are accumulated during streaming and saved when streaming completes.
+     */
+    private val streamingFramesBuffer: MutableMap<String, MutableList<StreamFrame>> = mutableMapOf()
+    private val streamingFramesBufferMutex = Mutex()
+
+    /**
+     * Configuration for the LongTermMemory feature.
+     *
+     * This class allows configuring:
+     * - Retrieval settings for RAG (prompt augmentation with context from the repository)
+     * - Ingestion settings for persisting messages to the repository
+     *
+     * Both [retrievalSettings] and [ingestionSettings] are null by default, meaning neither
+     * retrieval nor ingestion is active unless explicitly configured.
+     */
+    public class Config : FeatureConfig() {
+        /**
+         * Settings for retrieval-augmented generation (RAG).
+         * When null (default), no context augmentation from the memory record repository is performed.
+         *
+         * Can be set directly or configured via the [retrieval] DSL block.
+         */
+        public var retrievalSettings: RetrievalSettings? = null
+
+        /**
+         * Settings for ingesting (persisting) messages into the memory record repository.
+         * When null (default), no messages are persisted.
+         *
+         * Can be set directly or configured via the [ingestion] DSL block.
+         */
+        public var ingestionSettings: IngestionSettings? = null
+
+        /**
+         * Configure retrieval (RAG) settings via DSL block.
+         *
+         * Example usage:
+         * ```kotlin
+         * retrieval {
+         *     searchStrategy = SimilaritySearchStrategy(topK = 5)
+         *     promptAugmenter = UserPromptAugmenter()
+         * }
+         * ```
+         */
+        public fun retrieval(block: RetrievalSettingsBuilder.() -> Unit) {
+            retrievalSettings = RetrievalSettingsBuilder().apply(block).build()
+        }
+
+        /**
+         * Set pre-built retrieval settings directly.
+         */
+        public fun retrieval(settings: RetrievalSettings) {
+            retrievalSettings = settings
+        }
+
+        /**
+         * Configure ingestion settings via DSL block.
+         *
+         * Example usage:
+         * ```kotlin
+         * ingestion {
+         *     extractor = FilteringMemoryRecordExtractor(setOf(Message.Role.User))
+         * }
+         * ```
+         */
+        public fun ingestion(block: IngestionSettingsBuilder.() -> Unit) {
+            ingestionSettings = IngestionSettingsBuilder().apply(block).build()
+        }
+
+        /**
+         * Set pre-built ingestion settings directly.
+         */
+        public fun ingestion(settings: IngestionSettings) {
+            ingestionSettings = settings
+        }
+    }
+
+    /**
+     * Builder for [RetrievalSettings] used in the [Config.retrieval] DSL block.
+     */
+    public class RetrievalSettingsBuilder {
+        /**
+         * The retrieval storage to search for relevant memory records.
+         * Must be set explicitly in the retrieval { } block.
+         */
+        public var storage: RetrievalStorage? = null
+
+        /**
+         * The search strategy that defines how to search the retrieval storage.
+         *
+         * @see SearchStrategy
+         */
+        public var searchStrategy: SearchStrategy? = null
+
+        /**
+         * The augmenter that defines how retrieved context is inserted into the prompt.
+         * Defaults to [SystemPromptAugmenter].
+         *
+         * @see SystemPromptAugmenter
+         * @see ai.koog.agents.longtermmemory.retrieval.augmentation.UserPromptAugmenter
+         */
+        public var promptAugmenter: PromptAugmenter = SystemPromptAugmenter()
+
+        /**
+         * Fluent setter for [storage].
+         */
+        public fun withStorage(storage: RetrievalStorage): RetrievalSettingsBuilder = apply { this.storage = storage }
+
+        /**
+         * Fluent setter for [searchStrategy].
+         */
+        public fun withSearchStrategy(searchStrategy: SearchStrategy): RetrievalSettingsBuilder =
+            apply { this.searchStrategy = searchStrategy }
+
+        /**
+         * Fluent setter for [promptAugmenter].
+         */
+        public fun withPromptAugmenter(augmenter: PromptAugmenter): RetrievalSettingsBuilder =
+            apply { this.promptAugmenter = augmenter }
+
+        /**
+         * RetrievalSettings builder.
+         */
+        public fun build(): RetrievalSettings {
+            val retrievalStorage = storage ?: error("storage must be set in retrieval { } block")
+            return RetrievalSettings(
+                retrievalStorage,
+                searchStrategy,
+                promptAugmenter
+            )
+        }
+    }
+
+    /**
+     * Builder for [IngestionSettings] used in the [Config.ingestion] DSL block.
+     */
+    public class IngestionSettingsBuilder {
+        /**
+         * The ingestion storage where memory records will be persisted.
+         * Must be set explicitly in the ingestion { } block.
+         */
+        public var storage: IngestionStorage? = null
+
+        /**
+         * The extractor that defines how to transform messages into memory records.
+         *
+         * Pre-built ingesters are available:
+         * - [ai.koog.agents.longtermmemory.ingestion.extraction.FilteringMemoryRecordExtractor] - Filters messages by role
+         *
+         * Example usage:
+         * ```kotlin
+         * // Use pre-built extractor with parameters
+         * extractor = FilteringMemoryRecordExtractor(
+         *     messageRolesToExtract = setOf(Message.Role.User)
+         * )
+         *
+         * // Or use lambda for custom logic
+         * extractor = MemoryRecordExtractor { messages ->
+         *     messages.map { MemoryRecord.Plain(content = it.content) }
+         * }
+         * ```
+         */
+        public var extractor: MemoryRecordExtractor? = null
+
+        /**
+         * When to mapMessages messages. Defaults to [IngestionTiming.ON_LLM_CALL].
+         */
+        public var timing: IngestionTiming = IngestionTiming.ON_LLM_CALL
+
+        /**
+         * Fluent setter for [storage].
+         */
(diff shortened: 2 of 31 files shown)


serge-p7v:
So, this PR is about the feature, the PR with a low-level abstraction for vector databases is there - https://github.com/JetBrains/koog/pull/1474
The idea is that `VectorDatabase` abstraction can be an underlying repository for `RetrievalStorage`/`IngestionStorage`.

agents/agents-features/agents-features-longterm-memory/Module.md
```diff
@@ -0,0 +1,184 @@
...
+                    extractor = FilteringMemoryRecordExtractor()
+                    timing = IngestionTiming.ON_LLM_CALL
+                }
+            }
+        }
+```
+It has two independent group of settings: one for retrieval (RAG) and another is for ingestion.
+Let's see the corresponding flows.
+
+### Flow 1: Retrieval & Prompt Augmentation
```
serge-p7v: I added this "diagrams" primarily for the code review (because the PR is quite large). In the meantime, I'll add more examples and texts.

Rizzen (approved):

agents/agents-features/agents-features-longterm-memory/src/jvmTest/kotlin/ai/koog/agents/longtermmemory/feature/LongTermMemoryRetrievalTest.kt
```diff
@@ -0,0 +1,508 @@
...
+        val agent = AIAgent(
+            promptExecutor = executor,
+            strategy = nonStreamingStrategy,
+            agentConfig = defaultAgentConfig,
+            toolRegistry = ToolRegistry.EMPTY
+        ) {
+            install(LongTermMemory.Feature) {
+                retrieval {
+                    this.storage = storage
+                    searchRequestMaker = SearchRequestMaker { query ->
```
serge-p7v: After a second thought: perhaps it would be better to use a builder instead of this SearchRequestMaker so it becomes:
```
retrieval {
                    this.storage = storage
                    keywordSearch { topK = 5 }
                }
```
or
```
retrieval {
                    this.storage = storage
                    customSearch { _ -> KeywordSearchRequest(query = "Kotlin") }
                }
```

agents/agents-features/agents-features-longterm-memory/src/commonMain/kotlin/ai/koog/agents/longtermmemory/feature/LongTermMemory.kt
```diff
@@ -0,0 +1,541 @@
...
+         */
+        public fun ingestion(block: IngestionSettingsBuilder.() -> Unit) {
+            ingestionSettings = IngestionSettingsBuilder().apply(block).build()
+        }
+    }
+
+    /**
+     * Builder for [RetrievalSettings] used in the [Config.retrieval] DSL block.
+     */
+    public class RetrievalSettingsBuilder {
```
EugeneTheDev, line 133: Do we want to have builder classes here? I'm not against the idea of builders, they look nice. I'm only concerned about a slight maintenance overhead, maybe plain old classes would be fine here. And also don't forget about the Java API, Kotlin DSLs won't work there, so you need to provide a fluent API version as well (the same builder methods that would return a builder instance). If you think it's worth keeping the builders, it's fine to keep them. I'm just voicing a minor concern

agents/agents-features/agents-features-longterm-memory/src/commonMain/kotlin/ai/koog/agents/longtermmemory/feature/LongTermMemory.kt
```diff
@@ -0,0 +1,541 @@
...
+         * Example usage:
+         * ```kotlin
+         * retrieval {
+         *     customSearch { query ->
+         *         SimilaritySearchRequest(query, limit = 10, similarityThreshold = 0.8)
+         *     }
+         * }
+         * ```
+         */
+        public fun customSearch(block: (query: String) -> ai.koog.agents.longtermmemory.retrieval.SearchRequest) {
```
EugeneTheDev: nit: proper import

agents/agents-features/agents-features-longterm-memory/src/commonMain/kotlin/ai/koog/agents/longtermmemory/feature/LongTermMemory.kt
```diff
@@ -0,0 +1,541 @@
...
+                    ltmFeature.streamingFramesBuffer.remove(ctx.runId)
+                }
+                if (!frames.isNullOrEmpty()) {
+                    val fullResponse = frames.joinToString("")
+                    ingestMessages(
+                        ingestion,
+                        listOf(
+                            Message.Assistant(
+                                content = fullResponse,
+                                ResponseMetaInfo.Empty // TODO: pass the correct timestamp
```
EugeneTheDev: `StreamFrame.End` contains meta info, let's capture it too in the streaming interceptor and provide the actual meta info here

agents/agents-features/agents-features-longterm-memory/src/commonMain/kotlin/ai/koog/agents/longtermmemory/feature/LongTermMemory.kt
```diff
@@ -0,0 +1,541 @@
...
+@ExperimentalAgentsApi
+public class LongTermMemory(
+    private val retrievalSettings: RetrievalSettings? = null,
+    private val ingestionSettings: IngestionSettings? = null,
+) {
+    /**
+     * Buffer to accumulate streaming frames by runId.
+     * Frames are accumulated during streaming and saved when streaming completes.
+     */
+    private val streamingFramesBuffer: MutableMap<String, MutableList<String>> = mutableMapOf()
```
EugeneTheDev: Related to the comment above, let's capture all `StreamFrame`s as they are, not just `String`. We can then decide when ingesting these frames, how exactly we would do this, such as keeping `StreamFrame.Append` and `StreamFrame.End.metaInfo` only

agents/agents-features/agents-features-longterm-memory/src/commonMain/kotlin/ai/koog/agents/longtermmemory/feature/LongTermMemory.kt
```diff
@@ -0,0 +1,541 @@
...
+        /**
+         * Install ingestion interceptors for [IngestionTiming.ON_AGENT_COMPLETION] mode.
+         * All messages are saved at once when the agent completes.
+         */
+        private fun installOnAgentCompletionIngestion(
+            ingestion: IngestionSettings,
+            pipeline: AIAgentPipeline,
+        ) {
+            pipeline.interceptAgentCompleted(this) { ctx ->
+                val messages = ctx.context.llm.prompt.messages // TODO: switch to readSession
```
EugeneTheDev: Is it because of a deadlock with LLM sessions? Has it not been fixed yet?

agents/agents-features/agents-features-longterm-memory/src/commonMain/kotlin/ai/koog/agents/longtermmemory/feature/LongTermMemory.kt
```diff
@@ -0,0 +1,541 @@
...
+            messages: List<Message>,
+        ) {
+            val records = ingestion.memoryRecordExtractor.extract(messages)
+            if (records.isEmpty()) {
+                return
+            }
+
+            try {
+                ingestion.storage.add(records, ingestion.namespace)
+            } catch (e: Exception) {
```
EugeneTheDev, line 441: Don't forget to rethrow `CancellationException` to preserve structured concurrency

agents/agents-features/agents-features-longterm-memory/src/commonMain/kotlin/ai/koog/agents/longtermmemory/feature/LongTermMemory.kt
```diff
@@ -0,0 +1,541 @@
...
+            prompt: Prompt,
+            retrieval: RetrievalSettings,
+        ): Prompt? {
+            // TODO: the input for retrieval (last user message or custom) must be configurable.
+            val lastUserMessage = prompt.messages.lastOrNull { it.role == Message.Role.User } ?: return null
+
+            val searchResults = try {
+                val request = retrieval.searchRequestMaker.make(lastUserMessage.content)
+                retrieval.storage.search(request, retrieval.namespace)
+            } catch (e: Exception) {
```
EugeneTheDev, line 462: The same, rethrow `CancellationException`

agents/agents-features/agents-features-longterm-memory/src/commonMain/kotlin/ai/koog/agents/longtermmemory/model/MemoryRecord.kt
```diff
@@ -0,0 +1,29 @@
...
+
+    /**
+     * The main textual content to be embedded and searched
+     */
+    val content: String,
+
+    /**
+     * Flexible key-value metadata for filtering and custom fields
+     */
+    val metadata: Map<String, JsonPrimitive> = emptyMap()
```
EugeneTheDev: Why not just make it a `JsonObject`, it would be more flexible if we decide to add more complex structures in the future

agents/agents-features/agents-features-longterm-memory/src/commonMain/kotlin/ai/koog/agents/longtermmemory/prompt/PromptAugmenter.kt
```diff
@@ -0,0 +1,166 @@
...
+public object PromptAugmenter {
+
+    /**
+     * Default template for the system message when using [PromptAugmentationMode.SYSTEM_MESSAGE].
+     * Use {relevant_context} placeholder.
+     */
+    public const val DEFAULT_SYSTEM_PROMPT_TEMPLATE: String =
+        """Use the following information to answer the user's question.
+
+{relevant_context}
```
EugeneTheDev: nit: formatting, can you please adjust the indentation here and use `.trimIndent()`?

agents/agents-features/agents-features-longterm-memory/src/commonMain/kotlin/ai/koog/agents/longtermmemory/prompt/PromptAugmenter.kt
```diff
@@ -0,0 +1,166 @@
...
+     */
+    public fun augmentPrompt(
+        originalPrompt: Prompt,
+        relevantContext: List<String>,
+        promptAugmentationMode: PromptAugmentationMode,
+        systemPromptTemplate: String = DEFAULT_SYSTEM_PROMPT_TEMPLATE,
+        userPromptTemplate: String = DEFAULT_USER_PROMPT_TEMPLATE
+    ): Prompt {
+        val relevantContextText = if (relevantContext.isNotEmpty()) {
+            "Relevant information:\n" + relevantContext.mapIndexed { index, content ->
```
EugeneTheDev: Can we maybe make this part of the prompt configurable too? I think it's a good practice to avoid hardcoding the prompts as much as possible, to provide the necessary flexibility to the users

agents/agents-features/agents-features-longterm-memory/src/commonMain/kotlin/ai/koog/agents/longtermmemory/prompt/PromptAugmenter.kt
```diff
@@ -0,0 +1,166 @@
...
+    ): Prompt {
+        val contextMessage = formatTemplate(systemPromptTemplate, relevantContextText)
+        if (contextMessage.isBlank()) return originalPrompt
+
+        return originalPrompt.withMessages { messages ->
+            val systemMessage = Message.System(contextMessage, RequestMetaInfo.Empty)
+            // Insert system message at the beginning, or replace existing system message
+            val existingSystemIndex = messages.indexOfFirst { it is Message.System }
+            if (existingSystemIndex >= 0) {
+                // Combine with existing system message
```
EugeneTheDev: Is it necessary to concatenate system messages into a single system message? AFAIK we can have multiple system messages, and it should look even cleaner this way, IMHO

agents/agents-features/agents-features-longterm-memory/src/commonMain/kotlin/ai/koog/agents/longtermmemory/prompt/PromptAugmenter.kt
```diff
@@ -0,0 +1,166 @@
...
+ * with context retrieved from vector databases.
+ *
+ * @see PromptAugmentationMode
+ * @see ai.koog.agents.longtermmemory.feature.LongTermMemory
+ */
+public object PromptAugmenter {
+
+    /**
+     * Default template for the system message when using [PromptAugmentationMode.SYSTEM_MESSAGE].
+     * Use {relevant_context} placeholder.
```
EugeneTheDev: To make it a bit cleaner and robust, you can extract the placeholder value into a public constant too

agents/agents-features/agents-features-longterm-memory/src/commonMain/kotlin/ai/koog/agents/longtermmemory/storage/EphemeralMemoryStorage.kt
```diff
@@ -0,0 +1,104 @@
...
+ *
+ * ## Limitations:
+ * - Data is not persisted and will be lost on application restart
+ * - Supports only keyword search using simple substring matching
+ * - Filter expressions are not supported
+ *
+ * @param defaultNamespace The default namespace to use when none is specified in method calls.
+ *                         Defaults to "default".
+ */
+public open class EphemeralMemoryStorage(
```
EugeneTheDev: `InMemoryMemoryStorage` might sounds more clear (although two `Memory` in a row looks a bit weird, that's true). IMHO,  it's not claer enough what `Ephemeral` means without reading the KDoc

EugeneTheDev (changes requested):
Looks very clean, I like it. I left a number of comments/questions, please take a look. Also, let's not make `Module.md` that detailed and remove `README.md`. The information here and the code snippets might drift from the reality and they're not checked. This infromation belongs to `docs`, since we can verify code snippets here and it's easier to notice issues with outdated infromation

EugeneTheDev:
And a few more thoughts. Would it be possible to configure the storage for retrieval/ingestion, but disable the automatic behavior and allow manual memory management? E.g. if I don't want to access the memory on each user message since it's not needed for my use-case, but I want to code memory management in specific nodes manually. Something like this, similar to other existing features such as [A2A server](https://github.com/JetBrains/koog/blob/188f0f235375bf7f027629a34ef5f97ad946c4c0/agents/agents-features/agents-features-a2a-server/src/commonMain/kotlin/ai/koog/agents/a2a/server/feature/A2AAgentServer.kt#L112)

```kt
val checkMemory by node<String, String> { input ->
    withLongTermMemory {
       this.retrievalStorage.search(...)
       
       this.ingestionStorage.add(...)
    }

    input
}
```

IMHO keeping automatic behavior disabled by default should be more intuitive

agents/agents-features/agents-features-longterm-memory/src/commonMain/kotlin/ai/koog/agents/longtermmemory/storage/EphemeralMemoryStorage.kt
serge-p7v: I had the same concerns. There are also other options like "transient" or "temporary".

serge-p7v: Or `NonPersistentMemoryStorage`

agents/agents-features/agents-features-longterm-memory/src/commonMain/kotlin/ai/koog/agents/longtermmemory/feature/LongTermMemory.kt
serge-p7v: Thank you! I've added fluent setters for Java.

agents/agents-features/agents-features-longterm-memory/src/commonMain/kotlin/ai/koog/agents/longtermmemory/feature/LongTermMemory.kt
serge-p7v: Fixed

agents/agents-features/agents-features-longterm-memory/src/commonMain/kotlin/ai/koog/agents/longtermmemory/feature/LongTermMemory.kt
serge-p7v: Thanks! I used `toMessageResponses` for concatenation and metaInfo passing. `FilteringMemoryRecordExtractor` now uses timestamps.

agents/agents-features/agents-features-longterm-memory/src/commonMain/kotlin/ai/koog/agents/longtermmemory/feature/LongTermMemory.kt
serge-p7v: Done!

agents/agents-features/agents-features-longterm-memory/src/commonMain/kotlin/ai/koog/agents/longtermmemory/feature/LongTermMemory.kt
serge-p7v: Actually that code with TODO were misleading: it works with `readSession` too (I've switched to `readSession` and removed the TODO).
But it will deadlock with `writeSession` in the other place with prompt augmentation. No, it hasn't been fixed yet, I'm still working on it in https://github.com/JetBrains/koog/pull/1451

agents/agents-features/agents-features-longterm-memory/src/commonMain/kotlin/ai/koog/agents/longtermmemory/feature/LongTermMemory.kt
serge-p7v: Done

agents/agents-features/agents-features-longterm-memory/src/commonMain/kotlin/ai/koog/agents/longtermmemory/feature/LongTermMemory.kt
serge-p7v: Done

agents/agents-features/agents-features-longterm-memory/src/commonMain/kotlin/ai/koog/agents/longtermmemory/model/MemoryRecord.kt
serge-p7v: I switched to `JsonElement` so it will be possible to use both `JsonObject` and `JsonPrimitive`

agents/agents-features/agents-features-longterm-memory/src/commonMain/kotlin/ai/koog/agents/longtermmemory/prompt/PromptAugmenter.kt
serge-p7v: Fixed with `trimMargin()`

agents/agents-features/agents-features-longterm-memory/src/commonMain/kotlin/ai/koog/agents/longtermmemory/prompt/PromptAugmenter.kt
serge-p7v: Good catch! Done.

agents/agents-features/agents-features-longterm-memory/src/commonMain/kotlin/ai/koog/agents/longtermmemory/prompt/PromptAugmenter.kt
serge-p7v: True, no need to concatenate the system prompt. Fixed.

agents/agents-features/agents-features-longterm-memory/src/commonMain/kotlin/ai/koog/agents/longtermmemory/prompt/PromptAugmenter.kt
serge-p7v: Done

agents/agents-features/agents-features-longterm-memory/src/commonMain/kotlin/ai/koog/agents/longtermmemory/storage/EphemeralMemoryStorage.kt
serge-p7v: I renamed it to `NonPersistentMemoryStorage `.

serge-p7v:
> And a few more thoughts. Would it be possible to configure the storage for retrieval/ingestion, but disable the automatic behavior and allow manual memory management? E.g. if I don't want to access the memory on each user message since it's not needed for my use-case, but I want to code memory management in specific nodes manually. Something like this, similar to other existing features such as [A2A server](https://github.com/JetBrains/koog/blob/188f0f235375bf7f027629a34ef5f97ad946c4c0/agents/agents-features/agents-features-a2a-server/src/commonMain/kotlin/ai/koog/agents/a2a/server/feature/A2AAgentServer.kt#L112)
> 
> ```kotlin
> val checkMemory by node<String, String> { input ->
>     withLongTermMemory {
>        this.retrievalStorage.search(...)
>        
>        this.ingestionStorage.add(...)
>     }
> 
>     input
> }
> ```
> 
> IMHO keeping automatic behavior disabled by default should be more intuitive

In the previous PR I added node-related code (to use in custom strategies), but I decided to not include it in this PR due to the large size. But you're right, not being able to use the LTM feature in custom strategies is not great, so I returned some of the methods and a simple test.
I'll add memory nodes (e.g. 'extractMemoryRecords' with structured types) and tests for them in a future PR.

agents/agents-features/agents-features-longterm-memory/src/commonMain/kotlin/ai/koog/agents/longtermmemory/model/MemoryRecord.kt
```diff
@@ -0,0 +1,29 @@
...
+
+    /**
+     * The main textual content to be embedded and searched
+     */
+    val content: String,
+
+    /**
+     * Flexible key-value metadata for filtering and custom fields
+     */
+    val metadata: Map<String, JsonElement> = emptyMap()
```
EugeneTheDev: `Map<String, JsonElement>` is essentially a `JsonObject`, you can take a look at its definition. Sorry, I wasn't clear enough, this is what I meant initially - to have `metadata: JsonObject`

agents/agents-features/agents-features-longterm-memory/src/commonMain/kotlin/ai/koog/agents/longtermmemory/storage/EphemeralMemoryStorage.kt
EugeneTheDev: How about `InProcessMemoryStorage`?

agents/agents-features/agents-features-longterm-memory/src/commonMain/kotlin/ai/koog/agents/longtermmemory/feature/LongTermMemory.kt
```diff
@@ -0,0 +1,658 @@
...
+ *
+ * This is the primary way to access long-term memory storages from within strategy nodes.
+ * Inside the [action] block, you can use [LongTermMemory.retrievalSettings] and
+ * [LongTermMemory.ingestionSettings] to search and add memory records.
+ *
+ * Example usage:
+ * ```kotlin
+ * val myNode by node<String, Unit> {
+ *     withLongTermMemory {
+ *         ingestionSettings?.storage?.add(records, namespace)
```
EugeneTheDev: I don't think we want to expose settings here, it makes it look more complex. Let's keep these private and create public property getters in `LongTermMemory` class to be able to do `this.retrievalStorage` and `this.ingestionStorage`, like I proposed in my comment.

Or, maybe even better, separate storage configuration and automatic retrieval/ingestion behavior configuration. Then you won't need to create these public getters, you would already have two storages as public properties and optional ingestion/retrieval settings as private properties.

EugeneTheDev (approved):
I only have a question about disabling automatic retrieval/ingestion left. Other than that lgtm

Ololoshechkin (changes requested):
@serge-p7v please create one test in Java that uses all the settings (installs the feature, configures retrieval, also uses `keywordSearch`, `customSearch` and others).

Note: for instance, `keywordSearch` wouldn't work in Java because an extension lambda parameter of a function has no equivalent (`block: KeywordSearchBuilder.() -> Unit`).

Ololoshechkin:
@serge-p7v another point -- could you please also rebase on `develop` and inherit a `PlannerAIAgentFeature` (now it's a separate thing)

agents/agents-features/agents-features-longterm-memory/src/commonMain/kotlin/ai/koog/agents/longtermmemory/model/MemoryRecord.kt
serge-p7v: Thanks! Now it's JsonObject.

agents/agents-features/agents-features-longterm-memory/src/commonMain/kotlin/ai/koog/agents/longtermmemory/feature/LongTermMemory.kt
serge-p7v: Sure, I've made them private and added two property getters.

serge-p7v:
> @serge-p7v please create one test in Java that uses all the settings (installs the feature, configures retrieval, also uses `keywordSearch`, `customSearch` and others).
> 
> Note: for instance, `keywordSearch` wouldn't work in Java because an extension lambda parameter of a function has no equivalent (`block: KeywordSearchBuilder.() -> Unit`).

I've added `LongTermMemoryJavaConfigTest` and some fluent setters to `LongTermMemory`.

serge-p7v:
In the last commit `PromptAugmenter` became an interface to allow users to pass their own implementations with optional post-retrieval operations (e.g. to include only some of the search results in a desired manner into the Prompt).
But main changes in the last commit are about Java API.

agents/agents-features/agents-features-longterm-memory/src/commonMain/kotlin/ai/koog/agents/longtermmemory/storage/EphemeralMemoryStorage.kt
serge-p7v: It's `InMemoryRecordStorage` for now to include both "InMemory" and `MemoryRecord`.

Ololoshechkin (approved):

PR #1493, opened 2026-02-13
fix(prompt): Add Thought part and streaming 'thinking' support Ollama

JGuck:
Introduce a new ContentPart.Thought and StreamFrame.AppendThinking to represent models' internal "thinking" content. Add emitAppendThinking helper and wire parsing/emission through Ollama client (decode thinking, set request think=true) and update stream/message assembly to collect thinking frames. Extend converters and telemetry: A2A/ACP/Bedrock/Google converters and OpenTelemetry span attributes now handle Thought or include TODOs where direct mappings are unclear. Add tests and mock servers for Ollama thinking/streaming behavior and adjust integration test harness to accept an appendable for thinking frames.

My use case is that I am able to show the thinking process in the UI. Nevertheless, I am a little confused about the relation with the reasoning messages. Is Olama thinking the same as openAI reasoning? But in any case, we need to be able to stream the thoughts with StreamFrame. 

I am looking forward to your feedback.


tiginamaria:
Thank you for contribution!

I'm also currently working on the approach to allow reasoning streaming for all the clients we have and it's super great that you proposed the solution for the Ollama. So, if it's not a problem for you, I'd better wait until the general approach for reasoning streaming will be reviewed and merged, so it would be much easier to merge this on top of https://github.com/JetBrains/koog/pull/1479

JGuck:
@tiginamaria Thank you for your reply. This sounds like a good plan. I will wait for the merge and than refactor this pr.

JGuck:
@tiginamaria i messed up this branch, sorry. I created a new PR for a more easy rebase #1532.

PR #1497, opened 2026-02-14
fix(prompt): Stream Ollama response with preparePost.execute

JGuck:
Switch from client.post to client.preparePost(...).execute to stream and process the response body using HttpResponse and ByteReadChannel. Explicitly set Content-Type for the request, move the channel-reading loop into the execute block, and simplify malformed-JSON handling. Added necessary imports (HttpResponse, ByteReadChannel, preparePost) and removed the previous response variable usage.

This is needed because else the body is fully downloaded and then streamed. See https://api.ktor.io/ktor-client-core/io.ktor.client/-http-client.html section streaming

### DIFF
--- a/prompt/prompt-executor/prompt-executor-clients/prompt-executor-ollama-client/src/commonMain/kotlin/ai/koog/prompt/executor/ollama/client/OllamaClient.kt
+++ b/prompt/prompt-executor/prompt-executor-clients/prompt-executor-ollama-client/src/commonMain/kotlin/ai/koog/prompt/executor/ollama/client/OllamaClient.kt
@@ -45,13 +45,16 @@ import io.ktor.client.plugins.contentnegotiation.ContentNegotiation
 import io.ktor.client.plugins.defaultRequest
 import io.ktor.client.request.get
 import io.ktor.client.request.post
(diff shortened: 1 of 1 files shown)


JGuck:
@tiginamaria 
I renamed the pr to fix.
I double-checked my local tests. When I run jvmTests on global scope all tests pass. Are this tests flaky or is there something i miss.

I realized when I am on energy saving mode on my laptop, then there are random tests in A2AServerJsonRpcIntegrationTest failing due to timeouts.

tiginamaria:
Hi! Unfortunately, we don't have `ollama` in scopes, so you should put more global scope`prompt` to pass the conventional commit check https://github.com/JetBrains/koog/blob/develop/CONTRIBUTING.md#conventional-prs. Rerunning the `Simple Examples` helped to fix it, now it's green =)

JGuck:
Ok I renamed the PR; but btw checks are also failing in I don't know w on my pc the same tests are working

JGuck:
Hi @tiginamaria, is there something to-do from my side to get the PR merged?

tiginamaria (approved):
Thank you! All is fine here, just waiting for failing checks to pass

PR #1503, opened 2026-02-18
feat(prompt): router for LLM clients, allowing load balancing capabilities

Amaneusz:
Adds support for load balancing across multiple LLM clients through a new routing abstraction.

### New Module: `prompt-executor-router`
- **`LLMClientRouter` interface**: Abstraction for client selection strategies
- **`RoundRobinRouter`**: Default implementation providing round-robin load balancing per provider

### New Executor: `RoutingLLMPromptExecutor`
- Delegates client selection to pluggable `LLMClientRouter` implementations
- Supports fallback configuration when requested provider unavailable
- Uses `RoundRobinRouter` by default

### Test changes
- Added `MockLLMClient` - all other mocks for LLMClient are now using this


prompt/prompt-executor/prompt-executor-llms/src/commonMain/kotlin/ai/koog/prompt/executor/llms/RoutingLLMPromptExecutor.kt
```diff
@@ -0,0 +1,217 @@
...
+     * with the specified fallback provider.
+     *
+     * @property fallbackProvider The LLMProvider responsible for handling fallback requests.
+     * @property fallbackModel The LLModel instance to be used for fallback execution.
+     *
+     * @throws IllegalArgumentException If the provider of the fallback model does not match the
+     * fallback provider.
+     */
+    public data class FallbackPromptExecutorSettings(
+        val fallbackProvider: LLMProvider,
```
tiginamaria: Maybe we should only put a fallback model, as the provider is already defined inside it?

prompt/prompt-executor/prompt-executor-llms/src/commonMain/kotlin/ai/koog/prompt/executor/llms/RoutingLLMPromptExecutor.kt
```diff
@@ -0,0 +1,217 @@
...
+         * Logger instance used for logging messages within the RoutingLLMPromptExecutor class.
+         *
+         * This logger is used to provide debug logs during the execution of prompts and handling of streaming responses.
+         * It primarily tracks operations such as prompt execution initiation, tool usage, and responses received from the
+         * respective LLM clients.
+         *
+         * The logger can aid in debugging by capturing detailed information about the state and flow of operations within
+         * the class.
+         */
+        private val logger = KotlinLogging.logger("ai.koog.prompt.executor.llms.RoutingLLMPromptExecutor")
```
tiginamaria: nit: in order not to forget to change the logger name after the class rename, you could use `private val logger = KotlinLogging.logger { } `

prompt/prompt-executor/prompt-executor-llms/src/commonMain/kotlin/ai/koog/prompt/executor/llms/RoutingLLMPromptExecutor.kt
tiginamaria: Or you can autoset the provider after the model

prompt/prompt-executor/prompt-executor-router/Module.md
```diff
@@ -0,0 +1,37 @@
+# Module prompt-executor-router
```
tiginamaria: Could you also put the same (or almost the same) description in the `docs` module, so it would be visible in https://docs.koog.ai. I guess the right place is in `docs/prompts/prompt-executors.md`

tiginamaria (approved):
Very solid work, thank you! I added some nit comments and one important point about the docs (just to make sure I don't forget). You can add docs in a separate pr if you want

prompt/prompt-executor/prompt-executor-llms/src/commonMain/kotlin/ai/koog/prompt/executor/llms/RoutingLLMPromptExecutor.kt
Amaneusz: We could do that - I was following what was done in MultiLLMClient but honestly it makes sense to me to proivde only the model.

prompt/prompt-executor/prompt-executor-llms/src/commonMain/kotlin/ai/koog/prompt/executor/llms/RoutingLLMPromptExecutor.kt
Amaneusz: Nice, applied :) I did not know about this one

prompt/prompt-executor/prompt-executor-router/Module.md
Amaneusz: Included short description about Routing in docs

kpavlov (commented):
**My ⓹￠:**

While this is definitely a step forward, I think the MultiLLMPromptExecutor should be extended with routing and failover strategies rather than introducing a new executor. It would be a more generic and flexible solution. What should it provide:

1. dynamic switching on/off - dynamically enable/disable providers by internal (see 2) and external signals
2. availability check and circuit breaker behavior - enable/disable providers based on error rate, latency, health checks, custom signals
3. static routing - define provider priorities and select from available providers in the priority order (e.g., primary traffic goes to OpenAI, fallback to Anthropic)
4. Custom routing strategies - allow pluggable routing strategies, unblocking intelligent routing.

Meanwhile, I suggest merging it as  experimental API

Amaneusz:
Thanks for the review, @kpavlov — answering the comment: https://github.com/JetBrains/koog/pull/1503#pullrequestreview-3833955574
 GitHub does not allow replying directly to a review body, so doing it here.

Your points are valid and I agree with the direction. Extending `MultiLLMPromptExecutor` was actually my initial plan — the first commits on this branch went that way.

The problem I ran into: `MultiLLMPromptExecutor` uses `Map<LLMProvider, LLMClient>` (single client per provider), while routing requires `Map<LLMProvider, List<LLMClient>>`. Due to JVM type erasure, both map types erase to the same signature, so both constructors can't coexist. Trying to work around that ended up breaking the existing API and — more importantly — producing noticeably ugly code, which felt like a sign to step back.

Creating a separate `RoutingLLMPromptExecutor` keeps `MultiLLMPromptExecutor` stable and lets the routing API evolve independently — which also maps nicely to your points 3 and 4 (static routing + pluggable strategies), since `LLMClientRouter` is already a pluggable interface.

Points 1 and 2 (dynamic enable/disable, circuit breaker) are out of scope for this PR but could be follow-up implementations of `LLMClientRouter`. I already discussed it with @Ololoshechkin and we intend to proceed in that way.

Happy to mark it as experimental API — that's a reasonable middle ground.

prompt/prompt-executor/prompt-executor-llms/src/commonMain/kotlin/ai/koog/prompt/executor/llms/ExperimentalRoutingApi.kt
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * To use an annotated API, opt in explicitly at the call site or on the enclosing declaration:
+ * ```kotlin
+ * @OptIn(ExperimentalRoutingApi::class)
+ * fun myCode() { ... }
+ * ```
+ */
+@RequiresOptIn("This API is experimental and is likely to change in the future.")
+@Retention(AnnotationRetention.BINARY)
+public annotation class ExperimentalRoutingApi
```
kpavlov, line 18: Maybe, let's use `@ExperimentalAgentsApi` to avoid creating too many annotations

prompt/prompt-executor/prompt-executor-llms/src/commonMain/kotlin/ai/koog/prompt/executor/llms/LLMClientRouter.kt
```diff
@@ -0,0 +1,27 @@
...
+import ai.koog.prompt.llm.LLModel
+
+/**
+ * Routes requests to available LLM clients.
+ *
+ * Responsible for selecting which client should handle a request for a given model,
+ * based on factors like load distribution, availability, or health.
+ */
+@OptIn(ExperimentalRoutingApi::class)
+public interface LLMClientRouter {
```
kpavlov, line 13: Do we really need another module (`prompt-executor-router`) for client router? Shouldn't we put it in the same place as MultiLLMPromptExecutor (`prompt/prompt-executor/prompt-executor-llms`)?

prompt/prompt-executor/prompt-executor-llms/src/commonMain/kotlin/ai/koog/prompt/executor/llms/RoutingLLMPromptExecutor.kt
```diff
@@ -0,0 +1,207 @@
...
+     * @param tools A list of `ToolDescriptor` objects representing external tools available for use during execution.
+     **/
+    override fun executeStreaming(
+        prompt: Prompt,
+        model: LLModel,
+        tools: List<ToolDescriptor>
+    ): Flow<StreamFrame> {
+        logger.debug { "Executing streaming prompt: $prompt with model: $model" }
+
+        val (client, effectiveModel) = chooseClientAndModel(model)
```
kpavlov: `executeStreaming` throws eagerly from a non-suspending function

```kotlin
// RoutingLLMPromptExecutor.kt
override fun executeStreaming(
prompt: Prompt,
model: LLModel,
tools: List<ToolDescriptor>
): Flow<StreamFrame> {
val (client, effectiveModel) = chooseClientAndModel(model)   // throws here
return client.executeStreaming(prompt, effectiveModel, tools)
}
```

`chooseClientAndModel` throws `IllegalArgumentException` if no client is found. 

**BTW, why IllegalArgumentException, not IllegalStateException? It may pass and fail with the same argments, depending on routers availability.**

Since executeStreaming is not a suspend function, this exception escapes
synchronously at the call site, not during Flow collection. The test reflects this:

```kotlin

assertFailsWith<IllegalArgumentException> {
executor.executeStreaming(prompt, AnthropicModels.Sonnet_3_7).collect()
//                        ↑ throws here, .collect() never runs
}
```

The test is technically passing, but for the wrong reason. In Kotlin Flow, errors are expected to surface during collection, allowing callers to handle
them inside a catch operator or in collect's exception handler. An exception thrown by the factory function is uncatchable through Flow operators. The
idiomatic correction:
```kotlin

override fun executeStreaming(
prompt: Prompt,
model: LLModel,
tools: List<ToolDescriptor>
): Flow<StreamFrame> = flow {
val (client, effectiveModel) = chooseClientAndModel(model)
emitAll(client.executeStreaming(prompt, effectiveModel, tools))
}

```

Note: `MultiLLMPromptExecutor.executeStreaming` has the same defect and does not use the fallback for streaming at all (it skips the fallback logic
entirely).

prompt/prompt-executor/prompt-executor-router/src/commonMain/kotlin/ai/koog/prompt/executor/router/RoundRobinRouter.kt
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    private sealed class ClientPool {
+        abstract fun next(): LLMClient
+
+        data class Single(val client: LLMClient) : ClientPool() {
+            override fun next(): LLMClient = client
+        }
+
+        data class Multiple(val clients: List<LLMClient>) : ClientPool() {
```
kpavlov: data class generates equals, hashCode, copy, and toString based on declared properties. The counter field is not a constructor property, so copy() creates
a fresh Multiple with a zeroed counter, discarding the current rotation state. While ClientPool is private and callers do not currently invoke copy(),
mixing data class semantics with mutable internal state is a known trap. Use a plain class: 
```kotlin
class Multiple(val clients: List<LLMClient>) : ClientPool() { ... }
```

prompt/prompt-executor/prompt-executor-llms/src/commonMain/kotlin/ai/koog/prompt/executor/llms/RoutingLLMPromptExecutor.kt
```diff
@@ -0,0 +1,207 @@
...
+
+        return client.moderate(prompt, effectiveModel)
+    }
+
+    override suspend fun models(): List<LLModel> {
+        logger.debug { "Fetching available models from all clients" }
+
+        return clientRouter.clients
+            .flatMap { it.models() }
+            .distinct()
```
kpavlov, line 190: `models()` silently deduplicates; MultiLLMPromptExecutor.models() does not

prompt/prompt-executor/prompt-executor-llms/src/commonMain/kotlin/ai/koog/prompt/executor/llms/RoutingLLMPromptExecutor.kt
```diff
@@ -0,0 +1,207 @@
...
+    override fun close() {
+        clientRouter.clients.forEach { it.close() }
+    }
+
+    private fun chooseClientAndModel(requestedModel: LLModel): EffectiveExecutionSubject {
+        val lbClient = clientRouter.clientFor(requestedModel)
+        return when {
+            lbClient != null -> lbClient to requestedModel
+            fallback != null -> fallbackClient!! to fallback.fallbackModel
+            else -> throw IllegalArgumentException("No client found for provider: ${requestedModel.provider}")
```
kpavlov, line 202: use `error(message)`

IMPORTANT! Please enable detekt plugin in your IDE

prompt/prompt-executor/prompt-executor-llms/src/commonMain/kotlin/ai/koog/prompt/executor/llms/RoutingLLMPromptExecutor.kt
```diff
@@ -0,0 +1,207 @@
...
+    /**
+     * Fallback LLM client for interacting with a fallback LLM provider.
+     *
+     * Retrieves client specified in `fallback` settings from `clientRouter` if available.
+     * If multiple clients are found for the fallback provider, the first one is used as the fallback client.
+     * This client is intended to handle cases where no specific client is matched during prompt execution.
+     *
+     * Returns `null` if `fallback` is not specified or corresponding client is not found in `clientRouter`.
+     */
+    private val fallbackClient: LLMClient? = when {
```
kpavlov: In the `fallback != null` branch, `fallbackClient` is guaranteed non-null by the constructor invariant (the init block throws if the client is not found). But
the type is LLMClient?, so `!!` is necessary. This compiles but conceals the invariant from the reader. Structure the code so the type system carries the
guarantee:
```kotlin
private val fallbackClientAndModel: Pair<LLMClient, LLModel>? = fallback?.let { settings ->
val client = clientRouter.clients.firstOrNull { it.llmProvider() == settings.fallbackModel.provider }
?: error("Client for provider ${settings.fallbackModel.provider} not found in router")
client to settings.fallbackModel
}
```
Now `chooseClientAndModel` can use `fallbackClientAndMode` directly without `!!`, and the fallback model retrieval is co-located with the client lookup.

kpavlov (commented):
There are a few things we should probably take a closer look at.

prompt/prompt-executor/prompt-executor-llms/src/commonMain/kotlin/ai/koog/prompt/executor/llms/RoutingLLMPromptExecutor.kt
Amaneusz: This is intended - as `MultiLLMPromptExecutor` serves single client per provider, there is no logical way to get duplicates.
Routing assumes N clients for provider - if we drop `distinct` and, for example, we have 5 clients for OpenAI we will get 5 instances of each OpenAI model - which is IMO worse.
Or is there some case I am missing?

prompt/prompt-executor/prompt-executor-llms/src/commonMain/kotlin/ai/koog/prompt/executor/llms/LLMClientRouter.kt
Amaneusz: This was intended mainly because of:
1) Future growth. I expect routing capabilites to grow essentially - we will add routing based on latency, availability etc. I'd like not to pollute `prompt-executor-llms` with that, at least not directly.
2) We are open for separate consumption of `router` - choosing the correct client might not be concern of the `executor` only. Right now `router` only knows about clients, the execution part is out of its scope.

For me, having both in single module mean coupling. Having a separate module seems like a cost we can bare, considering clean separation of concerns it provides.

prompt/prompt-executor/prompt-executor-router/src/commonMain/kotlin/ai/koog/prompt/executor/router/RoundRobinRouter.kt
Amaneusz: Good remark, fixed

prompt/prompt-executor/prompt-executor-llms/src/commonMain/kotlin/ai/koog/prompt/executor/llms/RoutingLLMPromptExecutor.kt
Amaneusz: The use of `IllegalArgumentException` was based on what is going on in `MultiLLMPromptExecutor` - it behaves in this exact way and I wanted to preserve some compaitbility between them (aligned with the "least surprise principle" 😄 )

The `error` throws `IllegalStateException` - is really state illegal here? There is nothign wrong if we have specified specific clients in our router and executor - it's the responsibility of the call-site to provide the correct model for execution.
If the user can only use OpenAPI and tries to execute prompt on Gemini - IMO it's not the state that is invalid, rather the argument

prompt/prompt-executor/prompt-executor-llms/src/commonMain/kotlin/ai/koog/prompt/executor/llms/ExperimentalRoutingApi.kt
Amaneusz: IMO it's not a great fit. `ExperimentalAgentsApi` lives in `agents-core` under the `ai.koog.agents.core.annotation package` — it's semantically scoped to the `agent` layer. The router lives in the `prompt` layer, which is a separate concern.

Using `ExperimentalAgentsApi` on prompt executor classes would be a layering violation — the `prompt-*` modules don't depend on `agents-core`, and mixing those namespaces would be confusing.

prompt/prompt-executor/prompt-executor-llms/src/commonMain/kotlin/ai/koog/prompt/executor/llms/RoutingLLMPromptExecutor.kt
Amaneusz: Good remark, applied something similar - it's clenaer now

prompt/prompt-executor/prompt-executor-llms/src/commonMain/kotlin/ai/koog/prompt/executor/llms/RoutingLLMPromptExecutor.kt
Amaneusz: > why IllegalArgumentException, not IllegalStateException

see: https://github.com/JetBrains/koog/pull/1503#discussion_r2840098991
For now I am more for `IllegalArgumentException`

> executeStreaming throws eagerly from a non-suspending function

Really nice catch, fixed - both in new and old (Multi) executors.
I also slightly changed the test to be sure we fail in correct moment (on flow's collection)

prompt/prompt-executor/prompt-executor-llms/src/commonMain/kotlin/ai/koog/prompt/executor/llms/RoutingLLMPromptExecutor.kt
kpavlov: If we think of dynamic routing and circuit-breaker-like behavior, then it's a state that might change internally. The request is formally correct. I don't have a blocking opinion here - just an observation

prompt/prompt-executor/prompt-executor-router/src/commonMain/kotlin/ai/koog/prompt/executor/router/LLMClientRouter.kt
```diff
@@ -0,0 +1,45 @@
...
+     */
+    public val availableClients: List<LLMClient>
+
+    /**
+     * Selects a client to handle the given model.
+     *
+     * @param model The model to route
+     * @return A client capable of serving the model, or null if none available
+     */
+    public fun chooseRouteFor(model: LLModel): LLMClient?
```
kpavlov: We could even choose a route for expected model capabilities. This would allot to select the best provider for the job.

prompt/prompt-executor/prompt-executor-router/src/commonMain/kotlin/ai/koog/prompt/executor/router/RoutingStrategy.kt
```diff
@@ -0,0 +1,11 @@
+package ai.koog.prompt.executor.router
+
+/**
+ * Defines how requests are distributed across multiple clients.
+ */
+public enum class RoutingStrategy {
```
kpavlov: looks like an interface name

prompt/prompt-executor/prompt-executor-llms/src/commonMain/kotlin/ai/koog/prompt/executor/llms/MultiLLMPromptExecutor.kt
```diff
@@ -17,30 +17,19 @@ import kotlinx.coroutines.flow.Flow
...
+ * handle each request based on the requested model.
  *
- * @constructor Constructs an executor instance with a map of LLM providers associated with their respective clients.
- * @param llmClients A map containing LLM providers associated with their respective [LLMClient]s.
- * @param fallback Optional settings to configure the fallback mechanism in case a specific provider is not directly available.
+ * @param clientRouter Router responsible for selecting appropriate clients for each request
+ * @param fallback Optional fallback configuration when no client is available for the requested model
  */
 public open class MultiLLMPromptExecutor @JvmOverloads constructor(
-    private val llmClients: Map<LLMProvider, LLMClient>,
```
kpavlov, line 30: Let's keep an old constructor for compatibility. It will be a very breaking change for users

prompt/prompt-executor/prompt-executor-llms/src/commonMain/kotlin/ai/koog/prompt/executor/llms/RoutingLLMPromptExecutor.kt
```diff
@@ -0,0 +1,210 @@
...
+     *
+     * @param prompt The `Prompt` containing the content to be moderated.
+     * @param model The `LLModel` to use for moderation, including its ID and provider information.
+     * @return A `ModerationResult` representing the result of the moderation process.
+     * @throws IllegalArgumentException If no client is found for the model's provider.
+     */
+    override suspend fun moderate(prompt: Prompt, model: LLModel): ModerationResult {
+        logger.debug { "Moderating multi-modal content with model: ${model.id}" }
+
+        val (client, effectiveModel) = chooseClientAndModel(model)
```
kpavlov, line 180: I think we need to choose a client based on both Prompt and the Model. Prompt is required to implement sticky routing, e.g. to support stateful LLM conversations

kpavlov (commented):
Added couple of comments

prompt/prompt-executor/prompt-executor-router/src/commonMain/kotlin/ai/koog/prompt/executor/router/RoutingStrategy.kt
Amaneusz: This file does not exist anymore

prompt/prompt-executor/prompt-executor-llms/src/commonMain/kotlin/ai/koog/prompt/executor/llms/MultiLLMPromptExecutor.kt
Amaneusz: Yhis change is not part of the PR, this was during my attempt to make routing part of `MultiLLMPromptExecutor` - this comment also presents nicely why this was a wrong path ;)

prompt/prompt-executor/prompt-executor-router/src/commonMain/kotlin/ai/koog/prompt/executor/router/LLMClientRouter.kt
Amaneusz: This is more the case for Model Selection task - discussed on priv

prompt/prompt-executor/prompt-executor-llms/src/commonMain/kotlin/ai/koog/prompt/executor/llms/RoutingLLMPromptExecutor.kt
Amaneusz: This is more the case for Model Selection task - discussed on priv

Ololoshechkin:
@Amaneusz , regarding the Map:

>  The problem I ran into: MultiLLMPromptExecutor uses Map<LLMProvider, LLMClient> (single client per provider), while routing requires Map<LLMProvider, List<LLMClient>>. Due to JVM type erasure, both map types erase to the same signature, so both constructors can't coexist.

I think we can just go with a list of pairs (or vararg in the Kotlin case) constructor (which actually already exists). This way you can just pass list of pairs with repeating keys and then construct the right map on the implementation side.
WDYT?

Ololoshechkin:
One more note to keep in mind: we're planning to get rid of `MultiLLMPromptExecutor` and `SingleLLMPromptExecutor` classes, and make `PromptExecutor` an open class which is always multi-llm (single-llm is just a particular case of it)

Ololoshechkin:
Another point to consider: ideally for Java the API should be something like that:
```java
PromptExecutor.builder()
    .addLlmClient(client1)
    .addLlmClient(provider2, client2)
    .addLlmClient(provider2, client3)
    .withRouting(LLMClientRouter(...))
    .withFallback(...)
    .build()
```

prompt/prompt-executor/prompt-executor-llms/src/commonMain/kotlin/ai/koog/prompt/executor/llms/LLMClientRouter.kt
Ololoshechkin: Let's maybe keep the interface in the main module -- this way we can later on add nice chained builders for Java that would have rouring capability for `PromptExecutor`. And then all specific implementations we can move to separate modules.

Interface is thin enough :)

Ololoshechkin (approved):
@Amaneusz please check my comments, and don't forget to move the `LLMClientRouter` interface to the Main prompt-executor-llms  module (and also remove the dependency on `prompt-executor-router` from it, because currently it's a bit strange that we have `RoutingLLMPromptExecutor` in `executor-llms` module but it still requires  `LLMClientRouter` from the `executor-router` module => hence it's impossible to install only `executor-llms` without getting everything from  `executor-router` at the same time. So it doesn't make sense right now to have this separation)

PR #1505, opened 2026-02-19
fix(koog-ktor): parsing timeouts configuration

Amaneusz:
The `configure` function in `EnvConfigLoader` had an inverted condition and read from the root config instead of the provider's timeout sub-section, causing custom timeout values to be silently ignored. This fixes both issues and adds YAML-based tests covering the exact scenario reported in the issue.

The fix was originally proposed by community: https://github.com/JetBrains/koog/issues/807
This PR cherry picked these changes and added tests on top of it. Additionally it also fixes `application.yaml` example

Closes:
- https://youtrack.jetbrains.com/issue/KTOR-8881
- https://github.com/JetBrains/koog/issues/807

### DIFF
--- a/koog-ktor/src/commonMain/kotlin/ai/koog/ktor/utils/EnvConfigLoader.kt
+++ b/koog-ktor/src/commonMain/kotlin/ai/koog/ktor/utils/EnvConfigLoader.kt
@@ -157,7 +157,7 @@ private fun KoogAgentsConfig.TimeoutConfiguration.configure(key: String, config:
             ?.getString()
             ?.toLongOrNull()
             ?.let { connectTimeout = it.milliseconds }
-        
+
         providerTimeoutSection.propertyOrNull("socketTimeoutMillis")
             ?.getString()
             ?.toLongOrNull()
--- a/koog-ktor/src/commonMain/kotlin/ai/koog/ktor/utils/EnvConfigLoader.kt
+++ b/koog-ktor/src/commonMain/kotlin/ai/koog/ktor/utils/EnvConfigLoader.kt
@@ -145,17 +145,19 @@ private inline fun KoogAgentsConfig.config(
 }
 
 private fun KoogAgentsConfig.TimeoutConfiguration.configure(key: String, config: ApplicationConfig) {
-    if (config.propertyOrNull(key) == null) {
-        config.propertyOrNull("requestTimeoutMillis")
+    if (config.propertyOrNull(key) != null) {
+        val configKey = config.config(key)
+
+        configKey.propertyOrNull("requestTimeoutMillis")
             ?.getString()
             ?.toLongOrNull()
             ?.let { requestTimeout = it.milliseconds }
 
-        config.propertyOrNull("connectTimeoutMillis")
+        configKey.propertyOrNull("connectTimeoutMillis")
             ?.getString()
             ?.toLongOrNull()
             ?.let { connectTimeout = it.milliseconds }
-        config.propertyOrNull("socketTimeoutMillis")
+        configKey.propertyOrNull("socketTimeoutMillis")
             ?.getString()
             ?.toLongOrNull()
             ?.let { socketTimeout = it.milliseconds }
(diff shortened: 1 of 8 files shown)


koog-ktor/src/jvmTest/kotlin/ai/koog/ktor/YamlConfigurationLoadingTest.kt
```diff
@@ -0,0 +1,26 @@
+package ai.koog.ktor
+
+import io.ktor.server.config.yaml.YamlConfig
+import io.ktor.server.testing.testApplication
+import kotlin.test.Test
+
+class YamlConfigurationLoadingTest {
+
+    @Test
+    fun testYamlConfigWithoutTimeout() = testApplication {
```
sdubov, line 10: Does it make sense to check the configuration properties when you read it from the YAML to make sure you gonna run the app with expected config values?

koog-ktor/src/commonMain/kotlin/ai/koog/ktor/utils/EnvConfigLoader.kt
```diff
@@ -145,17 +145,20 @@ private inline fun KoogAgentsConfig.config(
 }
 
 private fun KoogAgentsConfig.TimeoutConfiguration.configure(key: String, config: ApplicationConfig) {
-    if (config.propertyOrNull(key) == null) {
-        config.propertyOrNull("requestTimeoutMillis")
+    if (config.propertyOrNull(key) != null) {
```
sdubov, line 148: Nice catch.

sdubov (approved):
@Amaneusz , would it be reasonable to add some more assertions in the tests to make sure the properties from config is received when starting the application?

koog-ktor/src/commonMain/kotlin/ai/koog/ktor/utils/EnvConfigLoader.kt
Amaneusz: Thanks, it's not mine :)

koog-ktor/src/jvmTest/kotlin/ai/koog/ktor/YamlConfigurationLoadingTest.kt
Amaneusz: Yes it does and I will work on that 😄 

At first I noticed that all related tests simply check if app is running and I followed the same pattern (to be fair the bug was about the app not starting which is covered).
We should be more strict about it though - I agree.

Amaneusz: tl;dr:
- testing this properly requires bigger refactoring
- I want to ship what we have now and work on making configs testable in other task

---

I explored this and it turns out the design makes it harder than it looks — timeout values get baked into private fields of each `LLMClient` during configuration and there's no way to read them back out through the public API. Verifying them would require either reflection or adding production code that serves no purpose beyond testing. Both options are not worth it IMO.

The current tests cover the regression at the right level: smoke tests confirm the full YAML path works without crashing for both cases (with and without timeout config), which shows that bug that was fixed.

I would suggest a follow-up to make configured state properly observable — likely by surfacing timeout config through the `LLMClient` abstraction — which would then make this kind of assertion straightforward and clean across all providers.

sdubov: Ok. That's fine with me. Just wanted to make sure that we verify a correct stuff. If you have a plan for fix this separately, that's totally good.

sdubov (approved):

Ololoshechkin (approved):

koog-ktor/src/jvmTest/kotlin/ai/koog/ktor/YamlConfigurationLoadingTest.kt
Amaneusz: Follow up task: https://youtrack.jetbrains.com/issue/KG-702/koog-ktor-tests-for-Koog-startup-config

PR #1185, opened 2025-11-26
Introduce Java API to allow using Koog from pure Java projects

Ololoshechkin:
## Breaking Changes

---

#### Type of the changes
- [x] New feature (non-breaking change which adds functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Tests improvement
- [x] Refactoring

#### Checklist
- [ ] The pull request has a description of the proposed change
- [ ] I read the [Contributing Guidelines](https://github.com/JetBrains/koog/blob/main/CONTRIBUTING.md) before opening the pull request
- [ ] The pull request uses **`develop`** as the base branch
- [ ] Tests for the changes have been added
- [ ] All new and existing tests passed

##### Additional steps for pull requests adding a new feature
- [ ] An issue describing the proposed change exists
- [ ] The pull request includes a link to the issue
- [ ] The change was discussed and approved in the issue
- [ ] Docs have been added / updated

PR #1508, opened 2026-02-19
fix(examples): 0.6.3 add missing code agents dir

Amaneusz:
Adding missing files (especially dir `step-05-history`) to 0.6.3 release as it is required by `code-agents-examples` workflow.
This was presumably left out by accident but it breaks PRs to release branch (ie: https://github.com/JetBrains/koog/actions/runs/22180137492/job/64139013325?pr=1507)


Amaneusz:
`step-05-history` depends on `singleRunStrategyWithHistoryCompression` which doesn't exist in `release/0.6.3` — backporting it would mean partially pulling in the Java API work (#1185) which is intentionally held back for `0.7.0.` For now, we should remove `step-05-history` from the CI matrix in `code-agent-examples.yml` on `release/0.6.3`. Once `0.7.0` ships with #1185, add it back

PR #1511, opened 2026-02-19
feat(agents)!: Chat memory

Rizzen:

BREAKING:
*

DEPRECATED:
*

closes

### DIFF
--- a/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/extension/AIAgentLLMActions.kt
+++ b/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/extension/AIAgentLLMActions.kt
@@ -126,6 +126,12 @@ public suspend fun AIAgentLLMWriteSession.replaceHistoryWithTLDR(
     strategy.compress(this, memoryMessages)
 }
 
+public fun AIAgentLLMWriteSession.replaceHistory(
+    messages: List<Message>
+) {
+    this.prompt = this.prompt.withMessages { messages }
+}
+
 /**
  * Drops all trailing tool call messages from the current prompt
  */
--- a/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/extension/AIAgentLLMActions.kt
+++ b/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/extension/AIAgentLLMActions.kt
@@ -126,12 +126,6 @@ public suspend fun AIAgentLLMWriteSession.replaceHistoryWithTLDR(
     strategy.compress(this, memoryMessages)
 }
 
-public fun AIAgentLLMWriteSession.replaceHistory(
-    messages: List<Message>
-) {
-    this.prompt = this.prompt.withMessages { messages }
-}
-
 /**
  * Drops all trailing tool call messages from the current prompt
  */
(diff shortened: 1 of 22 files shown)


agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/extension/AIAgentLLMActions.kt
```diff
@@ -126,6 +126,12 @@ public suspend fun AIAgentLLMWriteSession.replaceHistoryWithTLDR(
     strategy.compress(this, memoryMessages)
 }
 
+public fun AIAgentLLMWriteSession.replaceHistory(
```
Ololoshechkin: Probably this one is a bit redundant

agents/agents-features/agents-features-memory/src/commonMain/kotlin/ai/koog/agents/chatMemory/feature/ChatMemory.kt
```diff
@@ -0,0 +1,85 @@
...
+                val history = config.chatHistoryProvider.load(it.context.runId)
+
+                it.context.llm.writeSession {
+                    replaceHistory(history)
+                }
+            }
+
+            pipeline.interceptStrategyCompleted(this) {
+                val historyToSave = it.context.llm.prompt.messages
+                config.chatHistoryProvider.store(it.context.runId, historyToSave)
```
Ololoshechkin: TODO: as we discussed, current approach is to only save memory on synchronisation points (start/end of strategy) and advertise Persistence to handle the rest.
But let's also check how it's addressed in other frameworks

agents/agents-features/agents-features-memory/src/commonMain/kotlin/ai/koog/agents/chatMemory/feature/ChatMemoryConfig.kt
```diff
@@ -0,0 +1,42 @@
...
+ * for agent-user interactions.
+ */
+public class ChatMemoryConfig : FeatureConfig() {
+
+    /**
+     * A provider responsible for persisting and retrieving conversation history.
+     *
+     * Defaults to [InMemoryChatHistoryProvider].
+     */
+    public var chatHistoryProvider: ChatHistoryProvider = InMemoryChatHistoryProvider()
```
Ololoshechkin, line 19: If we want to have a windowed memory and use Postgres, do I need to instantiate PostgresProvider and then weap into a WindowedProvider(..., size)?

Ololoshechkin (approved):
LGTM in general, but let's think about sliding window beforehand, probably add to the example.

Having nested providers VS 2 config params : provider + window size is debatable:
Nested is easier to support.
2 Params are easier to use but then what if we want to add other types of filters?

Maybe provider + filter or provider + processor would be enough?

Processor can be als accumulative, and we can provide handy extensions like setWindowSize(...), filterMessages {...}, etc. that would update it in feature config

agents/agents-features/agents-features-memory/src/commonMain/kotlin/ai/koog/agents/chatMemory/feature/ChatMemoryConfig.kt
Rizzen: I'll implement window as a standart part of memory, not separate provider, it's just not here yet :D

Ololoshechkin: Please make sure it can be extended to non-window type of filters. In general should be something with `List<Message> -> List<Message>` capability one can override :)

Ololoshechkin:
@Rizzen please also rebase on develop and inherit your feature from PlannerFeature (it's now separate from Functional)

Ololoshechkin:
nit: `PreProcessor` -> `Preprocessor` maybe?

Ololoshechkin:
nice to have: maybe add an extra DSL method that works with lambda `Message -> Boolean` (should be a functional interface in order to work in Java) and is called `filterMessages` -- under the hood it would instantiate a filtering message processor (also add example + test). Also note that the order of adding preprocessors is important right now:
```kotlin
windowSize(10)
filterMessages { it.length <= 100 }
```
VS
```kotlin
filterMessages { it.length <= 100 }
windowSize(10)
```

First would give you no-more-than-10 messages which are short among the latest 10, second one would give you exactly 10 messages (last 10 among the short ones).

Maybe it's fine, just have to document and add to example

Ololoshechkin:
LGTM!

PR #1518, opened 2026-02-21
feat(prompt): add transparent PII anonymization modules and detectors

JGuck:
**prompt-pii** is a transparent privacy layer for prompt execution.

It wraps a `PromptExecutor` and does three things:

* Detects Personally Identifiable Information (PII) in prompt text.
* Replaces detected values with stable placeholder tags (for example `[[person 1]]`) before sending data to the model.
* Restores known tags back to original values in model responses.

It also enforces safety by validating tag usage in responses:

* unknown/invented tags are rejected (`UnknownPiiTagsException`)
* optional non-streaming repair is available via `PiiTagFixingParser`

Detection is pluggable:

* `RegexPiiDetector.default()` for deterministic regex/checksum-based detection
* `StructuredPiiDetector` for LLM-assisted typed extraction (substring + `PiiType`) with local span resolution

### What’s included

* Added new `prompt-pii` module with:

  * `prompt-pii-model`
  * `prompt-pii-detector-structured`
* Added core PII model/types and detector interface:

  * `PiiType`, `PiiDetection`, `PiiDetector`, `PiiTypeMapper`
* Added deterministic regex/checksum detector:

  * `RegexPiiDetector.default()`
* Added transparent prompt wrapper:

  * `PiiPromptExecutor` (anonymize before model call, deanonymize after)
* Added unknown-tag handling:

  * `UnknownPiiTagsException`
  * optional `PiiTagFixingParser` for non-streaming repair
* Added LLM-backed structured detector:

  * `StructuredPiiDetector` (+ config and prompt customization)
* Added module wiring/docs updates (settings.gradle.kts, Dokka/module docs, prompt module docs)
* Added JVM tests for executor behavior, regex detector, structured detector, and parser/mapping logic

### Why

Goal is to provide a reusable and transparent PII protection layer around `PromptExecutor` with:

* predictable anonymization tags
* strict safety around tag reuse
* pluggable detection (deterministic + LLM-structured)
* I think we should be pragmatic about the PiiType enumeration and extend types as we need them for integration.
* later integration of 
  * Presidio: [https://github.com/microsoft/presidio](https://github.com/microsoft/presidio)
  * https://www.private-ai.com


kpavlov:
@JGuck , thank you! This is really needed for enterprise users!

another candidate for integration is https://www.private-ai.com. Please check their api too.

After brief review, the approach looks good to me.

JGuck:
@kpavlov, thank you for your review. I also thought about my open question and decided that revealing tags might contradict the whole idea. I documented this also in the issue. Furthermore, I scanned through the service u mentioned and found that they provide the same thing what is implemented here as a service. Nevertheless, I would suggest merging this PR and add new integrations as new PRs. If we need new PiiTags we should extend the internal enum on the fly. What do you think?

JGuck:
Hay @kpavlov i just wanted to ping you about this PR. Is ther any thing i can do to get this PR merged?

PR #1526, opened 2026-02-23
fix!: retired and deprecated models cleanup

aozherelyeva:
Replace deprecated Anthropic/Bedrock model references with `Opus_4_6` and update related documentation, tests, and examples accordingly.

BREAKING:
*  Sonnet_3_7, Haiku_3_5, Sonnet_3_5, and Opus_3 profiles are deleted from the Anthropic models list 
* Opus_3, Haiku_4_5, AnthropicClaude3Sonnet, Sonnet_3_5, Haiku_3_5, AnthropicClaude21, AnthropicClaudeInstant, AI21JambaLarge, AI21JambaMini are deleted from the Bedrock models list 
* Removed the AI21 Jamba parser, data structure, and tests.
* In most of the places, Opus 4.6 has replaced the deprecated (or soon-to-be-deprecated) models (as recommended by Anthropic).

DEPRECATED:
* Haiku_3 is officially deprecated an soon-to-be-retired by Anthropic. Marked this profile as a deprecated one.


devcrocod (approved):
Thanks, lgtm

One question: which version is the default right now 4.6 or 4.5?
If it’s 4.5, I think it might make sense to use 4.5 as the more commonly used model.
If not, then we can leave it as is

tiginamaria (approved):

aozherelyeva:
@devcrocod 
> One question: which version is the default right now 4.6 or 4.5?

Thanks! The 4.6 is recommended by Anthropic so far: https://platform.claude.com/docs/en/about-claude/model-deprecations#2025-10-28-claude-sonnet-3-7-model

PR #1545, opened 2026-02-25
test: Add thinking model integration test

aozherelyeva:
Added an integration test to check thinking feature of Ollama models.

BREAKING:
* none

DEPRECATED:
* none

### DIFF
--- a/integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/executor/OllamaExecutorIntegrationTest.kt
+++ b/integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/executor/OllamaExecutorIntegrationTest.kt
@@ -20,13 +20,16 @@ import ai.koog.prompt.llm.LLMCapability.Tools
 import ai.koog.prompt.llm.LLMCapability.Vision
 import ai.koog.prompt.llm.LLModel
 import ai.koog.prompt.markdown.markdown
+import ai.koog.prompt.streaming.StreamFrame
 import io.kotest.assertions.withClue
 import io.kotest.matchers.booleans.shouldBeTrue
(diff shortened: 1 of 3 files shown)


tiginamaria (approved):
Looks good! Thank you!

integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/executor/OllamaExecutorIntegrationTest.kt
```diff
@@ -344,4 +348,35 @@ class OllamaExecutorIntegrationTest : ExecutorIntegrationTestBase() {
         val response = executor.execute(prompt, model).single()
         response.content.shouldNotBeBlank()
     }
+
+    @Test
+    fun `ollama_test thinking feature in streaming mode`() = runTest(timeout = 600.seconds) {
```
kpavlov, line 353: I think, using `runTest` in integration tests is very wrong because:

- It relies on virtual time and provides deterministic execution, which is unsuitable for real-world systems.
- Integration tests require actual dispatchers, real time, and asynchronous behavior.
- `runTest` is intended only for unit testing coroutine code in isolation.

With [JUnit 6](https://docs.junit.org/6.0.3/release-notes.html#v6.0.0) it could look like this.
```suggestion
    @Timeout(value = 10, unit = TimeUnit.MINUTES)
    suspend fun `ollama_test thinking feature in streaming mode`()  {
```
https://docs.junit.org/6.0.3/writing-tests/timeouts


But this is a different story. Let's address it separately

zarechneva-a: I'm implementing it in a separate branch, thanks!

PR #1547, opened 2026-02-25
fix(agents): KG-703 add trace attributes to all Langfuse spans, not j…

mihbor:
…ust invoke_agent

LangFuse trace attributes are currently not set on all spans but only set on invoke_agent span.
This leads to incorrectly labelled traces in LangFuse.
e.g. when setting "langfuse.environment" attribute the trace is still labelled as `env:default`.
This is becase the top level create_agent span (and all the remaining spans) is missing the "langfuse.environment" attribute and defaults environment to "default".

Screenshots showing the issue here: https://youtrack.jetbrains.com/issue/KG-703

closes KG-703

### DIFF
--- a/agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/integration/langfuse/LangfuseSpanAdapter.kt
+++ b/agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/integration/langfuse/LangfuseSpanAdapter.kt
@@ -86,6 +86,8 @@ internal class LangfuseSpanAdapter(
 
             else -> {}
         }
+        // adding attributes to all spans as per Langfuse recommendation for OTEL instrumentation:
+        // https://langfuse.com/integrations/native/opentelemetry#propagating-attributes
         traceAttributes.forEach { attribute ->
             span.addAttribute(attribute)
         }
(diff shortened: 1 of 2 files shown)


fergal-vcola:
Can confirm this is definitely an issue with the current KOOG-langfuse integration (environment Id is not properly set on top-level traces). 
Any change this PR could get review?

tiginamaria (approved):
Looks good to me! Thank you for your contribution!

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/integration/langfuse/LangfuseSpanAdapter.kt
```diff
@@ -90,6 +86,9 @@ internal class LangfuseSpanAdapter(
 
             else -> {}
         }
+        traceAttributes.forEach { attribute ->
```
sdubov, line 91: It would be nice to add some comment why do we add attributes for each span here. Maybe a link to the Langfuse documentation would be helpful - 
https://langfuse.com/integrations/native/opentelemetry#propagating-attributes

sdubov (approved):
@mihbor , thank you a bunch for fixing this! Indeed, according to the Langfuse documentation, this should goes to all spans instead (please see my comment in the PR):

> If you want to reliably filter or aggregate by these attributes, they need to be present on each span in the trace, not only on the root span.

agents/agents-features/agents-features-opentelemetry/src/jvmMain/kotlin/ai/koog/agents/features/opentelemetry/integration/langfuse/LangfuseSpanAdapter.kt
mihbor: thanks, added now

PR #1550, opened 2026-02-26
feat(agents): add planner-specific interceptors to the planner pipeline

antoniibelyshev:
added the following methods to the planning pipeline:
* `onPlanCreationStarting`
* `onPlanCreationCompleted`
* `onStepExecutionStarting`
* `onStepExecutionCompleted`
* `onPlanCompletionEvaluationStarting`
* `onPlanCompletionEvaluationCompleted`

closes KG-672

### DIFF
--- a/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/pipeline/AIAgentPlannerPipelineAPI.kt
+++ b/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/pipeline/AIAgentPlannerPipelineAPI.kt
@@ -1,0 +1,233 @@
+@file:Suppress("MissingKDocForPublicAPI")
+
+package ai.koog.agents.core.feature.pipeline
+
+import ai.koog.agents.core.agent.context.AIAgentContext
+import ai.koog.agents.core.agent.execution.AgentExecutionInfo
+import ai.koog.agents.core.annotation.InternalAgentsApi
+import ai.koog.agents.core.feature.AIAgentFeature
+import ai.koog.agents.core.feature.handler.planner.PlanCompletionEvaluationCompletedContext
+import ai.koog.agents.core.feature.handler.planner.PlanCompletionEvaluationStartingContext
+import ai.koog.agents.core.feature.handler.planner.PlanCreationCompletedContext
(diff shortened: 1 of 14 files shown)


agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/pipeline/AIAgentPlannerPipelineAPI.kt
```diff
@@ -0,0 +1,233 @@
+@file:Suppress("MissingKDocForPublicAPI")
```
EugeneTheDev: nit: Let's avoid suppressing `MissingKDocForPublicAPI` inspection for the whole files in the final version. This applies to all files, please check. It's fine to suppress it for ceratin symbols that objectively don't need KDoc that much (to keep it cleaner)

agents/agents-core/src/jvmMain/kotlin/ai/koog/agents/core/feature/pipeline/AIAgentPlannerPipeline.kt
```diff
@@ -0,0 +1,183 @@
...
+
+    /**
+     * Intercepts plan creation starting event to perform actions when a plan begins creation.
+     *
+     * JVM-friendly overload that accepts an async interceptor.
+     */
+    @JavaAPI
+    @InternalAgentsApi
+    @JvmName("interceptPlanCreationStarting")
+    public fun javaApiInterceptPlanCreationStarting(
```
EugeneTheDev, line 57: Why do we need these Java-friendly interceptors if it's internal API, not meant for public API use?

EugeneTheDev (approved):

agents/agents-core/src/jvmMain/kotlin/ai/koog/agents/core/feature/pipeline/AIAgentPlannerPipeline.kt
antoniibelyshev: Hm, good point. non-suspend interceptors for graph pipeline are also marked with JavaAPI and InternalAgentsApi, I just took it from there. I think, these should not be marked with InternalAgentsApi. I'll update graph pipeline as well

PR #1552, opened 2026-02-26
docs(agents)!: ChatMemory doc + Chat agent intro

Rizzen:

BREAKING:
*

DEPRECATED:
*

closes

### DIFF
--- a/docs/docs/chat-memory.md
+++ b/docs/docs/chat-memory.md
@@ -6,10 +6,6 @@ The ChatMemory feature gives AI agents persistent conversation history across mu
 When installed, the agent automatically loads previous messages at the start of each run and
 stores the updated conversation when the run completes — enabling natural multi-turn chat.
 
-> **ChatMemory vs AgentMemory**: ChatMemory stores raw message history (the actual user/assistant
-> messages). [AgentMemory](agent-memory.md) extracts and stores structured *facts* from conversations.
-> They serve different purposes and can be used together.
-
 ### Key capabilities
 
 - Automatic load/store of conversation history per session ID
--- a/docs/docs/chat-memory.md
+++ b/docs/docs/chat-memory.md
@@ -13,6 +13,45 @@ stores the updated conversation when the run completes — enabling natural mult
 - Built-in preprocessors to limit history size and filter messages
 - Custom preprocessor support for arbitrary message transformations
 
+## Install Koog and Memory feature
+
+=== "Gradle (Kotlin)"
+
+    ```kotlin title="build.gradle.kts"
+    dependencies {
+        implementation("ai.koog:koog-agents:0.7.0")
+        implementation("ai.koog:agents-features-memory:0.7.0")
+    }
+    ```
+
+=== "Gradle (Groovy)"
+
+    ```groovy title="build.gradle"
+    dependencies {
+        implementation 'ai.koog:koog-agents:0.7.0'
+        implementation 'ai.koog:agents-features-memory:0.7.0'
+    }
+    ```
+
+=== "Maven"
+
+    ```xml title="pom.xml"
+    <dependency>
+        <groupId>ai.koog</groupId>
+        <artifactId>koog-agents-jvm</artifactId>
+        <version>0.7.0</version>
+    </dependency>
+    <dependency>
+        <groupId>ai.koog</groupId>
+        <artifactId>agents-features-memory-jvm</artifactId>
+        <version>0.7.0</version>
+    </dependency>
+    ```
+
+!!! note
+    ChatMemory is available starting from version **0.7.0**, which has not been published to Maven Central yet.
+    In the meantime you can use it from a local build or a snapshot repository.
+
 ## Configuration and initialization
 
 ### Basic setup (Kotlin)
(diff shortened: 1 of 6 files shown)


docs/docs/quickstart/chat-agent.md
```diff
@@ -0,0 +1,247 @@
...
+    -->
+    ```kotlin
+    suspend fun main() {
+        val sessionId = "my-conversation"
+
+        simpleOpenAIExecutor(System.getenv("OPENAI_API_KEY")).use { executor ->
+            val agent = AIAgent(
+                promptExecutor = executor,
+                llmModel = OpenAIModels.Chat.GPT4oMini,
+                systemPrompt = "You are a friendly assistant. Keep your answers concise.",
```
Ololoshechkin: please also add tools here! otherwise the agent is just a single LLM-request and it looses the point :)

Ololoshechkin (commented):
Also @Rizzen please consider adding some simple 1-endpoint backend example (with Spring ideally) and referencing it from here like "Tipical use-case for the Chat Memory is to have a backend application that manages agent iterations (see example: <link>)"

Also please consider adding a chapter mentioning `Persistence` that would explain the difference between ChatMemory and Persistence: chat memory treats agent interactions as atomic self-contained loops (i.e. message history will be saved once the agent has fully finished), while if your agent has crashed in the middle the progress will be lost. In order to make agent recoverable from crashes right from where it stopped please consider using `Persistence` feature in addition to `ChatMemory` (or smth like that)

docs/docs/chat-memory.md
```diff
@@ -0,0 +1,159 @@
...
+    .build();
+```
+
+`MessageFilter` is a `fun interface`, so Java lambdas work directly.
+
+## Best practices
+
+1. **Always set a window size** — without one, prompt size grows unboundedly with conversation length.
+2. **Choose preprocessor order carefully** — filtering before windowing and windowing before filtering produce different results.
+3. **Use meaningful session IDs** — these are the keys for history isolation. User IDs, chat thread IDs, or UUIDs all work well.
```
serge-p7v, line 290: It would be great to mention somewhere that for multi-user storage the simple concatenation can do the trick (like a compound id "userId_sessionId"). And maybe for some users the parameter name "conversationId" is clearer than "sessionId".

docs/docs/chat-memory.md
```diff
@@ -0,0 +1,159 @@
...
+4. **Implement a persistent provider for production** — `InMemoryChatHistoryProvider` loses history on restart.
```
serge-p7v, line 291: It would be great to have an example of an implementation with a real database in the documentation or in the code. The most popular databases for chat memory seem to be PostgreSQL and Redis now.

docs/docs/chat-memory.md
```diff
@@ -0,0 +1,159 @@
...
+// Second turn — the agent sees the previous exchange
+agent.run("And what about Germany?", "session-1")
+```
+
+Different session IDs produce fully isolated histories.
+
+## Preprocessors
+
+Preprocessors transform the message list at both load time (before the agent sees it)
+and store time (before persisting). They run sequentially in the order they were added.
```
Amaneusz, line 111: > Preprocessors transform the message list at both load times (before the agent sees it)
and store time (before persisting)

nit: Consider adding some minimal explanation on why it is needed to apply this preprocessing twice. When I was reading this I got slightly confused – "why would I filter messages on load if I already did that before storing?"

_My guess_ is that filtering might be dynamic / based on externals:
- "filter out messages that are at least 7 days old"
- "filter out messages containing keyword incoming from API call to xyz.com/blacklist"

docs/docs/chat-memory.md
```diff
@@ -0,0 +1,159 @@
...
+```kotlin
+install(ChatMemory) {
+    addPreProcessor(RedactEmailsPreProcessor())
+    windowSize(50)
+}
+```
+
+## Custom history providers
+
+The default `InMemoryChatHistoryProvider` is thread-safe but non-persistent.
```
Amaneusz: nit: clarify "non-persistent"

```suggestion
The default `InMemoryChatHistoryProvider` is thread-safe but non-persistent (history is lost on restart).
```

docs/docs/chat-memory.md
Amaneusz, line 291: I got to read it after making this comment: https://github.com/JetBrains/koog/pull/1552/changes#r2863634902

I guess that makes that remark obsolete ;)

Rizzen: I'll implement Pg and Redis providers in next PR

docs/docs/chat-memory.md
Rizzen: Well, the idea behind filtering on store and load was "you get what you store considering filters". Because for example, if you filter only on store and already have something in chat memory that is not fitting into filters - that would end up in memory, because we filter only on store. The same issue occurs in reverse.
So my idea was if you for example set filter on messages up to 120 symbols - you store and load only filtered messages, no way around.

Amaneusz (approved):

Ololoshechkin (approved):

PR #1555, opened 2026-02-26
refactor(prompt)!: Make LLMClient and PromptExecutor Java-friendly with blocking APIs

Ololoshechkin:
Make LLMClient and PromptExecutor Java-friendly with blocking APIs:

BREAKING:
*

- LLMClient is now an abstract class (previously interface)
- PromptExecutor is now an abstract class (previously interface)

DEPRECATED:
*

closes


prompt/prompt-executor/prompt-executor-clients/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/LLMClient.kt
```diff
@@ -0,0 +1,102 @@
...
+     * @return List of response messages
+     */
+    @JavaAPI
+    @JvmOverloads
+    public fun execute(
+        prompt: Prompt,
+        model: LLModel,
+        tools: List<ToolDescriptor> = emptyList(),
+        executorService: ExecutorService? = null
+    ): List<Message.Response> = runOnIOBoundDispatcher { execute(prompt, model, tools) }
```
antoniibelyshev: shouldn't we pass the `executorService` here and in other `runOnIOBoundDispatcher` calls?

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/message/Message.kt
```diff
@@ -120,12 +120,14 @@ public sealed interface Message {
         /**
          * Single content part user message constructor
          */
+        @JvmOverloads
```
antoniibelyshev, line 123: do we need it when there are no default values for the arguments?

antoniibelyshev (approved):
Thank you, looks good! Do we also need to support async embedding methods? If so, can you please add it?

prompt/prompt-executor/prompt-executor-clients/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/LLMClient.kt
Ololoshechkin: Good catch! Thanks) Indeed, we should :)

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/message/Message.kt
Ololoshechkin: For some reason the test was failing with no-class-def without it...

EugeneTheDev (approved):
Please consider using `jvmCommonMain` / `nonJvmCommonMain` to split JVM vs KMP API. This should reduce code duplication

Ololoshechkin:
@EugeneTheDev is it already merged?

EugeneTheDev:
Yes, long time ago #1476

Ololoshechkin:
@EugeneTheDev ah, I see. The infrastructural part is merged, but it wasn't used anywhere so far. 
I'd prefer doing the split separately for all usages, then. Not in this PR -- and for everything at once.

PR #1563, opened 2026-02-27
refactor(rag)!: introduce store-storage separation and interfaces for VectorStore

serge-p7v:
`LongTermMemory` and Spring AI integration require a convenient abstraction for working with vector databases. The initial approach with the single `VectorDatabase` interface was not successful. So in this PR there are two main ideas.
1. Splitting `VectorDatabase` interface into four (for retrieval, ingestion, deletion and reading) for better composability.
2. Refactoring pre-existing code into two layers: the "storage" for low-level document operations and the "store" for higher-level operations that involve embeddings.

BREAKING:
* `RankedDocumentStorage` has become a non-inheriting interface `RetrievalStorage` and it has been moved to a `storage` submodule.

DEPRECATED:
* `rankDocuments` and `mostRelevantDocuments` are deprecated in favor of `search` method from `RetrievalStorage`


serge-p7v:
Screenshot 2026-03-03 at 12 36 13

docs/docs/retrieval-augmented-generation.md
```diff
@@ -0,0 +1,137 @@
+# Retrieval-augmented generation (RAG)
+
+Koog provides building blocks for retrieval-augmented generation (RAG): embedding text, storing embedded documents, and retrieving the most relevant results for a query.
+
+This page focuses on what is available in the current `rag` module and how to use it safely.
+
+## What Koog provides today
```
serge-p7v, line 7: This document is a draft. It should be rewritten when the new API is finalized based on the existing [ranked-document-storage.md](https://github.com/JetBrains/koog/blob/develop/docs/docs/ranked-document-storage.md) guide.

serge-p7v: I had to remove the old md-file, because it caused knit errors.

serge-p7v:
Initially, I added a `FilterExpressionBuilder` for passing metadata filters in `SearchRequest` implementations, but decided to remove it to reduce the size of the PR (otherwise it would be +1k lines more because of filter builder, parser and tests).

For now the "filterExpression" is just a string.

rag/rag-base/src/commonMain/kotlin/ai/koog/rag/base/storage/DeletionStorage.kt
```diff
@@ -0,0 +1,16 @@
...
+ */
+public interface DeletionStorage {
+    /**
+     * Deletes documents with the specified identifiers from the storage.
+     *
+     * @param ids The list of document identifiers to delete.
+     * @param namespace An optional namespace to scope the deletion. If null, the default namespace is used.
+     * @return The list of identifiers that were successfully deleted.
+     */
+    public suspend fun delete(ids: List<String>, namespace: String? = null): List<String>
```
Amaneusz, line 15: I am not sure how we should handle new public interfaces that include `suspend` functions.
If we want to be consistent with what has been done around `PromptExecutor` (see [original PR](https://github.com/JetBrains/koog/pull/1555/changes#diff-a56db53a45f434376c2b0102d8db221932432872c56e7fece2475364417acee7))
we should:
-  rename this interface to `DeletionStorageAPI` 
- add expected abstract class `DeletionStorage` that implements `DeletionStorageAPI`
- add empty implementation of `DeletionStorage` for each platform except jvm where we add an implementation containing blocking version of `delete` method that accepts `executorService` and calls `runOnIODispatcher(executor){ delete(ids, namespace) }`

That's only if we want to be consistent about suspend methods for Java callers.
Maybe we don't need to do that everywhere? IMO we should but I'd like to hear more opinions on the matter.

I recently faced the exact same issue in my [PR with Model Selection](https://github.com/JetBrains/koog/pull/1602) where I have couple of such interfaces. There, I decided to mimic the structure from the `PromptExecutor` & friends.

Amaneusz: This comment holds for all other public interfaces with `suspend` methods - I see you introduced a couple of them

rag/rag-base/src/commonMain/kotlin/ai/koog/rag/base/storage/search/ScoreMetric.kt
```diff
@@ -0,0 +1,14 @@
...
+ * Score metric semantics exposed by storage.
+ */
+public enum class ScoreMetric {
+    COSINE_SIMILARITY,
+    COSINE_DISTANCE,
+    DOT_PRODUCT,
+    EUCLIDEAN_DISTANCE,
+    BM25,
+    HYBRID,
+    CUSTOM
```
Amaneusz, line 52: nit: some docs about how each `ScoreMetric` behaves would be nice

rag/rag-vector/src/commonMain/kotlin/ai/koog/rag/vector/storage/EmbeddingStorage.kt
```diff
@@ -0,0 +1,110 @@
...
+                        document = document,
+                        score = Score(value = similarity, metric = ScoreMetric.COSINE_SIMILARITY)
+                    )
+                )
+            }
+        }
+
+        return results
+            .sortedByDescending { it.score.value }
+            .take(request.limit)
```
Amaneusz: What about `request.offset`? Currently it's being ignored

rag/vector-storage/src/commonMain/kotlin/ai/koog/rag/vector/store/EmbeddingStore.kt
```diff
@@ -0,0 +1,110 @@
...
+
+    /**
+     * Retrieves all documents from an underlying VectorStorage and does similarity search in memory.
+     */
+    override suspend fun search(
+        request: SearchRequest,
+        namespace: String?
+    ): List<SearchResult<Document>> {
+        val queryText = (request as? HasTextQuery)?.queryText
+            ?: error("EmbeddingStore requires a request implementing HasTextQuery")
```
Amaneusz: If we require `HasTextQuery` type maybe we could try to enforce it in compile time with some different typing?
If this wouldn't be a sensible option, we should at least document that the query must be of `HasTextQuery` type

rag/vector-storage/src/commonMain/kotlin/ai/koog/rag/vector/storage/VectorStorage.kt
```diff
@@ -0,0 +1,72 @@
...
+
+    /**
+     * Stores a document along with its pre-computed vector embedding under the specified ID.
+     * If a document with the given ID already exists, it is replaced.
+     *
+     * @param id The unique identifier to assign to the stored document.
+     * @param document The document to store.
+     * @param vector The pre-computed vector embedding for the document.
+     */
+    public suspend fun store(id: String, document: Document, vector: Vector)
```
Amaneusz: Maybe we could add some return type and signal whether the document already existed? This will make the API more consistent (see `delete` return type)

rag/rag-base/Module.md
```diff
@@ -6,78 +6,57 @@ A foundational module that provides core interfaces for document storage and ret
...
+        TextDocument("Another document about machine learning.")
+    )
+
+    // Store the documents and get their IDs
+    val documentIds = ingestion.add(documents)
+    println("Documents stored with IDs: $documentIds")
+
+    // Retrieve the documents using their IDs
+    val retrievedDocuments = reader.get(documentIds)
+    retrievedDocuments.forEach { (id, doc) ->
```
Amaneusz: Will this destructuring work as presented? The `get` returns `List<TextDocument>` not list of pairs (unless we assume TextDocument implements componentX methods, but I haven't seen this class anywhere)

rag/rag-base/Module.md
```diff
@@ -6,78 +6,57 @@ A foundational module that provides core interfaces for document storage and ret
...
+- The `RetrievalStorage` interface that provides ranking capabilities based on query relevance, returning `SearchResult` items with scores
+- The `SearchRequest` interface and `SimilaritySearchRequest` implementation for defining search parameters
+- The `DocumentWithPayload` data class for associating documents with metadata or payload
 - Support for generic document types, allowing flexibility in the types of documents that can be stored and retrieved
 
-This module serves as the base for all RAG submodules (e.g., rag-vector-storage) by providing a consistent API for document operations. It is designed to be implementation-agnostic, allowing different storage backends to be used interchangeably while maintaining a consistent interface for document management and retrieval.
+This module serves as the base for all RAG submodules (e.g., vector-storage) by providing a consistent API for document operations. It is designed to be implementation-agnostic, allowing different storage backends to be used interchangeably while maintaining a consistent interface for document management and retrieval.
 
 ### Example of usage
```
EugeneTheDev: nit: Move example of usage into `docs` where these snippets would be validated with knit. Otherwise we might forget to update this document when updating the API and these examples might diverge from the actual API

rag/rag-base/src/commonMain/kotlin/ai/koog/rag/base/storage/capability/CapabilityDetails.kt
```diff
@@ -0,0 +1,12 @@
+package ai.koog.rag.base.storage.capability
+
+import ai.koog.rag.base.storage.ScoreMetric
+
+/**
+ * Optional details for capability behavior.
+ */
+public data class CapabilityDetails(
```
EugeneTheDev: Just ouf of curiosity - why does capability struct look like this? Did you came up with these capability properties yourself or is it based on some popular existing approach?

rag/rag-base/src/commonMain/kotlin/ai/koog/rag/base/storage/IngestionStorage.kt
```diff
@@ -0,0 +1,26 @@
+package ai.koog.rag.base.storage
+
+/**
+ * Storage interface that provides the ability to ingest (add and update) documents.
+ *
+ * @param Document The type of the documents being stored.
+ */
+public interface IngestionStorage<Document> {
```
EugeneTheDev: `Ingestion` sounds a bit cumbersome when compared to `Read`. AFAIU `Ingestion` and `Read` should mirror each other. Why not just call `IngestionStorage` `WriteStorage` instead?

rag/rag-base/src/commonMain/kotlin/ai/koog/rag/base/storage/search/ScoreMetric.kt
```diff
@@ -0,0 +1,53 @@
+package ai.koog.rag.base.storage
+
+/**
+ * Score metric semantics exposed by storage.
+ *
+ * Defines how similarity or relevance scores are computed and interpreted when querying a vector or document store.
+ */
+public enum class ScoreMetric {
```
EugeneTheDev, line 8: nit: I like how you grouped capability-related entities under `capability` package. Let's also group `ScoreMetric.kt` , `SearchRequest.kt` and `SearchResult.kt` under `search` package, wdyt?

rag/vector-storage/src/commonMain/kotlin/ai/koog/rag/vector/store/VectorStore.kt
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * Interface for a vector store that combines document ingestion and retrieval.
+ *
+ * This is the primary user-facing abstraction for working with vector-based document storage.
+ * Implementations handle embedding documents into vectors and storing them for similarity-based retrieval.
+ *
+ * @param Document The type representing the document being stored.
+ */
+public interface VectorStore<Document> :
```
EugeneTheDev: General concern regarding naming: `Store` and `Storage` sound too similar, and it's confusing just from looking at these which one is more low-level and which one is a higher-level abstraction. Since this interface implements `...Storage` like interfaces, let's call it `VectorStorage` too. As for the lower-level `VectorStorage`, that only describes how vectors are stored and retrieved, let's come up with a different name that would be distinct enough to avoid confusion

rag/vector-storage/src/commonMain/kotlin/ai/koog/rag/vector/store/EmbeddingStore.kt
```diff
@@ -0,0 +1,111 @@
...
+ *
+ * The embedding step is handled by this class, while the actual persistence is delegated
+ * to the provided [VectorStorage]. This separation allows swapping backends (in-memory,
+ * file-based, real vector database) independently from the embedding model.
+ *
+ * @param Document The type of the document being stored and ranked.
+ * @property embedder A mechanism to generate vector embeddings for documents and queries.
+ * @property storage Underlying storage backend to hold documents and their corresponding vector embeddings.
+ */
+public open class EmbeddingStore<Document>(
```
EugeneTheDev: Why are this class and its implementations are `open`? I don't see any method overrides, maybe we should keep it closed?

rag/rag-base/src/commonMain/kotlin/ai/koog/rag/base/storage/search/ScoreMetric.kt
serge-p7v: Good idea, thanks! Added javadoc.

rag/rag-vector/src/commonMain/kotlin/ai/koog/rag/vector/storage/EmbeddingStorage.kt
serge-p7v: Good catch! Fixed:
```
.drop(request.offset)
.take(request.limit)
```

rag/vector-storage/src/commonMain/kotlin/ai/koog/rag/vector/storage/VectorStorage.kt
serge-p7v: Done

rag/rag-base/Module.md
serge-p7v: No, it won't 🙂 Thanks, fixed.

EugeneTheDev (changes requested):
The PR looks good overall, I only have some more minor questions regarding the naming. 

Also, let's rename the `vector-storage` module to the `rag-vector`, to be more consistent with `rag-base`?. Also, can we rename `rag-base` to `rag-core`, again, for better consistency with how our other modules are structured and called?

rag/rag-vector/src/commonMain/kotlin/ai/koog/rag/vector/storage/EmbeddingStorage.kt
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    /**
+     * Retrieves all documents from an underlying VectorStorage and does similarity search in memory.
+     */
+    override suspend fun search(
+        request: SearchRequest,
+        namespace: String?
+    ): List<SearchResult<Document>> {
+        require(request is SimilaritySearchRequest) {
```
EugeneTheDev: Sorry, missed this part. Runtime validation is indeed not the best choice if we can validate something at compile time. Can you make `SearchRequest` a generic parameter in the base `RetrievalStorage<Document, in TRequest : SearchRequest>` instead? And then each implementation can specify compatible request types explicitly, to validate these at compile time. If there are other similar places, can you take a look there too?

serge-p7v: Real vector databases support different types of search requests at the same: similarity, keyword and hybrid. And I think it's possible to "unlock" new search types in runtime by modifying indexes in vector databases. Also, users might want to define their own search requests (e.g. `VectorSearchRequest`).
So it's not easy to enforce that at compile-time. The `CapabilityAwareStorage` interface should help with checking which search types are available for callers: KEYWORD_SEARCH, SIMILARITY_SEARCH, HYBRID_SEARCH.

rag/vector-storage/src/commonMain/kotlin/ai/koog/rag/vector/store/EmbeddingStore.kt
serge-p7v: Because this class (and the original version of this class) supports only similarity search I switched to runtime type check without `HasTextQuery`. It would be great to enforce the type at compile-time, but here is the problem:
Real vector databases support different types of search requests at the same: similarity, keyword and hybrid. And I think it's possible to "unlock" new search types in runtime by modifying indexes in vector databases. Also, users might want to define their own search requests (e.g. VectorSearchRequest).
So it's not easy to enforce that at compile-time. The `CapabilityAwareStorage` interface should help with checking which search types are available for callers: KEYWORD_SEARCH, SIMILARITY_SEARCH, HYBRID_SEARCH.

rag/rag-vector/src/commonMain/kotlin/ai/koog/rag/vector/storage/EmbeddingStorage.kt
serge-p7v: I've added a generic type.
And for databases that definitely support multiple search types it will be possible to define:
`public interface MultiSearchRequest : SearchRequest`
Then to define custom search types if needed:
```
data class OneSearchRequest() : MultiSearchRequest
data class AnotherSearchRequest() : MultiSearchRequest
```
And use `MultiSearchRequest` instead of `SimilaritySearchRequest`.

rag/vector-storage/src/commonMain/kotlin/ai/koog/rag/vector/store/VectorStore.kt
serge-p7v: True, thanks. I added `VectorStorageBackend` instead of low-level `VectorStorage` and renamed `VectorStore` to `VectorStorage`.

rag/rag-base/src/commonMain/kotlin/ai/koog/rag/base/storage/search/ScoreMetric.kt
serge-p7v: Done!

rag/vector-storage/src/commonMain/kotlin/ai/koog/rag/vector/store/EmbeddingStore.kt
serge-p7v: Screenshot 2026-03-24 at 13 03 24
Unfortunately, I can't make it open while preserving the current hierarchy.

rag/rag-base/src/commonMain/kotlin/ai/koog/rag/base/storage/IngestionStorage.kt
serge-p7v: `IngestionStorage` is a `RetrievalStorage` counterpart, its name even has the same length 🙂 I think `ReadStorage` (and `WriteStorage`) is a bit generic, but I couldn't find a better name than `ReadStorage`.

rag/rag-base/src/commonMain/kotlin/ai/koog/rag/base/storage/capability/CapabilityDetails.kt
serge-p7v: Ah, it was an attempt to give more runtime information. I've removed it for now; when we have a real vector database implementation (such as pgvector) we can re-introduce it.

rag/rag-base/Module.md
serge-p7v: Good idea, thanks! The file in docs is also updated.

rag/vector-storage/src/commonMain/kotlin/ai/koog/rag/vector/store/EmbeddingStore.kt
serge-p7v: Now it accepts only SimilaritySearchRequest at compile time:
```
public open class EmbeddingStorage<Document>(
    private val embedder: DocumentEmbedder<Document>,
    private val storage: VectorStorageBackend<Document>
) : VectorStorage<Document, SimilaritySearchRequest>, CapabilityAwareStorage {
```

rag/rag-base/src/commonMain/kotlin/ai/koog/rag/base/storage/IngestionStorage.kt
EugeneTheDev: What does `ReadStorage` do and why there's no `WriteStorage`?

rag/vector-storage/src/commonMain/kotlin/ai/koog/rag/vector/store/VectorStore.kt
EugeneTheDev: TBH I don't like `StorageBackend` either, it's still a bit confusing and more cumbersome now, but I don't have any better ideas, so let it be

rag/rag-base/src/commonMain/kotlin/ai/koog/rag/base/storage/IngestionStorage.kt
EugeneTheDev: Just to clarify, it looks a bit confusing that write operations are handled in a single `IngestionStorage` interface, but read operations are split in two: `ReadStorage` and `IngestionStorage`

rag/vector-storage/src/commonMain/kotlin/ai/koog/rag/vector/store/EmbeddingStore.kt
EugeneTheDev: Sorry, missed it. Does it make sense then to make the methods `open` too, for a better extensibility?

rag/rag-base/src/commonMain/kotlin/ai/koog/rag/base/storage/DeletionStorage.kt
serge-p7v: Great point! Java API will be added later (when we pick the best approach), for now this PR  is not compatible with Java unfortunately.

rag/rag-base/src/commonMain/kotlin/ai/koog/rag/base/storage/IngestionStorage.kt
serge-p7v: `ReadStorage` is separate, because it most likely not very needed for vector databases, for example [VectorStore from Spring AI](https://docs.spring.io/spring-ai/docs/current/api/org/springframework/ai/vectorstore/VectorStore.html) has no getById methods.
Perhaps, LookupStorage, SearchStorage, WriteStorage, DeletionStorage would be a more descriptive combination of interfaces.

EugeneTheDev: Sounds decent enough, let's keep these names for now

EugeneTheDev (approved):

rag/vector-storage/src/commonMain/kotlin/ai/koog/rag/vector/store/EmbeddingStore.kt
serge-p7v: You're absolutely right! Fixed.

PR #1557, opened 2026-02-27
test!: upgrade JVM target, JUnit, and replace runBlocking with suspend

aozherelyeva:
1. Update JVM target and compatibility to Java 17 ([necessary](https://docs.junit.org/6.0.3/overview.html#java-versions) for JUnit 6);
2. Upgrade Java version in the CI checks to 21 (apparently required by one of the dependencies);
3. Upgrade JUnit to 6.0.3 (the latest version);
4. Replace `runBlocking` in tests with `suspend fun` where appropriate, delete `runBlocking` in other places;
5. Fix the persistency test by explicitly setting `automatic checkpoints = false`.

BREAKING:
* Update JVM target and compatibility to Java 17

DEPRECATED:
* none

PR #1570, opened 2026-03-02
deps!: Update min Java version to 11->17. Use JDK 21 on CI instead of JDK 17

aozherelyeva:
Update workflows and configurations to use Java 21 instead of Java 17. 
It's a prequel for the upcoming update to JUnit 6.0.3 (see #1557).

BREAKING:
* Update min Java version from 11 to 17; See #571
* Use JDK 21 on CI instead of JDK 17.

DEPRECATED:
* none


kpavlov (approved):
LGTM

Let's also update readme / docs with min JVM version

PR #1573, opened 2026-03-03
build: dispatch 'update-package' on koog-spm

Amaneusz:
Extends `build-xcframework.yml` to support the full SPM release flow — building the XCFramework, uploading it to the GitHub release, and triggering the `koog-spm` repository to update `Package.swift`.

Target workflow in `koog-spm`: https://github.com/JetBrains/koog-spm/blob/main/.github/workflows/update-package.yml
Github App used for authenticating the dispatch event: https://github.com/settings/apps/koog-spm-release

## Changes

**Structured manual trigger**
`workflow_dispatch` now accepts a `target` input (`artifact` / `release`, default `artifact`). `artifact` preserves the existing debug behavior (30-day artifact upload). `release` runs the full release flow and requires triggering from a version tag ref — the version is derived from `github.ref_name`, preventing version/code mismatch.

**Manual release trigger**
It's possible to run the workflow manually in "production mode" (releasing xcf and dispatching update-publish on koog-spm). This can be done only from ref tags.

**Replaced `softprops/action-gh-release` with `gh release upload`**
Eliminates the risk of accidentally creating a new release or tag — `gh release upload` is upload-only and targets the release explicitly by tag.

**`Dispatch koog-spm release` step**
After uploading the XCF zip, fires a `repository_dispatch` event to `koog-spm` passing `version`, `url`, and `checksum`. Requires `KOOG_SPM_TOKEN` secret — GH App token with write access to `koog-spm`.

⚠️  The [Koog SPM release App](https://github.com/settings/apps/koog-spm-release) is not yet transferred to JB and it's not installed on `koog-spm` - so if we don't get it in time of the next release, the combined workflow may fail - that is ok. We can always run manually once we have everything in place.

### DIFF
--- a/.github/workflows/build-xcframework.yml
+++ b/.github/workflows/build-xcframework.yml
@@ -54,13 +54,28 @@ jobs:
           echo "target=$TARGET" >> $GITHUB_OUTPUT
           echo "Using version: $VERSION, target: $TARGET"
 
-      - name: Verify koog-spm token
+      - name: Verify koog-spm app credentials
         if: steps.resolve.outputs.target == 'release'
         run: |
-          if [ -z "${{ secrets.KOOG_SPM_TOKEN }}" ]; then
-            echo "KOOG_SPM_TOKEN is required for release dispatch but is not set." >&2
+          if [ -z "${{ vars.KOOG_SPM_APP_ID }}" ]; then
+            echo "KOOG_SPM_APP_ID is required for release dispatch but is not set." >&2
             exit 1
           fi
+          if [ -z "${{ secrets.KOOG_SPM_APP_KEY }}" ]; then
+            echo "KOOG_SPM_APP_KEY is required for release dispatch but is not set." >&2
+            exit 1
+          fi
+
+      - name: Generate koog-spm app token
+        if: steps.resolve.outputs.target == 'release'
+        id: app-token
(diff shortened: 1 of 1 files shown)


.github/workflows/build-xcframework.yml
```diff
@@ -25,16 +33,47 @@ jobs:
...
+          echo "Using version: $VERSION, target: $TARGET"
+
+      - name: Verify koog-spm app credentials
+        if: steps.resolve.outputs.target == 'release'
+        run: |
+          if [ -z "${{ vars.KOOG_SPM_APP_ID }}" ]; then
+            echo "KOOG_SPM_APP_ID is required for release dispatch but is not set." >&2
+            exit 1
+          fi
+          if [ -z "${{ secrets.KOOG_SPM_APP_KEY }}" ]; then
```
devcrocod, line 64: I think it’s better not to use the secret directly, but to use via `env` instead. It’s safer and easier to manage

devcrocod (approved):
LGTM
Please note my comment

.github/workflows/build-xcframework.yml
Amaneusz: Applied!

Rizzen (approved):

PR #1583, opened 2026-03-04
feat(prompt): support cache blocks in bedrock converse API calls

nick-sparks:
This PR introduces a `CacheControl` property on Assistant, User and System messages within the Prompt; and integrates it in the Bedrock Converse API.

There appear to be several types of prompt caching supported by providers today:
1. Explicit cache blocks, as implemented here.
2. Explicit-but automatic caching, as a top level parameter on the request itself.
3. Implicit caching.
4. A stand-alone caching API, whose response you refer to later.

Things to consider:
* What is the best way to communicate the above differences - an `LLLMCapability`, an info log or is a silent ignore okay?
* I've put the cached tokens data in the metadata block for now rather than as a first-class property. I did this largely as different providers communicate it with different levels of granularity (i.e. `totalCachedTokens` rather than splitting into read and write).

### DIFF
--- a/agents/agents-tools/src/commonMain/kotlin/ai/koog/agents/core/tools/ToolDescriptor.kt
+++ b/agents/agents-tools/src/commonMain/kotlin/ai/koog/agents/core/tools/ToolDescriptor.kt
@@ -1,5 +1,7 @@
 package ai.koog.agents.core.tools
 
+import ai.koog.prompt.message.CacheControl
+
 /**
  * Represents a descriptor for a tool that contains information about the tool's name, description, required parameters,
  * and optional parameters.
@@ -10,12 +12,14 @@ package ai.koog.agents.core.tools
  * @property description The description of the tool.
  * @property requiredParameters A list of ToolParameterDescriptor representing the required parameters for the tool.
  * @property optionalParameters A list of ToolParameterDescriptor representing the optional parameters for the tool.
+ * @property cacheControl Optional cache control to apply after this tool definition.
  */
 public open class ToolDescriptor(
     public val name: String,
     public val description: String,
     public val requiredParameters: List<ToolParameterDescriptor> = emptyList(),
     public val optionalParameters: List<ToolParameterDescriptor> = emptyList(),
+    public val cacheControl: CacheControl? = null,
 ) {
     /**
      * Creates a copy of the current ToolDescriptor with the option to modify specific attributes.
@@ -26,6 +30,7 @@ public open class ToolDescriptor(
      * Defaults to the current required parameters if not provided.
      * @param optionalParameters A list of ToolParameterDescriptor representing the optional parameters for the tool.
      * Defaults to the current optional parameters if not provided.
+     * @param cacheControl Optional cache control to apply after this tool definition.
      * @return A new instance of ToolDescriptor with the updated attributes.
      */
     public fun copy(
@@ -33,12 +38,14 @@ public open class ToolDescriptor(
         description: String = this.description,
         requiredParameters: List<ToolParameterDescriptor> = this.requiredParameters.toList(),
         optionalParameters: List<ToolParameterDescriptor> = this.optionalParameters.toList(),
+        cacheControl: CacheControl? = this.cacheControl,
     ): ToolDescriptor {
         return ToolDescriptor(
             name = name,
             description = description,
             requiredParameters = requiredParameters,
             optionalParameters = optionalParameters,
+            cacheControl = cacheControl,
         )
     }
 
@@ -50,6 +57,7 @@ public open class ToolDescriptor(
         if (description != other.description) return false
         if (requiredParameters != other.requiredParameters) return false
         if (optionalParameters != other.optionalParameters) return false
+        if (cacheControl != other.cacheControl) return false
 
         return true
     }
@@ -74,6 +82,8 @@ public open class ToolDescriptor(
         appendParameters(optionalParameters)
         appendLine("  ]")
 
+        appendLine("  cacheControl=$cacheControl")
+
         append(")")
     }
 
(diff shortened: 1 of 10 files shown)


prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/dsl/PromptBuilder.kt
```diff
@@ -191,11 +199,12 @@ public class PromptBuilder internal constructor(
...
      *
+     * @param cacheControl Optional cache control to apply after this tool definition.
      * @param init The initialization block for the TextContentBuilder
      */
     @JavaAPI
-    public fun assistant(init: TextContentBuilder.() -> Unit): PromptBuilder = apply {
-        assistant(TextContentBuilder().apply(init).build())
+    public fun assistant(cacheControl: CacheControl? = null, init: TextContentBuilder.() -> Unit): PromptBuilder = apply {
+        assistant(TextContentBuilder().apply(init).build(), cacheControl)
     }
```
Rizzen: This one looks like breaking change for Java. As this param is optional I'd prefer to not to touch that method but introduce overload so we'd have two methods:
```
public fun assistant(cacheControl: CacheControl? = null, init: TextContentBuilder.() -> Unit)
public fun assistant(init: TextContentBuilder.() -> Unit)
```

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/dsl/PromptBuilder.kt
```diff
@@ -76,11 +80,12 @@ public class PromptBuilder internal constructor(
      * }
      * ```
      *
+     * @param cacheControl Optional cache control to apply after this tool definition.
      * @param init The initialization block for the TextContentBuilder
      */
     @JavaAPI
-    public fun system(init: TextContentBuilder.() -> Unit): PromptBuilder = apply {
-        system(TextContentBuilder().apply(init).build())
+    public fun system(cacheControl: CacheControl? = null, init: TextContentBuilder.() -> Unit): PromptBuilder = apply {
```
Rizzen: Same as previous, and few more places. We want to avoid breaking java code using Prompt builder, so I suggest to introduce overloads with CacheControl

Rizzen (changes requested):

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/dsl/PromptBuilder.kt
antoniibelyshev: can't we just use @JvmOverloads here, without manually adding the overload?

agents/agents-tools/src/commonMain/kotlin/ai/koog/agents/core/tools/ToolDescriptor.kt
```diff
@@ -10,12 +13,14 @@ package ai.koog.agents.core.tools
...
  * @property requiredParameters A list of ToolParameterDescriptor representing the required parameters for the tool.
  * @property optionalParameters A list of ToolParameterDescriptor representing the optional parameters for the tool.
+ * @property cacheControl Optional cache control to apply after this tool definition.
  */
 public open class ToolDescriptor(
     public val name: String,
     public val description: String,
     public val requiredParameters: List<ToolParameterDescriptor> = emptyList(),
     public val optionalParameters: List<ToolParameterDescriptor> = emptyList(),
+    public val cacheControl: CacheControl? = null,
```
antoniibelyshev, line 22: I think it should not be a part of the tool descriptor, it looks like a part of agent / client config. Is there a need for having cache control on a specific tool? Can we just have a flag `cacheTools` in the client?

prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/bedrock/converse/BedrockConverseConverters.kt
```diff
@@ -68,6 +75,20 @@ internal object BedrockConverseConverters {
         explicitNulls = false
     }
 
+    private fun Duration.toBedrockCacheTTL(): CacheTtl = when {
```
antoniibelyshev: let's have a `sealed interface BedrockCacheControl` with two implementations: `FiveMinutes` and `OneHour`, otherwise it's very easy to get a runtime exception. If you want to allow extending the cache control for other llm providers in the future, you can also do `sealed interface CacheControl`, `sealed interface Bedrock : CacheControl`, and define the available controls inside.

antoniibelyshev (changes requested):

agents/agents-tools/src/commonMain/kotlin/ai/koog/agents/core/tools/ToolDescriptor.kt
nick-sparks: You can technically put a cache block after any tool in the request to Bedrock, so this supports that. I imagine most use-cases will have it right at the end, but there is an argument that the flexibility to put it after any tool in the block means you could have a set of "core" tools that can be cached, and "dynamic" tools that you can add when you detect that you need them.

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/dsl/PromptBuilder.kt
nick-sparks: JVMOverloads would require a default for `init` I believe. I will add explicit overloads.

agents/agents-tools/src/commonMain/kotlin/ai/koog/agents/core/tools/ToolDescriptor.kt
antoniibelyshev: Ok, I see. Then can we specify the tool caching by passing something like `toolsCacheControl: Map<ToolDescriptor, CacheControl.Bedrock>` to the client? It looks to me like a more natural place to know about the tool caching.

nick-sparks: I think this API closely follows the API for message-level caching so it represents a more natural place to look for it. My other concern with that is that some LLM providers also have explicit caching at the request level - which is where I think your suggestion fits more naturally.

antoniibelyshev (approved):
Thank you, Looks good!

Rizzen (approved):

PR #1585, opened 2026-03-04
test: fix integration-tests CI workflow

aozherelyeva:
@antoniibelyshev fairly noted that we have a passed test in CI even if it fails locally. Fixed it by tinkering the config of the false-positive jobs.

BREAKING:
* none

DEPRECATED:
* none


antoniibelyshev (approved):
Thanks!

PR #1586, opened 2026-03-05
test: no duplicated messages in the history when calling requestLLMForceOneTool

aozherelyeva:
Add integration test for `requestLLMForceOneTool` to avoid duplicated messages.
A follow-up to the fix for [KG-663](https://youtrack.jetbrains.com/issue/KG-663) Extra messages in the history when calling requestLLMForceOneTool.

BREAKING:
* none

DEPRECATED:
* none

### DIFF
--- a/integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/agent/AIAgentIntegrationTest.kt
+++ b/integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/agent/AIAgentIntegrationTest.kt
@@ -44,6 +44,7 @@ import io.kotest.inspectors.shouldForAny
 import io.kotest.matchers.booleans.shouldBeTrue
 import io.kotest.matchers.collections.shouldBeEmpty
 import io.kotest.matchers.collections.shouldContain
+import io.kotest.matchers.collections.shouldHaveSize
 import io.kotest.matchers.collections.shouldNotBeEmpty
 import io.kotest.matchers.comparables.shouldBeLessThan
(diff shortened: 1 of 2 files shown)


integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/agent/AIAgentIntegrationTest.kt
```diff
@@ -1342,7 +1343,58 @@ class AIAgentIntegrationTest : AIAgentTestBase() {
...
+                    ),
+                    toolRegistry = ToolRegistry {
+                        tool(testTool)
+                    },
+                    installFeatures = {
+                        install(EventHandler.Feature, eventHandlerConfig)
+                    }
+                )
+
+                agent.run("Calculate 5 times 3") should contain("successfully")
```
kpavlov, line 1392: Shouldn't it contain "15"?

kpavlov, line 1392:
```suggestion
                agent.run("Calculate 5*3.Return only answer as integer") should contain("15")
```

antoniibelyshev: actually, no, since the output of the strategy is hard-coded: `"Tool call completed successfully without duplication"`. I'd say, assertion here is irrelevant, and it's better to remove it.

antoniibelyshev (approved):
Thank you, looks good!

PR #1590, opened 2026-03-06
fix(prompt): handle Dashscope streaming tool call argument merging

### LINKED ISSUE #1309: Questions about DashscopeLLMClient Streaming ToolCall
When using DashscopeLLMClient to execute [StreamingAgentWithTools](https://github.com/JetBrains/koog/tree/develop/examples/simple-examples/src/main/kotlin/ai/koog/agents/example/streaming)，

Image

An anomaly has occurred:

`Caused by: kotlinx.serialization.json.internal.JsonDecodingException: Unexpected JSON token at offset 0: Cannot read Json element because of unexpected end of the object '}' at path: $
JSON input: }`

After investigation, two potential issues were found in the code:

1. When the returned id is "", appendArgumentsDelta should be executed.but it is not.

  Image

Image

  
2. When function.arguments is null, the generated JSON has issues.

Image

Image

================================================================================
Since DashscopeLLMClient is final, it cannot be inherited.The only option is to write a similar "repeated" client, mimicking DashscopeLLMClient.  

I am puzzled, is there an issue with the data format returned by aliyun? Or do we need to adjust upsertToolCall? Thanks:)

devcrocod:
Fix streaming tool call handling for DashscopeLLMClient

fixes #1309 [KG-658](https://youtrack.jetbrains.com/issue/KG-658)


tiginamaria (approved):
Thank you for fixing! Looks good now

PR #1600, opened 2026-03-10
feat(agents): Java-friendly API for AIAgentStorage

EugeneTheDev:
Split `AIAgentStorage` into JVM/non-JVM source sets, provide Java-friendly methods

Also, added Claude skill to automated class splitting into JVM/non-JVM.


agents/agents-core/src/jvmCommonMain/kotlin/ai/koog/agents/core/agent/entity/AIAgentStorage.kt
```diff
@@ -0,0 +1,72 @@
...
+        delegate = AIAgentStorageImpl()
+    )
+
+    internal actual suspend fun copy(): AIAgentStorage {
+        return AIAgentStorage(delegate = delegate.copy())
+    }
+
+    @JavaAPI
+    @JvmName("set")
+    public fun <T : Any> setBlocking(key: AIAgentStorageKey<T>, value: T): Unit = runBlocking {
```
kpavlov: It diverges from the established pattern. The very same module contains CoroutineUtils.runBlockingIfRequired (in agents-core/src/jvmMain/) which exists explicitly to bridge Java callers into the coroutine world with deadlock prevention. The LLMClient and AIAgentStrategy blocking APIs in the same codebase use either runOnIOBoundDispatcher (with Dispatchers.IO) or runBlockingIfRequired — never bare runBlocking. A new reader of this code will not understand why the storage API is different.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/entity/AIAgentStorageImpl.kt
```diff
@@ -0,0 +1,46 @@
...
+    override suspend fun <T : Any> set(key: AIAgentStorageKey<T>, value: T): Unit = mutex.withLock {
+        storage[key] = value
+    }
+
+    @Suppress("UNCHECKED_CAST")
+    override suspend fun <T : Any> get(key: AIAgentStorageKey<T>): T? = mutex.withLock {
+        storage[key] as T?
+    }
+
+    override suspend fun <T : Any> getValue(key: AIAgentStorageKey<T>): T {
```
kpavlov, line 26: get(key) acquires mutex.withLock, releases it, then getValue checks the result. Between the two calls the lock is not held. A concurrent remove between get returning a non-null value and getValue returning it cannot occur (the ?: is pure Kotlin, not a second lock acquisition), so this specific implementation is safe. But semantically, the pattern is fragile: getValue delegates to get which holds the lock only for its own call. If getValue is ever refactored to re-read the key after the null check, there would be a real TOCTOU window.

Would it be safer to use:
```kotlin
override suspend fun <T : Any> getValue(key: AIAgentStorageKey<T>): T = mutex.withLock {
    @Suppress("UNCHECKED_CAST")
    storage[key] as T? ?: throw NoSuchElementException("Key $key not found in storage")
}
```

agents/agents-core/src/nonJvmCommonMain/kotlin/ai/koog/agents/core/agent/entity/AIAgentStorage.kt
```diff
@@ -0,0 +1,15 @@
+@file:Suppress("MissingKDocForPublicAPI", "EXPECT_ACTUAL_CLASSIFIERS_ARE_IN_BETA_WARNING")
```
kpavlov, line 1: Would it make sense to create KDoc instead of suppressing?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/entity/AIAgentStorage.kt
kpavlov, line 23: createStorageKey is now duplicated — one top-level, one companion

Files:
* commonMain/…/AIAgentStorage.kt:23: public fun <T : Any> createStorageKey(name: String)
* jvmCommonMain/…/AIAgentStorage.kt:70: AIAgentStorage.createStorageKey(name)

agents/agents-core/src/jvmCommonMain/kotlin/ai/koog/agents/core/agent/entity/AIAgentStorage.kt
EugeneTheDev: Good catch, thanks

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/entity/AIAgentStorageImpl.kt
EugeneTheDev: Thanks, will fix

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/entity/AIAgentStorage.kt
EugeneTheDev: This is expected, we want to provide top-level function for Kotiln and static function for Java. Since it's a one-liner, there's no need to delegate one to another

agents/agents-core/src/nonJvmCommonMain/kotlin/ai/koog/agents/core/agent/entity/AIAgentStorage.kt
EugeneTheDev: They are not needed here, because these are actual impelmentaiton that inheirit their docs from expect declarations

agents/agents-core/src/jvmCommonMain/kotlin/ai/koog/agents/core/agent/entity/AIAgentStorage.kt
EugeneTheDev: I fixed it by reusing `runBlockingIfRequired`. To access it from storage, I had to move `jvmMain` contents to `jvmCommonMain`. I didn't not change the files itself. Unfortunately, GitHub failed to recognize this as just a move and shows a giant diff now, when in fact this is not true

EugeneTheDev: Oh, my bad, I forgot to remove `jvmMain/kotlin`. Now the diff looks better

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/entity/AIAgentStorageImpl.kt
```diff
@@ -0,0 +1,46 @@
...
+    private val mutex = Mutex()
+    private val storage = mutableMapOf<AIAgentStorageKey<*>, Any>()
+
+    internal suspend fun copy(): AIAgentStorageImpl {
+        val newStorage = AIAgentStorageImpl()
+        newStorage.putAll(this.toMap())
+        return newStorage
+    }
+
+    @Suppress("UNCHECKED_CAST")
```
Amaneusz, line 16: This annotation can be removed :)

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/entity/AIAgentStorage.kt
Amaneusz: Agreed as long as it stays a one-liner (== we don't expect storage key generation to get more complex)

kpavlov (approved):
Thank you, LGTM

Amaneusz (approved):

.claude/skills/split-jvm-nonjvm/SKILL.md
```diff
@@ -0,0 +1,236 @@
...
+Third, the original class becomes `expect` class
+`MyClass` should extend `MyClassAPI` and override all public symbols from `MyClassAPI`, without method bodies and property values, since `expect` classes can't have any.
+If `MyClass` had any internal symbols, these should also be listed in the `expect` class along with their KDocs.
+If `MyClass` had a KDoc, add it on top of the expect class without parameters/properties description.
+If `MyClass` original KDoc described certain parameters/properties that it accepted in its constructor, add these to KDoc of the expect class secondary constructor.
+This constructor KDoc can't have `@property` tag for parameters, use only `@param` for all parameters, even if the original was `@property`.
+
+```kt
+// Required to put on top of the file containing expect classes to suppress IntelliJ warning about the experimental feature
+@file:Suppress("EXPECT_ACTUAL_CLASSIFIERS_ARE_IN_BETA_WARNING")
```
kpavlov, line 102: Could it be a [kotlin compiler flag](https://kotlinlang.org/docs/multiplatform/multiplatform-expect-actual.html#expected-and-actual-classes)?

```
kotlin {
    compilerOptions {
        // Common compiler options applied to all Kotlin source sets
        freeCompilerArgs.add("-Xexpect-actual-classes")
    }
}
```

.claude/skills/split-jvm-nonjvm/SKILL.md
```diff
@@ -0,0 +1,236 @@
...
+```
+
+Third, the original class becomes `expect` class
+`MyClass` should extend `MyClassAPI` and override all public symbols from `MyClassAPI`, without method bodies and property values, since `expect` classes can't have any.
+If `MyClass` had any internal symbols, these should also be listed in the `expect` class along with their KDocs.
+If `MyClass` had a KDoc, add it on top of the expect class without parameters/properties description.
+If `MyClass` original KDoc described certain parameters/properties that it accepted in its constructor, add these to KDoc of the expect class secondary constructor.
+This constructor KDoc can't have `@property` tag for parameters, use only `@param` for all parameters, even if the original was `@property`.
+
+```kt
```
kpavlov, line 100:
```suggestion
```kotlin
```

.claude/skills/split-jvm-nonjvm/SKILL.md
```diff
@@ -0,0 +1,236 @@
...
+commonMain/
+    kotlin/
+        com/example/
+            MyClass.kt
+            MyClassAPI.kt
+            MyClassImpl.kt
+jvmCommonMain/
+    kotlin/
+        com/example/
+            MyClass.kt
```
kpavlov, line 150: Maybe we should use jvm suffix in filename?
```suggestion
            MyClass.jvm.kt
```

PR #1604, opened 2026-03-10
feat(prompt): Add builders for PromptExecutor for Java

Ololoshechkin:


Ololoshechkin:
@EugeneTheDev could you please take a look

agents/agents-core/src/jvmMain/kotlin/ai/koog/agents/core/agent/config/AIAgentConfig.kt
```diff
@@ -173,7 +172,10 @@ public actual class AIAgentConfig actual constructor(
              * @throws IllegalStateException if required fields such as prompt or model are null
              */
             public fun build(): AIAgentConfig = AIAgentConfig(
-                model = model,
+                model = model ?: throw IllegalStateException(
```
EugeneTheDev: It's better to fail at compile-time if possible. Let's do the same approach as we have for strategies and nodes builders with `withInput`/`withOutput`. Let `builder` return some other initial builder class that has only `model` method and does not have a `build` method. And then this `method` would return the actual builder with the model preconfigured. This way, we would force users to set the model, otherwise they won't be able to call `build()` and get the final instance

```kt
AIAgentConfig
  .builder() // : AIAgentConfigInitialBuilder - has only `model` and no `build`
  .model(...) // : AIAgentConfigBuilder - the actual builder with the non-null model set and all methods
  .prompt(...)
  ...
  .build()
```

prompt/prompt-executor/prompt-executor-model/src/jvmMain/kotlin/ai/koog/prompt/executor/model/MultiLLMPromptExecutorBuilder.kt
```diff
@@ -0,0 +1,352 @@
...
+     *
+     * @param apiKey The API key for authenticating with the OpenAI API.
+     * @param settings Configuration settings for the OpenAI client. Defaults to [OpenAIClientSettings].
+     * @param baseClient The HTTP client used for API requests. Defaults to a new [HttpClient].
+     * @param clock The clock used for time-related operations. Defaults to [Clock.System].
+     * @param toolsConverter Tool descriptor schema generator for OpenAI. Defaults to [OpenAICompatibleToolDescriptorSchemaGenerator].
+     * @return This builder instance for chaining.
+     */
+    @JvmOverloads
+    public fun openAI(
```
EugeneTheDev, line 99: Do we need to have these dedicated methods for all available clients? This forces you to include all these clients as depenendices in `prompt-executor-model`, which is meant to stay lean, since it's a "model" module. And if I only plan to use OpenAI, for example, I don't want to pull all these other LLM clients and their dependencies in my project. E.g. Bedrock client depends on AWS SDK, and it's not a small dependency (and it has its own transitive dependencies too). So, let's just have an `addClient` or `client`. More lean, easier to maintain

prompt/prompt-executor/prompt-executor-model/src/jvmMain/kotlin/ai/koog/prompt/executor/model/RoutingPromptExecutorBuilder.kt
```diff
@@ -0,0 +1,345 @@
...
+     *
+     * @param apiKey The API key for authenticating with the OpenAI API.
+     * @param settings Configuration settings for the OpenAI client. Defaults to [OpenAIClientSettings].
+     * @param baseClient The HTTP client used for API requests. Defaults to a new [HttpClient].
+     * @param clock The clock used for time-related operations. Defaults to [Clock.System].
+     * @param toolsConverter Tool descriptor schema generator for OpenAI. Defaults to [OpenAICompatibleToolDescriptorSchemaGenerator].
+     * @return This builder instance for chaining.
+     */
+    @JvmOverloads
+    public fun openAI(
```
EugeneTheDev, line 99: The same

EugeneTheDev (changes requested):

prompt/prompt-executor/prompt-executor-model/src/jvmMain/kotlin/ai/koog/prompt/executor/model/MultiLLMPromptExecutorBuilder.kt
Ololoshechkin: Otherwise, how we can provide handy API for building PromptExecutors with popular providers?

EugeneTheDev: That's too big of a trade off. We didn't do this in Kotlin before and it was fine. We can just document available clients rather than pulling all the dependencies every time

Amaneusz: I agree with @EugeneTheDev here.
Having
```
PromptExecutor.builder()
   .client(OpenAIClient(key1))
   .client(OpenAIClient(ky2))
```

is not much worse than
```
PromptExecutor.builder()
   .openAI(key1)
   .openAI(key2)
```

Especially considering the cost. Now you have to import all available client modules just to use PromptExecutor. 
Think how that will affect clients that 
- want to use Koog and have some memory restrictions when it comes to package size (ie mobile)
- don't care about mainstream providers and have some custom hipster model, so they need to implement custom `LLMClient` anyway - why would we pollute their namespace with specific vendors?

From a maintenance perspective: now when adding a new provider, we always need to update the builder code and add a dedicated method for it - minor, but annoying 😄

prompt/prompt-executor/prompt-executor-model/src/jvmMain/kotlin/ai/koog/prompt/executor/model/PromptExecutor.kt
```diff
@@ -115,4 +117,74 @@ public actual abstract class PromptExecutor actual constructor() : PromptExecuto
...
+         *     .addClient(openAIClient)
+         *     .addClient(anthropicClient)
+         *     .build();
+         * ```
+         *
+         * @return A new [InitialMultiLLMPromptExecutorBuilder] instance.
+         */
+        @JvmStatic
+        @JavaAPI
+        public fun builder(): InitialMultiLLMPromptExecutorBuilder = multiExecutorBuilder()
```
Amaneusz: @Ololoshechkin , @EugeneTheDev - let's decide which "builder approach" do we want to use. There are two variants:

**Approach A — Single "automatic" builder** (`PromptExecutor.builder()`)

One entry point that automatically picks the executor type at `build()` time:
- `MultiLLMPromptExecutor` when each provider has exactly one client
- `RoutingLLMPromptExecutor` when any provider has more than one client

```java
PromptExecutor executor = PromptExecutor.builder()
    .addClient(openAIClient)
    .addClient(openAIClientBackup) // triggers routing
    .build();
```

✅ Minimal API — one method to remember
✅ User doesn't need to understand the difference between executor types upfront
❌ Implicit — it's not obvious which executor you'll get without reading the docs
❌ The heuristic (duplicate provider → routing) may surprise users who add a second client not realising it silently upgrades them to an experimental executor type
❌ Routing executor is currently experimental - offering it by the main `builder()` method does enforce something that we declared as opt-in.

**Approach B — Explicit builder per executor** (`PromptExecutor.multiExecutorBuilder()` / `PromptExecutor.routingExecutorBuilder()`)

Two dedicated entry points, one per executor type. `PromptExecutor.builder()` kept as a convenience alias for `multiExecutorBuilder()`.

```java
// Explicit multi-provider setup
PromptExecutor executor = PromptExecutor.multiExecutorBuilder()
    .addClient(openAIClient)
    .addClient(anthropicClient)
    .build();

// Explicit load-balanced setup
PromptExecutor executor = PromptExecutor.routingExecutorBuilder()
    .addClient(openAIClientPrimary)
    .addClient(openAIClientSecondary)
    .build();
```

✅ Explicit — user declares intent upfront, no surprises
✅ Clear boundary between stable (`multiExecutorBuilder`) and experimental (`routingExecutorBuilder`) — easy to enforce or relax later
✅ Easier to drop `@ExperimentalRoutingApi` independently per builder when the time comes
❌ User needs to know the difference between `Multi` and `Routing` before choosing
❌ Slightly larger API surface (though `builder()` alias softens this)

---

My vote is for **Approach B** - I like expressive things but then again, maybe I spent too much time in Java 😅 
That being said, I think **Approach A** is more present in the codebase - I'll be fine with either.

prompt/prompt-executor/prompt-executor-model/src/jvmMain/kotlin/ai/koog/prompt/executor/model/RoutingPromptExecutorBuilder.kt
Amaneusz: fixed :)

prompt/prompt-executor/prompt-executor-model/src/jvmMain/kotlin/ai/koog/prompt/executor/model/PromptExecutor.kt
Amaneusz: I have both ready to go and sitting in different commits, so the workload is none.
A: https://github.com/JetBrains/koog/pull/1604/changes/25e55359c4552fd79d13a8d16803cdc5ca7ea510
B: https://github.com/JetBrains/koog/pull/1604/changes/2f2ac5d807b7d770c77421653d5c63b1cb6a0116

prompt/prompt-executor/prompt-executor-model/src/jvmMain/kotlin/ai/koog/prompt/executor/model/MultiLLMPromptExecutorBuilder.kt
Amaneusz: Aligned code with the approach suggested above

prompt/prompt-executor/prompt-executor-model/src/jvmMain/kotlin/ai/koog/prompt/executor/model/PromptExecutor.kt
Ololoshechkin: A

EugeneTheDev: I like A

prompt/prompt-executor/prompt-executor-model/src/jvmMain/kotlin/ai/koog/prompt/executor/model/MultiLLMPromptExecutorBuilder.kt
EugeneTheDev: We discussed it with Vadim, and it seems it would be the problem with Bedrock client only, since it uses AWS SDK. The rest of the clients perform direct HTTP requests to the LLM provider, so they only have ktor client dependency, which is common anyways among all these clients. 

So, lets implement a middle-ground solution for now. Lets provide all these builder methods per client (openAI, anthropic, etc.) but without bedrock. This would allow to discover major thin clients easilly, while keeping the bulky Bedrock away from the default dependencies

Amaneusz: Ok, will implement 👍 

I'd love to talk about this afterwards, though – I am still slightly skeptical about coupling with specific implementations:
1) The discoverability of clients seems equally convenient in both cases
2) We provide all clients but not really, we have a special case right at the start - the Bedrock is an exception, so we treat it worse than other clients. What if something similar happens to an existing client? Let's say we need some "Google SDK" for Google models and a Google client will become heavy. Do we remove the convenience method – adding more exceptions and violating backward compatibility? 
3) The assumption of an HTTP client as "being used always anyway" might be wrong for some users (local inference can happen without HTTP call – we might end up with JNI-based LLMClient for some mobile apps that will not need Ktor at all)
4) The argument of maintainability for each new provider is still there (even though it is minor)

IMO especially arguments no 2 and 3 are not worth the added value

EugeneTheDev: I agree with you, I also share these concerns. However, I also realize that we want to make Java API as fluent as possible initially, to make more Java developers interested. Considering that, it doesn't seem like too bad of a trade-off. But later on, as we understand the actual usage patterns and adoption from the Java side, we can gradually deprecate and remove these methods. Sure, we can talk about it

prompt/prompt-executor/prompt-executor-model/src/jvmMain/kotlin/ai/koog/prompt/executor/model/PromptExecutor.kt
Amaneusz: Included in https://github.com/JetBrains/koog/pull/1604/commits/0d35c7b298c0a6c29b8b9f2a3de404eeb763acc7

EugeneTheDev (approved):

PR #1617, opened 2026-03-13
refactor(agents)!: Change Java API for graphs according to the agreed design

Ololoshechkin:
Make Java graph API more convenient to use

BREAKING:
* Removed `node`, `subgraph`, `parallel` methods on `AIAgentSubgraphBuilderBase`. Top-level functions with the same name should be used instead.


agents/agents-core/src/androidMain/kotlin/ai/koog/agents/core/agent/entity/AIAgentGraphStrategy.kt
```diff
@@ -0,0 +1,22 @@
+@file:Suppress("EXPECT_ACTUAL_CLASSIFIERS_ARE_IN_BETA_WARNING")
```
kpavlov: Could it be a [kotlin compiler flag](https://kotlinlang.org/docs/multiplatform/multiplatform-expect-actual.html#expected-and-actual-classes)?

```
kotlin {
    compilerOptions {
        // Common compiler options applied to all Kotlin source sets
        freeCompilerArgs.add("-Xexpect-actual-classes")
    }
}

```

EugeneTheDev (approved):

PR #1623, opened 2026-03-13
fix: hardening kotlin compiler checks, fix compile error

kpavlov:
### 1. Enable progressive mode, strict warnings, and additional compiler arguments in Kotlin configuration

  - Set `allWarningsAsErrors=false`, `extraWarnings`, and `progressiveMode` in Kotlin compiler options.
  - Add custom free compiler arguments for enhanced warnings and debugging (`-Xexpect-actual-classes`, `-Xmulti-dollar-interpolation`).

### 2. fix: Compilation Error in AIAgentPipelineImpl 
 
  Root cause: AIAgentPipelineImpl.installSystemFeature() checked when (this) for AIAgentGraphPipeline/AIAgentFunctionalPipeline, but this is always AIAgentPipelineImpl (the delegate), so these checks were always false — a warning promoted to error by -Werror.
Fix applied across 5 files:

  1. AIAgentPipelineAPI.kt — Added prepareFeatures(actualPipeline: AIAgentPipelineAPI) with a default impl that delegates to prepareFeatures()
  2. AIAgentPipelineImpl.kt:
      - prepareFeatures() now calls prepareFeatures(this)
      - New prepareFeatures(actualPipeline) override holds the real logic
      - installFeaturesFromSystemConfig and installSystemFeature now accept and forward actualPipeline
      - when (actualPipeline) replaces when (this) — the always-false checks are gone
  3. commonMain/AIAgentPipeline.kt (expect) — Added prepareFeatures(actualPipeline) declaration
  4. nonJvmCommonMain/AIAgentPipeline.kt — Fixed the broken override: now correctly overrides prepareFeatures() and calls prepareFeatures(this) where this is the concrete subclass

  5. jvmCommonMain/AIAgentPipeline.kt — Added the same actual override suspend fun prepareFeatures() calling prepareFeatures(this)

  Bonus fix: HistoryCompressionStrategiesTest.kt — removed a redundant outer strategy { return ... } wrapper that caused a "return prohibited in expression body" error under -Werror.

### docs: Update footer copyright year and external documentation links 

  - Extend footer copyright to 2026.
  - Adjust `ktor` documentation URL.
  - Add new external documentation links for `kotlinx-schema`.

### fix: Update deprecation for Haiku_3 model and suppress warnings 
  - Mark Haiku_3 as deprecated with replacement pointing to Haiku_4_5.
  - Add `@Suppress("DEPRECATION")` to model lists and version map to resolve warnings.

### fix: Change var to val for codeBlockContent in MarkdownParser 
  - Use immutable `val` instead of `var` for improved code safety and clarity.


convention-plugin-ai/src/main/kotlin/ai.kotlin.dokka.gradle.kts
```diff
@@ -8,7 +8,7 @@ dokka {
             includes.from("Module.md")
 
             pluginsConfiguration.html {
-                footerMessage = "Copyright © 2000-2025 JetBrains s.r.o."
+                footerMessage = "Copyright © 2000-2026 JetBrains s.r.o."
```
EugeneTheDev: Can we just set current year programmatically?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/pipeline/AIAgentPipelineAPI.kt
```diff
@@ -413,6 +413,9 @@ public interface AIAgentPipelineAPI {
     @InternalAgentsApi
     public suspend fun prepareFeatures()
 
+    @InternalAgentsApi
+    public suspend fun prepareFeatures(actualPipeline: AIAgentPipelineAPI): Unit = prepareFeatures()
```
EugeneTheDev: That's a good catch, thank you. However, do we really need two `prepareFeatures` methods? What if we leave the parameterless one, and create a property instead
```kt
protected val actualPipeline: AIAgentPipelineAPI
```

Then, actual pipeline classes would set it to `this` and `prepareFeatures` would use this `actualPipeline`. IMHO this looks less confusing than having two similar methods

EugeneTheDev (changes requested):
Nice fixes, thank you

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/pipeline/AIAgentPipelineAPI.kt
kpavlov: @EugeneTheDev, I agree, it was not the most suitable solution. This is [the fix](https://github.com/JetBrains/koog/pull/1623/changes/34d29a8f7c455b737434c428ce1e13d3d5dbbab8).

kpavlov:
waiting for PR #1653 to go first

kpavlov:
that's too outdated

PR #1626, opened 2026-03-13
feat(prompt): allow customized field name for AdditionalPropertiesFlatteningSerializer

### LINKED ISSUE #1624: OpenRouter's `additionalProperties` is not flattened
Hi,

Since `OpenRouterParams` doesn't provide the way to set reasoning efforts, I tried to use `additionalProperties`. According to what I traced in my IDE, eventually the content of this property should be put under the root object. So in theory, if I wrote:

```
    val parameters = OpenRouterParams(
        temperature = 0.4,
        topP = 0.95,
        additionalProperties = mapOf(
            "reasoning" to JsonObject(mapOf(
                "effort" to JsonPrimitive("high")
            ))
        )
    )
```

And use it to create a prompt and send it, the request body should be:

```
{
  "messages": [
    {
      "role": "system",
      "content": "......"
    },
    {
      "role": "user",
      "content": "......"
    }
  ],
  "model": "openai/gpt-5.1",
  "stream": true,
  "temperature": 0.4,
  "tools": [],
  "top_p": 0.95,
  "reasoning": {
      "effort": "high"
  }
}
```

However, this is what I actually got (find out by using mitmproxy):

```
{
  "messages": [
    {
      "role": "system",
      "content": "......"
    },
    {
      "role": "user",
      "content": "......"
    }
  ],
  "model": "openai/gpt-5.1",
  "stream": true,
  "temperature": 0.4,
  "tools": [],
  "top_p": 0.95,
  "additional_properties": {
    "reasoning": {
      "effort": "high"
    }
  }
}
```

For some reason, the `AdditionalPropertiesFlatteningSerializer` expects the field `additionalProperties`, but for OpenRouter, it's actually `additional_properties`. This mismatch causes the OpenRouter client failed to flatten the additional properties, which blocking me from setting the reasoning effort field.

A screenshot of my IDE:

Image

Also the reason why it becomes snake case:

```
    protected val json: Json = Json {
        ignoreUnknownKeys = true
        isLenient = true
        encodeDefaults = true
        explicitNulls = false
        namingStrategy = JsonNamingStrategy.SnakeCase
    }
```
From `AbstractOpenAILLMClient`

hurui200320:
This PR addresses issue #1624 with the following changes:

+ Instead of hardcoding `additionalPropertiesField` to `additionalProperties`, now it's a parameter of the constructor with a default value. This allows the implementations to define their own additional properties field name, which for OpenAI and its compatible APIs, it's the snake case.

The fix only applies to OpenRouter impl since I'm blocking by this issue.

This PR has no breaking change.

closes #1624

### DIFF
--- a/prompt/prompt-executor/prompt-executor-clients/prompt-executor-openrouter-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openrouter/models/OpenRouterChatCompletion.kt
+++ b/prompt/prompt-executor/prompt-executor-clients/prompt-executor-openrouter-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openrouter/models/OpenRouterChatCompletion.kt
@@ -198,4 +198,8 @@ public class OpenRouterError(
 )
 
 internal object OpenRouterChatCompletionRequestSerializer :
-    AdditionalPropertiesFlatteningSerializer<OpenRouterChatCompletionRequest>(OpenRouterChatCompletionRequest.serializer())
+    AdditionalPropertiesFlatteningSerializer<OpenRouterChatCompletionRequest>(
+        OpenRouterChatCompletionRequest.serializer(),
+        // OpenRouter uses snake case
+        additionalPropertiesField = "additional_properties"
+    )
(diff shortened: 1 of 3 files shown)


hurui200320:
I think it's not fully resolved KG-531. If the JSON contains a field like `top_p`, it will go to the additional properties field instead of the real `topP` field. However, considering the serializer is only used for serialization, mostly for turning a request object into JSON string, I think it's not going to be an issue, or at least the changes in this PR will not cause more damage.

Updated PR description, this PR does not fix KG-531 at all.

kpavlov (approved):
Thank you, @hurui200320

aozherelyeva (approved):
Thanks!

PR #1637, opened 2026-03-16
fix(agents): KG-741. Support functional agent event for strategy handler in Debugger feature

sdubov:
Allow to use Debugger feature with functional agents as well as with graph-based agents. 

- Extended Debugger feature to handle both graph-based and functional strategies by dispatching appropriate strategy starting events;
- Refactored test utilities into TestAgentFactory object for better organization.

closes: KG-741 (https://youtrack.jetbrains.com/issue/KG-741)


agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/debugger/Debugger.kt
```diff
@@ -229,18 +230,25 @@ public class Debugger(public val port: Int, public val awaitInitialConnectionTim
...
+                val strategyName = eventContext.strategy.name
+                val timestamp = pipeline.clock.now().toEpochMilliseconds()
+
+                val event = when (strategy) {
+                    is AIAgentGraphStrategy<*, *> -> {
+                        @OptIn(InternalAgentsApi::class)
+                        val graph = strategy.startNodeToGraph()
+                        GraphStrategyStartingEvent(eventId, executionInfo, runId, strategyName, graph, timestamp)
+                    }
+                    else -> {
```
tiginamaria, line 247: Am I correct that, as for now, we have only two options, so it is not the `when` case but just `if else`?

tiginamaria (approved):
Looks good! Just a small clarification question to make sure we don't miss anything

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/debugger/Debugger.kt
sdubov: You're right. Graph strategy event context provides an info about the graph structure that we add further to the event. We check for graph strategy type here for this purpose. `when` and `if` should work identically here.

PR #1643, opened 2026-03-17
fix: serialization in tools/subgraphs/subtasks, ToolDescriptor generation, simple-examples-java

EugeneTheDev:


examples/simple-examples-java/src/main/java/ai/koog/agents/example/strategies/FunctionalStrategyExample.java
```diff
@@ -1,8 +1,11 @@
-package ai.koog.agents.example.strategies.functional;
```
serge-p7v, line 1: The blogpost refers to this example with "functional" in the path

tiginamaria: If it's a link, it's easy to update in blogpost

PR #1646, opened 2026-03-17
docs(agents): Add java snippets for Graph-based and Planner agents

innateteniuk:


docs/docs/agents/graph-based-agents.md
```diff
@@ -146,45 +200,110 @@ effectively transforming `Message.Assistant` to `String`, because `nodeFinish` e
...
+        .build());
+    calculatorAgentStrategy.edge(AIAgentEdge.builder()
+        .from(nodeSendToolResult)
+        .to(nodeExecuteTool)
+        .onCondition(msg -> msg instanceof Message.Tool.Call)
+        .transformed(msg -> (Message.Tool.Call) msg)
+        .build());
+
+    AIAgent<String, String> mathAgent = AIAgent.builder()
+        .promptExecutor(simpleOllamaAIExecutor("http://localhost:11434"))
```
tiginamaria: nit: Maybe it will be clearer if we create an executor separately from the builder

docs/docs/agents/graph-based-agents.md
```diff
@@ -146,45 +200,110 @@ effectively transforming `Message.Assistant` to `String`, because `nodeFinish` e
...
+    <!--- SUFFIX
+        }
+    }
+    -->
+    ```java
+    var calculatorAgentStrategy = AIAgentGraphStrategy.builder("Simple calculator")
+        .withInput(String.class)
+        .withOutput(String.class);
+
+    var nodeSendInput = AIAgentNode.llmRequest(true, "nodeSendInput");
```
tiginamaria, line 266: nit: I need advice here from someone from the Java community, but now it seems to me that Java developers like to put the explicit type AIAgentNodeBase<String, List<Message.Response>>. Or maybe I'm wrong, let's discuss and decide in the chat after, and then use the same approach in all the examples

tiginamaria (approved):
Thank you! I have only one major future question (but not really important for now) about whether we should or shouldn't put explicit types in Java. Let's discuss it this week

docs/docs/agents/graph-based-agents.md
innateteniuk: Updated as follows
```kotlin
var promptExecutor = PromptExecutor.builder()
        .ollama("http://localhost:11434")
        .build();

AIAgent<String, String> mathAgent = AIAgent.builder()
      .promptExecutor(promptExecutor)
```

tiginamaria:
Thank you! Have to merge it due to the 0.7.1 release tonight!

PR #1650, opened 2026-03-17
refactor(agents): replace API + Impl with one abstract class for functional agent context

antoniibelyshev:
Removed `AIAgentFunctionalContextBaseAPI`, renamed `AIAgentFunctionalContextBaseImpl` -> `AIAgentFunctionalContextBaseCommon`, made it public and abstract, allowing to inherit `AIAgentFunctionalContextBase` from it. That way, we do not have to repeat multiple times the declarations of methods common across targets. Additionally, it allows to define `inline` methods in the common class, rather than having them defined in the actual classes.


agents/agents-core/src/jvmCommonMain/kotlin/ai/koog/agents/core/agent/context/AIAgentFunctionalContextBase.kt
Amaneusz: nit: not the scope of the PR but this doc is cut mid-sentence
> * @return A [Message.Response] object containing the*/

Amaneusz (approved):

PR #1660, opened 2026-03-17
refactor(agents): replace API + Impl with a simpler logic, introduce base agent builder

antoniibelyshev:
Replaced API + Impl pattern with a base class with methods common across targets for agent builders and agent service builders. Introduced `AIAgentBuilderBase` containing fields used in multiple builders (`promptExecutor`, `config`, config elements, etc.), removing code duplication.


agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgentBuilderBase.kt
```diff
@@ -0,0 +1,196 @@
...
+            maxAgentIterations = 50,
+            serializer = serializer
+        ),
+        Clock.System,
+    )
+
+    private companion object {
+        val NoLLMProvider = object : LLMProvider("None", "Provider is not set") {}
+
+        val ModelNotSet = LLModel(
```
kpavlov: Shooldn't we add `@JvmField` for all static fields?

kpavlov (approved):

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgentBuilderBase.kt
antoniibelyshev: I think we do not need it for a private companion object

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgentBuilderBase.kt
```diff
@@ -0,0 +1,196 @@
...
+            id = "model_not_set"
+        )
+    }
+
+    protected abstract fun self(): Self
+
+    protected val validatedPromptExecutor: PromptExecutor
+        get() = requireNotNull(promptExecutor) { "PromptExecutor must be set" }
+
+    protected val validatedConfig: AIAgentConfig
```
Amaneusz: I'm not sure if this remark is about this PR but recently I worked on a PR with similar approach towards builder validation and got this comment: https://github.com/JetBrains/koog/pull/1604#discussion_r2918904640

The main idea:
> It's better to fail at compile-time if possible

In `AIAgentBuilder` there are two required values - if we follow the same path as we did in the PR above, we end up with three classes:
1) `initial builder 1` that has a single method accepting `PromptExecutor` and returns `initial builder 2`
2) `initial builder 2` that requires `PromptExecutor` at construction time and that has a single method accepting `LLModel` that returns the `final builder`
3) The `final" builder` that requires both executor and model at construction time and offers all other non-required setters

--- 

I'm not sure if it would be elegant here as we're playing with abstract type 🤷‍♂️ 
But currently we are incosistent when it comes to builders - maybe it's ok.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgentBuilderBase.kt
```diff
@@ -0,0 +1,196 @@
...
+        get() = when (config.model) {
+            ModelNotSet -> throw IllegalArgumentException("model must be set, plase use .model() on AIAgentBuilder or set AIAgentConfig")
```
Amaneusz: Typo in exception msg: `model()` ➡️ `llmModel()`

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgentBuilderBase.kt
```diff
@@ -0,0 +1,196 @@
...
+     * Sets the `LLModel` instance to be used by the builder.
+     *
+     * This method configures the builder with a specified Large Language Model (LLM),
+     * representing the model's provider, identifier, capabilities, and constraints such as
+     * context length or maximum output tokens.
+     *
+     * @param model The [LLModel] instance representing the large language model to set.
+     * @return The current builder instance for chaining additional configurations.
+     */
+    public fun llmModel(model: LLModel): Self = self().apply {
```
Amaneusz: Not the fault of the PR but the name of this one is funny - `llmModel` has `m` repeated, it's like "large language model model" 🎵

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgentBuilderCommon.kt
```diff
@@ -0,0 +1,143 @@
...
+     *
+     * This method finalizes the current configuration and constructs an AI agent. The agent is
+     * equipped with the specified execution strategy, tool registry, identifier, prompt, language
+     * model, and other optional configurations.
+     *
+     * @return An instance of [AIAgent] with the configured input and output types as `String`.
+     */
+    public fun build(): AIAgent<String, String> {
+        return AIAgent(
+            promptExecutor = requireNotNull(promptExecutor) { "promptExecutor must be set" },
```
Amaneusz: Isn't the `validatedPromptExecutor` doing the null-check?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgentServiceBuilder.kt
```diff
@@ -78,117 +46,18 @@ public class GraphAgentServiceBuilder<Input, Output> internal constructor(
...
+    promptExecutor: PromptExecutor?,
+    toolRegistry: ToolRegistry,
+    config: AIAgentConfig,
     private var featureInstallers: MutableList<FeatureContext.() -> Unit> = mutableListOf(),
+) : AIAgentBuilderBase<GraphAgentServiceBuilder<Input, Output>>(
+    promptExecutor = promptExecutor,
+    toolRegistry = toolRegistry,
+    id = null,
+    config = config,
+    clock = Clock.System,
```
Amaneusz, line 58: `GraphAgentServiceBuilder` and `FunctionalAgentServiceBuilder` extend `AIAgentBuilderBase`, so they inherit all its public setters including `id()` and `clock()`. A caller can write:

```                                                                                                                         
AIAgentServiceBuilder()
      .id("my-service")                                                                                                  
      .clock(testClock)      
      .llmModel(...)                                                                                                     
      .build()
```

This compiles and runs without error. But the `id` and `clock` get overwritten by default values here.
It's more of a design thing - `AIAgentBuilderBase` is meant for building agents, but services share the same methods.
The user ends up with the illusion from our API that they can control `id` and `clock`

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgentBuilderBase.kt
antoniibelyshev: I agree with the idea to have compile errors when a required property is not provided. Let's discuss that on porka, and create a separate pr for that

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgentBuilderBase.kt
antoniibelyshev: true, we have this incosistency in multiple places, but I'd say it's worth updating all at once

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgentServiceBuilder.kt
antoniibelyshev: you are right. I will add a separate `AIAgentServiceBuilder` without id and clock

PR #1662, opened 2026-03-17
dependencies: Bump kotlinx-schema version to 0.4.3

kpavlov:
Bump kotlinx-schema version to 0.4.3

Reflection schema generation fix for java records


EugeneTheDev (approved):
Thanks, will include in 0.7.2 asap

PR #1663, opened 2026-03-17
Cherry-pick #1643 to 0.7.1

EugeneTheDev:


gradle/libs.versions.toml
```diff
@@ -19,7 +19,7 @@ kotlinx-coroutines = "1.10.2"
 kotlinx-datetime = "0.7.1"
 kotlinx-io = "0.7.0"
 kotlinx-serialization = "1.10.0" # check with IJ
-kotlinx-schema = "0.4.1"
+kotlinx-schema = "0.4.2"
```
kpavlov, line 22:
```suggestion
kotlinx-schema = "0.4.3"
```

EugeneTheDev: This will be part of 0.7.2, which I'll prepare and release after 0.7.1 asap. For now, we want to have a working release with all the major changes

PR #1666, opened 2026-03-17
feat(agents): resurrect `LLMDescription.description` ; support `@kotlinx.schema.Description` annotation in reflection schemas 

kpavlov:
## Support `@kotlinx.schema.Description` annotation in reflection schemas, improve documentation links

- Revert `LLMDescription.description` for compatibility, add `effectiveValue`. It was broken in 0.7.x. The `description` parameter removed in 0.7.0 broke existing callers. It is restored as a deprecated field with effectiveValue routing reads through `value.ifBlank { description }.ifBlank { null }`, preserving the migration path without breaking the API surface.
  - Introduced `effectiveValue` property in `LLMDescription` to prioritize `value` over deprecated `description`.
  - Simplified and refactored reflection logic to improve error messaging and validation checks.
  - Updated test cases to cover `effectiveValue` usage and enhanced exceptions for schema generation.
- Support both `@LLMDescription` and `@kotlinx.schema.Description` annotations for reflection function schema generation. Matches `@LLMDescription` first, then falling back to `@Description`.
- Dokka: Added external documentation link for kotlinx-schema.
- Updated Gradle dependencies with kotlinx-schema annotations.


prompt/prompt-structure/src/commonMain/kotlin/ai/koog/prompt/structure/json/generator/JsonSchemaGenerator.kt
```diff
@@ -69,17 +65,21 @@ public abstract class JsonSchemaGenerator {
...
-                .firstOrNull()
-                ?.value
 
-            return elementDescriptionOverride ?: elementDescriptionAnnotation
+            return elementDescriptionOverride ?: getDescriptionFromAnnotations(descriptor.getElementAnnotations(index))
         }
+
+        private fun getDescriptionFromAnnotations(annotations: List<Annotation>): String? = annotations
+            .firstNotNullOfOrNull { annotation ->
+                return when (annotation) {
```
Amaneusz: `firstNotNullOfOrNull` is an inline function, so `return` inside the lambda is a non-local `return` that exits `getDescriptionFromAnnotations` **immediately** — not just the lambda

Dropping the `return` keyword will fix it :)

Amaneusz: Confirmed with silly snippet:
```
        val annotations = listOf(
            object : Annotation {}, // some non-description annotation first
            Description("description"),
            LLMDescription("llm description"),
        )

        val description = getDescriptionFromAnnotations(annotations)
        
```

convention-plugin-ai/src/main/kotlin/ai.kotlin.dokka.gradle.kts
```diff
@@ -44,6 +44,10 @@ dokka {
                 url("https://modelcontextprotocol.github.io/kotlin-sdk/")
                 packageListUrl("https://modelcontextprotocol.github.io/kotlin-sdk/package-list")
             }
+
+            externalDocumentationLinks.register("kotlinx-schema") {
+                url("https://kotlin.github.io/kotlinx-schema/")
```
Amaneusz, line 49: nit: consider adding package list
I confirmed it exists: https://kotlin.github.io/kotlinx-schema/package-list

Amaneusz (approved):
Thanks for the fix! Please see comments - there is one critical thing and one minor nit.
Assuming the critical one will be addressed - approved 👍

convention-plugin-ai/src/main/kotlin/ai.kotlin.dokka.gradle.kts
kpavlov: I thought it takes package list by convention?

prompt/prompt-structure/src/commonMain/kotlin/ai/koog/prompt/structure/json/generator/JsonSchemaGenerator.kt
kpavlov: Added test and fixed, thank you for heads up

agents/agents-tools/src/commonTest/kotlin/ai/koog/agents/core/tools/LLMDescriptionUsageTest.kt
```diff
@@ -103,4 +103,36 @@ class LLMDescriptionUsageTest {
         assertEquals("Nested class property desc", props.getValue("street").description)
         assertEquals("", props.getValue("number").description)
     }
+
+    // -------------------------------------------------------------------------
+    // Tests verifying finding #1: effectiveValue backward-compat and priority
+    // -------------------------------------------------------------------------
+
+    // Finding #1a: the deprecated `description` field is still honoured via effectiveValue.
```
Amaneusz, line 111: Are we sure abiut keeping these comments?

agents/agents-tools/src/commonTest/kotlin/ai/koog/agents/core/tools/LLMDescriptionUsageTest.kt
```diff
@@ -103,4 +103,36 @@ class LLMDescriptionUsageTest {
...
+    fun deprecated_description_field_is_read_via_effective_value() {
+        val desc = getToolDescriptor(typeToken<LegacyAnnotated>(), "legacy")
+        assertEquals("Legacy class description", desc.description)
+        assertEquals(
+            "Legacy property description",
+            desc.requiredParameters.single { it.name == "x" }.description
+        )
+    }
+
+    // Finding #1b: when both `value` and the deprecated `description` are set, `value` wins.
```
Amaneusz, line 128: Potential comment to be removed

prompt/prompt-structure/src/commonTest/kotlin/ai/koog/prompt/structure/json/generator/JsonSchemaGeneratorTest.kt
```diff
@@ -177,12 +196,35 @@ class JsonSchemaGeneratorTest {
         val value: JsonElement? = null
     )
 
+    // Finding #1: empty @LLMDescription (effectiveValue == null) must not block @Description fallback.
```
Amaneusz, line 199: And here

prompt/prompt-structure/src/commonTest/kotlin/ai/koog/prompt/structure/json/generator/JsonSchemaGeneratorTest.kt
```diff
@@ -177,12 +196,35 @@ class JsonSchemaGeneratorTest {
...
+    @LLMDescription
+    @Description("Fallback description from @Description")
+    data class DescriptionFallbackClass(
+        @property:LLMDescription
+        @property:Description("Fallback property description")
+        val prop: String
+    )
+
+    // Finding #2: @LLMDescription must beat @Description regardless of declaration order.
+    // Here @Description is declared first to provoke the ordering-dependent bug.
```
Amaneusz, line 211: and here

prompt/prompt-structure/src/commonTest/kotlin/ai/koog/prompt/structure/json/generator/JsonSchemaGeneratorTest.kt
```diff
@@ -879,7 +901,77 @@ class JsonSchemaGeneratorTest {
...
+    // -------------------------------------------------------------------------
+    // Tests verifying findings #1 and #2 from PR review
+    // -------------------------------------------------------------------------
+
+    /**
+     * Finding #1: when @LLMDescription has empty value and description, effectiveValue must be
+     * treated as absent (null) so the @Description annotation is used as fallback.
+     * An effectiveValue of "" (non-null empty string) would short-circuit the fallback and
+     * produce a schema with no description at all.
+     */
```
Amaneusz: and here

prompt/prompt-structure/src/commonTest/kotlin/ai/koog/prompt/structure/json/generator/JsonSchemaGeneratorTest.kt
```diff
@@ -879,7 +901,77 @@ class JsonSchemaGeneratorTest {
...
+        """.trimIndent()
+    }
 
-        assertEquals(expectedSchema, schema)
+    /**
+     * Finding #2: @LLMDescription must take explicit priority over @Description regardless of
+     * annotation declaration order. Here @Description is intentionally declared before
+     * @LLMDescription on both the class and the property to verify that iteration order cannot
+     * cause @Description to win.
+     */
```
Amaneusz: and here

Amaneusz (approved):
Looks legit, I would just get rid of the comments here and there

kpavlov:
There is a similar bug with non-local return in kotlinx-schema 0.4.3, which should be addressed first

agents/agents-tools/src/commonMain/kotlin/ai/koog/agents/core/tools/annotations/LLMDescription.kt
```diff
@@ -16,4 +16,18 @@ import kotlinx.serialization.SerialInfo
...
+        replaceWith = ReplaceWith("value"),
+    )
+    val description: String = "",
+)
+
+/**
+ * Gets the value from [LLMDescription.value] or [LLMDescription.description].
+ */
+@Suppress("DEPRECATION")
+public val LLMDescription.effectiveValue: String?
```
EugeneTheDev, line 32: Do you think it's worth it? I assume that Kotlin users didn't specify parameter name for description explicitly, so for them nothing is broken. It looks more cumbersome than useful, tbh. I don't think we need this depreciation

EugeneTheDev (changes requested):

agents/agents-tools/src/commonMain/kotlin/ai/koog/agents/core/tools/annotations/LLMDescription.kt
kpavlov: @EugeneTheDev, unfortunately, there is a breaking change [in 0.7.1](https://github.com/Kotlin/kotlinx-schema/actions/runs/23248928601/job/67585133232?pr=267#step:6:1266). Automatic upgrade is not possible. I was, probably, the first reporter

EugeneTheDev: Did you check that it works from Java as before? The goal was to be able to write `@LLMDescription("foo")` without having to specify parameter name explicitly, because as default Java only understands `value` parameter

Rizzen (changes requested):
Could you please provide both Java and Kotlin examples and/or tests just to healthcheck this API?
Btw maybe it also worth to add some tests on how it works with Jackson, considering latest issues?
@EugeneTheDev what do you think?

EugeneTheDev:
It's not related to serialization directly, so I don't think such tests are needed

agents/agents-tools/src/commonMain/kotlin/ai/koog/agents/core/tools/annotations/LLMDescription.kt
kpavlov: There are java tests in agent-tools module. They are green

agents/agents-tools/src/commonMain/kotlin/ai/koog/agents/core/tools/annotations/LLMDescription.kt
```diff
@@ -16,4 +16,18 @@ import kotlinx.serialization.SerialInfo
...
     AnnotationTarget.FUNCTION
 )
-public annotation class LLMDescription(val value: String)
+public annotation class LLMDescription(
+    val value: String = "",
+    @Deprecated(
+        message = "Use 'value' instead.",
+        replaceWith = ReplaceWith("value"),
+    )
+    val description: String = "",
```
Ololoshechkin, line 25: very strange. I could've understand if it was `= value` by default -- then fine (though, obviously, annotations don't support non-constant default values). But now this way you define `description` field as an empty string `""` -- so older kotlinx-schema (or whatever library that expected "description" field to be present) would not break but would see an empty string and cause even less obvious bugs.
Am I missing smth?

Ololoshechkin (changes requested):

PR #1668, opened 2026-03-18
fix(prompt): add correct copy overloads in LLMParams subclasses

Amaneusz:
`Prompt.withUpdatedParams` internally calls `params.copy(…)` with the 8-parameter base signature. All `LLMParams` subclasses (`GoogleParams`, `AnthropicParams`, `OpenAIChatParams`, etc.) defined their own `copy()` with additional provider-specific parameters, but did not `override` the base `copy()`. 
Since the signatures differ, JVM virtual dispatch fell through to `LLMParams.copy()`, which always returned a plain `LLMParams` — silently dropping provider-specific fields like `GoogleParams.thinkingConfig`, `AnthropicParams.thinking`, etc.

Fixed by adding `override fun copy(8 base params)` to each subclass. Each override delegates to the subclass's full `copy()`, preserving all provider-specific fields from `this`. The integration test for `requestLLMForceOneTool` that was previously skipping Google models is now unblocked.

Additionally fixed `BedrockConverseParams.copy()` which was missing `performanceConfig`, `promptVariables`, and `requestMetadata` parameters, and `DashscopeParams` which had an override that incorrectly called `super.copy()`.

closes KG-742

### DIFF
--- a/integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/agent/AIAgentIntegrationTest.kt
+++ b/integration-tests/src/jvmTest/kotlin/ai/koog/integration/tests/agent/AIAgentIntegrationTest.kt
@@ -1379,11 +1379,6 @@ class AIAgentIntegrationTest : AIAgentTestBase() {
         assumeTrue(model.supports(LLMCapability.Tools), "Model $model does not support tools")
 
         Models.assumeEnumToolCallsAreStable(model, "force-one-tool integration with calculator enum arguments")
-        assumeTrue(
-            model.provider.id != LLMProvider.Google.id,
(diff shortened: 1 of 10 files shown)


kpavlov (approved):
@Amaneusz thank you

Maybe add KDoc about OpenRouterParams is an open class — the new override returns OpenRouterParams, so a subclass that forgets to override copy() inherits the same bug.

Amaneusz:
I planned some follow up work targetting `LLMParams` subclasses safety: https://youtrack.jetbrains.com/issue/KG-752/LLMParams-subtype-safety-enforce-copy-override-at-compile-time-and-remove-toParams

antoniibelyshev (approved):

PR #1671, opened 2026-03-18
fix(agents): preserve prompt messages, id, and params when adding systemMessage in agent builder

antoniibelyshev:
Updated `systemPrompt` method in agent builders to preserve previously configured messages, id, and params in the prompt.

closes [KG-747](https://youtrack.jetbrains.com/issue/KG-747)

### DIFF
--- a/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgentBuilder.kt
+++ b/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgentBuilder.kt
@@ -18,6 +18,7 @@ import ai.koog.agents.planner.AIAgentPlannerStrategyBuilder
 import ai.koog.agents.planner.PlannerAIAgent
 import ai.koog.agents.planner.TypedAgentPlannerStrategyBuilder
 import ai.koog.prompt.dsl.Prompt
+import ai.koog.prompt.dsl.prompt
 import ai.koog.prompt.executor.model.PromptExecutor
 import ai.koog.prompt.llm.LLModel
 import ai.koog.serialization.TypeToken
@@ -198,7 +199,7 @@ public class GraphAgentBuilder<Input, Output>(
      * @return The updated instance of GraphAgentBuilder with the system prompt applied.
      */
     public fun systemPrompt(systemPrompt: String): GraphAgentBuilder<Input, Output> =
-        prompt(ai.koog.prompt.dsl.prompt(id = "agent") { system(systemPrompt) })
+        prompt(prompt(config.prompt) { system(systemPrompt) })
 
     /**
      * Sets the prompt for the GraphAgentBuilder and returns the builder instance for further configuration.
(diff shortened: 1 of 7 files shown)


agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgentBuilder.kt
```diff
@@ -375,7 +376,7 @@ public class FunctionalAgentBuilder<Input, Output>(
      * @return The instance of FunctionalAgentBuilder with the updated system-level prompt.
      */
     public fun systemPrompt(systemPrompt: String): FunctionalAgentBuilder<Input, Output> =
```
Amaneusz, line 378: nit:
There's a slight behavioral change that might (?) be worth documenting

Before: each call to systemPrompt() replaced the entire prompt → only the last call mattered.                          
After: each call appends a new system message to the existing prompt → multiple calls accumulate.
                                                                                                                         
IMO the new way is better, but a user calling systemPrompt() twice thinking it overrides may be surprised.

agents/agents-core/src/jvmTest/kotlin/ai/koog/agents/core/agent/JavaAPIAgentBuilderTest.kt
```diff
@@ -249,4 +251,29 @@ class JavaAPIAgentBuilderTest {
...
+        val id = "original-prompt-id"
+        val temperature = 0.42
+        val maxTokens = 42
+        val originalSystemPrompt = "Original system prompt"
+        val systemPrompt = "System prompt"
+
+        val agent = AIAgent.builder()
+            .promptExecutor(getMockExecutor(serializer) { })
+            .llmModel(OpenAIModels.Chat.GPT4o)
+            .prompt(prompt(id, LLMParams(maxTokens = maxTokens)) { system(originalSystemPrompt) })
```
Amaneusz, line 266: This test is fine but I wonder if we shouldn't cover a case where we have a builder without any kind of prompt being configured before - just to document the fact that `systemPrompt` without any prior prompt-calls will produce a Prompt with empty id.

We could also consider covering prompt stacking that was mentioned here: https://github.com/JetBrains/koog/pull/1671/changes#r2952782148

Amaneusz (approved):
LGTM, thanks!

Added some nit-comments but nothing super important tbh

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgentBuilder.kt
antoniibelyshev: I'll add it to the kdoc after I merge the refactoring of agent builders to not have to modify multiple identical kdocs 🙂

agents/agents-core/src/jvmTest/kotlin/ai/koog/agents/core/agent/JavaAPIAgentBuilderTest.kt
kpavlov: BTW, shouldn't we always generate some id for prompt? e.g.UUID?

antoniibelyshev: I think, we do not need unique prompt ids. I thought they are just for observability, but I am not totally sure here tbh

PR #1680, opened 2026-03-18
fix(agents): serch for a tool call in responses when forcing a tool call

antoniibelyshev:
Previously `requestLLMForceOneTool` returned the first of the received messages.
This approach fails to return a tool call message if the model returns a reasoning message followed by a tool call.
To avoid this, this pr updates the method by first searching a tool call in the responses and returning it.


aozherelyeva (approved):
Thank you!

PR #1689, opened 2026-03-19
feat: Add support for Claude Sonnet 4.6

aozherelyeva:
Support Claude Sonnet 4.6
See https://platform.claude.com/docs/en/about-claude/models/overview


tiginamaria (approved):
Thank you! The fastest hand in the Old West!

PR #1691, opened 2026-03-19
feat(agents): add `JavaTool` class for idiomatic class-based tools implementation from Java

EugeneTheDev:
`Tool.execute` is suspendable, which does not look clean when implemented from Java, since it exposes continuation parameter. Introduce dedicated `JavaTool` class that provides the same API shape, but hides all suspendable machinery, allowing for idiomatic Java implementations. 

Also, class-based tools docs page is updated accordingly to include Java examples.

Class-based tool would look like this now in Java:
```java
class CalculatorTool extends JavaTool<CalculatorTool.Args, Integer> {
    public CalculatorTool() {
        super(
            TypeToken.of(Args.class),
            TypeToken.of(Integer.class),
            "calculator",
            "A simple calculator that can add two integers"
        );
    }

    // Arguments for the calculator tool
    record Args(
        @JsonProperty("number1")
        @LLMDescription("The first integer to add")
        int number1,
        @JsonProperty("number2")
        @LLMDescription("The second integer to add")
        int number2
    ) {}
    
    // Function to add two numbers
    @Override
    public Integer execute(Args args) {
        return args.number1 + args.number2;
    }

    // Optional custom result string representation for the LLM
    @Override
    public String encodeResultToString(Integer result, JSONSerializer serializer) {
        return "Calculation result: " + result;
    }
}

```


agents/agents-tools/src/jvmMain/java/ai/koog/agents/core/tools/JavaTool.java
```diff
@@ -0,0 +1,117 @@
...
+*/
+
+/**
+ * Base class representing a tool in Java that can be invoked by the LLM.
+ * Tools are usually used to return results, make changes to the environment, or perform other actions.
+ *
+ * @param <TArgs>   The type of arguments the tool accepts.
+ * @param <TResult> The type of result the tool returns.
+ */
+@SuppressWarnings("unused")
```
Ololoshechkin, line 26: Add JavaAPI annotation

agents/agents-tools/src/jvmMain/java/ai/koog/agents/core/tools/JavaTool.java
```diff
@@ -0,0 +1,117 @@
...
+    ) {
+        super(argsType, resultType, name, description);
+    }
+
+    @Override
+    final public @Nullable TResult execute(
+        TArgs args,
+        @NotNull Continuation<? super TResult> $completion
+    ) {
+        return execute(args);
```
Ololoshechkin: Submit the non-suspended work to some executor instead.
Deafault by default or some executor from constructor (must be optional of course).

Let's not just block coroutines

Ololoshechkin: Submit the non-suspended work to some executor instead.
Deafault by default or some executor from constructor (must be optional of course).

Let's not just block coroutines

Ololoshechkin (approved):
@EugeneTheDev also please make sure that documentation for tools has this for Java!

kpavlov (commented):
Shouldn't we also annotate kotlin suspend methods with `@JvmSynthetic`?

>  Synthetic targets become inaccessible for Java sources at compile time while still being accessible for Kotlin sources.
> Marking target as synthetic is a binary compatible change, already compiled Java code will be able to access such target.
>
> This annotation is intended for *rare cases* when API designer needs to hide Kotlin-specific target from Java API
> while keeping it a part of Kotlin API so the resulting API is idiomatic for both languages.

agents/agents-tools/src/jvmMain/java/ai/koog/agents/core/tools/JavaTool.java
EugeneTheDev: Good point, thanks. I had to move useful coroutine utils from `agents-core` to `utils` module to make them accesible globally and reimplemented this part to use `runBlockingIfRequired` with respect to the `$completion` parameter. Please take a look

EugeneTheDev: It's now implemented like this:
```kotlin
@Override
final public @Nullable TResult execute(
    TArgs args,
    @NotNull Continuation<? super TResult> $completion
) {
    return CoroutineUtilsKt.runBlockingIfRequired(
        $completion.getContext(),
        ($completionLambda) ->  execute(args)
    );
}

```

EugeneTheDev:
@kpavlov I can't make `execute` `JvmSynthetic` since overriding it from the Java is the whole purpose of `JavaTool`

kpavlov:
> I can't make execute JvmSynthetic since overriding it from the Java is the whole purpose of JavaTool

@EugeneTheDev , no I mean make it `@JvmSynthetic suspend execute(...)` in Tool.kt

EugeneTheDev:
> no I mean make it @JvmSynthetic suspend execute(...) in Tool.kt

Yes, I'm talking about it too. I need to override it from Java, so it can't be synthetic

agents/agents-tools/src/jvmMain/java/ai/koog/agents/core/tools/JavaTool.java
EugeneTheDev: Reimplemented this part to actually provide non-blocking API by offloading tool execution to a dedicated `Executor`

agents/agents-tools/src/jvmMain/java/ai/koog/agents/core/tools/JavaTool.java
```diff
@@ -0,0 +1,220 @@
...
+ * You can specify your own executor by using constructors accepting {@link Executor}.
+ * By default, each tool instance uses its own single-threaded executor.
+ *
+ * @param <TArgs>   The type of arguments the tool accepts.
+ * @param <TResult> The type of result the tool returns.
+ */
+@SuppressWarnings("unused")
+@JavaAPI
+public abstract class JavaTool<TArgs, TResult> extends Tool<TArgs, TResult> {
+    private Executor executor = Executors.newSingleThreadExecutor();
```
kpavlov, line 29: let's make it final with assignment in constructor

agents/agents-tools/src/jvmMain/java/ai/koog/agents/core/tools/JavaTool.java
```diff
@@ -0,0 +1,220 @@
...
+
+    /**
+     * Get {@link Executor} instance used to run the tool execute method.
+     */
+    public Executor getExecutor() {
+        return executor;
+    }
+
+    @Override
+    final public @Nullable Object execute(
```
kpavlov, line 195: This method is visible in java, isn't it?

kpavlov (changes requested):
@EugeneTheDev,  Please, remove `$continuations` from public java api

PR #1694, opened 2026-03-19
fix(agents): FIx AIAcgentConfig.copy for jvmCommon to copy jvmCommon-spesific fields

tiginamaria:
Replaced AIAcgentConfig.copy extension with an interface to copy jvmCommon-spesific fields fields

closes KG-749


aozherelyeva (approved):
LGTM, thanks!

PR #1699, opened 2026-03-20
docs(agents): update wording in History compression and Predefined nodes and components

marinkovicmarko:
Update wording on the following pages:
* History compression
* Pre-defined nodes and components

To account for the introduction of the Java API and Java code samples in the docs.

### DIFF
--- a/docs/docs/history-compression.md
+++ b/docs/docs/history-compression.md
@@ -32,12 +32,20 @@ There are two main approaches to implementing history compression in your agent:
 
 ### History compression in a strategy graph
 
-To compress the history in a strategy graph, you need to use the `nodeLLMCompressHistory` node in Kotlin or `AIAgentNode.llmCompressHistory()` in Java.
+To compress the history in a strategy graph, you need to use the pre-defined node that compresses the current message history into a concise summary:
+
+* **Kotlin**: `nodeLLMCompressHistory`
+* **Java**: `AIAgentNode.llmCompressHistory()`
+
+For more information and specific examples, see [History compression node](nodes-and-components.md#history-compression-node).
+
 Depending on which step you decide to perform compression, the following scenarios are available: 
 
 * To compress the history when it becomes too long, check the message count in your edge 
-  conditions and add a history compression node. In Kotlin, you can define a helper extension 
-  function. In Java, use inline lambda expressions in `.onCondition()`:
+  conditions and add a history compression node. To check the history length, do the following:
+
+* **Kotlin**: Define a helper extension.
+* **Java**: Use inline lambda expressions in `.onCondition()`.
 
 === "Kotlin"
 
@@ -285,7 +293,11 @@ To learn more about custom nodes, see [Custom nodes](custom-nodes.md).
 
 ## History compression strategies
 
-You can customize the compression process by passing an optional `strategy` parameter to `nodeLLMCompressHistory(strategy=...)` or `replaceHistoryWithTLDR(strategy=...)` (Kotllin), or `.compressionStrategy()` (Java).
+You can customize the compression process using the optional `strategy` parameter:
+
+* **Kotlin**: Pass the strategy to `nodeLLMCompressHistory(strategy=...)` or `replaceHistoryWithTLDR(strategy=...)`.
+* **Java**: Use the `.compressionStrategy()` builder method.
+
 The framework provides several built-in strategies.
 
 ### WholeHistory (Default)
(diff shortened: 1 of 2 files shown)


docs/docs/history-compression.md
```diff
@@ -24,15 +24,20 @@ History compression is performed at specific steps in the agent workflow:
 
 There are two main approaches to implementing history compression in your agent:
 
-- In a strategy graph.
-- In a custom node.
+- In a strategy graph
+- In a custom node (Kotlin)
+
+!!! warning 
+    Custom nodes are available only in Kotlin.
```
innateteniuk: I'd change this to something like 
"_History compression inside custom node logic is available only in Kotlin._"

Custom nodes can be created in Java https://github.com/JetBrains/koog/blob/develop/docs/docs/custom-nodes.md

@Ololoshechkin please correct me if I'm wrong

docs/docs/history-compression.md
```diff
@@ -248,7 +253,7 @@ In this example, the history is compressed after completing the information coll
 
 ### History compression in a custom node
 
-If you are implementing a custom node, you can compress history using the `replaceHistoryWithTLDR()` function as follows:
+If you are implementing a custom node, you can compress history using the `replaceHistoryWithTLDR()` function (Kotlin) as follows:
```
innateteniuk, line 267: It might be useful to add [a similar warning](https://github.com/JetBrains/koog/pull/1699/changes#r2976586189) about the availability of history compression in custom nodes

docs/docs/history-compression.md
```diff
@@ -24,15 +24,20 @@ History compression is performed at specific steps in the agent workflow:
...
+- In a strategy graph
+- In a custom node (Kotlin)
+
+!!! warning
+    History compression inside custom node logic is available only in Kotlin.
 
 ### History compression in a strategy graph
 
-To compress the history in a strategy graph, you need to use the `nodeLLMCompressHistory` node.
+To compress the history in a strategy graph, you need to use the `nodeLLMCompressHistory` node in Kotlin or `AIAgentNode.llmCompressHistory()` in Java.
```
innateteniuk: As an alternative, we could transform the first sentence into a general description, something like this:
`To compress the history in a strategy graph, you need to use the pre-defined node that compresses the current message history into a concise summary. `

And then provide the node names in Kotlin and Java as a list + add links to the corresponding chapter in `nodes-and-components.md. `
* **Kotlin**: `nodeLLMCompressHistory` 
* **Java**: `AIAgentNode.llmCompressHistory()`

This approach would also work for the similar paragraphs below.

docs/docs/history-compression.md
```diff
@@ -24,15 +24,20 @@ History compression is performed at specific steps in the agent workflow:
...
 Depending on which step you decide to perform compression, the following scenarios are available: 
 
-* To compress the history when it becomes too long, you can define a helper function and add the `nodeLLMCompressHistory` node to your strategy graph with the following logic:
+* To compress the history when it becomes too long, check the message count in your edge 
+  conditions and add a history compression node. In Kotlin, you can define a helper extension 
+  function. In Java, use inline lambda expressions in `.onCondition()`:
```
innateteniuk: Here you could also use a list or move the relevant sentences to the corresponding tabs.

docs/docs/nodes-and-components.md
```diff
@@ -1060,8 +1062,9 @@ You can use this subgraph for the following purposes:
 
 The API allows you to fine‑tune execution with optional parameters:
 
-- runMode: controls how tool calls are executed during the task (sequential by default). Use this to switch between different tool execution strategies when supported by the underlying model/executor.
-- assistantResponseRepeatMax: limits how many assistant responses are allowed before concluding the task cannot be completed (defaults to a safe internal limit if not provided).
+- [runMode](api:agents-core::ai.koog.agents.core.agent.entity.SubgraphWithTaskBuilder.runMode): controls how tool calls are executed during the task (sequential by default). Use this to switch between different tool execution strategies when supported by the underlying model/executor.
```
innateteniuk, line 1065: We'll need to clarify what these links will look like after the Java API reference is generated.

innateteniuk (approved):

docs/docs/nodes-and-components.md
marinkovicmarko: WDYT would be the right approach for the links for now? Keep the links to the current reference?

innateteniuk: Yes, for now, I would keep links to the current Kotlin API reference. This is just a note for the future.

PR #1700, opened 2026-03-21
fix(prompt): add is_error flag for failed tool calls in Anthropic client

### LINKED ISSUE #1598: When assembling the request for the Anthropic large model interface, the is_error field is not used for tool call exception messages.
The ToolResult object does not have the is_error attribute
Image
but the official documentation includes this field.
Image

omarraf:
The Anthropic API supports an is_error field on tool results to indicate when a tool call has failed, but the framework wasn't using it. This PR adds support for that field so that when a tool execution fails, the model is properly informed it was an error rather than a successful result.

closes #1598


omarraf:
Hello! While working on this, I noticed the Bedrock Anthropic client may need a similar update. The [Bedrock docs](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ToolResultBlock.html) use a status field (success/error) instead of is_error, but it serves the same purpose. Should I add it to this PR or open a separate issue?

prompt/prompt-executor/prompt-executor-clients/prompt-executor-anthropic-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/anthropic/AnthropicLLMClient.kt
```diff
@@ -389,7 +389,8 @@ public open class AnthropicLLMClient @JvmOverloads constructor(
                             content = listOf(
                                 AnthropicContent.ToolResult(
                                     toolUseId = message.id ?: "",
-                                    content = message.content
+                                    content = message.content,
+                                    isError = message.isError.takeIf { it }
```
antoniibelyshev: why do you need `.takeIf { it }`?

prompt/prompt-executor/prompt-executor-clients/prompt-executor-anthropic-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/anthropic/models/AnthropicChatMessages.kt
```diff
@@ -210,12 +210,15 @@ public sealed interface AnthropicContent {
...
+     * @property isError Whether this tool result represents an error. When true, Anthropic will treat
+     *   the content as an error message. Null means not an error (field omitted from the request).
      */
     @Serializable
     @SerialName("tool_result")
     public data class ToolResult(
         val toolUseId: String,
-        val content: String
+        val content: String,
+        val isError: Boolean? = null
```
antoniibelyshev: Can we remove the default here to avoid not providing this argument by accident? Also I think we can make it not nullable

antoniibelyshev (changes requested):
@omarraf Thank you, looks good! Just a couple small comments.

Regarding the Bedrock client, indeed the `isError` could also be added there. It would be cool to have as well, but if you do not add it in this pr, it's also fine.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-anthropic-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/anthropic/AnthropicLLMClient.kt
omarraf: Yea looking back we probably don't need this. I thought it'd be cleaner to omit fields that weren't needed by converting false -> null  but the API treats a missing `isError` the same as false so it makes no functional difference. I addressed this alongside your comment below this one

omarraf:
> the `isError` could also be added there

Sounds good! I'll work on adding it to this PR

antoniibelyshev (approved):
@omarraf thank you!

PR #1614, opened 2026-03-12
fix(prompt): ensure `reasoningContent` present with tool calls in DeepSeek

Stream29:
This PR keeps `reasoningContent` in tool calls and merge them with tool calls to satisfy DeepSeek API requirements.

closes #1599

PR #1701, opened 2026-03-22
fix(prompt): ensure `reasoningContent` present with tool calls in Anthropic

kevinherron:
## Summary

- When the Anthropic model returns both reasoning (thinking) and tool calls, the reasoning content was being filtered out
- Keep `Message.Reasoning` alongside `Message.Tool.Call` instead of discarding everything except tool calls
- Mirrors the fix applied for DeepSeek in #1614

## Test plan

- [x] Verify reasoning content is preserved when Anthropic model returns tool calls with thinking
- [x] Run `./gradlew :prompt:prompt-executor:prompt-executor-clients:prompt-executor-anthropic-client:jvmTest`

🤖 Generated with [Claude Code](https://claude.com/claude-code)


kevinherron:
Same thing `GoogleLLMClient.kt` does as well.

PR #1703, opened 2026-03-22
feat(prompt): expose cached input token counts in ResponseMetaInfo

### LINKED ISSUE #1275: Expose cached token counts in ResponseMetaInfo
# Feature Request: Expose cached token counts in ResponseMetaInfo

## Problem

`ResponseMetaInfo` does not include cached input token counts, even though Koog already parses this data from provider responses.

I'm building a chat application that tracks costs per conversation. OpenAI returns `prompt_tokens_details.cached_tokens` in responses, and Koog correctly deserializes this into `PromptTokensDetails.cachedTokens`. However, this data is dropped in `createMetaInfo()` and never reaches my application code.

## Current Behavior

**AbstractOpenAILLMClient.kt:482-487**
```kotlin
protected fun createMetaInfo(usage: OpenAIUsage?): ResponseMetaInfo = ResponseMetaInfo.create(
    clock,
    totalTokensCount = usage?.totalTokens,
    inputTokensCount = usage?.promptTokens,
    outputTokensCount = usage?.completionTokens
    // promptTokensDetails.cachedTokens is not passed
)
```

The `metadata` parameter exists in `ResponseMetaInfo.create()` but is never used.

## Data That Gets Lost

**OpenAIDataModels.kt:916-920** - Parsed but not exposed:
```kotlin
@Serializable
public class PromptTokensDetails(
    public val audioTokens: Int? = null,
    public val cachedTokens: Int? = null,  // This field exists
)
```

**OpenAIDataModels.kt:901-907** - Also parsed but not exposed:
```kotlin
@Serializable
public class CompletionTokensDetails(
    public val acceptedPredictionTokens: Int? = null,
    public val audioTokens: Int? = null,
    public val reasoningTokens: Int? = null,
    public val rejectedPredictionTokens: Int? = null,
)
```

## Proposed Solution

Populate the existing `metadata: JsonObject?` field in `ResponseMetaInfo`:

```kotlin
protected fun createMetaInfo(usage: OpenAIUsage?): ResponseMetaInfo {
    val tokenDetails = usage?.let {
        buildJsonObject {
            it.promptTokensDetails?.cachedTokens?.let { put("cachedInputTokens", it) }
            it.completionTokensDetails?.reasoningTokens?.let { put("reasoningTokens", it) }
        }.takeIf { it.isNotEmpty() }
    }

    return ResponseMetaInfo.create(
        clock,
        totalTokensCount = usage?.totalTokens,
        inputTokensCount = usage?.promptTokens,
        outputTokensCount = usage?.completionTokens,
        metadata = tokenDetails
    )
}
```

This is non-breaking since `metadata` already exists and accepts `JsonObject?`.

## Why This Matters

Cached tokens cost 50-90% less than regular input tokens depending on provider:
- OpenAI: 50% discount
- Anthropic: 90% discount
- Gemini: 75% discount

Without this data, applications either overcharge users or absorb the difference. For a typical chat session with 5K cached tokens and 1K new tokens using Claude Sonnet, the cost difference is ~$0.013 per request.

## Files to Modify

1. `AbstractOpenAILLMClient.kt` - Update `createMetaInfo()`
2. Similar changes needed in Anthropic and Google clients for their cache fields

## References

- OpenAI API: https://platform.openai.com/docs/api-reference/chat/object#chat/object-usage
- Koog source: `OpenAIDataModels.kt:916-920`, `Message.kt:398`

mvanhorn:
`ResponseMetaInfo` does not include cached input token counts, even though providers return this data. This adds a `cachedInputTokensCount: Int?` field (default null for backward compatibility) and populates it from each provider's response:

- **Anthropic:** Sums `cache_creation_input_tokens` and `cache_read_input_tokens` from the usage object
- **OpenAI:** Reads `prompt_tokens_details.cached_tokens`
- **Bedrock (Anthropic/Nova):** Sums `cache_read_input_tokens` and `cache_creation_input_tokens`

Also adds a `cachedInputTokensCount()` setter to the Java `ResponseMetaInfoBuilder`.

Files changed:
- `prompt/prompt-model/.../Message.kt` - New field on `ResponseMetaInfo`
- `prompt/prompt-model/.../MessageBuilder.kt` - Java builder setter
- `prompt/prompt-executor/.../AnthropicLLMClient.kt` - Parse cached tokens from Anthropic
- `prompt/prompt-executor/.../OpenAILLMClient.kt` - Parse cached tokens from OpenAI
- `prompt/prompt-executor/.../BedrockAnthropicClaudeSerialization.kt` - Parse cached tokens from Bedrock

Fixes #1275

This contribution was developed with AI assistance (Claude Code + Codex).

### DIFF
--- a/prompt/prompt-executor/prompt-executor-clients/prompt-executor-anthropic-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/anthropic/AnthropicLLMClient.kt
+++ b/prompt/prompt-executor/prompt-executor-clients/prompt-executor-anthropic-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/anthropic/AnthropicLLMClient.kt
@@ -193,10 +193,14 @@ public open class AnthropicLLMClient @JvmOverloads constructor(
         return buildStreamFrameFlow {
             var inputTokens: Int? = null
             var outputTokens: Int? = null
+            var cacheReadInputTokensCount: Int? = null
+            var cacheWriteInputTokensCount: Int? = null
 
             fun updateUsage(usage: AnthropicUsage) {
                 inputTokens = usage.inputTokens ?: inputTokens
                 outputTokens = usage.outputTokens ?: outputTokens
+                cacheReadInputTokensCount = usage.cacheReadInputTokens ?: cacheReadInputTokensCount
+                cacheWriteInputTokensCount = usage.cacheCreationInputTokens ?: cacheWriteInputTokensCount
             }
 
             fun getMetaInfo(): ResponseMetaInfo = ResponseMetaInfo.create(
@@ -204,6 +208,8 @@ public open class AnthropicLLMClient @JvmOverloads constructor(
                 totalTokensCount = inputTokens?.plus(outputTokens ?: 0) ?: outputTokens,
                 inputTokensCount = inputTokens,
                 outputTokensCount = outputTokens,
+                cacheReadInputTokensCount = cacheReadInputTokensCount,
+                cacheWriteInputTokensCount = cacheWriteInputTokensCount,
             )
 
             try {
@@ -665,12 +671,15 @@ public open class AnthropicLLMClient @JvmOverloads constructor(
         val inputTokensCount = response.usage?.inputTokens
         val outputTokensCount = response.usage?.outputTokens
         val totalTokensCount = response.usage?.let { it.inputTokens?.plus(it.outputTokens ?: 0) ?: it.outputTokens }
-        val cacheCreationInputTokens = response.usage?.cacheCreationInputTokens
-        val cacheReadInputTokens = response.usage?.cacheReadInputTokens
+        val cacheWriteInputTokensCount = response.usage?.cacheCreationInputTokens
+        val cacheReadInputTokensCount = response.usage?.cacheReadInputTokens
 
+        // The typed counts above are the supported way to read these. The
+        // metadata object is still populated because it was the only path
+        // before those fields existed and consumers depend on it.
         val cacheMetadata = buildJsonObject {
-            cacheCreationInputTokens?.let { put("cacheCreationInputTokens", it) }
-            cacheReadInputTokens?.let { put("cacheReadInputTokens", it) }
+            cacheWriteInputTokensCount?.let { put("cacheCreationInputTokens", it) }
+            cacheReadInputTokensCount?.let { put("cacheReadInputTokens", it) }
         }.takeIf { it.isNotEmpty() }
 
         val metaInfo = ResponseMetaInfo.create(
@@ -678,6 +687,8 @@ public open class AnthropicLLMClient @JvmOverloads constructor(
             totalTokensCount = totalTokensCount,
             inputTokensCount = inputTokensCount,
             outputTokensCount = outputTokensCount,
+            cacheReadInputTokensCount = cacheReadInputTokensCount,
+            cacheWriteInputTokensCount = cacheWriteInputTokensCount,
             metadata = cacheMetadata,
         )
 
--- a/prompt/prompt-executor/prompt-executor-clients/prompt-executor-anthropic-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/anthropic/AnthropicLLMClient.kt
+++ b/prompt/prompt-executor/prompt-executor-clients/prompt-executor-anthropic-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/anthropic/AnthropicLLMClient.kt
@@ -203,9 +203,7 @@ public open class AnthropicLLMClient @JvmOverloads constructor(
                 cacheWriteInputTokensCount = usage.cacheCreationInputTokens ?: cacheWriteInputTokensCount
             }
 
-            fun getMetaInfo(): ResponseMetaInfo = ResponseMetaInfo.create(
-                clock = clock,
-                totalTokensCount = inputTokens?.plus(outputTokens ?: 0) ?: outputTokens,
+            fun getMetaInfo(): ResponseMetaInfo = createResponseMetaInfo(
                 inputTokensCount = inputTokens,
                 outputTokensCount = outputTokens,
                 cacheReadInputTokensCount = cacheReadInputTokensCount,
@@ -667,29 +665,11 @@ public open class AnthropicLLMClient @JvmOverloads constructor(
         }
 
     private fun processAnthropicResponse(response: AnthropicResponse): Message.Assistant {
-        // Extract token count from the response
-        val inputTokensCount = response.usage?.inputTokens
-        val outputTokensCount = response.usage?.outputTokens
-        val totalTokensCount = response.usage?.let { it.inputTokens?.plus(it.outputTokens ?: 0) ?: it.outputTokens }
-        val cacheWriteInputTokensCount = response.usage?.cacheCreationInputTokens
-        val cacheReadInputTokensCount = response.usage?.cacheReadInputTokens
-
-        // The typed counts above are the supported way to read these. The
-        // metadata object is still populated because it was the only path
-        // before those fields existed and consumers depend on it.
-        val cacheMetadata = buildJsonObject {
-            cacheWriteInputTokensCount?.let { put("cacheCreationInputTokens", it) }
-            cacheReadInputTokensCount?.let { put("cacheReadInputTokens", it) }
-        }.takeIf { it.isNotEmpty() }
-
-        val metaInfo = ResponseMetaInfo.create(
-            clock,
-            totalTokensCount = totalTokensCount,
-            inputTokensCount = inputTokensCount,
-            outputTokensCount = outputTokensCount,
-            cacheReadInputTokensCount = cacheReadInputTokensCount,
-            cacheWriteInputTokensCount = cacheWriteInputTokensCount,
-            metadata = cacheMetadata,
+        val metaInfo = createResponseMetaInfo(
+            inputTokensCount = response.usage?.inputTokens,
+            outputTokensCount = response.usage?.outputTokens,
+            cacheReadInputTokensCount = response.usage?.cacheReadInputTokens,
+            cacheWriteInputTokensCount = response.usage?.cacheCreationInputTokens,
         )
 
         val parts = response.content.map { content ->
@@ -728,6 +708,24 @@ public open class AnthropicLLMClient @JvmOverloads constructor(
             metaInfo = metaInfo
         )
     }
+
+    private fun createResponseMetaInfo(
+        inputTokensCount: Int?,
+        outputTokensCount: Int?,
+        cacheReadInputTokensCount: Int?,
+        cacheWriteInputTokensCount: Int?,
+    ): ResponseMetaInfo = ResponseMetaInfo.create(
+        clock = clock,
+        totalTokensCount = inputTokensCount?.plus(outputTokensCount ?: 0) ?: outputTokensCount,
+        inputTokensCount = inputTokensCount,
+        outputTokensCount = outputTokensCount,
+        cacheReadInputTokensCount = cacheReadInputTokensCount,
+        cacheWriteInputTokensCount = cacheWriteInputTokensCount,
+        metadata = buildJsonObject {
+            cacheWriteInputTokensCount?.let { put("cacheCreationInputTokens", it) }
+            cacheReadInputTokensCount?.let { put("cacheReadInputTokens", it) }
+        }.takeIf { it.isNotEmpty() },
+    )
 
     /**
      * Helper function to get the type map for a parameter type without using smart casting
--- a/prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/bedrock/modelfamilies/amazon/BedrockAmazonNovaSerialization.kt
+++ b/prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/bedrock/modelfamilies/amazon/BedrockAmazonNovaSerialization.kt
@@ -190,6 +190,8 @@ internal object BedrockAmazonNovaSerialization {
         totalTokensCount = novaUsage?.totalTokens,
         inputTokensCount = novaUsage?.inputTokens,
         outputTokensCount = novaUsage?.outputTokens,
+        cacheReadInputTokensCount = novaUsage?.cacheReadInputTokenCount,
+        cacheWriteInputTokensCount = novaUsage?.cacheWriteInputTokenCount,
         metadata = buildJsonObject {
             put("cacheReadInputTokenCount", JsonPrimitive(novaUsage?.cacheReadInputTokenCount))
             put("cacheWriteInputTokenCount", JsonPrimitive(novaUsage?.cacheWriteInputTokenCount))
--- a/prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/message/Message.kt
+++ b/prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/message/Message.kt
@@ -490,6 +490,8 @@ public data class RequestMetaInfo @JvmOverloads constructor(
  * @property totalTokensCount The total number of tokens involved in the response, including both input and output tokens, or null if not available.
  * @property inputTokensCount The number of tokens used in the input, or null if not available.
  * @property outputTokensCount The number of tokens generated in the output, or null if not available.
+ * @property cacheReadInputTokensCount The number of input tokens served from cache (cache read), or null if not available.
+ * @property cacheWriteInputTokensCount The number of input tokens written to cache (cache creation), or null if not available.
  * @property metadata Additional metadata as a JSON object.
  *                    This can be used to store custom metadata that doesn't fit into the standard fields.
  * @property timestamp The timestamp indicating when the response was created.
@@ -504,6 +506,10 @@ public data class ResponseMetaInfo @JvmOverloads constructor(
     public val totalTokensCount: Int? = null,
     public val inputTokensCount: Int? = null,
     public val outputTokensCount: Int? = null,
+    /** The number of input tokens served from cache (cache read), or null if not available. */
+    public val cacheReadInputTokensCount: Int? = null,
+    /** The number of input tokens written to cache (cache creation), or null if not available. */
+    public val cacheWriteInputTokensCount: Int? = null,
     public val modelId: String? = null,
     override val metadata: JsonObject? = null,
 ) : MessageMetaInfo {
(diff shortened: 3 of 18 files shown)


prompt/prompt-executor/prompt-executor-clients/prompt-executor-anthropic-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/anthropic/AnthropicLLMClient.kt
```diff
@@ -552,11 +558,15 @@ public open class AnthropicLLMClient @JvmOverloads constructor(
         val inputTokensCount = response.usage?.inputTokens
         val outputTokensCount = response.usage?.outputTokens
         val totalTokensCount = response.usage?.let { it.inputTokens?.plus(it.outputTokens ?: 0) ?: it.outputTokens }
+        val cachedInputTokensCount = response.usage?.let { usage ->
+            listOfNotNull(usage.cacheCreationInputTokens, usage.cacheReadInputTokens).takeIf { it.isNotEmpty() }?.sum()
+        }
```
devcrocod: There are no unit tests covering this

prompt/prompt-executor/prompt-executor-clients/prompt-executor-anthropic-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/anthropic/AnthropicLLMClient.kt
```diff
@@ -187,17 +187,23 @@ public open class AnthropicLLMClient @JvmOverloads constructor(
...
 
             fun updateUsage(usage: AnthropicUsage) {
                 inputTokens = usage.inputTokens ?: inputTokens
                 outputTokens = usage.outputTokens ?: outputTokens
+                cachedInputTokensCount = usage.cachedInputTokensCount() ?: cachedInputTokensCount
             }
 
+            fun AnthropicUsage.cachedInputTokensCount(): Int? =
+                listOfNotNull(cacheCreationInputTokens, cacheReadInputTokens).takeIf { it.isNotEmpty() }?.sum()
+
```
devcrocod: This seems a bit inconsistent to me: there is a utility function here, but not for streaming

devcrocod (commented):
Thank you for your contribution

I have a question about the behavior of anthropic and openai, since they use different approaches to token counting. For Anthropic, cache reads and cache writes are being added together, although the statement only refers to cached input tokens. As I understand it, input creation tokens represent tokens written to the cache, not tokens read from it. If that is correct, the cost tracking here would be inaccurate

cc: @tiginamaria 
if I remember correctly, you worked on a similar task

mvanhorn:
Good catch, @devcrocod. You're right that cache_creation_input_tokens and cache_read_input_tokens represent different things - creation tokens are written to the cache (billed at 1.25x) while read tokens are served from it (billed at 0.1x). Summing them into a single cachedInputTokensCount loses that cost distinction.

I can split this into two separate fields (cachedCreationTokensCount and cachedReadTokensCount) in ResponseMetaInfo so downstream consumers can differentiate. Or if you'd prefer a different approach, happy to adjust.

devcrocod:
Hi @mvanhorn

>I can split this into two separate fields (cachedCreationTokensCount and cachedReadTokensCount) in ResponseMetaInfo so downstream consumers can differentiate.

Is this specific to Anthropic only? I’d rather avoid adding another public field to a shared class if it is effectively used by just one provider.

Also, could you please take a look at CI? It seems there are currently some code errors there.

mvanhorn:
Good questions.

> Is this specific to Anthropic only?

No - `cachedInputTokensCount` is populated by multiple providers in this PR:
- Anthropic: sum of `cache_creation_input_tokens` + `cache_read_input_tokens`
- OpenAI: `usage.promptTokensDetails.cachedTokens` (and `inputTokensDetails.cachedTokens` for the Responses API)
- Bedrock Anthropic: same as Anthropic
- Bedrock Amazon Nova: sum of `cacheReadInputTokenCount` + `cacheWriteInputTokenCount`

The field captures a common cross-provider concept (input tokens served from cache without re-processing), so a shared `ResponseMetaInfo` field is appropriate. Agreed on not adding `cachedCreationTokensCount` / `cachedReadTokensCount` as public fields - those are Anthropic-specific and consumers who need that granularity can read the raw response.

> could you please take a look at CI?

Fixed in 1b61e274. The local extension function `AnthropicUsage.cachedInputTokensCount()` was declared after its call site in `updateUsage()`, which failed Kotlin Native metadata compilation with the unresolved-reference error. Inlined the `listOfNotNull(usage.cacheCreationInputTokens, usage.cacheReadInputTokens).takeIf { it.isNotEmpty() }?.sum()` expression so it no longer depends on declaration order; matches the existing pattern in `processAnthropicResponse` and `BedrockAnthropicClaudeSerialization`.

No local Kotlin/JVM toolchain here, so relying on CI to confirm the fix.

aozherelyeva:
Hi @mvanhorn, thanks for the contribution! Could you please rebase onto `develop` and resolve the conflicts? Thanks!

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/message/Message.kt
```diff
@@ -419,6 +420,8 @@ public data class ResponseMetaInfo @JvmOverloads constructor(
     public val totalTokensCount: Int? = null,
     public val inputTokensCount: Int? = null,
     public val outputTokensCount: Int? = null,
+    /** The number of input tokens served from cache, or null if not available. */
+    public val cachedInputTokensCount: Int? = null,
```
tiginamaria: Does it make sense to store the cache write tokens count as well?

tiginamaria: I'd suggest adding `Read` or something more identifiable to the name (input is not really says that we read a cache)

prompt/prompt-executor/prompt-executor-clients/prompt-executor-anthropic-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/anthropic/AnthropicLLMClient.kt
```diff
@@ -187,17 +187,22 @@ public open class AnthropicLLMClient @JvmOverloads constructor(
         return buildStreamFrameFlow {
             var inputTokens: Int? = null
             var outputTokens: Int? = null
+            var cachedInputTokensCount: Int? = null
 
             fun updateUsage(usage: AnthropicUsage) {
                 inputTokens = usage.inputTokens ?: inputTokens
                 outputTokens = usage.outputTokens ?: outputTokens
+                cachedInputTokensCount =
```
tiginamaria: It should be `cacheReadInputTokens` only

prompt/prompt-executor/prompt-executor-clients/prompt-executor-anthropic-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/anthropic/AnthropicLLMClient.kt
```diff
@@ -552,11 +557,15 @@ public open class AnthropicLLMClient @JvmOverloads constructor(
         val inputTokensCount = response.usage?.inputTokens
         val outputTokensCount = response.usage?.outputTokens
         val totalTokensCount = response.usage?.let { it.inputTokens?.plus(it.outputTokens ?: 0) ?: it.outputTokens }
+        val cachedInputTokensCount = response.usage?.let { usage ->
+            listOfNotNull(usage.cacheCreationInputTokens, usage.cacheReadInputTokens).takeIf { it.isNotEmpty() }?.sum()
```
tiginamaria: Same, only `cacheReadInputTokens`

prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/bedrock/modelfamilies/amazon/BedrockAmazonNovaSerialization.kt
```diff
@@ -162,6 +162,10 @@ internal object BedrockAmazonNovaSerialization {
         totalTokensCount = novaUsage?.totalTokens,
         inputTokensCount = novaUsage?.inputTokens,
         outputTokensCount = novaUsage?.outputTokens,
+        cachedInputTokensCount = listOfNotNull(
```
tiginamaria: Let's have different fields for read and write cache tokens?

tiginamaria (changes requested):
Thank you for your contribution! In general, it looks good, but a little bit not in time as only anthropic and bedrock clients in Koog support caching (but I hope every client will do it soon). I'd suggest splitting the read and write cache token, counting in metadata, and giving it a better name

mvanhorn:
Addressed in 07af4234:

- Renamed `cachedInputTokensCount` to `cacheReadInputTokensCount` so direction is explicit.
- Added `cacheWriteInputTokensCount` as a separate `Int?`. Read and write tokens are now reported independently; summing hid the cost difference.
- Anthropic, Bedrock Anthropic Claude, and Bedrock Amazon Nova populate both fields from `cache_read_input_tokens` and `cache_creation_input_tokens` in streaming and non-streaming paths.
- OpenAI populates the read field only - it doesn't expose a separate cache-creation count.
- Bedrock Nova drops the `additionalInfo` workaround now that structured fields exist.
- Unit tests cover the split in Bedrock Anthropic (parse + stream) and Bedrock Nova (parse). Verified with `./gradlew :prompt:prompt-model:jvmTest :prompt:prompt-executor:prompt-executor-clients:prompt-executor-{anthropic,openai,bedrock}-client:jvmTest :agents:agents-features:agents-features-event-handler:jvmTest` - 449 tests pass.

prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/message/Message.kt
mvanhorn: Added `cacheWriteInputTokensCount` as a separate `Int?` in 07af4234. Anthropic and Bedrock populate it from `cache_creation_input_tokens`.

mvanhorn: Renamed to `cacheReadInputTokensCount` in 07af4234.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-anthropic-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/anthropic/AnthropicLLMClient.kt
mvanhorn: Changed to use `cacheReadInputTokens` only for the read field; `cacheCreationInputTokens` populates the new write field (07af4234).

prompt/prompt-executor/prompt-executor-clients/prompt-executor-anthropic-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/anthropic/AnthropicLLMClient.kt
mvanhorn: Same fix applied to the non-streaming path (07af4234).

prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/bedrock/modelfamilies/amazon/BedrockAmazonNovaSerialization.kt
mvanhorn: Split in 07af4234: `cacheReadInputTokenCount` populates `cacheReadInputTokensCount`, `cacheWriteInputTokenCount` populates `cacheWriteInputTokensCount`. Also dropped the `additionalInfo` workaround now that structured fields exist.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-anthropic-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/anthropic/AnthropicLLMClient.kt
mvanhorn: Added unit tests in 07af4234. Bedrock Anthropic covers parse + stream (`cache_read_input_tokens` -> read, `cache_creation_input_tokens` -> write); Bedrock Nova covers parse. The Anthropic client shares the same `AnthropicUsage` data class so the same parsing logic is exercised. Let me know if dedicated Anthropic-client tests would be useful.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-anthropic-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/anthropic/AnthropicLLMClient.kt
mvanhorn: Streaming and non-streaming paths now use the same pattern: `cacheReadInputTokensCount = usage.cacheReadInputTokens ?: cacheReadInputTokensCount`, `cacheWriteInputTokensCount = usage.cacheCreationInputTokens ?: cacheWriteInputTokensCount`. The summing logic that required a helper is gone, so both paths are trivially identical (07af4234).

mvanhorn:
@aozherelyeva happy to rebase, but there's a design overlap worth flagging first.

Since I opened this PR, `develop` added a `metadata: JsonObject?` field on `ResponseMetaInfo`, and the Anthropic client packs `cacheCreationInputTokens` and `cacheReadInputTokens` into it via `buildJsonObject`. My PR adds typed `cacheReadInputTokensCount` and `cacheWriteInputTokensCount` fields instead. Alternate solutions to the same problem for Anthropic.

They aren't equivalent though. `metadata` on develop is populated only by Anthropic. OpenAI, Bedrock Anthropic, and Bedrock Nova don't put cache tokens anywhere. My typed fields are populated by all four (OpenAI via `usage.promptTokensDetails.cachedTokens`, Nova via `cacheReadInputTokenCount` and `cacheWriteInputTokenCount`, and so on).

Three ways to go:

1. Close as superseded. If Anthropic-only cache token visibility via `metadata` is fine, this PR is redundant.
2. Rebase and keep the typed fields. Resolve conflicts by keeping `cacheReadInputTokensCount` and `cacheWriteInputTokensCount` alongside `metadata`. Anthropic would populate both.
3. Migrate the metadata packing into typed fields. Pull `cacheCreationInputTokens` and `cacheReadInputTokens` out of the opaque `metadata` blob into typed `ResponseMetaInfo` fields. Larger refactor but avoids two code paths.

Leaning (2) or (3), since losing OpenAI and Bedrock cache token coverage would be a regression. Happy to close if (1) is the direction you want. Which would you prefer?

aozherelyeva:
@tiginamaria could you please help with the questions above? 😅

danteoh-toast:
Bumping this - @mvanhorn @tiginamaria whatever support is needed here to get this across the line would be epic.

mvanhorn:
Thanks for the bump @danteoh-toast - sorry this stalled. Started the rebase onto current `develop` and got 5 of the conflicts cleanly resolved (`Message.kt` data class, `MessageBuilder.kt`, `AnthropicLLMClient.kt`, `BedrockAmazonNovaSerialization.kt`). The remaining work is on `EventHandlerTest.kt` - the test refactor on develop means the `toString()` fixtures need updating to include the new typed cache fields, and that's a gradle-iteration job rather than a one-shot. Setting aside dedicated time for it shortly. Strategy: option 2 from my 2026-04-23 comment (keep both typed fields + metadata, Anthropic populates both). Plan filed locally at `~/.osc/plans/2026-05-17-001-jetbrains-koog-1703-rebase-onto-develop.md`.

@aozherelyeva @tiginamaria if option 2 is the wrong direction, easiest to say so now before I re-push - happy to switch to option 1 (close as superseded) or option 3 (full metadata-to-typed refactor) instead.

danteoh-toast:
@aozherelyeva @tiginamaria @mvanhorn  how are others pulling cached input token counts out today? are they just intercepting the underlying request object to pull those counts out?

mvanhorn:
@danteoh-toast Today on `develop` the only built-in path is the `metadata: JsonObject?` field on `ResponseMetaInfo`, and only the Anthropic client populates it — it packs `cacheCreationInputTokens` and `cacheReadInputTokens` in via `buildJsonObject`. So a caller pulls them out of the opaque blob by string key, e.g. `responseMetaInfo.metadata?.get("cacheReadInputTokens")`.

For OpenAI and the Bedrock clients (Anthropic + Nova) there's no built-in surface at all — you'd have to intercept the raw provider response before koog discards it (`usage.prompt_tokens_details.cached_tokens` for OpenAI, `cacheReadInputTokenCount`/`cacheWriteInputTokenCount` for Nova).

That's the gap this PR closes: typed `cacheReadInputTokensCount` / `cacheWriteInputTokensCount` on `ResponseMetaInfo`, populated uniformly across all four clients — no JSON spelunking or request interception. The rebase onto current `develop` (option 2, keeping the typed fields alongside `metadata`) is in progress; the only remaining conflict is the `EventHandlerTest.kt` toString fixtures.

mvanhorn:
1d9440a rebases this onto current develop and resolves the drift. @danteoh-toast, thanks for the patience.

@tiginamaria, the counts are split as you asked: `cacheReadInputTokensCount` and `cacheWriteInputTokensCount` are separate fields rather than summed, since a cache write and a cache hit are different costs. Anthropic maps `cache_read_input_tokens` to read and `cache_creation_input_tokens` to write, and Bedrock Nova and Bedrock Anthropic map their equivalents.

One decision I would like a second opinion on: develop now populates cache counts into the `metadata` JsonObject, and that is what people are reading today. I kept it populated alongside the new typed fields rather than removing it, so existing consumers do not break. If you would rather this PR be the moment that goes away, say so and I will drop it.

The rebase also picked up two upstream changes: the Java builders moved into `ResponseMetaInfoBuilder.kt`, so the new setters live there, and the Bedrock parse helpers now return a single `Message.Assistant`, so the tests were updated accordingly. prompt-model and the Bedrock client compile and their jvmTest suites pass.

mvanhorn:
CI is green now. Three separate things were stacked behind that one red check, and each only became visible once the previous was fixed:

- `BedrockAnthropicUsage` gained the cache token fields, but its jvm ABI dump was never regenerated, so `checkLegacyAbi` failed on a signature change from an earlier commit rather than anything in the review.
- Same for `ResponseMetaInfo` in `prompt-model`, across all three dumps (jvm, android, klib).
- The new `AnthropicLLMClientTest` uses `MockEngine`, which was not on this module's test classpath. Added `ktor-client-mock` to `jvmTest`, matching the openai client which declares it alongside `http-client-ktor` and `ktor-client-cio`.

The review change itself is unchanged: the duplicated `ResponseMetaInfo` construction in the streaming and non-streaming paths is now one private helper, with a test asserting cache read and write input tokens split correctly and stay null when the usage block omits them.

danteoh-toast:
@tiginamaria can you or someone on the core team give us an indicator of when this is going to be reviewed/hope to land in a release?   We've already had to intercept http requests for other reasons Koog doesn't support (strict mode for tools) and we'd really prefer to not have to keep doing that.

PR #1715, opened 2026-03-23
feat: #176 Support for Google GenAI SDK: LLMClient and EmbeddingProvider

### LINKED ISSUE #176: Add support for Vertex Executor
Google cloud Vertex doesn't support API keys for accessing it's services. It works via Service account access. Would it be possible to add a Vertex executor to run against Gemini in Google Cloud?

kpavlov:
Fixes #176, [KG-110](https://youtrack.jetbrains.com/issue/KG-110)

Add Google GenAI LLM client using the official [google-genai Java SDK](https://github.com/googleapis/java-genai), with a standalone
  embedding provider and shared provider-validation infrastructure.

  - **`GoogleGenaiLLMClient`**: Full `LLMClient` implementation supporting chat completions, streaming, tool calling, structured output, thinking/reasoning models, multimodal content, and model listing
  - **`GoogleGenaiEmbeddingProvider`**: Separate `LLMEmbeddingProvider` for text embeddings with configurable `EmbedContentConfig` and Java-friendly `ExecutorService` constructor
  - **`LLMProviderAware`**: New common interface extracted from `LLMClientAPI` and `LLMEmbeddingProvider`, sharing `llmProvider()` and `requireMatchingProvider()` validation logic
  - **Internal helpers**: Shared error handling (`GoogleGenaiApiHelper`), request/response converters, tool conversion, and JSON/schema utilities
  - **Test suite**: Black-box unit tests (chat, streaming, tools, models, embedding, lifecycle) and integration tests for both executor and embedding  

Updating documentation is **out of this PR's scope**


kpavlov:
The integration test infrastructure requires some adjustments to support two clients for Google LLMProvider. This can be addressed separately.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-genai-client/src/main/kotlin/ai/koog/prompt/executor/clients/google/genai/GoogleGenaiLLMClient.kt
```diff
@@ -0,0 +1,865 @@
...
+
+    private companion object {
+        private val logger = KotlinLogging.logger { }
+
+        /**
+         * Default thought signature used for thinking models when no signature is available.
+         *
+         * See https://ai.google.dev/gemini-api/docs/thought-signatures
+         */
+        public const val DEFAULT_THOUGHT_SIGNATURE = "context_engineering_is_the_way_to_go"
```
zarechneva-a: 👍🏼

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-genai-client/build.gradle.kts
```diff
@@ -0,0 +1,30 @@
+import ai.koog.gradle.publish.maven.Publishing.publishToMaven
+
+group = rootProject.group
+version = rootProject.version
+
+plugins {
+    id("ai.kotlin.jvm")
+    alias(libs.plugins.kotlin.serialization)
+}
```
Ololoshechkin, line 10: Add jvm publication plugin here -- otherwise Maven publication wouldn't hapen for this module. 
See how it's done in spring-boot-starter module, for ex.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-genai-client/src/main/kotlin/ai/koog/prompt/executor/clients/google/genai/GoogleGenaiLLMClient.kt
```diff
@@ -0,0 +1,865 @@
...
+
+    override fun getBasicJsonSchemaGenerator(): GoogleBasicJsonSchemaGenerator {
+        return GoogleBasicJsonSchemaGenerator
+    }
+
+    override fun getStandardJsonSchemaGenerator(): GoogleStandardJsonSchemaGenerator {
+        return GoogleStandardJsonSchemaGenerator
+    }
+
+    override fun llmProvider(): LLMProvider = LLMProvider.Google
```
Ololoshechkin: Hmm... that's a tricky one.
We have Google LLM provider for Gemini models, and KMP multiplatform GoogleLLMClient that works with them (without this library)
Then we're adding this JVM-only module that can work both with Gemini and with Vertex also.

But we don't have Vertex models yet in the `GoogleModels` object...

Considerations:
1. We can't fully replace the older client due to Multiplatform thing (this one is not).
2. We can't add Vertex models just to the existing `GoogleModels` object and keep `LLMProvider.Google` for them because then the older (existing) GoogleLLMClient wouldn't work with them (and users would learn about it in runtime)
3. We still want this client to be able to work both with existing Google models and also with some new VertexAIModels (and respected unique LLMProvider).

Suggestions:
1. LLMProvider is already an abstract class. Let's add `GoogleLLMProviderFamily` class with 2 impls: (`GoogleLLMProvider`) (same as `LLMProvider.Google`) -- existing one, and `VertexAILLMProvider). 
2. Also add `LLMProvider.VertexAI` getter for it.
3. Add VertexAI mention to docs (https://docs.koog.ai/llm-providers/) and README (https://github.com/JetBrains/koog/blob/develop/README.md) 
4. Add VertexAI usage example (not just tests) to examples. I think let's have an AIAgent that uses it, not just a prompt executor/llm cline.

Ololoshechkin: see https://docs.koog.ai/examples/BedrockAgent/ -- same one but for Vertex

Ololoshechkin: Also, do we have integration tests for Vertex?
Let's add them before merging this PR 🙏

cc @aozherelyeva

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-genai-client/src/main/kotlin/ai/koog/prompt/executor/clients/google/genai/GoogleGenaiLLMClient.kt
```diff
@@ -0,0 +1,865 @@
...
+        require(model.supports(LLMCapability.Completion)) {
+            "Model ${model.id} does not support chat completions"
+        }
+
+        val (contents, systemInstruction) = buildSdkContents(prompt, model)
+        val config = buildConfig(prompt.params, model, tools, systemInstruction).build()
+
+        callApi {
+            val stream = client.async.models.generateContentStream(model.id, contents, config).await()
+            withContext(ioDispatcher) {
```
Ololoshechkin: Let's not use some fixed dispatcher (especially, IO) hard-coded without configuration.
That's very important because users want to be able to provide their own.
Also, for IntelliJ specifically -- that would just trigger a build-time inspection and they won't be able to use our library because of this :(

Easy fix -- unless we can go completely without this part -- let's just add CoroutineDispatcher to `GoogleGenaiLLMClient` parameters -- then, Dispatchers.IO can be a default one

Ololoshechkin: UPD: sorry, my bad, that's exactly what you did already :)
Then, great :)

Ololoshechkin: Also, maybe we can consider a separate constructor overload for Java users -- so that they could provide ExecutorService instead of kotlin CoroutineDispatcher (they're convertible into each other)

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-genai-client/src/main/kotlin/ai/koog/prompt/executor/clients/google/genai/GoogleGenaiLLMClient.kt
```diff
@@ -0,0 +1,865 @@
...
+                    for (chunk in responseStream) {
+                        val meta = extractResponseMetaInfo(chunk)
+
+                        chunk.candidates().orElse(null)?.firstOrNull()?.let { candidate ->
+                            candidate.content().orElse(null)?.parts()?.orElse(null)
+                                ?.forEachIndexed { index, part ->
+                                    val functionCall = part.functionCall().orElse(null)
+                                    val text = part.text().orElse(null)
+
+                                    when {
```
Ololoshechkin: And images/videos/documents?

Let's also add integration tests for all those data types.
and all models of course.

Also, does VertexAI support embeddings or moderation?

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-genai-client/src/main/kotlin/ai/koog/prompt/executor/clients/google/genai/GoogleGenaiLLMClient.kt
```diff
@@ -0,0 +1,865 @@
...
+
+        return contents to systemInstruction
+    }
+
+    private fun buildUserContent(message: Message.User, model: LLModel): Content {
+        val parts = message.parts.map { part ->
+            when (part) {
+                is ContentPart.Text -> Part.fromText(part.text)
+
+                is ContentPart.Image -> {
```
Ololoshechkin: Ok, I see. Does it mean that Vertex only supports non-text INPUT from the user, but doesn't support non-text OUTPUT from LLMs?

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-genai-client/src/main/kotlin/ai/koog/prompt/executor/clients/google/genai/GoogleGenaiLLMClient.kt
```diff
@@ -0,0 +1,865 @@
...
+                    require(model.supports(LLMCapability.Vision.Image)) {
+                        "Model ${model.id} does not support images"
+                    }
+                    blobPart(part.content, part.mimeType)
+                }
+
+                is ContentPart.Audio -> {
+                    require(model.supports(LLMCapability.Audio)) {
```
Ololoshechkin: Please, also add all VertexAI models into `VertexAIModels` (same as GoogleModels) with actual capabilities

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-genai-client/src/main/kotlin/ai/koog/prompt/executor/clients/google/genai/GoogleGenaiLLMClient.kt
```diff
@@ -0,0 +1,865 @@
...
+     */
+    protected open fun buildSdkTools(tools: List<ToolDescriptor>): List<Tool.Builder>? {
+        if (tools.isEmpty()) return null
+
+        val declarations = tools.map { tool ->
+            val properties = (tool.requiredParameters + tool.optionalParameters)
+                .associate { it.name to buildParamSchema(it) }
+
+            val schema = mapOf(
+                "type" to "object",
```
Ololoshechkin: just "object" ?

Ololoshechkin (changes requested):
Most importantly, let's:
- Split Vertex and Gemini providers (and add inheritance) as proposed
- Add integration tests to real models (both vertex and gemini)
- Add vertex example / guide

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-genai-client/src/main/kotlin/ai/koog/prompt/executor/clients/google/genai/GoogleGenaiLLMClient.kt
kpavlov: @Ololoshechkin, Vertex integration is out of scope.
GenAI API abstracts Vertex and Gemini. I see the reason to test Koog-GenAI SDK integration, but not the SDK itself.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-genai-client/src/main/kotlin/ai/koog/prompt/executor/clients/google/genai/GoogleGenaiLLMClient.kt
```diff
@@ -0,0 +1,890 @@
...
+    override suspend fun executeMultipleChoices(
+        prompt: Prompt,
+        model: LLModel,
+        tools: List<ToolDescriptor>
+    ): List<LLMChoice> {
+        logger.debug { "Executing prompt with multiple choices: $prompt with tools: $tools and model: $model" }
+        require(model.supports(LLMCapability.Completion)) {
+            "Model ${model.id} does not support chat completions"
+        }
+        require(model.supports(LLMCapability.Tools) || tools.isEmpty()) {
```
zarechneva-a: I failed to find it in the docs, but are tools really crucial for the MCQs?

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-genai-client/src/main/kotlin/ai/koog/prompt/executor/clients/google/genai/GoogleGenaiLLMClient.kt
```diff
@@ -0,0 +1,890 @@
...
+
+                is Message.User -> {
+                    flushAll()
+                    contents.add(buildUserContent(message, model))
+                }
+
+                is Message.Assistant -> {
+                    flushAll()
+                    contents.add(
+                        Content.builder().role("model").parts(Part.fromText(message.content)).build()
```
zarechneva-a: Maybe we could add a helper method `buildAssistantContent` just like the `buildUserContent` above for clarity?

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-genai-client/src/main/kotlin/ai/koog/prompt/executor/clients/google/genai/GoogleGenaiLLMClient.kt
kpavlov: @Ololoshechkin LLMProvider abstraction is broken. Gemini or Bedrock allows running different models, e.g. Anthropic models. This grouping by provider is artificial today. We should focus on the Model(i.e. capabilities) and LLMClient (transport/api shape)

kpavlov:
@Ololoshechkin 

> Split Vertex and Gemini providers (and add inheritance) as proposed

Nope, there is LLM client i.e. API client, connecting to the API endpoint and using Models. Provider is an artificial construct here and should be removed. I don't see a point to change something here.

> Add integration tests to real models (both vertex and gemini)

Added integration test, verifying that we can call GenAI SDK and receive response. Cofiguring SDK flag is user responsibility and is out of scope.

> Add vertex example / guide

This is separate documentation task

aozherelyeva (commented):
Overall LGTM, but just a couple of comments are below

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-genai-client/src/main/kotlin/ai/koog/prompt/executor/clients/google/genai/GoogleGenaiLLMClient.kt
kpavlov: absolutely! The json schema element has type "object" with "properties" and "required" fields

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-genai-client/src/main/kotlin/ai/koog/prompt/executor/clients/google/genai/GoogleGenaiLLMClient.kt
kpavlov: You are right. If user wants to use embedding model, which is supported by genai SDK, they don't need tools or completion. Such checks should go to execute methods

zarechneva-a: Could you please put them into the execute method, or this would require a lot of changes?
Or maybe we could just omit the tools support check...

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-genai-client/src/main/kotlin/ai/koog/prompt/executor/clients/google/genai/GoogleGenaiLLMClient.kt
kpavlov: Unit tests have been added for response conversion. Integration tests are not considered appropriate as the Genai SDK is not the System-Under-Test. This would result in an unnecessary combinatorial expansion of the test suite.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-genai-client/src/main/kotlin/ai/koog/prompt/executor/clients/google/genai/GoogleGenaiLLMClient.kt
kpavlov: This is beyond the scope of this PR.

andruhon:
Looks really promising! I'm excited to try this tomorrow.

aozherelyeva (approved):
LGTM from the tests' side! Thanks

andruhon:
Before going too far with testing real agent, I tried modifying `step-01-minimal-agent` to use the new connector. 
Works fine with Gemini 2.5 models, but falls apart with Gemini 3 and above

```
ai.koog.prompt.executor.clients.LLMClientException: Error from client: GoogleGenaiLLMClient
     Status code: 400 (). 400 . Request contains an invalid argument.
```

This this is most likely something to do with thought signature. I will try to debug and will provide more details.

Otherwise the Gemini 3 is still a `preview`, so probably improvements for Gemini 3 models can go as a follow-up.

**UPD**: This is the thought signature issue: the first API call succeeds (model returns a tool call for __list_directory__). The 400 happens on the second call when sending the tool result back.

andruhon:
Bug 1: Thought signature byte corruption (GoogleGenaiConversionUtils.kt)
  - Thought signatures are opaque binary data, not valid UTF-8
  - signatureFromBytes used decodeToString() which corrupted bytes (271 bytes → 258 chars)
  - signatureToBytes then produced different bytes → API rejected the invalid signature
  - Fix: Use Base64 encoding for lossless round-trip of binary data

  Bug 2: Signature not propagated from Reasoning with content (GoogleGenaiRequestConverter.kt)
  - When Message.Reasoning had both content and a signature, the code created the thought part but didn't set lastSignature
  - The subsequent Tool.Call had no signature to use
  - Fix: Always set lastSignature = message.encrypted regardless of whether the Reasoning has content

**UPD**: there was a diff fixing these two, I've added another diff below, which also fixes "parallel" or batch tool calling.

andruhon:
Found another gap, Gemini 3.1 likes to call multiple tools in parallel.

For this prompt `List current dir and read ./LICENSE.txt'` 2.5 would have called one tool, and then receiving response from tools it would have called the second.

3.1 Does this:
```
{
      "role" : "model",
      "parts" : [ {
        "functionCall" : {
          "name" : "__list_directory__",
          "args" : {
            "absolutePath" : "/Users/user/wrk/_koog/koog",
            "depth" : 1
          }
        },
        "thoughtSignature" : "abracadabra=="
      }, {
        "functionCall" : {
          "name" : "__read_file__",
          "args" : {
            "path" : "/Users/user/wrk/_koog/koog/LICENSE.txt"
          }
        }
      } ]
    }
```

It also falls apart. Looking.

andruhon:
According to Google Doco in https://docs.cloud.google.com/vertex-ai/generative-ai/docs/thought-signatures only first tool call in the batch needs a signature. The diff for all three fixes

```
Index: prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-genai-client/src/main/kotlin/ai/koog/prompt/executor/clients/google/genai/GoogleGenaiRequestConverter.kt
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-genai-client/src/main/kotlin/ai/koog/prompt/executor/clients/google/genai/GoogleGenaiRequestConverter.kt b/prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-genai-client/src/main/kotlin/ai/koog/prompt/executor/clients/google/genai/GoogleGenaiRequestConverter.kt
--- a/prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-genai-client/src/main/kotlin/ai/koog/prompt/executor/clients/google/genai/GoogleGenaiRequestConverter.kt	(revision cc9f74e6a313fbe0b31dc56e567a801bcdf96062)
+++ b/prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-genai-client/src/main/kotlin/ai/koog/prompt/executor/clients/google/genai/GoogleGenaiRequestConverter.kt	(date 1774506818331)
@@ -84,9 +84,9 @@
                         contents.add(
                             Content.builder().role("model").parts(listOf(partBuilder.build())).build()
                         )
-                    } else {
-                        lastSignature = message.encrypted
-                    }
+                    }
+                    // Always propagate the signature so subsequent Tool.Call parts can use it
+                    lastSignature = message.encrypted
                 }
 
                 is Message.Tool.Result -> {
@@ -99,17 +99,21 @@
                 }
 
                 is Message.Tool.Call -> {
-                    if (pendingCalls.isEmpty()) {
+                    val isFirstCallInBatch = pendingCalls.isEmpty()
+                    if (isFirstCallInBatch) {
                         flushResults()
                     }
 
                     val signature = lastSignature
                     lastSignature = null
 
-                    val effectiveSignature = signature ?: if (isThinkingModel) {
-                        fallbackThoughtSignature
+                    // Only the first functionCall part in a parallel batch needs the signature.
+                    // Subsequent parts must not have one (per Google API spec).
+                    // See https://docs.cloud.google.com/vertex-ai/generative-ai/docs/thought-signatures
+                    val effectiveSignature = if (isFirstCallInBatch) {
+                        signature ?: if (isThinkingModel) fallbackThoughtSignature else null
                     } else {
-                        null
+                        signature
                     }
 
                     val args = conversionUtils.parseJsonToMap(message.content)
Index: prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-genai-client/src/main/kotlin/ai/koog/prompt/executor/clients/google/genai/GoogleGenaiConversionUtils.kt
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-genai-client/src/main/kotlin/ai/koog/prompt/executor/clients/google/genai/GoogleGenaiConversionUtils.kt b/prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-genai-client/src/main/kotlin/ai/koog/prompt/executor/clients/google/genai/GoogleGenaiConversionUtils.kt
--- a/prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-genai-client/src/main/kotlin/ai/koog/prompt/executor/clients/google/genai/GoogleGenaiConversionUtils.kt	(revision cc9f74e6a313fbe0b31dc56e567a801bcdf96062)
+++ b/prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-genai-client/src/main/kotlin/ai/koog/prompt/executor/clients/google/genai/GoogleGenaiConversionUtils.kt	(date 1774505556858)
@@ -3,6 +3,7 @@
 import com.google.genai.types.Schema
 import io.github.oshai.kotlinlogging.KLogger
+import java.util.Base64
 import kotlinx.serialization.json.JsonArray
 import kotlinx.serialization.json.JsonElement
 import kotlinx.serialization.json.JsonNull
@@ -22,13 +23,11 @@
     /**
      * Converts a signature string to byte[] for the SDK.
      *
-     * The Google API returns thought signatures as JSON strings. The SDK expects `byte[]` and
-     * serializes them back to JSON. We use UTF-8 encoding (string -> bytes) on requests and
-     * UTF-8 decoding (bytes -> string) on responses, preserving the original string value.
-     * This assumes signatures are valid UTF-8 strings (which holds for both base64 and plaintext tokens).
+     * Thought signatures are opaque binary data. We use Base64 encoding to safely
+     * round-trip the raw bytes through the String-based [Message.Reasoning.encrypted] field.
      */
-    fun signatureToBytes(value: String): ByteArray = value.encodeToByteArray()
-    fun signatureFromBytes(value: ByteArray): String = value.decodeToString()
+    fun signatureToBytes(value: String): ByteArray = Base64.getDecoder().decode(value)
+    fun signatureFromBytes(value: ByteArray): String = Base64.getEncoder().encodeToString(value)
 
     // endregion
 
```

Tested on examples/code-agent/step-01-minimal-agent with

- Gemini 2.5 Pro
- Gemini 3 Pro Preview
- Gemini 3.1 Pro Prefiew. 

Prompts
- List all the README files
- List current dir and read ./LICENSE.txt

kpavlov:
@andruhon, thank you for your patch. I applied it along with some unit tests

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-genai-client/src/main/kotlin/ai/koog/prompt/executor/clients/google/genai/GoogleGenaiLLMClient.kt
```diff
@@ -0,0 +1,471 @@
...
+    @OptIn(InternalLLMClientApi::class)
+    override suspend fun embed(text: String, model: LLModel): List<Double> {
+        requireMatchingProvider(model)
+        require(model.supports(LLMCapability.Embed)) {
+            "Model ${model.id} does not support embedding."
+        }
+
+        logger.debug { "Embedding text with model: ${model.id}" }
+
+        val response = callApi { client.async.models.embedContent(model.id, text, null).await() }
```
serge-p7v: It seems the third argument of the `embedContent` method might be important for users - https://youtrack.jetbrains.com/issue/KG-780 (they wanted to pass the taskType parameter: RETRIEVAL_DOCUMENT, RETRIEVAL_QUERY, CODE_RETRIEVAL_QUERY, etc.)
As a workaround, is it possible to add an optional constructor parameter to GoogleGenaiLLMClient with `EmbedContentConfig` or somehow else allow to pass the taskType?

kpavlov: Thank you for your suggestion. However, we should consider a holistic approach: the `LLMEmbeddingProvider` requires extension to support:

1. Non-text content (documents and fragments)
2. Embedding model options

Let's implement the interface extension separately.

But, I am fine to extract EmbeddingProvider specific methods from the GoogleGenaiLLMClient to a separate class with a property `EmbedContentConfig`

kpavlov: done

prompt/prompt-executor/prompt-executor-clients/src/commonMain/kotlin/ai/koog/prompt/executor/clients/LLMProviderAware.kt
```diff
@@ -0,0 +1,31 @@
...
+import ai.koog.prompt.llm.LLMProvider
+import ai.koog.prompt.llm.LLModel
+
+/**
+ * Common interface for components that are associated with a specific [LLMProvider].
+ *
+ * Implemented by both [LLMClientAPI] and [LLMEmbeddingProvider] to enable shared
+ * provider-validation logic (e.g. [requireMatchingProvider]).
+ */
+public interface LLMProviderAware {
```
tiginamaria: I'm not sure we need a special interface for it?

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-genai-client/src/main/kotlin/ai/koog/prompt/executor/clients/google/genai/GoogleGenaiLLMClient.kt
```diff
@@ -0,0 +1,420 @@
...
+
+    override fun llmProvider(): LLMProvider = llmProvider
+
+    private suspend fun <T> callApi(block: suspend () -> T): T = callGoogleGenaiApi(clientName, block)
+
+    // region Execute
+
+    @OptIn(InternalAgentToolsApi::class, InternalLLMClientApi::class)
+    override suspend fun execute(prompt: Prompt, model: LLModel, tools: List<ToolDescriptor>): List<Message.Response> {
+        requireMatchingProvider(model)
```
tiginamaria: We need to add it to all clients; otherwise, the behavior is different. And I don't really like the interface approach. Why can't we just make `model.requireProvider(llmProvider)` like requireCapability (or another name, but same approach)?

prompt/prompt-llm/src/commonMain/kotlin/ai/koog/prompt/llm/LLModel.kt
```diff
@@ -27,3 +27,12 @@ public data class LLModel(
      */
     public fun supports(capability: LLMCapability): Boolean = capabilities?.contains(capability) ?: false
 }
+
+/**
+ * Checks if the Large Language Model (LLM) supports a specific capability.
+ * @param capability The capability to check for, represented by an instance of [LLMCapability].
+ */
+public fun LLModel.requireCapability(capability: LLMCapability): Unit =
```
tiginamaria, line 35: If we change the message in one provider, it's better to change it in all, so it's better to replace the require(support) with `requireCapability` in all other providers

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-genai-client/src/main/kotlin/ai/koog/prompt/executor/clients/google/genai/GoogleGenaiLLMClient.kt
```diff
@@ -0,0 +1,420 @@
...
+
+    @OptIn(InternalLLMClientApi::class)
+    override fun executeStreaming(
+        prompt: Prompt,
+        model: LLModel,
+        tools: List<ToolDescriptor>
+    ): Flow<StreamFrame> = buildStreamFrameFlow {
+        requireMatchingProvider(model)
+        logger.debug { "Executing streaming prompt: $prompt with model: $model" }
+        require(model.supports(LLMCapability.Completion)) {
```
tiginamaria: should be `requireCapability`

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-genai-client/src/main/kotlin/ai/koog/prompt/executor/clients/google/genai/GoogleGenaiLLMClient.kt
```diff
@@ -0,0 +1,420 @@
...
+
+    @OptIn(InternalAgentToolsApi::class, InternalLLMClientApi::class)
+    override suspend fun executeMultipleChoices(
+        prompt: Prompt,
+        model: LLModel,
+        tools: List<ToolDescriptor>
+    ): List<LLMChoice> {
+        requireMatchingProvider(model)
+        logger.debug { "Executing prompt with multiple choices: $prompt with tools: $tools and model: $model" }
+        require(model.supports(LLMCapability.Completion)) {
```
tiginamaria: should be `requireCapability`

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-genai-client/src/main/kotlin/ai/koog/prompt/executor/clients/google/genai/GoogleGenaiLLMClient.kt
```diff
@@ -0,0 +1,420 @@
...
+        // System instruction
+        systemInstruction?.let { builder.systemInstruction(it) }
+
+        // Tools
+        buildSdkTools(tools)?.map { it.build() }?.let { builder.tools(it) }
+        buildSdkToolConfig(googleParams.toolChoice)?.let { builder.toolConfig(it) }
+
+        // Response format (structured output schema)
+        googleParams.schema?.let { schema ->
+            require(model.supports(schema.capability)) {
```
tiginamaria: should be `requireCapability`

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-genai-client/src/main/kotlin/ai/koog/prompt/executor/clients/google/genai/GoogleGenaiRequestConverter.kt
```diff
@@ -0,0 +1,194 @@
...
+        return Content.builder().role("model").parts(Part.fromText(message.content)).build()
+    }
+
+    private fun buildUserContent(message: Message.User, model: LLModel): Content {
+        val parts = message.parts.map { part ->
+            when (part) {
+                is ContentPart.Text -> Part.fromText(part.text)
+
+                is ContentPart.Image -> {
+                    require(model.supports(LLMCapability.Vision.Image)) {
```
tiginamaria, line 154: should be `requireCapability`

tiginamaria (approved):
Thank you! Looks like a lot of work!
I have some minor comments about the new method and interfaces that introduce inconsistencies for clients' behavior/organization. I'd suggest thinking it through, maybe rolling back some of them and opening a separate PR, and updating all clients there (as the current one is already huge)

prompt/prompt-llm/src/commonMain/kotlin/ai/koog/prompt/llm/LLModel.kt
kpavlov: Yes, but I wanted to keep this PR focused on Google only. I was thinking we could switch from copying and pasting checks to using a single, consistent helper method.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-genai-client/src/main/kotlin/ai/koog/prompt/executor/clients/google/genai/GoogleGenaiLLMClient.kt
kpavlov: we should

andruhon:
This PR seems to be stuck. Is there anything I can help with?

kpavlov:
@tiginamaria , would you please help with syncing and resolving a conflict?

oliverhenlich:
Hoping this PR can be revived soon.

micahsmith:
Any chance of this moving forward? I would be glad to contribute help if any is needed to get this unstuck.

PR #1716, opened 2026-03-23
fix(agents): KG-750. Fix deadlock when agent.run() is called from within executor.submit

sdubov:
Fix deadlock when agent.run() is called from within executor.submit()

- When agent.run() was invoked from a thread that is itself a worker of the configured ExecutorService runBlocking(context) would dispatch the coroutine back onto that executor and park the calling thread, but no other thread existed to run it.

closes: [KG-750](https://youtrack.jetbrains.com/issue/KG-750)


aozherelyeva (approved):
Thanks for the fix!

PR #1717, opened 2026-03-23
fix(prompt): support Anthropic adaptive thinking with tool calls

kevinherron:
## Summary

- Add `AnthropicThinking.Adaptive` class for the adaptive thinking mode introduced in Opus 4.6+ and Sonnet 4.6+, where the model dynamically decides when and how much to think
- Fix coalescing of consecutive assistant-role content blocks (Reasoning, Assistant, Tool.Call) into a single `AnthropicMessage.Assistant`, as required by the Anthropic API when extended thinking is enabled with tool use
- Preserve `Message.Reasoning` alongside `Message.Tool.Call` in response filtering instead of discarding reasoning content when the model returns both

## Test plan

- [x] Verify `./gradlew :prompt:prompt-executor:prompt-executor-clients:prompt-executor-anthropic-client:jvmTest` passes
- [x] Verify existing Anthropic client tests still pass with the coalescing changes
- [x] Manual integration test with an Anthropic model using adaptive thinking + tool calls to confirm thinking and tool_use blocks are sent in the same assistant message

### DIFF
--- a/prompt/prompt-executor/prompt-executor-clients/prompt-executor-anthropic-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/anthropic/AnthropicLLMClient.kt
+++ b/prompt/prompt-executor/prompt-executor-clients/prompt-executor-anthropic-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/anthropic/AnthropicLLMClient.kt
@@ -349,41 +349,51 @@ public open class AnthropicLLMClient @JvmOverloads constructor(
         val systemMessage = mutableListOf<SystemAnthropicMessage>()
         val messages = mutableListOf<AnthropicMessage>()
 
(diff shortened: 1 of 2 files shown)


kevinherron:
For my agent, vs `0.7.2`, this is the difference between seeing a single thinking block at the beginning of the session, vs interleaved thinking blocks as the model deems necessary.

I first tried without introducing the adaptive thinking parameter and it made no difference - I think the 4.6 models ignore `thinking=enabled` and now require `thinking=adaptive` to interleave thinking/reasoning messages.

kevinherron:
Not sure if this is `fix(...)` or `feat(...)` at this point... I went into this thinking "fix whatever prevents me from receiving thinking output from Anthropic when it works as expected with Gemini".

PR #1718, opened 2026-03-23
fix(prompt): pass tools parameter in JVM PromptExecutor.execute() overload

Rizzen:
The JVM-specific execute() method was ignoring the `tools` argument and always passing emptyList(), breaking Java interop for tool-based calls. Re-enable the integration test that was disabled for [KG-725](https://youtrack.jetbrains.com/issue/KG-725).

closes
KG-725

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -111,6 +111,19 @@ subprojects {
         outputToConsole = true
         coloredOutput = true
     }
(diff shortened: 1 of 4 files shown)


Ololoshechkin (approved):
Thanks!

aozherelyeva (approved):
Thanks!!!

PR #1722, opened 2026-03-23
build: remove stale coreLibrariesVersion override (#1697)

### LINKED ISSUE #1697: Published artifacts declare kotlin-stdlib 2.1.21 but require Kotlin 2.2+ due to 2.3.0 metadata
The published POMs for **v0.7.1** declare `kotlin-stdlib:2.1.21` as a compile dependency (e.g. [agents-core-jvm](https://repo1.maven.org/maven2/ai/koog/agents-core-jvm/0.7.1/agents-core-jvm-0.7.1.pom)), suggesting compatibility with Kotlin 2.1. However the artifacts are compiled with Kotlin 2.3.10, which stamps metadata version 2.3.0 into the class files.

Kotlin 2.1 compilers can only read metadata up to version 2.2.0 (one minor version ahead), so any project using Kotlin 2.1 gets:
```
Module was compiled with an incompatible version of Kotlin.
The binary version of its metadata is 2.3.0, expected version is 2.1.0.
```

The convention plugins set `coreLibrariesVersion = "2.1.21"` which controls the stdlib dependency in the published POM, but the actual compiler version is 2.3.10 (from libs.versions.toml). Is that intentional?

These two values are out of sync, so I would suggest either compile with Kotlin 2.1 to match the declared stdlib floor, or remove the `coreLibrariesVersion` override so the POM accurately reflects the minimum Kotlin version required.

Affected files:
- [gradle/libs.versions.toml](https://github.com/JetBrains/koog/blob/5af51d759a5952626690b34198b5086059ebdb9b/gradle/libs.versions.toml#L17)
- [convention-plugin-ai/src/main/kotlin/ai.kotlin.multiplatform.gradle.kts](https://github.com/JetBrains/koog/blob/5af51d759a5952626690b34198b5086059ebdb9b/convention-plugin-ai/src/main/kotlin/ai.kotlin.multiplatform.gradle.kts#L124-L126)
- [convention-plugin-ai/src/main/kotlin/ai.kotlin.multiplatform.server.gradle.kts](https://github.com/JetBrains/koog/blob/5af51d759a5952626690b34198b5086059ebdb9b/convention-plugin-ai/src/main/kotlin/ai.kotlin.multiplatform.server.gradle.kts#L21-L23)

hxreborn:
The convention plugins were setting `coreLibrariesVersion = "2.1.21"` which made the published POMs declare kotlin-stdlib 2.1.21 as a dependency. But the actual compiler version is 2.3.x (from libs.versions.toml), so the compiled bytecode has 2.3.0 metadata and breaks for anyone on Kotlin 2.1.

Removed the `coreLibrariesVersion` override from both `ai.kotlin.multiplatform.gradle.kts` and `ai.kotlin.multiplatform.server.gradle.kts` so the POM picks up the real compiler version instead.

Tried lowering `kotlinApiVersion` to `KOTLIN_2_1` and `KOTLIN_2_2` per maintainer feedback, but both break compilation because `kotlin.time.ExperimentalTime` only became stable in 2.3.

Fixes #1697

### DIFF
--- a/convention-plugin-ai/src/main/kotlin/ai.kotlin.configuration.gradle.kts
+++ b/convention-plugin-ai/src/main/kotlin/ai.kotlin.configuration.gradle.kts
@@ -15,7 +15,7 @@ val libs = the<LibrariesForLibs>()
   See: https://youtrack.jetbrains.com/issue/KT-66755/Native-non-JVM-targets-add-support-for-languageVersion
  */
 val kotlinLanguageVersion = KotlinVersion.KOTLIN_2_3
-val kotlinApiVersion = KotlinVersion.KOTLIN_2_3
+val kotlinApiVersion = KotlinVersion.KOTLIN_2_1
 val kotlinBomVersion = requireNotNull(libs.kotlin.bom.get().version)
 
 extensions.getByType<KotlinProjectExtension>().apply {
--- a/convention-plugin-ai/src/main/kotlin/ai.kotlin.configuration.gradle.kts
+++ b/convention-plugin-ai/src/main/kotlin/ai.kotlin.configuration.gradle.kts
@@ -15,7 +15,7 @@ val libs = the<LibrariesForLibs>()
   See: https://youtrack.jetbrains.com/issue/KT-66755/Native-non-JVM-targets-add-support-for-languageVersion
  */
 val kotlinLanguageVersion = KotlinVersion.KOTLIN_2_3
-val kotlinApiVersion = KotlinVersion.KOTLIN_2_1
+val kotlinApiVersion = KotlinVersion.KOTLIN_2_3
 val kotlinBomVersion = requireNotNull(libs.kotlin.bom.get().version)
 
 extensions.getByType<KotlinProjectExtension>().apply {
--- a/convention-plugin-ai/src/main/kotlin/ai.kotlin.multiplatform.gradle.kts
+++ b/convention-plugin-ai/src/main/kotlin/ai.kotlin.multiplatform.gradle.kts
@@ -121,9 +121,6 @@ kotlin {
         }
     }
 
-    compilerOptions {
-        coreLibrariesVersion = "2.1.21"
-    }
 }
 
 android {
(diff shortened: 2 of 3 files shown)


serge-p7v:
Hi @hxreborn! Thank you for the PR. It would be great to allow Kotlin versions earlier than 2.3, because some users might prefer not to upgrade to Kotlin 2.3 right now.
It seems it can be done without removal of coreLibrariesVersion by setting in `ai.kotlin.configuration.gradle.kts`:
`val kotlinApiVersion = KotlinVersion.KOTLIN_2_1`

hxreborn:
Hey @serge-p7v, good point. Restored coreLibrariesVersion and set kotlinApiVersion to KOTLIN_2_1 in the configuration script so users on Kotlin 2.1 can still consume the artifacts. Let me know if that looks right

hxreborn:
@serge-p7v I tried both `KOTLIN_2_1` and `KOTLIN_2_2` for kotlinApiVersion but both break compilation. `kotlin.time.ExperimentalTime` (used in `TimeUtils.kt` and elsewhere) only became stable in 2.3, so anything below that needs opt-in annotations across the codebase. Lowering apiVersion alone won't work without source changes. I've reverted to the original approach of removing the `coreLibrariesVersion` override so the POM reflects the actual compiler version. Let me know how you want to handle this.

EugeneTheDev:
Thanks for the fix. These hard-coded `coreLibraryVersion` properties where indeed a leftover from a previous build configs version and shouldn't be there anymore, so good catch. 

> Tried lowering kotlinApiVersion to KOTLIN_2_1 and KOTLIN_2_2 per maintainer feedback, but both break compilation because kotlin.time.ExperimentalTime only became stable in 2.3.

This wouldn't've been the right approach anyway, since it won't work properly for KMP, see [this comment](https://github.com/JetBrains/koog/blob/8d5ae552b9a0927767997ea7ceb47d58b6793a72/convention-plugin-ai/src/main/kotlin/ai.kotlin.configuration.gradle.kts#L12)

EugeneTheDev (approved):

PR #1727, opened 2026-03-24
docs(agents): update wording in Custom nodes

marinkovicmarko:
Update the texts on the Custom nodes page to also cover the newly introduced Java code samples and concepts in the Java API.


innateteniuk (approved):
Looks good to me.
If you decide to go with [this approach](https://github.com/JetBrains/koog/pull/1699#discussion_r2993502606), consider using it here as well.

PR #1740, opened 2026-03-25
fix(agents): Do not throw error for custom MCP transport type in MCPToolRegistryProvider

Ololoshechkin:
Was causing runtime crashes for users with non-default custom MCP transports


sdubov (approved):
Nice! Thank you! @Ololoshechkin, I would just add some debug logging here to make sure we do not miss this case. Also, I would rename the commit - "Remove check for MCP transport type ..." to "Do not throw ... ".

PR #1743, opened 2026-03-25
ci: CI build optimisation

kpavlov:
1. Fix cancel-in-progress bug: `|| → && `so builds on main/develop aren't incorrectly cancelled
  2. Fix memory config: Moved Kotlin daemon/Gradle JVM args to GRADLE_OPTS where they actually take effect; kept JAVA_OPTS at 8GB for the runner
  3. Introduced save/restore steps in GH Workflow for Gradle's build cache to optimize workflow execution time.
  4. Replace assemble with targeted compilation: Single invocation compiling JVM, JS, Android, WASM + lint + test classes — skips only iOS/native targets (validated on macOS)
  5. Fix _checkSplitPackages_: Added `inputs.files(cpConf)` so Gradle auto-depends on upstream JARs, eliminating the separate jvmJar invocation
  6. Add JVM-only mode for faster Dokka builds with draft PRs via `-Pdokka.jvmOnly=true`.
  7. Update `checks.yml` to modify `dokka` dependencies and use JVM-only mode for draft PRs.
  8. Disable Gradle configuration cache explicitly in `gradle.properties` for compatibility.

Results:
- Total duration dropped from ~[30-50-70 minutes](https://github.com/JetBrains/koog/actions/runs/23534402316) down to [19 minutes](https://github.com/JetBrains/koog/actions/runs/23556166141/usage)


.github/workflows/checks.yml
```diff
@@ -16,12 +16,13 @@ on:
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
   # Cancel only when the run is not on main or develop
-  cancel-in-progress: ${{ github.ref != 'refs/heads/main' || github.ref != 'refs/heads/develop' }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' && github.ref != 'refs/heads/develop' }}
```
EugeneTheDev, line 19: nit: I think we can remove `main` from this predicate, since we switched to `develop` a long time ago, and `main` is stale now

.github/workflows/checks.yml
```diff
@@ -66,12 +67,17 @@ jobs:
       # See: https://github.com/gradle/actions/blob/main/setup-gradle/README.md
       - name: Setup Gradle
         uses: gradle/actions/setup-gradle@v5
+        with:
+          cache-read-only: false
 
-      - name: Assemble with Gradle Wrapper
-        run: ./gradlew assemble ktlintCheck
+      - name: Compile and lint
+        run: ./gradlew ktlintCheck compileKotlinJvm compileKotlinJs compileDebugKotlinAndroid compileKotlinWasmJs jvmTestClasses jsTestClasses
```
EugeneTheDev, line 74: I understand that it might save some time, but `assemble` is still the most reliable way to verify that everything builds and that we can publish a new version without any problems. I think we should prioritize stability over speed here, so I'd like to keep `assemble`

.github/workflows/checks.yml
```diff
@@ -133,6 +139,15 @@ jobs:
       # See: https://github.com/gradle/actions/blob/main/setup-gradle/README.md
       - name: Setup Gradle
         uses: gradle/actions/setup-gradle@v5
+        with:
+          cache-read-only: true
+
+      - name: Restore Gradle build cache
+        if: matrix.os == 'ubuntu-latest'
```
EugeneTheDev: Why do you restore the cache on `ubuntu` only? I see that corresponding `Save Gradle build cache` step doesn't have any conditions

gradle.properties
```diff
@@ -9,6 +9,7 @@ kotlin.js.yarn=false
 org.gradle.jvmargs=-Xmx4096M -Dfile.encoding=UTF-8
 org.gradle.parallel=true
 org.gradle.caching=true
+org.gradle.configuration-cache=false
```
EugeneTheDev, line 12: Why did you disable configuraiton cache?

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.14.3-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.14.4-bin.zip
```
EugeneTheDev, line 3: When updating Gradle version in the main Koog project, please don't forget to update Gradle version in examples projects that use composite build too, to avoid weird compatibility problems (`simple-examples`, `simple-examples-java`, all coding agent steps)

EugeneTheDev (changes requested):

gradle.properties
kpavlov: because it is not working with current plugin set. It was de-facto disabled, made it explicit

gradle/wrapper/gradle-wrapper.properties
kpavlov: These are separate projects and should be upgraded separately.

.github/workflows/checks.yml
kpavlov: good catch, make sense to restore it everywhere

.github/workflows/checks.yml
kpavlov: Added `assemble` after tests. Build should fail fast. 18 minutes for compilation is not acceptable.

Rizzen:
Thanks for the improvements here — this looks promising.

One small note on the performance measurements: right now it’s based on a single “before” run (~32 min) and a single “after” run (~19 min). CI timings can vary quite a bit depending on cache state, runner conditions, network, etc., so it’s hard to draw a firm conclusion from just one comparison.

It might be more reliable to compare sequences rather than individual points — e.g., several consecutive runs before vs several after (ideally under similar conditions, with **cache state** noted) and then look at the typical/median duration (or even better - compare distributions or percentiles).

Alternatively, the wording could be softened to reflect that this is an observed improvement rather than a strictly measured one.

Not blocking, just a suggestion 👍

gradle/wrapper/gradle-wrapper.properties
Rizzen: While they are technically separate projects - we wan to use single gradle version across whole repo. We often load these projects along in same IDE instance and it's better to preserve same gradle version in order to save developer's time and mental health.

kpavlov: @Rizzen, you can upgrade the Gradle version in the plugins in a separate pull request. This PR is not about samples.

kpavlov:
@EugeneTheDev, if the current CI process isn’t quite working for you, could we perhaps allocate your time to fix the CI workflow so it completes in around 20 minutes instead of the usual 40-60?

PR #1760, opened 2026-03-27
docs: add a hook to remove blank lines from HTML comments to avoid breaking tab groups and leaking a comment onto a page

EugeneTheDev:
When you have a tab group, e.g. a Kotlin/Java toggle, and you put a codeblock inside, you can't have blank lines inside `INCLUDE` and `SUFFIX` comments. These blank lines are interpreted as a tab content termination, thus breaking a tab making a comment leak onto a page. It's hard to always remember not to have any blank lines in your includes when writing the docs, which leads to UI bugs described above. To address this, an mkdocs hook was added that looks for and removes blank lines from such comments.


innateteniuk (approved):
Thank you!

marinkovicmarko (approved):

PR #1764, opened 2026-03-27
feat(prompt): introduce PromptExecutorHooks and align ContextualPromp…

Amaneusz:
## Background

`ContextualPromptExecutor` wraps any user-provided executor and fires pipeline lifecycle callbacks (`onLLMCallStarting`, `onLLMCallCompleted`, streaming equivalents) consumed by tracing, event-handler, and memory features. Before this PR those callbacks received the model *as requested by the caller* — but `MultiLLMPromptExecutor` and `RoutingLLMPromptExecutor` can silently swap models (fallback, load balancing, routing). So tracing could report "Model A called" while Model B actually served the request.

Root cause: model resolution happens *inside* the executor, after `ContextualPromptExecutor` has already committed what to pass to the callbacks.

## What this PR does

**Introduces per-call hooks on `PromptExecutorAPI`.**

Each method now accepts an optional `hook` parameter:

```kotlin
execute(prompt, model, tools, hook: ExecuteHook? = null)
executeStreaming(prompt, model, tools, hook: StreamingExecutorHook? = null)
executeMultipleChoices(prompt, model, tools, hook: MultipleChoicesHook? = null)
moderate(prompt, model, hook: ModerationHook? = null)
```

Hook interface hierarchy:
- `ExecutorHook` (base) — `onModelChoiceFailed`, `beforeExecution`, `onFailure`
- `SimpleExecutorHook<T>` — adds `onCompleted(result: T)`
- `StreamingExecutorHook` — adds `onFrame`, `onCompleted`

Concrete executors call `beforeExecution` **after** resolving the effective model, so every callback receives accurate data. `ExecutorHooksHelper.executeWithHook` / `streamingWithHook` eliminate the lifecycle boilerplate from each implementation.

**Refactors `ContextualPromptExecutor` to use hooks.**

Instead of wrapping each `execute` call with before/after logic, `ContextualPromptExecutor` now constructs an inline hook per call that fires the pipeline callbacks with the executor-resolved model. Any outer hook passed by the caller is chained after.

**Makes prompt modification explicit.**

Previously `ContextualPromptExecutor` detected pipeline-driven prompt changes by snapshotting `context.llm.prompt` before `onLLMCallStarting` and comparing after. Now `beforeExecution` returns `ExecutionArgOverrides.UseDifferentPrompt` to substitute the prompt. The contract is visible in the type signature rather than relying on mutable shared state.


agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/session/AIAgentLLMReadSessionImpl.kt
```diff
@@ -0,0 +1,231 @@
...
+        return executeStreaming(prompt, tools)
+    }
+
+    override suspend fun requestModeration(moderatingModel: LLModel?): ModerationResult {
+        validateSession()
+        val preparedPrompt = preparePrompt(prompt, emptyList())
+        return executor.moderate(
+            prompt = preparedPrompt,
+            model = moderatingModel ?: model,
+            hooks = null
```
kpavlov: Let’s take out `hooks = null` to ensure this change works with older versions. We need overload methods if it fails

Amaneusz: Updated & fixed :)

prompt/prompt-executor/prompt-executor-model/src/commonMain/kotlin/ai/koog/prompt/executor/model/PromptExecutorAPI.kt
```diff
@@ -21,12 +21,14 @@ public interface PromptExecutorAPI : AutoCloseable {
...
      * @param tools An optional list of `ToolDescriptor` objects that define the tools available for the execution.
+     * @param hooks Optional executor hooks invoked around model resolution and execution lifecycle events.
      * @return A list of `Message.Response` objects representing the responses generated by the LLM.
      */
     public suspend fun execute(
         prompt: Prompt,
         model: LLModel,
-        tools: List<ToolDescriptor> = emptyList()
+        tools: List<ToolDescriptor> = emptyList(),
+        hooks: PromptExecutorHooks? = null
```
sdubov: @Amaneusz, I like the new design for the hooks! I have one question/concern about this approach that I want to discuss. Could you please clarify why do you use hooks as parameter for every method instead of including hooks into the PromptExecutor API itself? I mean something like this:
```
public interface PromptExecutorAPI : AutoCloseable {

    public suspend fun onModelChoiceFailed(intent: InitialExecutionIntent, error: Throwable)

    public suspend fun beforeExecution(intent: InitialExecutionIntent, effectiveModel: LLModel): ExecutionArgOverrides

    public suspend fun onFailure(intent: ResolvedExecutionIntent, effectiveModel: LLModel, error: Throwable)

    public suspend fun execute(...): List<Message.Response>

    public fun executeStreaming(...): Flow<StreamFrame>
```

Updated API should enforce any derived class to add implementation for hooks. Also, it does not need to handle the outer + inner hooks through the methods. You can just wrap the original hooks into a pipeline logic:
```
public class ContextualPromptExecutor(private val executor: PromptExecutor, ... ) {
    ...
    override suspend fun onCompleted(...) {
        context.pipeline.onLLMCallCompleted(...)
        executor.onCompleted() // executor instance is a base prompt executor passed to the contextual one as a parameter
    }
}

```

I looked through current implementations and I found that it might be quite hard to know that you need to wrap your logic into hooks when you make your own `PromptExecutor` implementation. It might be I miss some important design points here. Very welcome to discuss this.

prompt/prompt-executor/prompt-executor-model/src/commonMain/kotlin/ai/koog/prompt/executor/model/PromptExecutorHooks.kt
```diff
@@ -0,0 +1,275 @@
...
+ * Each helper encapsulates the repetitive scaffolding around a single LLM call:
+ * resolving [ExecutionArgOverrides] from [ExecutorHook.beforeExecution], constructing
+ * [ResolvedExecutionIntent], running the actual call inside a try/catch, and dispatching
+ * the appropriate completion or failure callback.
+ *
+ * Executors call these helpers rather than duplicating the lifecycle sequence themselves,
+ * keeping each `execute` / `executeStreaming` override focused on client selection and
+ * the LLM call itself.
+ */
+public object ExecutorHooksHelper {
```
sdubov: Up to your decision - maybe it worth extracting this into a separate file to not mix helper methods and the core hooks logic?

prompt/prompt-executor/prompt-executor-model/src/commonMain/kotlin/ai/koog/prompt/executor/model/PromptExecutorHooks.kt
```diff
@@ -0,0 +1,275 @@
...
+     * Runs [block] with the full [SimpleExecutorHook] lifecycle:
+     * - resolves overrides via [SimpleExecutorHook.beforeExecution]
+     * - builds [ResolvedExecutionIntent],
+     * - calls the block, and notifies [SimpleExecutorHook.onCompleted] or [SimpleExecutorHook.onFailure].
+     *
+     * The [effectiveModel] parameter should be populated by [PromptExecutor] instances that support dynamic model selection
+     * (ie [ai.koog.prompt.executor.llms.MultiLLMPromptExecutor.fallback]).
+     * If given [PromptExecutor] implementation does not support dynamic model selection, it should use initially provided model - [InitialExecutionIntent.model]
+     */
+    public suspend fun <T> executeWithHook(
```
sdubov: It seems for me that this method should be internal or even private. As far as I understand, hooks are very internal logic that should not be exposed outside. Please let me know if I miss anything.

sdubov (commented):
I like the clean design! @Amaneusz , thank you a bunch. I raised a question about the API to discuss. Please let me know what do you think.

prompt/prompt-executor/prompt-executor-model/src/commonMain/kotlin/ai/koog/prompt/executor/model/PromptExecutorAPI.kt
Amaneusz: Hey @sdubov, really good question - I had similar idea at the beginning but it won't work out for us. Let me explain the reasoning behind the design.

tl;dr - there are two main rules to follow:
1) The caller must define the logic of the hook (it's the `Contextutal` that knows about the `pipeline`)                 
2) The executor must define the moment of hook trigger (it's the `Multi`/`Routing` that know when the prompt is actually executed - `execute*` methods are not only about executing, there's more arbitrary logic to them that we can't control)

Longer take:

The core problem this solves: **the caller needs to observe execution facts that only the executor knows** — specifically, which model actually ran. The caller requests a model, but executors can silently substitute it (fallback, routing). By the time `execute()` returns, that information is gone.

Solving this requires two things simultaneously:
- the **caller** defines what to do with those facts (pipeline callbacks, tracing — context the executor has no access to)
- the **executor** fires at the moment the facts exist (after model resolution — a moment invisible from outside)

This is why hooks can't live on the interface as methods. With your proposed design (where `hooks` are owned by executor):

```
contextual.execute(prompt, model, tools)
	nested.execute(prompt, model, tools)
		// `nested` logic before model choice
		val effectiveModel = chooseModel()
		// `nested` logic after model choice
	    nested.onCompleted(effectiveModel, result)
    contextual.OnCompleted // no way to get `effectiveModel` here
```

The inner executor fires its own methods, not the wrapper's. To fix that, you'd need the inner executor to hold a reference back to the outer wrapper — that would require some runtime injection mechanism - just as the on in this PR.

Hooks as a call parameter satisfy both constraints cleanly:

```
contextual.execute(prompt, model, tools, outerHooks)
	nested.execute(..., hooks = ContextualHooks(pipeline, outerHooks))
		// `nested` logic before model choice
		val effectiveModel = chooseModel()
		// `nested` logic after model choice
	    hooks.onCompleted(effectiveModel, result) // hooks allow executors to communicate
	        pipeline.onLLMCallCompleted(effectiveModel, ...)
            outerHooks?.onCompleted(...)
```

The caller provides the logic, the executor decides when to invoke it.

prompt/prompt-executor/prompt-executor-model/src/commonMain/kotlin/ai/koog/prompt/executor/model/PromptExecutorHooks.kt
```diff
@@ -0,0 +1,275 @@
...
+ * The executor invokes the appropriate sub-hook at each lifecycle point, allowing callers
+ * to observe or alter the effective prompt and model without wrapping the executor.
+ *
+ * By default, all hooks are No-Ops. Overriding each hook is opt-in.
+ */
+public interface PromptExecutorHooks {
+
+    /** Hooks for [PromptExecutorAPI.execute]. */
+    public val execute: SimpleExecutorHook<List<Message.Response>>
+        get() = object : SimpleExecutorHook<List<Message.Response>> {}
```
sdubov: Here and below. It would be better to keep these objects in static variables. Currently, they are created on every `get()` invocation.

prompt/prompt-executor/prompt-executor-model/src/commonMain/kotlin/ai/koog/prompt/executor/llms/SingleLLMPromptExecutor.kt
```diff
@@ -35,36 +39,52 @@ public open class SingleLLMPromptExecutor(
...
+        tools: List<ToolDescriptor>,
+        hooks: PromptExecutorHooks?
+    ): List<Message.Response> {
         logger.debug { "Executing prompt: $prompt with tools: $tools and model: $model" }
-        val response = llmClient.execute(prompt, model, tools)
-        logger.debug { "Response: $response" }
-
-        return response
+        return executeWithHook(InitialExecutionIntent(prompt, tools, model), hook = hooks?.execute) { finalIntent ->
+            llmClient.execute(finalIntent.prompt, model, finalIntent.tools)
```
sdubov: Here and below, should not it be `finalIntent.model` in the `llmClient.execute(...)` call?

prompt/prompt-executor/prompt-executor-model/src/commonMain/kotlin/ai/koog/prompt/executor/model/PromptExecutorAPI.kt
sdubov: @Amaneusz, thank you for detailed explanation of the approach and a new API changes. As we discussed it might be worth extracting the `Hookable` interface separately. So, we can use it for `PromptExecutorAPI` and append extra logic to set/get hooks that are used internally. We do not need to change the current PromptExecutor API in that case and append `hooks` parameter for every method. I'm open for discussion here. Please let me know wdyt?

prompt/prompt-executor/prompt-executor-model/src/commonMain/kotlin/ai/koog/prompt/executor/model/PromptExecutorHooks.kt
Amaneusz: Good call, done 👍

prompt/prompt-executor/prompt-executor-model/src/commonMain/kotlin/ai/koog/prompt/executor/model/PromptExecutorHooks.kt
Amaneusz: I thought about it and I think it should stay public. Here is why:
- this PR assumes hooks become a public thing
- this method is just an opt-in helper but it's quite handy to compose the correct lifecycle of PromptExecutor 

On more pragmatic note - I tried to internalize the whole `PromptExecutorHelper` but this require much bigger effort mainly becuase of the structur of our tests (`MockPromptExecutor` specifically). We can consider it but I'd suggest doing that in separate PR if we decide to proceed.

Amaneusz:
I refreshed this PR after a longer PTO and a couple of other interruptions. After getting back to it, I was analyzing if hooks are the right way for us to proceed - again, I was unable to find better alternatives, but I'm open to suggestions and won't mind if we decide to close this without merging if a better design is found

prompt/prompt-executor/prompt-executor-model/src/commonMain/kotlin/ai/koog/prompt/executor/model/PromptExecutorAPI.kt
Amaneusz: I prepared a small doc summarising all alternatives I considered and decided not to choose. The one you're suggesting (if I get it correctly) is the first one (`Separate hookable executor type`).

## Alternative approaches that were not chosen

### Separate hookable executor type

Explored in commit `996b1d5`, reverted in `6963350`. The idea was a distinct interface (e.g. `HookablePromptExecutor`) implemented only by executors that support hooks.

The problem: every executor used inside an agent needs hooks, so every concrete implementation would end up implementing the new interface anyway. It adds a type boundary without adding any real structure — just propagates the required implementation to each class individually with no benefit.

### Hooks as constructor fields

The obvious alternative: attach hooks to the executor at construction time so the nested executor already "knows" about the pipeline.

This doesn't fit Koog's lifecycle. The user constructs `MultiLLMPromptExecutor` independently, before any agent exists. `ContextualPromptExecutor` is created per-run at `prepareContext()`, not at agent construction. Injecting hooks as fields would require mutating the user's executor after construction — once per agent run. That is unsafe under concurrent agent runs sharing the same executor instance: runs would overwrite each other's hooks.

### Event / observer stream

Publish a `PromptExecutorEvent` stream from each executor; `ContextualPromptExecutor` subscribes. Good for decoupling, but event correlation becomes a problem: each call needs a stable identity that survives prompt modification (`Prompt` identity changes when its content changes). An explicit `eventId` would have to be threaded through, adding as much API surface as hooks with worse locality and harder testability.

### Returning the effective model inside the response

Have `LLMClient` embed the resolved model in each response object. That gives `ContextualPromptExecutor` accurate data for completion callbacks.

But `onLLMCallStarting` fires *before* the client call — there is no response to read yet, so the effective model is still unknown at that point. The starting callback would remain inaccurate. Beyond that: this approach requires changing the client contract across every provider integration, and conflates response data with execution metadata.

### Coroutine context injection

The inner executor sets the effective model into a coroutine context key; `ContextualPromptExecutor` reads it on the way out. No new API surface on `PromptExecutorAPI`.

Two problems. First, `onLLMCallStarting` must fire *after* model resolution but *before* the actual client call — the correct ordering for tracing and prompt-modification features. Coroutine context is set inside the nested executor and can only be observed from the outside *after* the call returns, so the "starting" callback would either fire with the wrong model or be dropped entirely. Second, the coupling is implicit: invisible at call sites, non-obvious propagation semantics under structured concurrency (especially with streaming flows), and unit-testing the interaction requires setting up coroutine context state rather than passing a plain hook object.

PR #1593, opened 2026-03-08
feat(prompt): add native structured output (JSON Schema) support for Claude 4.5+ series models across three providers

grantlerduck:
Adds native structured output (JSON Schema) support for Claude 4.5+ series models across three providers.

**Anthropic direct API:**
- Added `output_config` with `json_schema` format to `AnthropicMessageRequest`
- Created `AnthropicBasicJsonSchemaGenerator` and `AnthropicStandardJsonSchemaGenerator`
- Added `LLMCapability.Schema.JSON.Standard` to Haiku 4.5, Sonnet 4.5, Sonnet 4.6, Opus 4.5, Opus 4.6
- Removed the `require(schema == null)` block that previously blocked structured output

**AWS Bedrock:**
- InvokeModel: added `output_config` field to `BedrockAnthropicInvokeModel`
- Converse API: wired `OutputConfig` / `OutputFormat` / `JsonSchemaDefinition` SDK types into request builders

 **OpenRouter:**
 - Added `Claude4_6Sonnet` and `Claude4_6Opus` model definitions with structured output capability

**Documentation:**
- Updated model capabilities table, provider comparison, structured output guide, and Anthropic module docs

closes #1592

PR #1772, opened 2026-03-30
feat(prompt): support structured output for bedrock anthropic models

nathaniellombardi-toast:
Anthropic models on Bedrock have added [support for structured output](https://docs.aws.amazon.com/bedrock/latest/userguide/structured-output.html). In this PR, we add the Json structured output capability to the relevant models and update the prompt to Bedrock Anthropic request conversion to support it.

### DIFF
--- a/prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/bedrock/BedrockModels.kt
+++ b/prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/bedrock/BedrockModels.kt
@@ -125,6 +125,11 @@ public object BedrockModels : LLModelDefinitions {
         LLMCapability.Embed
     )
 
+    private val structuredOutputCapabilities: List<LLMCapability> = listOf(
+        LLMCapability.Schema.JSON.Basic,
+        LLMCapability.Schema.JSON.Standard
+    )
+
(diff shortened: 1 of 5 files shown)


prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/bedrock/BedrockModels.kt
```diff
@@ -134,6 +139,13 @@ public object BedrockModels : LLModelDefinitions {
         LLMCapability.Tools,
     )
 
+    // Add JSON structured output capabilities to model if not present
+    private fun LLModel.withStructuredOutputCapabilities(): LLModel {
```
EugeneTheDev, line 143: Wouldn't it be simpler to just add it manually to each model capabilities list? IMHO such a slight verbosity would make it cleaner and more straightforward

EugeneTheDev (approved):
LGTM, verified that it works with `ai.koog.integration.tests.executor.BedrockConverseApiIntegrationTest#integration_testStructuredOutputNative`. Only a small nit comment

nathaniellombardi-toast:
I see this [other PR](https://github.com/JetBrains/koog/pull/1593) that does the same thing, but with better documentation and across other modes of invocation, like converse api and anthropic itself. Will let the maintainers decide which to go forward with.

EugeneTheDev:
Indeed, I'll close it in favor of #1593 then

PR #1781, opened 2026-03-31
fix(prompt): RetryingLLMClient delegates for json schema generators

nathaniellombardi-toast:
Fixes an issue where the correct JsonSchemaGenerator implementation is lost when using the RetryingLLMClient.


tiginamaria (approved):
Looks good, thank you for the contribution!

PR #1786, opened 2026-04-01
feat(agents): support persistense feature for planner agents

antoniibelyshev:
This PR extends the persistense feature to be available for the planner agents.

Example:

```
val agent = AIAgent(
    promptExecutor = promptExecutor,
    strategy = AIAgentPlannerStrategy("my-strategy", planner),
    agentConfig = agentConfig
) {
    install(Persistence) {
        this.storage = provider
    }
}
```

note: `data class AgentCheckpointData` was updated: fields `nodePath`, `lastInput`, and `lastOutput` were removed (they are now stored in the `properties: JSONObject` which was there before). Since the intended use-case for this class from the user perspective is to only serialize-deserialize it, the users should not be affected by that, therefore I do not mark this as a braking change.

closes [KG-673](https://youtrack.jetbrains.com/issue/KG-673)


agents/agents-features/agents-features-snapshot/src/commonMain/kotlin/ai/koog/agents/snapshot/feature/Persistence.kt
```diff
@@ -322,17 +414,78 @@ public class Persistence(
...
-                messageHistory = prompt.messages,
-                nodePath = agentContext.executionInfo.path(),
-                lastOutput = outputJson,
-                createdAt = Clock.System.now(),
-                version = version,
-            )
+        val checkpoint = AgentCheckpointData(
+            checkpointId = checkpointId ?: Uuid.random().toString(),
+            messageHistory = agentContext.getHistory(),
+            createdAt = Clock.System.now(),
```
Rizzen: please use clock provided by ctor above

agents/agents-features/agents-features-snapshot/src/commonMain/kotlin/ai/koog/agents/snapshot/feature/AgentCheckpointData.kt
```diff
@@ -26,139 +39,210 @@ import kotlin.uuid.Uuid
...
+        )
     )
 
-    init {
-        if (nodePath != PersistenceUtils.TOMBSTONE_CHECKPOINT_NAME) {
-            require(lastInput == null || lastOutput == null) { "`lastInput` and `lastOutput` cannot be both set" }
-            require(lastInput != null || lastOutput != null) { "`lastInput` (until 0.6.0) or `lastOutput` (since 0.6.1) must be set" }
+    @Deprecated("nodePath is deprecated, use properties[\"nodePath\"] instead")
+    public val nodePath: String
+        get() = properties.entries["nodePath"]!!.toKotlinxJsonElement().jsonPrimitive.content
```
Rizzen: Please make proper error message (?: error(message)) on these entries instead of !! operator

Rizzen (changes requested):
Could you please test backward compatibility with JDBC providers as well? We want to be sure that people with old tables be good.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/planner/AIAgentPlanner.kt
```diff
@@ -80,36 +76,64 @@ public abstract class AIAgentPlanner<State : Any, Plan : Any>(
...
                 state.iterations
             }
 
-            val plan = context.with(partName = "buildPlan-${stepIndex + 1}") { executionInfo, eventId ->
-                context.pipeline.onPlanCreationStarting(eventId, executionInfo, context, state, previousPlan, stepIndex + 1)
-                val newPlan = buildPlan(context, state, previousPlan)
-                context.pipeline.onPlanCreationCompleted(eventId, executionInfo, context, state, newPlan, stepIndex + 1)
-                newPlan
+            if (executionPoint == null) {
+                executionPoint = null
```
Rizzen: I dont understand why

agents/agents-features/agents-features-snapshot/src/commonMain/kotlin/ai/koog/agents/snapshot/feature/Persistence.kt
```diff
@@ -133,7 +143,11 @@ public class Persistence(
                 val checkpoint = persistence.rollbackToLatestCheckpoint(ctx.context)
 
                 if (checkpoint != null) {
-                    logger.info { "Restoring checkpoint: ${checkpoint.checkpointId} to node ${checkpoint.nodePath}" }
+                    val nodePath = ctx.context.config.serializer.decodeFromJSONElement<GraphCheckpointProperties>(
+                        checkpoint.properties,
+                        typeToken<GraphCheckpointProperties>()
```
Rizzen: Do we want to always deserialize whole untyped properties json into this particular GraphCheckpointProperties type? That's kinda risky and fragile, because we don't have such contract anywhere.

Maybe we use separate subobject? With json like
```
{
  // any other properties
  ...
  graphCheckpointProperties: {  our GraphCheckpointProperties object  },
  ...
  // any other properties
}
```

Rizzen (approved):

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/planner/AIAgentPlanner.kt
antoniibelyshev: `executionPoint = null` is indeed redundant. Removed it

agents/agents-features/agents-features-snapshot/src/commonMain/kotlin/ai/koog/agents/snapshot/feature/Persistence.kt
antoniibelyshev: Good point, I'll update it

PR #1800, opened 2026-04-05
fix(prompt)!: restore LLMProvider singletons and fix reified type inference

### LINKED ISSUE #1755: Migration challenge: Changes in 4bbe1d6 affect reified type inference and serialization ergonomics
### Description
In commit 4bbe1d6, LLM providers (e.g., `OpenAI`, `Anthropic`) were changed from `object` to `val`. While this addresses some implementation issues, it introduces a significant migration overhead and some API ergonomics issues for downstream users:

1. **Inference Degradation & Reified Type Loss**:
Previously, each provider was a unique singleton `object` type. Passing a provider instance into an `inline` function allowed `reified T` to capture the specific implementation class. Now, because providers are explicitly declared as `val Provider: LLMProvider`, the compiler infers `T` as the abstract class `LLMProvider` itself. This prevents existing code from correctly capturing the concrete type identity at compile-time.

2. **Migration Cost & Implementation Exposure**:
The change from `object` to `val` requires a non-trivial migration for polymorphic serialization:
   - **Serial Name Mismatch**: The underlying class names have changed (e.g., from the object's name to a new implementation class like `SomeProviderClass`). This requires manual mapping to remain compatible with existing persisted data.
   - **Leaky Abstraction**: To maintain type-safe registration, users are now prompted to bypass the public API and manually reference or instantiate internal implementation classes (e.g., `SomeProviderClass()`) instead of using the stable provider properties.

### Examples
Previously, providers could be registered elegantly using type inference:

```kotlin
// This worked when providers were 'object'
inline fun <reified T : LLMProvider> SerializersModuleBuilder.register(provider: T) {
    polymorphic(LLMProvider::class, T::class, serializer<T>())
}

// Usage was clean and correctly inferred:
module.register(LLMProvider.SomeProvider) // T was inferred as the specific object type
```

**Current Migration Requirement:**
The same call now infers `T` as `LLMProvider` (the abstract class). To achieve the same result, users must now expose internal details:
```kotlin
// Now requires referencing the implementation class directly:
module.register(SomeProviderClass()) 
```

### Questions / Suggestions
I respect that this change was made to address some implementation issues. Since this is still in the 0.x stage, I would like to discuss if there are ways to reduce the migration impact:

- **Type Declaration**: Would removing the explicit `: LLMProvider` base class declaration from the `val` properties break any of the intended fixes? Removing it would allow the compiler to infer the concrete implementation type and restore `reified` functionality.
- **Compatibility**: Would restoring the original class names and `object` declarations negatively impact the current implementation? This would maintain stable serial names and avoid forcing users to call internal constructors.
- **Alternative**: Could the library provide a built-in `SerializersModule` that pre-registers these standard instances to shield users from internal implementation changes?

Note: This issue was generated with AI assistance and manually audited for technical correctness. I apologize for any potential errors or phrasing issues; please feel free to point them out.

### Update: Structural changes in serialization
Upon further investigation, I’ve identified an additional structural change: 
Previously, as an `object`, these providers were serialized as empty objects (e.g., `{}`). However, the new `class` implementations expect these fields to be present or handled during the deserialization process. Since existing persisted data does not contain these identifiers, it cannot be directly mapped into the new class structure.

This means the migration would involve more than a name mapping; it would require a custom deserializer to manually handle the transition from the old "empty object" representation to the new class-based structure. This adds a layer of manual logic for anyone maintaining existing persistent storage.

### Update 2: Equality and Identity issues
I’ve also noticed that changing these providers from `object` to `class` affects code relying on equality checks. 

Previously, as a singleton `object`, `LLMProvider.OpenAI == someVariable` worked reliably for identity comparison. Now that it is a `class`, every instantiation (whether through a constructor call or deserialization) creates a new instance. Unless `equals` and `hashCode` are manually overridden to account for the `id` or other properties, existing logic that relies on these equality checks will no longer behave as expected.

xfqwdsj:
This PR refactors `LLMProvider` implementations from `class` back to `object` and introduces explicit type declarations for the provider constants in the companion object.

**Motivation:**
1. **Restore Reified Type Inference**: By declaring constants with their concrete implementation types (e.g., `val OpenAI: OpenAILLMProvider`), the Kotlin compiler can again correctly capture specific types in `inline` functions with `reified` parameters.
2. **Ensure Identity Equality**: Reverting to `object` ensures that identity checks (`==`) work reliably across the system, especially when comparing static constants against instances recreated via deserialization. This prevents potential bugs where `HashMap` lookups or equality checks fail due to different instance references.
3. **Serialization Ergonomics**: Singletons in `kotlinx.serialization` result in smaller payloads (empty objects) and more predictable behavior for shared logic.

BREAKING:
* Existing direct constructor calls to provider classes (e.g., `OpenAILLMProvider()`) must be replaced with the singleton reference (`OpenAILLMProvider`).
* The underlying class structure changes from `class` to `object`, which may require a migration for persisted data if the previous version's serialized form used class-specific fields.

closes #1755

### DIFF
--- a/prompt/prompt-llm/src/commonMain/kotlin/ai/koog/prompt/llm/LLMProvider.kt
+++ b/prompt/prompt-llm/src/commonMain/kotlin/ai/koog/prompt/llm/LLMProvider.kt
@@ -29,7 +29,7 @@ public abstract class LLMProvider(public val id: String, public val display: Str
          * between different LLM providers within the system.
          */
         @JvmField
-        public val OpenAI: LLMProvider = OpenAILLMProvider()
+        public val OpenAI: OpenAILLMProvider = OpenAILLMProvider
 
         /**
          * Represents the Anthropic Large Language Model (LLM) provider.
@@ -39,7 +39,7 @@ public abstract class LLMProvider(public val id: String, public val display: Str
          * Large Language Models and their associated operations.
          */
         @JvmField
-        public val Anthropic: LLMProvider = AnthropicLLMProvider()
+        public val Anthropic: AnthropicLLMProvider = AnthropicLLMProvider
 
         /**
          * Represents the Google provider for Large Language Models (LLMs).
@@ -50,7 +50,7 @@ public abstract class LLMProvider(public val id: String, public val display: Str
          * an LLM instance.
          */
         @JvmField
-        public val Google: LLMProvider = GoogleLLMProvider()
+        public val Google: GoogleLLMProvider = GoogleLLMProvider
 
         /**
          * Represents the Meta large language model (LLM) provider.
@@ -59,7 +59,7 @@ public abstract class LLMProvider(public val id: String, public val display: Str
          * It can be used to configure or interact with Meta-based language models.
          */
         @JvmField
-        public val Meta: LLMProvider = MetaLLMProvider()
+        public val Meta: MetaLLMProvider = MetaLLMProvider
 
         /**
          * Represents the Alibaba Large Language Model (LLM) provider.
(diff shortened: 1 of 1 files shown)


EugeneTheDev (approved):

Rizzen (approved):
Thank you for addressing it!!

PR #1808, opened 2026-04-06
docs: sessions Java snippets

EugeneTheDev:
Add Java code snippets on LLM sessions docs page

### DIFF
--- a/docs/docs/sessions.md
+++ b/docs/docs/sessions.md
@@ -37,7 +37,7 @@ The key difference is that write sessions can modify the conversation history, w
 
 Sessions have a defined lifecycle:
 
-1. **Creation**: a session is created using `llm.writeSession { ... }` or `llm.readSession { ... }`.
+1. **Creation**: a session is created, e.g., using `llm.writeSession { ... }` or `llm.readSession { ... }`.
 2. **Active phase**: the session is active while the lambda block is executing.
 3. **Termination**: the session is automatically closed when the lambda block completes.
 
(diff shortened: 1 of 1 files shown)


docs/docs/sessions.md
```diff
@@ -91,54 +124,119 @@ This ensures that the conversation history is not corrupted by concurrent modifi
...
+                .build();
+        }
+    }
+    -->
+    ```java
+    ctx.getLlm().writeSession(session -> {
+        // Modify the prompt
+        session.appendPrompt(prompt -> {
+            prompt.user("New user message");
+            return null;
```
Ololoshechkin, line 234: Is it required?
Can we do without return null?

docs/docs/sessions.md
```diff
@@ -91,54 +124,119 @@ This ensures that the conversation history is not corrupted by concurrent modifi
...
+        });
+
+        // Modify the tools
+        var updatedTools = session.getTools().stream().toList();
```
Ololoshechkin: Please hide this line in suffix. Even updatedTools = List.of(); // empty list

docs/docs/sessions.md
```diff
@@ -244,137 +413,281 @@ For more advanced use cases, the platform provides methods for structured and st
...
+    }
+    -->
+    ```java
+    ctx.getLlm().writeSession(session -> {
+        // Make a non-tool request
+        var responseWithoutTools = session.requestLLMWithoutTools();
+
+        // Make a streaming request
+        var responseStream = session.requestLLMStreaming();
+        // Process chunks from Flow.Publisher<StreamFrame>
```
Ololoshechkin, line 473: Show this here?

docs/docs/sessions.md
```diff
@@ -244,137 +413,281 @@ For more advanced use cases, the platform provides methods for structured and st
...
+                })
+                .build();
+        }
+    }
+    -->
+    ```java
+    ctx.getLlm().writeSession(session -> {
+        session.appendPrompt(prompt -> {
+            // Add a system message
+            prompt.system("You are a helpful assistant.");
```
Ololoshechkin: Maybe rename "prompt" variable to "promptBuilder" -- it would be less confusing. Here and in other similar snippets

Ololoshechkin (approved):
@EugeneTheDev please check my comments

docs/docs/sessions.md
EugeneTheDev: Unfortunately no, because it's Java and return value is required here, since this lambda has a return type. It can't assume `Unit` implicitly, like Kotlin does

docs/docs/sessions.md
EugeneTheDev: We already explain it on this page (I've added Java docs in a separate PR, they were not deployed yet) https://docs.koog.ai/streaming-api/#working-with-frames-directly. I'll add a link here to this page

PR #1812, opened 2026-04-07
feat(prompt): Support cache control for anthropic provider

tiginamaria:
Anthropic: https://platform.claude.com/docs/en/build-with-claude/prompt-caching

* Supported automatic prompt caching by adding "cache_control" into the request parameters
* Supported explicit cache breakpoints caching by adding "cache_control" into messages
* Added cache tokens to the usage https://platform.claude.com/docs/en/build-with-claude/prompt-caching#tracking-cache-performance

Breaking:
* Removed deprecated `user` message builders to allow one with cacheControl

partially fix KG-707


prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/dsl/PromptBuilder.kt
```diff
@@ -120,38 +120,9 @@ public class PromptBuilder internal constructor(
...
-     *
-     * @param content Content of the user message
-     * @param attachments Attachments to be added to the message
-     */
-    @JavaAPI
-    @Deprecated("Use user(block: ContentPartsBuilder.() -> Unit instead.")
-    public fun user(content: String, attachments: List<ContentPart> = emptyList()): PromptBuilder = apply {
-        user(listOf(ContentPart.Text(content)) + attachments)
+    @JvmOverloads
+    public fun user(content: String, cacheControl: CacheControl? = null): PromptBuilder = apply {
```
antoniibelyshev, line 124: don't we also need `user(cacheControl: CacheControl, block: ContentPartsBuilder.() -> Unit)`?

prompt/prompt-executor/prompt-executor-clients/prompt-executor-anthropic-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/anthropic/AnthropicLLMClient.kt
```diff
@@ -343,6 +349,13 @@ public open class AnthropicLLMClient @JvmOverloads constructor(
         }.requireEndFrame()
     }
 
+    internal fun CacheControl.toAnthropicCacheControl(): AnthropicCacheControlBlock {
+        return when (this.require<AnthropicCacheControl>()) {
```
antoniibelyshev, line 358: will this throw an error when an incorrect cachecontrol is used? if yes, let's switch it to a warning

note: I think, we'll also have to design how to allow setting a cache control for multiple providers at once. when a user wants to use models of anthropic and bedrock in one run and specify caching for both on one message.

antoniibelyshev (approved):
Thank you, looks good! Just a few small comments

prompt/prompt-executor/prompt-executor-clients/prompt-executor-anthropic-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/anthropic/AnthropicLLMClient.kt
tiginamaria: So true, I think we should be able to define LLM params separately for each provider (or even model)

PR #1820, opened 2026-04-08
docs(agents): Add missing Java snippets for AgentEvents

Rizzen:


Ololoshechkin (approved):

sdubov:
Agh. Sorry, it seems I missed that there is already existing PR with this Java doc. I was looking into the same task. I've added a bit more extended API for events that uses more native Java types (I found it is quite strange to relay on generated `Continuation` types from suspend Kotlin functions in Java API):
```
 protected Object processMessage(FeatureMessage message, Continuation<? super Unit> completion)
```
I've fixed this on top of adding and fixing java examples for the `Tracing` feature. @Ololoshechkin, @Rizzen, could you please check this PR as well - https://github.com/JetBrains/koog/pull/1833 ?

aozherelyeva:
Will be covered by changes #1833.

PR #1821, opened 2026-04-08
docs(agents): KG-786. Add Java code snippets for tracing feature

sdubov:
KG-786. Add Java code snippets for tracing feature

- Add convenient Java API for adding message processors in Tracing feature;
- Update the Tracing feature MD file with Java code snippets for corresponding code example blocks.

closes [KG-786](https://youtrack.jetbrains.com/issue/KG-449/issue/KG-786/Add-missing-Java-examples-in-Tracing-documentation)

### DIFF
--- a/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/remote/server/config/DefaultServerConnectionConfig.kt
+++ b/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/remote/server/config/DefaultServerConnectionConfig.kt
@@ -3,6 +3,7 @@ package ai.koog.agents.core.feature.remote.server.config
 import ai.koog.agents.core.feature.remote.server.config.DefaultServerConnectionConfig.Companion.DEFAULT_AWAIT_INITIAL_CONNECTION
 import ai.koog.agents.core.feature.remote.server.config.DefaultServerConnectionConfig.Companion.DEFAULT_HOST
 import ai.koog.agents.core.feature.remote.server.config.DefaultServerConnectionConfig.Companion.DEFAULT_PORT
+import kotlin.jvm.JvmOverloads
 import kotlin.time.Duration
 import kotlin.time.Duration.Companion.seconds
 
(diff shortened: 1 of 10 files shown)


agents/agents-features/agents-features-trace/src/jvmMain/kotlin/ai/koog/agents/features/tracing/writer/TraceFeatureMessageLogWriterJvm.kt
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * This avoids the need to use `KotlinLogging` or `KLogger` directly from Java code.
+ *
+ * Example usage from Java:
+ * ```java
+ * var logger = LoggerFactory.getLogger("tracing");
+ * var writer = TraceFeatureLogWriters.create(logger);
+ * ```
+ */
+public object TraceFeatureMessageLogWriterJvm {
```
tiginamaria: Maybe it's minor, but I don't really like Jvm in the name. Could we also use the expect/actual approach here so we have the same naming but for different targets?

Ololoshechkin: +1, let's remove the "JVM" and move to expect/actual

Ololoshechkin (approved):

agents/agents-features/agents-features-trace/src/jvmMain/kotlin/ai/koog/agents/features/tracing/writer/TraceFeatureMessageLogWriterJvm.kt
sdubov: Folks, it is actually not related to target implementations ( `expected/actual`s ). 

The problem with Java is that there are Java-specific vs Kotlin specific types. For example, the `Sink` is not something typical for Java users. They mostly not use this at all in the code. Instead they work with `OutputStream`. However, our implementation uses the KMP `Sink` class. 

This `TraceFeatureMessageLogWriterJvm` is just a collection of more native Java methods that might be useful for Java users. So, you in Java code will you no Kotlin compiled packages, but instead will use a native Java OutputStream.

@Ololoshechkin , what do you think about this approach?

PR #1825, opened 2026-04-08
fix(prompt): Opus 4.6 maxOutputTokens from 1M to 128K

### LINKED ISSUE #1824: Bug: Opus 4.6 having wrong maxOutputTokens = 1_000_000
Anthropic Opus 4.6 does have the wrong maxOutput tokens set.
When performing a LLM call with the models default it fails with a 400 and very explicit message that the limit is 128,000 tokens (128K).

Anthropic reference, see [Models Overview](https://platform.claude.com/docs/en/about-claude/models/overview)

I will open a quick fix for this in a PR.

grantlerduck:
- Fix `maxOutputTokens` for Claude Opus 4.6 from `1_000_000` to `128_000` in `AnthropicModels` and `OpenRouterModels`
- The incorrect value causes 400 errors from the API when users rely on the model's max output tokens
- Bedrock inherits from `AnthropicModels.Opus_4_6` so it is fixed transitively

closes #1824


aozherelyeva (approved):
Thanks for the catch!

PR #1828, opened 2026-04-09
feat(agents): Moved Persistence to AIAgentStorage

Rizzen:
- Storage-based persistence: Migrated the Persistence feature to use AIAgentStorage (on AIAgentContext) instead of direct context fields, making persistence state accessible through the standard storage API
 
- External storage injection: Added AIAgentSessionInputs — an extensible container that allows pre-populating AIAgentStorage before session execution via session.run(input, AIAgentSessionInputs(storage =    
  ...))

- runFromCheckpoint API: New extension functions on AIAgent and AIAgentRunSession that restore agent execution from a saved AgentCheckpointData without requiring the Persistence feature to be installed. 

- Persistence setExecutionPoint* methods are now suspend to align with the storage API migration


agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/context/AIAgentContext.kt
```diff
@@ -118,6 +118,7 @@ public interface AIAgentContext {
      * @param key A uniquely identifying key of type `AIAgentStorageKey` used to store the feature.
      * @param value The feature to be stored, which can be of any type.
      */
+    @Deprecated("Use context.storage.set() instead", level = DeprecationLevel.WARNING)
```
Ololoshechkin, line 121: Do you think we should deprecate this?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/session/AIAgentRunSession.kt
```diff
@@ -1,10 +1,23 @@
...
 
+/**
+ * Additional inputs that can be provided when running an [AIAgentRunSession].
+ *
+ * This class serves as an extensible container for optional parameters that influence
+ * session execution beyond the primary input.
+ *
+ * @property storage Optional pre-populated storage to use for the session execution.
+ */
+public data class AIAgentSessionInputs(
```
Ololoshechkin: nit: Let's either rename to `AIAgentSessionAdditionalInputs` or alternative move `input: T` to this class as well

Ololoshechkin: Probably, renaming to "AdditionalInputs" is a better option

Ololoshechkin: Also, maybe let's extend to the whole AIAgentContext that one can provide as input (not just storage that is actually a part of it)?
Then it will be also possible to share context between CLI Agents (Claude -> Junie -> Codex), WDYT @sdubov ?

agents/agents-core/src/jvmTest/kotlin/ai/koog/agents/core/agent/AIAgentStoragePassingTest.kt
```diff
@@ -0,0 +1,259 @@
...
+            promptExecutor = getMockExecutor(serializer) { },
+            strategy = storageReadingStrategy(),
+            agentConfig = agentConfig,
+            toolRegistry = ToolRegistry.EMPTY,
+        )
+
+        val session = agent.createSession("test-session")
+        val output = session.run(
+            input = "ignored",
+            sessionInputs = AIAgentSessionInputs(storage = AIAgentStorage()),
```
Ololoshechkin: looking at the usage side, I'd probably name the parameter `additionalContext` or similar.

Also, have you updated the Java API for it? Let's please add a Java test too!

agents/agents-features/agents-features-snapshot/src/commonMain/kotlin/ai/koog/agents/snapshot/feature/Persistence.kt
```diff
@@ -317,7 +317,7 @@ public class Persistence(
      * @param messageHistory The message history to set for the agent
      * @param input The input data to set for the agent
      */
-    public fun setExecutionPoint(
+    public suspend fun setExecutionPoint(
```
Ololoshechkin, line 405: And how to use it from Java now?

agents/agents-features/agents-features-snapshot/src/commonMain/kotlin/ai/koog/agents/snapshot/feature/RunFromCheckpoint.kt
```diff
@@ -0,0 +1,56 @@
...
+ * so that the agent's graph strategy restores execution from the checkpoint's position.
+ * The [Persistence] feature does **not** need to be installed on the agent for this to work.
+ *
+ * @param agentInput The input to provide to the agent.
+ * @param checkpoint The checkpoint data to restore from.
+ * @param rollbackStrategy The strategy to use when restoring state. Defaults to [RollbackStrategy.Default].
+ * @param sessionId Optional session identifier. A random UUID is generated if not provided.
+ * @return The output produced by the agent after resuming from the checkpoint.
+ */
+public suspend fun <Input, Output> AIAgent<Input, Output>.runFromCheckpoint(
```
Ololoshechkin: And how to use it from Java? Let's consider moving this either to `AIAgent` class (or one of it's sub-classes), or creating some helper object or static method for Java (I am thinking about something like `Persistence.runFromCheckpoint(agent, input, checkpoint, ...)` )

Ololoshechkin: Plase also create Java example or/and test

agents/agents-features/agents-features-snapshot/src/jvmTest/java/ai/koog/agents/snapshot/feature/RunFromCheckpointJavaTest.java
```diff
@@ -0,0 +1,200 @@
...
+                new Message.User("User message", new RequestMetaInfo(time, null)),
+                new Message.Assistant("Assistant message", new ResponseMetaInfo(time, null, null, null))
+            ),
+            0L,
+            null
+        );
+
+        AIAgent<String, String> agent = createAgent();
+
+        String output = RunFromCheckpointJvm.runFromCheckpoint(
```
Ololoshechkin: Let's move it from `RunFromCheckpointJvm` (which looks a bit too specific for an object name)  to `Persistence` (as a JvmStatic method) so that it's  `Persistence.runFromCheckpoint(...)` .

Or let's think of some other better place for it. `RunFromCheckpointJvm` looks a bit like a glue code coming from Kotlin equivalent

Ololoshechkin (approved):
@Rizzen please consider my comments about naming and Java.

Also please add documentation about the new functionality in this PR (both Kotlin and Java) to our Persistence docs -- that's also very important

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/context/AIAgentContext.kt
Rizzen: My main concern here - is the fact we have two ways to store something - `context.store()` and `context.storage.store()`. That's a bit confusing for users because they need to think what to choose, but in that case it even doesn't matter because this method is basically delegating storage to AIAgentStorage.

agents/agents-features/agents-features-snapshot/src/commonMain/kotlin/ai/koog/agents/snapshot/feature/Persistence.kt
Rizzen: Basically most of the Persistence methods are suspend. So if we want to make it Java-accessible I'd introduce blocking versions separately

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/session/AIAgentRunSession.kt
Ololoshechkin: upd: let's proceed with just `storage` for now as it's already done here. Later on we can add more fields when (and if) requested by users

Issue #1826, opened 2026-04-09
ChatMemory doesn't use system prompt


kobe2000:
```kotlin
val agent = AIAgent(
        promptExecutor = executor,
        llmModel = OpenAIModels.Chat.GPT4oMini,
        systemPrompt = "You are a helpful assistant.",
        toolRegistry = toolRegistry,
    ) {
        install(ChatMemory) {
            chatHistoryProvider = MyDatabaseProvider() // persistent storage
            windowSize(50)
        }
    }
```

systemPrompt will not be sent to LLM

PR #1832, opened 2026-04-09
test: chat memory prompt preservation

aozherelyeva:
Add a test for chat memory prompt preservation (see #1826).


aozherelyeva:
Covered in #1835

PR #1835, opened 2026-04-09
fix(agents): Fix missed prompt messages when chat memory feature enabled

### LINKED ISSUE #1826: ChatMemory doesn't use system prompt

tiginamaria:
Reason:
When the `ChatMemory` feature was enabled, only messages from history were appended to the initial prompt, while system prompts from `AIAgent` or prompts from `AIAgentConfig` were ignored.

Solution:
When the history is empty (the first-ever agent run), the prompts from the agent configuration are added; only the history is added (assuming this prompt was somewhere near the beginning). Although the solution is debatable, as the other option is always to add this prompt on top of the history, I consider this case unreasonable.

closes https://github.com/JetBrains/koog/issues/1826

### DIFF
--- a/agents/agents-features/agents-features-memory/src/commonMain/kotlin/ai/koog/agents/chatMemory/feature/ChatMemory.kt
+++ b/agents/agents-features/agents-features-memory/src/commonMain/kotlin/ai/koog/agents/chatMemory/feature/ChatMemory.kt
@@ -88,12 +88,17 @@ public class ChatMemory {
         }
 
         private fun installInternal(config: ChatMemoryConfig, pipeline: AIAgentPipeline) {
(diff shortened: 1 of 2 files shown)


Ololoshechkin (approved):
Best

PR #1838, opened 2026-04-10
test: fix streaming test

aozherelyeva:
Use non-reasoning parameters in the streaming test.


tiginamaria (approved):
Thank you for fixing it! It's because I've added checks so that non-reasoning models cannot have reasoning params and haven't thought about the integration tests

PR #1843, opened 2026-04-11
fix(agents): preserve system prompt in ChatMemory across conversation turns

Renanse:
ChatMemory.installInternal replaces all prompt messages with loaded history via withMessages, which drops the system prompt set on the current agent. After the first turn of any conversation with persistent history, the model receives an empty system field in the API request.

Found while building a multi-agent app using Koog + H2JdbcChatHistoryProvider. Specialist agents with detailed system prompts appeared to ignore them entirely after the first conversation turn.

## Fix

**Load side (unconditional):** on subsequent turns, the prompt becomes `currentSystemMessages + historyWithoutSystem`. The live agent's system prompt always takes precedence, and any system messages carried in stored history are ignored on load. First run (empty history) is unchanged.

This is framed as "the live agent owns the system prompt" rather than "delete stored data." The store side never deletes anything by default.

**Store side (opt-in):** added `DropSystemMessagesPreProcessor` and a `ChatMemoryConfig.dropSystemMessages()` config DSL method, in the same shape as the existing `WindowSizePreProcessor` / `windowSize()` and `FilterMessagesPreProcessor` / `filterMessages()`. Users who want stored history free of system messages opt in; everyone else gets the default behavior.

The develop branch has a partial workaround (`historyMessages.ifEmpty { initialMessages + historyMessages }`) that preserves initial messages on the first run but still drops the system prompt on all subsequent runs when history is non-empty.

## Tests

No existing test assertions are modified.

Four new tests added:
- regression: system prompt survives when loaded history has no system message
- live agent's system prompt replaces a stale one carried in history
- `DropSystemMessagesPreProcessor` unit test
- `dropSystemMessages()` keeps stored history free of system messages while multi-turn conversation still works end-to-end

All 43 tests in `ChatMemoryTest` pass locally.

## Revision history

The first revision of this PR mandatorily stripped system messages on store. Per reviewer feedback that this contradicted the feature description, the store side is now untouched by default and the filtering is an opt-in preprocessor.

### DIFF
--- a/agents/agents-features/agents-features-memory/src/commonMain/kotlin/ai/koog/agents/chatMemory/feature/ChatMemory.kt
+++ b/agents/agents-features/agents-features-memory/src/commonMain/kotlin/ai/koog/agents/chatMemory/feature/ChatMemory.kt
@@ -36,6 +36,12 @@ import ai.koog.prompt.message.Message
  *     }
  * }
  * ```
+ *
+ * The system prompt is owned by the live agent: when history is loaded it is restored after the
+ * current agent's system messages and any system messages present in the loaded history are
+ * ignored, so the agent's configured system prompt is never lost across turns. Use
+ * [ChatMemoryConfig.dropSystemMessages] if you also want to keep system messages out of stored
+ * history.
  */
 public class ChatMemory {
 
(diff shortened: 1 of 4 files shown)


agents/agents-features/agents-features-memory/src/commonMain/kotlin/ai/koog/agents/chatMemory/feature/ChatMemory.kt
```diff
@@ -90,20 +90,28 @@ public class ChatMemory {
...
                         }
                     }
                 }
             }
 
             pipeline.interceptStrategyCompleted(this) {
-                val history = it.context.llm.prompt.messages
+                // Strip system messages before storing — the system prompt is set fresh
+                // on each agent creation and should not accumulate in persistent history.
+                val history = it.context.llm.prompt.messages.filter { msg -> msg !is Message.System }
```
tiginamaria: Not sure we want to skip system messages, I think it's better to provide this filtration as an extension to the ChatMemory feature rather than do it mandatory here

tiginamaria (changes requested):
Thank you for your contribution! This is a very important issue you addressed, but maybe the solution should be more extensible (by a special message processor in the feature), as deleting some messages from history contradicts the feature description

agents/agents-features/agents-features-memory/src/commonMain/kotlin/ai/koog/agents/chatMemory/feature/ChatMemory.kt
Renanse: Got it - updates incoming!

Renanse:
Force-pushed a revised approach addressing the review feedback. PR description updated to match.

Highlights:
- Store side is untouched by default. No more mandatory `filter { !is Message.System }` in `interceptStrategyCompleted`.
- Filtering is opt-in via a new `DropSystemMessagesPreProcessor` and a `dropSystemMessages()` config DSL method, mirroring `WindowSizePreProcessor` / `windowSize()`.
- Load side is the real fix and applies unconditionally, but framed as precedence ("live agent owns the system prompt") rather than deletion. Stored history is read-only on load.
- Rebased onto current `develop`, no longer conflicting.
- No existing test assertions modified. Four new tests added.

Ready for another look when you have a chance.

PR #1844, opened 2026-04-12
fix(prompt): use channelFlow in buildStreamFrameFlow to fix Flow inva…

### LINKED ISSUE #1775: IllegalStateException: Flow invariant is violated when using `OllamaClient.executeStreaming`
When attempting to stream a response using `OllamaClient.executeStreaming`, the library throws IllegalStateException due to a Flow invariant violation.
The stack trace indicates that the flow is emitting from an undispatched Ktor continuation context (UndispatchedMarker).

Here's a simple example to reproduce

``` kotlin
val client = OllamaClient()
val response = client.executeStreaming(
    prompt("1") {
        user("Hello! How can you help me?")
    },
    model = OllamaModels.Meta.LLAMA_3_2
)

// The frames are not received here
response.collect { frame ->
    when (frame) {
        is StreamFrame.TextDelta -> println(frame.text)
        else -> {} 
    }
}
```

Every StreamFrame emission fails internally with the following exception, preventing frames from being received by the collector

```
java.lang.IllegalStateException: Flow invariant is violated:
		Flow was collected in [CoroutineId(15895), "coroutine#15895":DispatchedCoroutine{Active}@443f0b13, Dispatchers.IO],
		but emission happened in [CoroutineId(15895), kotlinx.coroutines.UndispatchedMarker@6f6e5114, "coroutine#15895":UndispatchedCoroutine{Active}@484d029d, Dispatchers.IO].
		Please refer to 'flow' documentation or use 'flowOn' instead
```

**Environment**

- Library Version: ai.koog:koog-agents:0.7.3
- Kotlin Version: 2.3.0
- Coroutines Version: 1.10.2
- Platform: Compose Multiplatform (Desktop)

Thanks in advance!

Yelagandula:
## Summary

Fixes #1775

## Problem

`OllamaClient.executeStreaming` throws `IllegalStateException: Flow invariant is violated` when collecting stream frames. The `buildStreamFrameFlow` function delegates to `streamFrameFlow`, which uses the `flow { }` builder. The `flow { }` builder enforces strict context preservation — emission must happen in the same coroutine context as collection. When LLM clients like `OllamaClient` call `client.preparePost(...).execute { }` inside this flow, Ktor's streaming HTTP execution uses an undispatched continuation (`UndispatchedMarker`), creating a different coroutine context and violating the invariant.

## Fix

Replaced `flow { }` with `channelFlow { }` in `buildStreamFrameFlow`. `channelFlow` uses a `Channel` internally and supports emission from any coroutine context. A `FlowCollector` adapter bridges `emit()` calls to `send()` on the channel, so `StreamFrameFlowBuilder` works without changes.

## Impact

This is shared infrastructure used by all LLM clients. The change is safe because `channelFlow` is strictly more permissive than `flow` — any code that worked with `flow` will work with `channelFlow`.

### DIFF
--- a/prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/streaming/StreamFrameFlowBuilder.kt
+++ b/prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/streaming/StreamFrameFlowBuilder.kt
@@ -4,6 +4,8 @@ import ai.koog.prompt.message.ResponseMetaInfo
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.FlowCollector
 import kotlinx.coroutines.flow.asFlow
+import kotlinx.coroutines.channels.ProducerScope
+import kotlinx.coroutines.flow.channelFlow
 import kotlinx.coroutines.flow.flow
 import kotlinx.coroutines.flow.map
 import kotlin.concurrent.atomics.AtomicReference
@@ -115,8 +117,9 @@ public suspend fun FlowCollector<StreamFrame>.emitToolCallComplete(
  * Should be used only in case model does not produce completion events.
  */
 public fun buildStreamFrameFlow(block: suspend StreamFrameFlowBuilder.() -> Unit): Flow<StreamFrame> =
-    streamFrameFlow {
-        val builder = StreamFrameFlowBuilder(this)
+    channelFlow {
+        val collector = FlowCollector<StreamFrame> { frame -> send(frame) }
+        val builder = StreamFrameFlowBuilder(collector)
         block(builder)
     }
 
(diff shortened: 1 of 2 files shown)


tiginamaria (approved):
Thank you for your contribution! Looks good to me, now emit could be done concurrently!

devcrocod (approved):
lgtm

Could you please take care of the formatting issues? You can run `./gradlew ktlintFormat`

Also, if possible, could you please add a regression test

Yelagandula:
Thanks for the review! I've addressed both:

Ran ./gradlew ktlintFormat (removed an unused ProducerScope import)
Added a regression test testBuildStreamFrameFlowSupportsCrossContextEmission that reproduces the original #1775 scenario by emitting from withContext(Dispatchers.Default) inside buildStreamFrameFlow. It passes with the fix and would fail with the old flow { } builder.

PR #1846, opened 2026-04-13
example: Add Kotlin and Java examples for Spring I/O 2026

Ololoshechkin:


examples/spring-io-2026/java/src/main/java/org/example/koog/spring/java/agents/BankingAgentService.java
```diff
@@ -0,0 +1,284 @@
...
+                    config.windowSize(50);
+                })
+                .install(Persistence.Feature, config -> {
+                    config.setStorage(new PostgresJdbcPersistenceStorageProvider(dataSource));
+                    config.setEnableAutomaticPersistence(true);
+                })
+                .install(LongTermMemory.Feature, config -> {
+                    config.retrieval(new LongTermMemory.RetrievalSettingsBuilder()
+                            .withStorage(knowledgeBase)
+                            .withSearchStrategy(new SimilaritySearchStrategy.Builder()
```
Ololoshechkin, line 155: @serge-p7v we should probably add `LongTermMemory.RetrievalSettings.builder()` static method (so that we can use it without `new` as everywhere else) as well as `SimilaritySearchStrategy.builder()`

tiginamaria (approved):

PR #1849, opened 2026-04-13
docs: add Quality Gates guidance to AGENTS.md

aozherelyeva:
1. Add mandatory Quality Gates guidance to AGENTS.md (as discussed on the retro)
2. Update MediaTestScenarios and RetryUtils

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -104,6 +104,67 @@ Features have unique storage keys and can intercept agent lifecycle events.
 - **Type Safety**: Generics ensure compile-time correctness for tool arguments/results
 - **Builder Patterns**: Fluent APIs for configuration throughout the framework
 
+## Quality Gates (MANDATORY)
+
+These rules are non-negotiable and apply to **every** change that touches production code.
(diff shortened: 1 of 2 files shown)


devcrocod:
agents.md needs a complete rewrite because it has become significantly outdated over the past year

It should stay short, since it is included in every request. Once the file grows to 300+ lines, it starts consuming too much context, which negatively affects coding agent results

It is also better not to include behavioral instructions there. Those instructions usually cover only specific use cases, while coding agents are used much more broadly. As a result, such instructions may be useful in some scenarios, but harmful in others: they can reduce output quality, lead to worse decisions, and produce unexpected behavior. With the introduction of skills, behavioral instructions in agents.md have become obsolete

aozherelyeva:
@devcrocod fair point! I moved quality gates to TESTING.md and added a note about it in the AGENTS.md. I think rewriting this file should be done as a separate task. I also removed some of the instructions, so now they're more general.

devcrocod (approved):
lgtm

PR #1860, opened 2026-04-16
feat(agents): KG-793. Extend agent run API to receive a custom AgentResult type

sdubov:
`AIAgent#run()` method return the output type without a context. It limits the ability for chaining Koog agents and sharing context between execution. This PR extend agent run API with additional `runWithResult()` method that return a custom `AgentResult` type that includes `AIAgentContext` property.

closes: [KG-793](https://youtrack.jetbrains.com/issue/KG-793)


agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/session/AIAgentRunSession.kt
```diff
@@ -38,6 +39,17 @@ public interface AIAgentRunSession<Input, Output, TContext : AIAgentContext> {
...
     ): Output
 
+    /**
+     * Executes the agent pipeline with the given context and input, producing a result
+     * that includes both the output and the execution context.
+     *
+     * @param input The input provided to the pipeline during execution.
+     * @return The [AIAgentContextualResult] containing the output and execution context.
+     */
+    public suspend fun runWithResult(
```
antoniibelyshev, line 49: `runWithResult` sounds odd. I would suggest renaming it, e.g. `runWithContextualResult`?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgentResult.kt
```diff
@@ -0,0 +1,16 @@
+package ai.koog.agents.core.agent
+
+/**
+ * Represents the result produced by an AI agent after processing.
+ * This interface provides a generic structure to encapsulate the output generated by the agent.
+ *
+ * @param TOutput The type of the output produced by the AI agent.
+ */
+public interface AIAgentResult<TOutput> {
```
antoniibelyshev, line 9: Why do we need this interface?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgentResultBase.kt
```diff
@@ -0,0 +1,21 @@
...
+ * and the contextual information associated with the agent's execution.
+ *
+ * @param TOutput The type of the output generated by the AI agent.
+ * @param TContext The type of the context, which must extend [AIAgentContext]. This provides details about
+ * the environment, state, configuration, and other metadata during the agent's lifecycle.
+ *
+ * @property output The result output produced by the AI agent after processing.
+ * @property context The contextual information associated with the AI agent's execution.
+ */
+@Serializable
```
antoniibelyshev: context is not serializable. if we want serialization here, we have to extract the serializable properties from the context: message history. however, I would suggest removing the serialization annotation

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
```diff
@@ -37,10 +37,20 @@ public expect abstract class AIAgent<Input, Output> constructor() : Closeable {
...
     public abstract suspend fun run(agentInput: Input, sessionId: String? = null): Output
 
+    /**
+     * Executes the AI agent with the given input and retrieves the result including the execution context.
+     *
+     * @param agentInput The input for the agent.
+     * @param sessionId An optional session ID for the agent run.
+     * @return The [AIAgentResult] containing the output and execution context.
+     */
+    public abstract suspend fun runWithResult(agentInput: Input, sessionId: String? = null): AIAgentResult<Output>
```
antoniibelyshev, line 52: for an abstract AIAgent there's no `Context` (at least, the AIAgent interface does not require to use any context / message history/ etc in the run, it's very generic (and also basically all agents which use a context will be inherited from the AIAgentBase). So what do you expect to see in the result except the output, which can be obtained by the `run`?

antoniibelyshev (changes requested):
Thank you! Left a few comments to discuss

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgentResultBase.kt
tiginamaria: Actually, I'd suggest extracting serializable fields rather than returning the whole context, wdyt? Because we probably don't want to give access to the pipeline or environment, but only agentId, runId, and LLM context. Or I have another suggestion here: I think it makes sense to return +- the same object as we serialize in ChatMemory

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgentResult.kt
sdubov: This interface describe the general API semantic for agent - the agent should return an object of `AgentResult` type. `AIAgentContextualResult` type is a concrete implementation. @antoniibelyshev, please let me know if you have any concerns about it.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgentResultBase.kt
sdubov: I would agree with @antoniibelyshev here. I think the initial idea was that we can pass this result to the outside and store it, but I think it can be fixed separately (cc @Rizzen ).

@tiginamaria , what do you think? Am I miss something from your message?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
sdubov: Currently, the interface contract define that you will receive the property `output` with the object of defined output type.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/session/AIAgentRunSession.kt
sdubov: Hm. The `runWithContextualResult` name describe the implementation details. It is not necessarily your output contains context. Please see the comments below.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
antoniibelyshev: As far as I understand, if you receive a general `AgentResult` from the `agent.run`, you'll have to do something like `result as? AIAgentContextualResult` to then get to the context. It seems strange to me, but I guess there's no better way to do it for now

antoniibelyshev (approved):

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/session/AIAgentRunSession.kt
Rizzen: `runWithX` sounds like you put input X into agent and run it. If you run agent and get X I suggest to use runAndGetX (`runAndGetResult` in that case). It's not the most elegant name but it's straightforward.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgentResult.kt
Rizzen: The only use of this interface as of now - is to cast it to concrete implementation, which is not the best use of interface and creates confusing situations for users

Rizzen (changes requested):
As far as I see - you want to get context after run of the session/agent. 
Why not use following?
```
session.run()
session.context()
```
Or you can introduce ext/util methods for this.

I'd avoid to introduce such method in very core interface.

sdubov:
> As far as I see - you want to get context after run of the session/agent. Why not use following?
> 
> ```
> session.run()
> session.context()
> ```
> 
> Or you can introduce ext/util methods for this.
> 
> I'd avoid to introduce such method in very core interface.

Good point. I think this can works for my case as well. I will close the PR as obsolete then.

PR #1863, opened 2026-04-20
test: fix image checks

aozherelyeva:
1. Rewritten image integration assertions to validate image understanding without relying on exact response wording. 
2. Fixed the JPEG test asset to avoid MIME mismatch failures.


Rizzen (approved):
Maybe it's worth to download images which we store remotely with some retries to avoid possible flakiness due to network?

PR #1864, opened 2026-04-20
refactor(agents): Use SearchRequest interface in LongTermMemory feature 

serge-p7v:
Having concrete type SimilaritySearchRequest in LongTermMemory feature prevents using storages supporting various and/or multiple search types (e.g. keyword, hybrid, etc).

BREAKING:
*

DEPRECATED:
*

closes


tiginamaria (approved):
Thank you! Looks good to me, but now `SimilaritySearch` is not tested (having in-memory does not support it). Any plans for adding tests for that? Or maybe it's better to support it in some way (e.g., with a simple Levenshtein or distance similarity) and test? (can be addressed in a different PR)

serge-p7v:
> Thank you! Looks good to me, but now `SimilaritySearch` is not tested (having in-memory does not support it). Any plans for adding tests for that? Or maybe it's better to support it in some way (e.g., with a simple Levenshtein or distance similarity) and test? (can be addressed in a different PR)

Thanks! There are two shallow tests `similaritySearch builder retrieves matching records` and `ingested data is retrievable in subsequent agent run` that have test-only `InMemorySimilaritySearchStorage`. This storage computes Jaccard coefficient, but the tests don't actually send permutated strings, only exact match.
There will be a much more realistic test with pgvector database in a future PR!

PR #1867, opened 2026-04-20
feat(agents): serializable AIAgentStorage

Rizzen:
Aim of this PR is to make possible to serialize AIAgentStorage to json.

### DIFF
--- a/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/entity/AIAgentStorage.kt
+++ b/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/entity/AIAgentStorage.kt
@@ -2,6 +2,12 @@
 
 package ai.koog.agents.core.agent.entity
 
+import kotlinx.serialization.KSerializer
+import kotlinx.serialization.json.Json
+import kotlinx.serialization.json.JsonElement
+import kotlinx.serialization.json.JsonObject
+import kotlinx.serialization.serializer
+
 /**
  * Represents a storage key used for identifying and accessing data associated with an AI agent.
  *
@@ -14,7 +20,7 @@ package ai.koog.agents.core.agent.entity
  *
  * @param name The human-readable name of the storage key, used only for its string representation.
  */
-public class AIAgentStorageKey<T : Any>(public val name: String) {
+public open class AIAgentStorageKey<T : Any>(public val name: String) {
     override fun toString(): String = "${super.toString()}(name=$name)"
 }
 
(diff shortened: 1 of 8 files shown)


agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/entity/AIAgentStorage.kt
```diff
@@ -22,10 +28,56 @@ public class AIAgentStorageKey<T : Any>(public val name: String) {
  */
 public fun <T : Any> createStorageKey(name: String): AIAgentStorageKey<T> = AIAgentStorageKey(name)
 
+/**
+ * A storage key that carries a [KSerializer] for its associated value type [T].
+ *
+ * When storage is serialized via [AIAgentStorageAPI.serializeToJson], only entries whose key is
+ * a [SerializableStorageKey] are included; all other entries are silently skipped.
+ *
+ * @param name The string identifier that uniquely represents the storage key.
```
serge-p7v: @param name The human-readable name of the storage key, used only for its string representation.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/entity/AIAgentStorage.kt
```diff
@@ -22,10 +28,56 @@ public class AIAgentStorageKey<T : Any>(public val name: String) {
...
+        json.encodeToJsonElement(serializer, value as T)
+
+    internal fun decodeValue(json: Json, element: JsonElement): T =
+        json.decodeFromJsonElement(serializer, element)
+}
+
+/**
+ * Creates a [SerializableStorageKey] with an explicit serializer.
+ *
+ * @param name The name of the storage key, used to uniquely identify it.
```
serge-p7v: @param name The human-readable name of the storage key, used only for its string representation.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/entity/AIAgentStorage.kt
```diff
@@ -22,10 +28,56 @@ public class AIAgentStorageKey<T : Any>(public val name: String) {
...
+public fun <T : Any> createSerializableStorageKey(
+    name: String,
+    serializer: KSerializer<T>,
+): SerializableStorageKey<T> = SerializableStorageKey(name, serializer)
+
+/**
+ * Creates a [SerializableStorageKey] using the reified serializer for [T].
+ * [T] must be annotated with [@Serializable][kotlinx.serialization.Serializable].
+ *
+ * @param name The name of the storage key, used to uniquely identify it.
```
serge-p7v: @param name The human-readable name of the storage key, used only for its string representation.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/entity/AIAgentStorage.kt
```diff
@@ -22,10 +28,56 @@ public class AIAgentStorageKey<T : Any>(public val name: String) {
...
+        json.decodeFromJsonElement(serializer, element)
+}
+
+/**
+ * Creates a [SerializableStorageKey] with an explicit serializer.
+ *
+ * @param name The name of the storage key, used to uniquely identify it.
+ * @param serializer The [KSerializer] for the value type [T].
+ */
+public fun <T : Any> createSerializableStorageKey(
```
Ololoshechkin, line 63: Maybe consider adding an optional `serializer` param to `createStorageKey` (and then either creating non-serializable or serializable one, correspondingly).
And for inline version of `createStorageKey` in Kotlin we can always make it serializable.

For Java -- apparently we also need to make a sibling API (I thought @EugeneTheDev did that previously, so we should just extend it)

Ololoshechkin: Please make sure that JavaUtils.kt in tests are not needed :)

Ololoshechkin (changes requested):
@Rizzen please check my comments above about Java and defaults

Also please create a test (both in Kotlin and Java) that:
1. defines an agent that uses storage normally (using inline `createStorageKey` in kotlin, that deligates to creating serializable storage key; in Java - passes serializer manually).
2. has a strategy that saves to storage and then fails in the middle.
3. uses `install(Persistence) { ... }`
4. recovers again, and then asserts that storage is recovered correctly.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/entity/AIAgentStorage.kt
Rizzen: > Maybe consider adding an optional serializer param to createStorageKey (and then either creating non-serializable or serializable one, correspondingly).
And for inline version of createStorageKey in Kotlin we can always make it serializable.

From my point of view that would create unclear situations for API users. 
We'd have something like that
```
public fun <T : Any> createStorageKey(
    name: String,
    serializer: KSerializer<T>? = null ,
): AIAgentStorageKey<T>
```
And it would create a bit confusing situations
1. If we pass serializer - it return `SerializableStorageKey` but hidden behind `AIAgentStorageKey`
2. If we do not pass serializer - it returns non-serializable key, again as `AIAgentStorageKey`

It could lead to confusing situations when you remove one parameter from method call and suddenly your key is lost during serialization. And you even cannot typecheck it, as they both returned as interface.

I'd prefer to stay on two methods `createStorageKey` and `createSerializableStorageKey` with both returning concrete implementations so you can see the difference in types and names and you also see options when type `AIAgentStorage.(...)`

Rizzen:
TODO redesign

Issue #1866, opened 2026-04-20
OpenAILLMClient.executeResponsesStreaming crashes on 'keepalive' events from Responses API


ren-diao-quikbot:
## Summary

OpenAI's Responses API streams `{"type":"keepalive","sequence_number":N}` events during periods of long-running reasoning (GPT-5 family when `reasoning.effort = medium` or `high`). Koog's `OpenAIStreamEvent` sealed class has no `@SerialName("keepalive")` subtype, so kotlinx-serialization's polymorphic deserializer throws on the first keepalive event and the entire agent run dies with:

```
ai.koog.http.client.KoogHttpClientException: Error from client: OpenAILLMClient
Message: Unexpected JSON token at offset 0: Serializer for subclass 'keepalive' is not found in the polymorphic scope of 'OpenAIStreamEvent' at path: $
Check if class with serial name 'keepalive' exists and serializer is registered in a corresponding SerializersModule.
To be registered automatically, class 'keepalive' has to be '@Serializable', and the base class 'OpenAIStreamEvent' has to be sealed and '@Serializable'.
JSON input: {"type":"keepalive","sequence_number":3}
Status code: 200
```

## Reproduction

Using Koog 0.8.0 (also confirmed on `develop` HEAD `51fe749` at time of writing):

```kotlin
val agent = AIAgent(
    promptExecutor = SingleLLMPromptExecutor(OpenAILLMClient(apiKey = env("OPENAI_API_KEY"))),
    strategy = chatAgentStrategy(),
    agentConfig = AIAgentConfig(
        prompt = prompt(
            id = "keepalive-repro",
            params = OpenAIResponsesParams(
                reasoning = ReasoningConfig(
                    effort = ReasoningEffort.MEDIUM,
                    summary = ReasoningSummary.AUTO,
                ),
            ),
        ) { system("You are a careful mathematician. Work step by step.") },
        model = OpenAIModels.Chat.GPT5Mini,
        maxAgentIterations = 5,
    ),
    toolRegistry = ToolRegistry.EMPTY,
)

agent.run(
    "Carefully calculate 947 × 823 × 467 step by step. " +
    "Show every intermediate calculation and verify by alternative methods."
)
```

Fires within ~3-10 seconds as OpenAI sends the first keepalive event during the reasoning phase.

## Current behaviour

[...]

PR #1868, opened 2026-04-20
fix(prompt): Fix reasoning for openai and google clients

### LINKED ISSUE #1865: GoogleLLMClient.executeStreaming drops Gemini thought parts and emits them as TextDelta (inconsistent with non-streaming path and OpenAI streaming)
_Note: CONTRIBUTING asks for YouTrack for bug reports, but since closely-related issues (#238, #1148, #1264) were filed and resolved here on GitHub, I'm following precedent. Happy to re-file on YouTrack if preferred._

## Summary

When a Gemini thinking model (e.g. `gemini-3-pro-preview`, `gemini-3.1-pro-preview`) is invoked via `GoogleLLMClient.executeStreaming` with `GoogleThinkingConfig(includeThoughts = true)`, the Google API returns `parts` with `thought: true` interleaved among the visible output. The current streaming handler has **no branch for thought-flagged parts** — it silently drops them or emits them as regular `StreamFrame.TextDelta`, mixing model reasoning into the visible assistant reply.

This is inconsistent with:
1. **Koog's own non-streaming path** (`processGoogleCandidate`), which correctly produces `Message.Reasoning` from thought-flagged parts.
2. **The OpenAI equivalent** (issue #1264), where the fix added `StreamFrame.ReasoningDelta` emission for `ResponseReasoningSummaryTextDelta` events — so `gpt-5`-family models already stream reasoning correctly since that fix landed.

Gemini 3 Pro models declare `LLMCapability.Thinking`, and with #238 + #1148 already fixed, thought serialization and thought-signature multi-turn correctness both work. The remaining gap is **surfacing thought content as `ReasoningDelta` on the streaming path**, so UIs can render thinking separately from the final reply.

## Reproduction

Using Koog 0.8.0 (confirmed same code on `develop` at time of writing):

```kotlin
val agent = AIAgent(
    promptExecutor = SingleLLMPromptExecutor(GoogleLLMClient(apiKey = env("GEMINI_API_KEY"))),
    strategy = chatAgentStrategy(),  // or any strategy using requestLLMStreaming()
    agentConfig = AIAgentConfig(
        prompt = prompt(id = "thinking-demo", params = GoogleParams(
            thinkingConfig = GoogleThinkingConfig(
                thinkingLevel = GoogleThinkingLevel.HIGH,
                includeThoughts = true,  // <-- asks Gemini to return thoughts
            ),
        )) { system("You are helpful.") },
        model = GoogleModels.Gemini3_Pro_Preview,
        maxAgentIterations = 5,
    ),
    toolRegistry = ToolRegistry.EMPTY,
) {
    handleEvents {
        onLLMStreamingFrameReceived { ctx ->
            when (val f = ctx.streamFrame) {
                is StreamFrame.ReasoningDelta -> println("REASONING: ${f.text}")
                is StreamFrame.TextDelta       -> println("TEXT:      ${f.text}")
                else -> {}
            }
        }
    }
}

agent.run("Explain step by step why the sky is blue.")
```

**Expected**: `REASONING: …` lines during the thinking phase, then `TEXT: …` lines for the visible answer.

**Actual**: All output arrives as `TEXT: …` lines — the thought text is interleaved with the answer text, because the thought flag is discarded. If only `thinkingLevel` is set (no `includeThoughts`), nothing thought-related is visible at all.

## Current behaviour

In [`prompt-executor-google-client/.../GoogleLLMClient.kt`](https://github.com/JetBrains/koog/blob/develop/prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClient.kt) `executeStreaming`:

```kotlin
candidate.content?.parts?.forEachIndexed { index, part ->
    when (part) {
        is GooglePart.FunctionCall -> {
            emitToolCallDelta(
                id = part.functionCall.id,
                name = part.functionCall.name,
                args = part.functionCall.args?.toString() ?: "{}",
                index = index,
            )
        }

        is GooglePart.Text -> {
            emitTextDelta(part.text, index)   // thought flag is ignored here
        }

        else -> Unit                          // (and Thought Signatures land here silently)
    }
}
```

`GooglePart.Text` carries `val thought: Boolean? = null` and `val thoughtSignature: String? = null`. Both are dropped on the streaming path.

## Expected behaviour

Mirror `processGoogleCandidate` (non-streaming) — when `part.thought == true`, emit `StreamFrame.ReasoningDelta` carrying the text (and thoughtSignature as id if useful), not `StreamFrame.TextDelta`.

## Proposed fix

Add a thought branch at the top of the `when (part)`:

```kotlin
when (part) {
    is GooglePart.Text -> {
        if (part.thought == true) {
            emitReasoningDelta(
                id = part.thoughtSignature,
                text = part.text,
                index = index,
            )
        } else {
            emitTextDelta(part.text, index)
        }
    }

    is GooglePart.FunctionCall -> {
        emitToolCallDelta(
            id = part.functionCall.id,
            name = part.functionCall.name,
            args = part.functionCall.args?.toString() ?: "{}",
            index = index,
        )
    }

    else -> Unit
}
```

Optionally also emit a terminal `StreamFrame.ReasoningComplete` once `candidate.finishReason` arrives, mirroring the OpenAI Responses streaming shape that #1264's fix landed.

## Why this matters

Without this fix, consumer UIs that model reasoning as a separate visual channel cannot render Gemini's thinking. Enabling `includeThoughts = true` silently pollutes the assistant reply with reasoning text; leaving it off means no visible thinking at all. There's no middle ground with the current streaming handler.

## Environment

- Koog version: `0.8.0` (also confirmed on `develop` HEAD `51fe749` as of 2026-04-20)
- Kotlin: 2.3.20 / JVM 21
- Tested models: `gemini-3-pro-preview`, `gemini-3.1-pro-preview`
- Platform: JVM

## References

- #238 — Support thought message from Google models (serialization) — fixed 2025-11
- #1148 — Support Gemini 3 (thought_signature on function calls) — fixed 2025-12
- #1264 — OpenAI Responses Reasoning Summary streaming — fixed 2026-02; **direct analogue for OpenAI, this issue is the Google-side equivalent gap**
- Gemini thought-signature docs: https://ai.google.dev/gemini-api/docs/thought-signatures
- Gemini thinking docs: https://ai.google.dev/gemini-api/docs/thinking

## Happy to PR

If the fix shape is acceptable, I can open a PR against `develop` with the thought branch + an `executeStreaming` integration test along the lines of the `Gemini 2.5 Pro preview` test profile referenced in #238.

tiginamaria:
* Support keepalive event from openai streaming
* Add condition to check if the text from google is reasoning or plain text?

closes #1865 #1866

### DIFF
--- a/prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClient.kt
+++ b/prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClient.kt
@@ -210,17 +210,25 @@ public open class GoogleLLMClient @JvmOverloads constructor(
                 response.candidates.firstOrNull()?.let { candidate ->
                     candidate.content?.parts?.forEachIndexed { index, part ->
                         when (part) {
+                            is GooglePart.Text -> {
+                                if (part.thought == true) {
+                                    emitReasoningDelta(
+                                        id = part.thoughtSignature,
+                                        text = part.text,
+                                        index = index,
+                                    )
+                                } else {
+                                    emitTextDelta(part.text, index)
+                                }
+                            }
+
                             is GooglePart.FunctionCall -> {
                                 emitToolCallDelta(
                                     id = part.functionCall.id,
                                     name = part.functionCall.name,
                                     args = part.functionCall.args?.toString() ?: "{}",
-                                    index = index
+                                    index = index,
                                 )
-                            }
-
-                            is GooglePart.Text -> {
-                                emitTextDelta(part.text, index)
                             }
 
                             else -> Unit
--- a/prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/jvmTest/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClientTest.kt
+++ b/prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/jvmTest/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClientTest.kt
@@ -3,12 +3,15 @@ package ai.koog.prompt.executor.clients.google
 import ai.koog.agents.core.tools.ToolDescriptor
 import ai.koog.agents.core.tools.ToolParameterDescriptor
 import ai.koog.agents.core.tools.ToolParameterType
+import ai.koog.http.client.KoogHttpClient
 import ai.koog.prompt.dsl.Prompt
 import ai.koog.prompt.executor.clients.google.models.GoogleCandidate
 import ai.koog.prompt.executor.clients.google.models.GoogleContent
 import ai.koog.prompt.executor.clients.google.models.GoogleData
 import ai.koog.prompt.executor.clients.google.models.GoogleFunctionCallingMode
 import ai.koog.prompt.executor.clients.google.models.GooglePart
+import ai.koog.prompt.executor.clients.google.models.GoogleRequest
+import ai.koog.prompt.executor.clients.google.models.GoogleResponse
 import ai.koog.prompt.executor.clients.google.models.GoogleThinkingConfig
 import ai.koog.prompt.message.AttachmentContent
 import ai.koog.prompt.message.ContentPart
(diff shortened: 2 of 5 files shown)


antoniibelyshev (approved):
Thank you, the fix looks good! Can we also add a test for the reasoning?

aozherelyeva:
I'll add the tests in a bit 

UPD: done

PR #1869, opened 2026-04-20
example: Remove Calculator V2 Example References

horatiothomas:
This example seems to no longer exist but is still referenced. Removing it to cleanup and prevent confusion.


aozherelyeva (approved):
Good catch, thank you!

PR #1871, opened 2026-04-21
fix(prompt): withPrompt method now uses writeLock instead of readLock

serge-p7v:
withPrompt method now uses writeLock instead of readLock

BREAKING:
*

DEPRECATED:
*

closes


antoniibelyshev (approved):

aozherelyeva (approved):
Thanks!

PR #1882, opened 2026-04-23
refactor(agents): replace invoke constructors with factory methods

antoniibelyshev:
Replaced the pattern
```
class A {
    companion object {
        operator fun invoke(...) = /* create an instance of A */
    }
}
```
with
```
class A { ... }
fun A(...) = /* create an instance of A */
```

the usage does not change:
```
import x.y.z.A

val a = A()
```
but the resolution is clearer with the updated factories.

updated classes:
`AIAgent`, `AIAgentService`, `ToolRegistry`, `RollbackToolRegistry`, `AIAgentPlannerStrategy`.

BREAKING:
The unintended usage will break, e.g. these would compile previously, but will not compile now
```
A.Companion.invoke()
A.invoke()
A.Companion()
etc.
```
now only `A(...)` will be allowed.


agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
```diff
@@ -53,239 +43,15 @@ public expect abstract class AIAgent<Input, Output> constructor() : Closeable {
...
-         * @param agentConfig The configuration for the AI agent, including the prompt, model, and other parameters.
-         * @param strategy The strategy for executing the AI agent's graph logic, including workflows and decision-making.
-         * @param toolRegistry The registry of tools available for use by the agent. Defaults to an empty registry.
-         * @param id Unique identifier for the agent. Random UUID will be generated if set to null.
-         * @param clock The clock to be used for time-related operations. Defaults to the system clock.
-         * @param installFeatures A lambda expression to install additional features in the agent's feature context. Defaults to an empty implementation.
-         * @return An instance of an AI agent configured with the specified parameters and capable of executing its logic.
-         */
-        @OptIn(ExperimentalUuidApi::class)
-        public inline operator fun <reified Input, reified Output> invoke(
```
devcrocod, line 84: Did you intentionally remove these operators without deprecating them?

agents/agents-tools/src/commonMain/kotlin/ai/koog/agents/core/tools/ToolRegistry.kt
```diff
@@ -31,6 +31,13 @@ import kotlin.jvm.JvmStatic
  */
 public class ToolRegistry internal constructor(tools: List<Tool<*, *>> = emptyList()) {
 
+    /**
+     * Creates a new ToolRegistry using the provided builder initialization block.
+     *
+     * @param init A lambda that configures the registry by adding tools
+     */
+    public constructor(init: ToolRegistryBuilder.() -> Unit) : this(ToolRegistryBuilder().apply(init).build().tools)
```
devcrocod, line 39: some specific logic in `build`?
I think you can take tools without calling `build`?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgentServiceFactories.kt
```diff
@@ -0,0 +1,127 @@
...
+ * @param Output The output type that the service produces.
+ * @param promptExecutor The executor responsible for processing AI prompts and responses.
+ * @param agentConfig Configuration parameters for the AI agent.
+ * @param strategy A strategy defining the graph structure for AI agent interactions and processing.
+ * @param toolRegistry The registry of tools available to the agent. Defaults to an empty registry.
+ * @param installFeatures A lambda expression to install additional features in the agent's feature context.
+ * @return A [GraphAIAgentService] instance configured with the provided parameters.
+ */
+@OptIn(InternalAgentsApi::class)
+public inline fun <reified Input, reified Output> AIAgentService(
```
devcrocod: How would this look from Java?
Maybe it would be worth adding `@JvmName`

devcrocod (commented):
Good PR 👍 
left a few comments

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
antoniibelyshev: There is only a minor break in API, related to unintended uses of explicit `invoke` or `Companion`, so I decided not to keep deprecations here.

devcrocod: I'm not sure I fully understand about minor break
invoke operator is imported, so removing it would introduce source incompatibility for all users

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgentServiceFactories.kt
antoniibelyshev: These factories were not intended for use from java anyway (they were not marked @JvmStatic). Also, they are very inconvenient in java. So I think we should keep it this way and only expose `AIAgent.builder` for java

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
antoniibelyshev: I did not think about this, I'll keep them as deprecated then

agents/agents-tools/src/commonMain/kotlin/ai/koog/agents/core/tools/ToolRegistry.kt
antoniibelyshev: Actually, `builder.tools` is private and not defined in the `commonMain`, so let's keep the `build()` here, as it was here before the refactoring

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
EugeneTheDev: But we're going to remove deprecated before 1.0 anyway, so IMHO it's fine to remove them now

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
```diff
@@ -81,6 +80,7 @@ public expect abstract class AIAgent<Input, Output> constructor() : Closeable {
          * @return An instance of an AI agent configured with the specified parameters and capable of executing its logic.
          */
         @OptIn(ExperimentalUuidApi::class)
+        @Deprecated("Use AIAgent(...)")
```
EugeneTheDev: Let's remove deprecated instead of keeping them, since we're going to remove them before 1.0 anyway

EugeneTheDev (approved):

PR #1883, opened 2026-04-23
fix(agents): consider tool's customName in tools from callable

### LINKED ISSUE #1881: @Tool(customName = ...) is ignored when using ToolSet.asTools() / ToolRegistryBuilder.tools(ToolSet)
## `@Tool(customName = ...)` is ignored when using `ToolSet.asTools()` / `ToolRegistryBuilder.tools(ToolSet)`

**Version:** `ai.koog:agents-tools-jvm:0.8.0`

---

### Description

The `customName` property of the `@Tool` annotation is silently ignored when tools are registered via `ToolSet.asTools()` or `ToolRegistryBuilder.tools(instance: ToolSet)`.

When two or more `ToolSet` implementations declare a method with the same name (a very common pattern — e.g. `execute`, `search`, `run`), even though each method carries a distinct `@Tool(customName = "...")`, they all end up registered under the raw Kotlin function name. This causes a crash at application startup:

```
java.lang.IllegalArgumentException: Tool "execute" is already defined
    at ai.koog.agents.core.tools.ToolRegistryBuilderKt.addTool(ToolRegistryBuilder.kt:27)
```

---

### Steps to Reproduce

```kotlin
class ToolA : ToolSet {
    @Tool(customName = "tool_a_search")
    @LLMDescription("Searches source A")
    fun execute(query: String): String = "result A"
}

class ToolB : ToolSet {
    @Tool(customName = "tool_b_search")
    @LLMDescription("Searches source B")
    fun execute(query: String): String = "result B"
}

// Throws: IllegalArgumentException: Tool "execute" is already defined
val registry = ToolRegistry {
    tools(ToolA())
    tools(ToolB())
}
```

---

### Root Cause

**File:** `jvmCommonMain/ai/koog/agents/core/tools/reflect/ToolFromCallableExtensions.kt`  
**Function:** `KClass<T>.asTools(thisRef)`

The function correctly filters methods annotated with `@Tool`, but **never reads `customName`** when constructing the `ToolFromCallable`:

```kotlin
// Current (broken) implementation
public fun <T : Any> KClass<out T>.asTools(thisRef: T? = null): List<ToolFromCallable<*>> {
    return this.functions
        .filter { m -> m.getPreferredToolAnnotation() != null }
        .map {
            it.asTool(thisRef = thisRef)   // ← name is always null, customName is lost
        }
}
```

`asTool()` passes `name = null` to `ToolFromCallable`, which then calls:

```kotlin
// File: jvmCommonMain/ai/koog/agents/core/tools/schema/SchemaGenerator.jvmCommon.kt
// Function: getToolDescriptor(...)
return ToolDescriptor(
    name = toolName ?: schema.name,   // toolName is null → falls back to the raw function name
    ...
)
```

So `schema.name` — the raw Kotlin function name (e.g. `"execute"`) — is used instead of `customName`.

---

### Expected Behavior

`customName` (when non-empty) should be extracted from the `@Tool` annotation and forwarded to `asTool()`.

**Suggested fix in `ToolFromCallableExtensions.kt`:**

```kotlin
public fun <T : Any> KClass<out T>.asTools(thisRef: T? = null): List<ToolFromCallable<*>> {
    return this.functions
        .filter { m -> m.getPreferredToolAnnotation() != null }
        .map {
            val customName = it.getPreferredToolAnnotation()?.customName?.ifEmpty { null }
            it.asTool(thisRef = thisRef, name = customName)
        }
}
```

Note: `getPreferredToolAnnotation()` is already a private function in the same file, so no additional infrastructure is needed.

---

### Workaround

Until this is fixed, `customName` can be applied manually by bypassing `tools(ToolSet)`:

```kotlin
val registry = ToolRegistry {
    listOf(ToolA(), ToolB()).forEach { toolSet ->
        toolSet::class.functions
            .filter { it.findAnnotation<Tool>() != null }
            .forEach { func ->
                val customName = func.findAnnotation<Tool>()?.customName?.ifEmpty { null }
                tool(func, toolSet, customName)
            }
    }
}
```

---

### Files to Fix

| File | Change needed |
|------|---------------|
| `jvmCommonMain/ai/koog/agents/core/tools/reflect/ToolFromCallableExtensions.kt` | Read `customName` from `@Tool` annotation and pass it as `name` to `asTool()` in `KClass<T>.asTools()` |
| `jvmCommonMain/ai/koog/agents/core/tools/reflect/ToolFromCallableExtensions.kt` | Same fix applies to the inline `asToolsByClass<T>()` extension if it delegates to `KClass.asTools()` |

aozherelyeva:
Fix the big "@Tool(customName = ...) is ignored when using ToolSet.asTools() / ToolRegistryBuilder.tools(ToolSet)" and add a test.

closes #1881

### DIFF
--- a/agents/agents-tools/src/jvmCommonMain/kotlin/ai/koog/agents/core/tools/reflect/ToolFromCallableExtensions.kt
+++ b/agents/agents-tools/src/jvmCommonMain/kotlin/ai/koog/agents/core/tools/reflect/ToolFromCallableExtensions.kt
@@ -65,7 +65,8 @@ public fun <T : Any> KClass<out T>.asTools(
     return this.functions.filter { m ->
         m.getPreferredToolAnnotation() != null
     }.map {
-        it.asTool(thisRef = thisRef)
+        val customName = it.getPreferredToolAnnotation()?.customName?.ifEmpty { null }
+        it.asTool(thisRef = thisRef, name = customName)
     }.apply {
         require(isNotEmpty()) { "No tools found in ${this@asTools}" }
     }
--- a/agents/agents-tools/src/jvmTest/kotlin/ai/koog/agents/core/tools/reflect/ToolSetAsToolsTest.kt
+++ b/agents/agents-tools/src/jvmTest/kotlin/ai/koog/agents/core/tools/reflect/ToolSetAsToolsTest.kt
@@ -1,5 +1,6 @@
 package ai.koog.agents.core.tools.reflect
 
+import ai.koog.agents.core.tools.ToolRegistry
 import ai.koog.agents.core.tools.annotations.InternalAgentToolsApi
 import ai.koog.agents.core.tools.annotations.LLMDescription
 import ai.koog.agents.core.tools.annotations.Tool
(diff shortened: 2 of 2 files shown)


antoniibelyshev (approved):
Thank you! LGTM

PR #1884, opened 2026-04-23
fix(prompt): stop additionalProperties leaking as additional_properties to OpenAI

### LINKED ISSUE #1878: OpenAI Responses: additional_properties request field causes 400 unknown_parameter
Firstly, thank you for adding support for gpt-5.4-nano with the last issue i opened.

## Summary

When our repo image-analysis adapter uses Koog's OpenAI Responses path, the outbound request includes a top-level `additional_properties` field and OpenAI rejects it with `400 invalid_request_error` / `unknown_parameter`. The same image prompt sent through a bare Koog path passes, which makes this look like a Koog serialisation issue triggered by the adapter's `LLMParams(additionalProperties = ...)` usage rather than an OpenAI model-specific behaviour.

Possibly relevant prior Koog issues touching request shape or additional-properties handling: #1624, #1626, #998, and #1022.

## Environment

- Koog dependency: `ai.koog:prompt-executor-openai-client:0.8.0`
- Runtime: Kotlin/JVM 21, Gradle
- Provider: OpenAI Responses API
- Models exercised in the live parity test: `gpt-4.1-mini` and `gpt-5.4-nano`

## What we tested

We ran a live parity check against the same provider/base URL with the same PNG bytes:

1. **Bare Koog image path**: direct `OpenAILLMClient.execute(...)` call with an image prompt and no `LLMParams(additionalProperties = ...)`.
2. **Repo adapter path**: `OpenAiImageAnalysisClient.analyzeImage(...)`, which builds the same image prompt but sets:

```kotlin
params =
    LLMParams(
        additionalProperties =
            mapOf(
                "reasoning" to
                    buildJsonObject {
                        put("effort", JsonPrimitive("none"))
                    },
            ),
    )
```

## Reproduction matrix

| Path | Transport | Endpoint | Result |
| --- | --- | --- | --- |
| Bare Koog image path | `OpenAILLMClient.execute(...)` | OpenAI Responses | Pass |
| Repo adapter path | `OpenAiImageAnalysisClient.analyzeImage(...)` | OpenAI Responses | Fail: `400 unknown_parameter` for `additional_properties` |

In our live parity test, the configured repo model showed `bareKoog=Pass` and `repoAdapter=AdapterUnknownParameter`. Control-model runs behaved the same whenever the bare Koog path was provisioned and passed.

## Actual result

OpenAI rejects the repo adapter request with:

```json
{
  "error": {
    "message": "Unknown parameter: 'additional_properties'.",
    "type": "invalid_request_error",
    "param": "additional_properties",
    "code": "unknown_parameter"
  }
}
```

The adapter surfaces this as `IllegalStateException: OpenAI image analysis request failed` wrapping the Koog HTTP client exception.

## Expected result

Koog should not serialise unsupported top-level request fields for the OpenAI Responses API. If `LLMParams.additionalProperties` is intended only for internal prompt metadata or for a different endpoint shape, it should be ignored, mapped differently, or rejected locally before sending the request. The repo adapter path should behave consistently with the bare Koog path and complete successfully.

## Why this looks like Koog serialisation rather than OpenAI/model behaviour

- The provider is rejecting the request shape, not the prompt content.
- The failing parameter name is the serialised top-level field `additional_properties`.
- The same image bytes and same provider path pass through a bare Koog call.
- The failure reproduces across more than one model whenever the bare Koog control is provisioned, so it does not look model-specific.
- The only relevant behavioural difference in the repo adapter path is the use of `LLMParams(additionalProperties = mapOf("reasoning" to ...))`, which appears to be emitted into the OpenAI request body.

## Minimal reproduction

1. Create an OpenAI Responses-capable `OpenAILLMClient`.
2. Build an image prompt with `LLMParams(additionalProperties = mapOf("reasoning" to {"effort":"none"}))`.
3. Execute it against OpenAI Responses.
4. Observe that the outgoing request is rejected for unknown top-level parameter `additional_properties`.

Removing the `additionalProperties` usage, or using the bare Koog path that does not set it, avoids the failure in our live test.

zxuhan:
## Summary
- `AdditionalPropertiesFlatteningSerializer` looked up the flatten key by its declared form only, so under `JsonNamingStrategy.SnakeCase` (used by every OpenAI-compatible client) the key became `additional_properties` before `transformSerialize` ran, the filter missed it, and the whole map leaked to the request root — OpenAI then rejected the call with `400 unknown_parameter`.
- Match both camelCase and snake_case forms on encode/decode and emit the additional-properties key in the form the payload uses, so round-trips work under either naming strategy without touching any request class.

Fixes #1878. Also addresses the KG-531 note left in `OpenRouterSerializationTest`.

## Test plan
- [x] New `OpenAIRequestSnakeCaseSerializationTest` — fails against the old serializer (leaks `additional_properties`, drops fields on round-trip) and passes with the fix.
- [x] `./gradlew :prompt:prompt-executor:prompt-executor-clients:prompt-executor-openai-client:jvmTest` and all other clients that consume the serializer (anthropic, google, openrouter, mistralai, deepseek, dashscope, ollama, openai-client-base, prompt-executor-clients) — green.
- [x] `compileKotlinJs` and `compileKotlinWasmJs` for `prompt-executor-clients` — green (change lives in `commonMain`).

### DIFF
--- a/prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client/src/jvmTest/kotlin/ai/koog/prompt/executor/clients/openai/models/OpenAIRequestSnakeCaseSerializationTest.kt
+++ b/prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client/src/jvmTest/kotlin/ai/koog/prompt/executor/clients/openai/models/OpenAIRequestSnakeCaseSerializationTest.kt
@@ -1,0 +1,114 @@
+package ai.koog.prompt.executor.clients.openai.models
+
+import io.kotest.assertions.json.shouldEqualJson
+import io.kotest.matchers.collections.shouldContain
+import io.kotest.matchers.collections.shouldNotContain
+import io.kotest.matchers.nulls.shouldNotBeNull
+import io.kotest.matchers.shouldBe
+import kotlinx.serialization.json.Json
+import kotlinx.serialization.json.JsonNamingStrategy
+import kotlinx.serialization.json.JsonPrimitive
+import kotlinx.serialization.json.buildJsonObject
+import kotlinx.serialization.json.jsonObject
+import kotlinx.serialization.json.jsonPrimitive
+import kotlin.test.Test
+
+/**
+ * Reproduces the snake_case serialization path used by [OpenAILLMClient] to guard the fix
+ * for an additional-properties leak in the outbound request body.
+ */
+class OpenAIRequestSnakeCaseSerializationTest {
+
+    private val snakeCaseJson = Json {
+        ignoreUnknownKeys = true
+        isLenient = true
+        encodeDefaults = true
+        explicitNulls = false
+        namingStrategy = JsonNamingStrategy.SnakeCase
+    }
+
+    @Test
+    fun `responses request flattens additionalProperties under snake_case naming`() {
+        // Regression for #1878: with JsonNamingStrategy.SnakeCase the additionalProperties field
+        // was being emitted verbatim as `additional_properties` at the top level of the request,
+        // which OpenAI rejects with `400 unknown_parameter`.
+        val request = OpenAIResponsesAPIRequest(
+            model = "gpt-4o",
+            instructions = "Please help with this task",
+            temperature = 0.7,
+            additionalProperties = mapOf(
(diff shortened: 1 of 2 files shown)


aozherelyeva (approved):
Thank you for the contribution!

PR #1885, opened 2026-04-23
fix(embeddings): process Ollama batch embeddings

### LINKED ISSUE #1874: Calculation of embeddins with Ollama doesn't work
I've tried using the method "embed" for Ollama and it doesn't work. The result is:

> Caused by: kotlinx.serialization.MissingFieldException: Field 'embeddings' is required for type with serial name 'ai.koog.prompt.executor.ollama.client.dto.EmbeddingResponseDTO', but it was missing at path: $

The JSON answer from Ollama is: 

> {"embedding":[]}

The cause seems to be that OllamaClient is using the old call "/api/embeddings".
I've tried the current call "/api/embed" and it correctly returns an embeddings array. 
It still doesn't work because it's a 2-dimensional array:

> {"model":"jina/jina-embeddings-v2-base-es","embeddings":[[0.0131005775,...,-0.019281294],"index":0}],"model":"jina/jina-embeddings-v2-base-es","usage":{"prompt_tokens":12,"total_tokens":12}}

t is necessary to reduce it to a 1-dimensional array:

> public class OllamaClient @JvmOverloads constructor(
> ...
> override suspend fun embed(text: String, model: LLModel): List<Double> {
> ...
> return embeddingResponse.embeddings.firstOrNull() ?: emptyList()
> }
> ...
> }

In my case, it works with these 2 changes (URL and 'return') with Ollama 0.21.0 (linux)

aozherelyeva:
Fix Ollama embeddings compatibility with both current and legacy API response formats.

closes #1874

### DIFF
--- a/prompt/prompt-executor/prompt-executor-clients/prompt-executor-ollama-client/src/commonMain/kotlin/ai/koog/prompt/executor/ollama/client/dto/OllamaModels.kt
+++ b/prompt/prompt-executor/prompt-executor-clients/prompt-executor-ollama-client/src/commonMain/kotlin/ai/koog/prompt/executor/ollama/client/dto/OllamaModels.kt
@@ -122,9 +122,19 @@ internal data class EmbeddingBatchRequestDTO(
  */
 @Serializable
 internal data class EmbeddingBatchResponseDTO(
-    @SerialName("embeddings") val embeddings: List<List<Double>>,
+    @SerialName("embeddings") val embeddings: List<List<Double>>? = null,
+    @SerialName("embedding") val embedding: List<Double>? = null,
     @SerialName("model") val modelId: String? = null
-)
+) {
+    /**
+     * Normalizes Ollama's two embedding response shapes into a single batch representation.
+     *
+     * Returning `List<List<Double>>` lets the rest of the embedding pipeline treat single-item and
+     * multi-item requests the same way, which keeps downstream similarity/search code independent
+     * from transport-specific response details.
+     */
+    fun normalizedEmbeddings(): List<List<Double>> = embeddings ?: embedding?.let(::listOf) ?: emptyList()
+}
 
 /**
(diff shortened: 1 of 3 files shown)


prompt/prompt-executor/prompt-executor-clients/prompt-executor-ollama-client/src/commonMain/kotlin/ai/koog/prompt/executor/ollama/client/dto/OllamaModels.kt
```diff
@@ -122,9 +122,12 @@ internal data class EmbeddingBatchRequestDTO(
  */
 @Serializable
 internal data class EmbeddingBatchResponseDTO(
-    @SerialName("embeddings") val embeddings: List<List<Double>>,
+    @SerialName("embeddings") val embeddings: List<List<Double>>? = null,
+    @SerialName("embedding") val embedding: List<Double>? = null,
     @SerialName("model") val modelId: String? = null
-)
+) {
+    fun normalizedEmbeddings(): List<List<Double>> = embeddings ?: embedding?.let(::listOf) ?: emptyList()
```
Amaneusz, line 136: nth: Consider adding some small kdocs explaining why we need to normalize embeddings

Amaneusz (approved):
Good fix, thank you! 👏 
I left one minor comment :)

prompt/prompt-executor/prompt-executor-clients/prompt-executor-ollama-client/src/commonMain/kotlin/ai/koog/prompt/executor/ollama/client/dto/OllamaModels.kt
zarechneva-a: Good point, thanks

PR #1886, opened 2026-04-23
feat(agents): add ToolCallMetadata side channel for tool execution

### LINKED ISSUE #1777: Pass metadata to execute tool
Use-case: I want to be able to pass metadata to a toolcall. This metadata is based on the execution of the agent (in my case, I want to pass the observation id of a trace to the tool call, so we can do distributed tracing)

Right now, only arguments are passed in the `public abstract suspend fun execute(args: TArgs): TResult`  [Tool.kt](https://github.com/JetBrains/koog/blob/develop/agents/agents-tools/src/commonMain/kotlin/ai/koog/agents/core/tools/Tool.kt#L105).

The proposal is to extend this with `public abstract suspend fun execute(args: TArgs, metadata: Map<String, Any? = emptyMap()): TResult`, so that extra metadata can be passed (so in my case, that observation id of the trace) 

Idea of implementation:

1. add this in [AIAgentEnvironment](https://github.com/JetBrains/koog/blob/develop/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/environment/AIAgentEnvironment.kt#L22)  - `public suspend fun executeTool(toolCall: Message.Tool.Call, metadata: Map<String, Any? = emptyMap()): ReceivedToolResult`

This way, one can make a custom AIAgentEnvironment (loaded through a feature and pipeline.interceptEnvironmentCreated) and have the flexibility now to add extra metadata of whatever they want to inject.

2. Add a way to provide metadata for a tool call within a `Feature`

Something like:
```
public override fun provideMetadataForToolCall(
    feature: AIAgentFeature<*, *>,
    handle: suspend (AgentEnvironmentToolCallStartingContext) -> Map<String, Any?>
)
```

qflen:
Closes #1777.

## Summary

Tools sometimes need per-call context that has no business in their argument schema: a trace span id, a run-scoped correlation id, a feature flag, or the live agent context itself. 

This PR adds `ToolCallMetadata`, an immutable side channel threaded into tool execution that is never serialized to or from the LLM and never participates in tool selection. It is purely additive: every existing tool keeps working untouched.

## How it works

Tool execution is unified under a new base, `ToolBase`, whose `execute(args, metadata)` is what the runtime always calls. Three shapes sit on top of it, so each tool opts into exactly as much as it needs:

- **`Tool`** — the conventional `execute(args)` shape. Existing tools already subclass
  this; their behavior is unchanged and any supplied metadata is simply discarded.
- **`AgentContextAwareTool`** (in `agents-core`) — receives the live `AIAgentContext`
  as a typed `execute(args, context)` parameter, for tools that need the agent's full
  state.
- **`ToolBase`** directly — for tools that want to read raw metadata entries contributed
  by features, such as a tracing span id.

`ToolCallMetadata` implements `Map<String, Any?>` by delegation (the same trick `JsonObject` uses), so all read-only map operations work directly on it; `EMPTY`, `of(...)`, and `plus` cover construction and merging.

Features contribute entries through the pipeline: `AIAgentPipeline.provideToolCallMetadata { ... }` registers a contributor, and `ContextualAgentEnvironment` aggregates them at call time. Caller-supplied values win over feature contributions on key collision. The framework then injects the live `AIAgentContext` *last*, under the reserved `AgentContextKey`, so neither callers nor features can spoof it. `AgentContextAwareTool` reads that key for you; tools that extend `ToolBase` can use the `ToolCallMetadata.agentContext` extension to get it typed.

## Scope

**agents-tools**: new `ToolBase` and `ToolCallMetadata`; `Tool` reparented onto `ToolBase` with a final `execute(args, metadata)` that forwards to `execute(args)`.

**agents-core**: new `AgentContextAwareTool` and the `agentContext` accessor; metadata-aware overloads on `AIAgentEnvironment.executeTool/executeTools` and `SafeTool.execute/executeUnsafe`; `GenericAgentEnvironment` threads merged metadata to the tool; `ContextualAgentEnvironment` performs the merge-and-inject; and the pipeline gains `provideToolCallMetadata` / `collectToolCallMetadata`, backed by a new `ToolCallMetadataContributing` lifecycle event.

**agents-mcp-server**: `McpServer.addTool` widened to accept `ToolBase`.

**docs**: usage and precedence rules added to `docs/docs/class-based-tools.md` (where snippets
are compile-verified). `Module.md` files carry only short pointers, per review feedback.

## Backwards compatibility

Every new signature is additive with a default that delegates to the existing one, and no existing method changed or was removed. Tools that override only `execute(args)` need no changes.

## Tests

Unit coverage lives in `ToolCallMetadataTest` (agents-tools), and `AgentContextAwareToolTest`, `ContextualAgentEnvironmentMetadataTest`, `GenericAgentEnvironmentMetadataTest`, and
`SafeToolMetadataTest` (agents-core), covering the merge precedence, caller- and feature-spoof rejection on the reserved key, and the standalone-null case for direct test usage. `ToolCallMetadataEndToEndTest` exercises the full feature-author path with no stubs in the middle: feature install -> pipeline aggregation -> environment merge -> real `Tool.execute`.


qflen:
Oh, missed two mechanical things on the CI pass... Just pushed a fixup:

Rizzen:
@qflen I suggest 
`package ai.koog.agents.core.agent.tools` just to keep it separate

Rizzen (approved):
Solid work, thank you!

croemmich:
Thanks @qflen and @Rizzen, looking forward to using this as soon as it's available in the nightly build!

agents/agents-tools/Module.md
```diff
@@ -14,6 +14,28 @@ The agents-tools module provides a comprehensive framework for creating and mana
 
 Tools are designed to be executed within an environment context, ensuring proper handling of events, feature pipelines, and testing capabilities.
 
+### Passing metadata to a tool
```
EugeneTheDev, line 17: Let's avoid putting code snippets and detailed descriptions in the `Module.md`, since these are not verified. It's better to put docs in the `docs`, since code snippets there are verified

agents/agents-tools/src/commonMain/kotlin/ai/koog/agents/core/tools/ToolCallMetadata.kt
```diff
@@ -0,0 +1,93 @@
...
+ * serialized to or from the LLM, and must not be relied on for routing or tool selection. Typical use
+ * cases are cross-cutting concerns such as a distributed-tracing span identifier, a run-scoped correlation
+ * id, or a per-call feature flag contributed by an installed feature.
+ *
+ * Instances can be constructed from a `Map`, built via [of], or combined with [plus]. The [EMPTY] singleton
+ * represents the absence of metadata and is the default passed through the framework.
+ *
+ * @property values The underlying key-value map. Null values are permitted and retained as-is.
+ */
+public class ToolCallMetadata(
```
EugeneTheDev, line 18: Looks like it could've been a simple map instead of a custom class - it implements all the map methods anyway. Is there a specific reason to use a custom class?

EugeneTheDev (approved):
Looks really good, I only have a couple of small questions, can you take a look?

agents/agents-tools/src/commonMain/kotlin/ai/koog/agents/core/tools/ToolCallMetadata.kt
qflen: The nominal type carries the side-channel contract on every signature it appears in.

In this case: not in the argument schema, not LLM-visible, reserved-key semantics around `agentContext`. 

A raw `Map<String, Any?>` would erase that signal across `Tool`, `AIAgentEnvironment`, `SafeTool`, and the pipeline. 

I thought it also provides
1. An anchor for typed accessors like `metadata.agentContext`, since extending raw `Map<String, Any?>` would attach them to every unrelated map in scope.
2. Room to add reserved-key checks or instrumentation later without callers that break source.

If the concern is the wrapper allocation, it can be switched to
`@JvmInline value class ToolCallMetadata(public val values: Map<String, Any?>)`

That can keep all of the above with zero JVM overhead. Feel free to let me know.

EugeneTheDev: Got it, makes sense. But to avoid imlementing map-like methods manually, how about using this neat trick with delegation that `JsonObject` from kotlinx-serialization uses - implementing `Map` interface and delegating to an actual underlying map. You should still be able to override map methods if needed in the future
https://github.com/Kotlin/kotlinx.serialization/blob/b51076f4d612f7e81518a180e5cb0ebb78763c1e/formats/json/commonMain/src/kotlinx/serialization/json/JsonElement.kt#L186

qflen: I see, thanks. How does it look now? [57587d1](https://github.com/JetBrains/koog/pull/1886/commits/57587d10ed4e4079725ba2161e2459bad4afb908)

EugeneTheDev: Nice, thanks. I'll merge it once the checks are green

qflen: By the way the only current red failed on develop too at https://github.com/JetBrains/koog/actions/runs/25808105438.

PR doesn't touch `A2AServerJsonRpcIntegrationTest.test`.

Issue #1252, opened 2025-12-10
`TraceFeatureMessageRemoteWriterTest` is flaky


EugeneTheDev:
Similar to #1124 and #1223, it runs a web server during unit tests, making it flaky.

Issue #1223, opened 2025-12-04
Flaky "system feature" tests in `AIAgentPipelineJvmTest`


antoniibelyshev:
`test duplicate system features provided in config` fails with the following stacktrace:
```
AIAgentPipelineJvmTest[jvm] > test duplicate system features provided in config()[jvm] FAILED
    kotlinx.coroutines.test.UncompletedCoroutinesError: After waiting for 10s, the test body did not run to completion
        at app//kotlinx.coroutines.test.TestBuildersKt__TestBuildersKt$runTest$2$1$2.invokeSuspend$lambda$0(TestBuilders.kt:353)
        at app//kotlinx.coroutines.InvokeOnCancelling.invoke(JobSupport.kt:1571)
        at app//kotlinx.coroutines.JobSupport.notifyCancelling(JobSupport.kt:1604)
        at app//kotlinx.coroutines.JobSupport.tryMakeCancelling(JobSupport.kt:826)
        at app//kotlinx.coroutines.JobSupport.makeCancelling(JobSupport.kt:786)
        at app//kotlinx.coroutines.JobSupport.cancelImpl$kotlinx_coroutines_core(JobSupport.kt:702)
        at app//kotlinx.coroutines.JobSupport.cancelCoroutine(JobSupport.kt:689)
        at app//kotlinx.coroutines.TimeoutCoroutine.run(Timeout.kt:157)
        at app//kotlinx.coroutines.EventLoopImplBase$DelayedRunnableTask.run(EventLoop.common.kt:505)
        at app//kotlinx.coroutines.EventLoopImplBase.processNextEvent(EventLoop.common.kt:263)
        at app//kotlinx.coroutines.DefaultExecutor.run(DefaultExecutor.kt:105)
        at java.base@17.0.17/java.lang.Thread.run(Thread.java:840)
```

`test known system feature in config set by vm option`
```
AIAgentPipelineJvmTest[jvm] > test known system feature in config set by vm option()[jvm] FAILED
    kotlinx.coroutines.test.UncompletedCoroutinesError: After waiting for 10s, the test body did not run to completion
        at app//kotlinx.coroutines.test.TestBuildersKt__TestBuildersKt$runTest$2$1$2.invokeSuspend$lambda$0(TestBuilders.kt:353)
        at app//kotlinx.coroutines.InvokeOnCancelling.invoke(JobSupport.kt:1571)
[...]

Issue #1124, opened 2025-11-14
Debugger feature tests are flaky


EugeneTheDev:
`test read port from vm option` example stacktrace:
```
DebuggerConfigTest[jvm] > test read port from vm option()[jvm] FAILED
    org.opentest4j.AssertionFailedError: Client or server did not finish in time ==> expected: not <null>
        at app//kotlin.test.junit5.JUnit5Asserter.assertNotNull(JUnitSupport.kt:48)
        at app//kotlin.test.AssertionsKt__AssertionsKt.assertNotNull(Assertions.kt:145)
        at app//kotlin.test.AssertionsKt.assertNotNull(Unknown Source)
        at app//ai.koog.agents.core.system.feature.DebuggerTestAPI$runAgentPortConfigThroughSystemVariablesTest$2.invokeSuspend(DebuggerTestAPI.kt:239)
        at app//kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
        at app//kotlinx.coroutines.internal.ScopeCoroutine.afterResume(Scopes.kt:35)
        at app//kotlinx.coroutines.AbstractCoroutine.resumeWith(AbstractCoroutine.kt:101)
        at app//kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:46)
        at app//kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:98)
        at app//kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:586)
        at app//kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:829)
        at app//kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:717)
        at app//kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:704)

```

`test read default port when not set by property or env variable or vm option` example stacktrace:
```
DebuggerConfigTest[jvm] > test read default port when not set by property or env variable or vm option()[jvm] FAILED
    org.opentest4j.AssertionFailedError: Client or server did not finish in time ==> expected: not <null>
        at app//kotlin.test.junit5.JUnit5Asserter.assertNotNull(JUnitSupport.kt:48)
[...]

PR #1887, opened 2026-04-23
fix(prompt): accept text/plain Content-Type on Ollama non-streaming responses

### LINKED ISSUE #1237: ollamaExecutor throws NoTransformationFoundException due to unhandled text/plain Content-Type from Ollama API
When using ollamaExecutor (specifically the OllamaClient), the execution fails with io.ktor.client.call.NoTransformationFoundException. The error occurs because the local Ollama instance returns the response with the header Content-Type: text/plain; charset=utf-8, even though the body is valid JSON. The internal Ktor client in koog is strictly expecting application/json and fails to find a suitable transformation for text/plain, returning a raw SourceByteReadChannel instead of the expected OllamaChatResponseDTO.

Stack Trace

```
io.ktor.client.call.NoTransformationFoundException: Expected response body of the type 'class ai.koog.prompt.executor.ollama.client.dto.OllamaChatResponseDTO' but was 'class io.ktor.utils.io.SourceByteReadChannel'
In response from `http://127.0.0.1:11434/api/chat`
Response status `200 OK`
Response header `ContentType: text/plain; charset=utf-8` 
Request header `Accept: application/json`
...
at io.ktor.client.call.HttpClientCall.bodyNullable(HttpClientCall.kt:104)
at ai.koog.prompt.executor.ollama.client.OllamaClient.execute(OllamaClient.kt:524)

```

zxuhan:
Ollama sometimes replies to non-streaming `/api/chat` requests with `Content-Type: text/plain; charset=utf-8` even though the body is valid JSON. Ktor's `ContentNegotiation` plugin was only registered for `application/json`, so the response failed with `NoTransformationFoundException` before reaching `OllamaChatResponseDTO`.

### Change
Register the same JSON deserializer for `text/plain` alongside `application/json` in the Ollama client's `ContentNegotiation` config. Streaming responses are unaffected — they read the body as a raw channel and bypass content negotiation.

### Test
- Extended `MockOllamaChatServer` to allow overriding the response `Content-Type`.
- Added `OllamaContentTypeTest` that sends a `text/plain; charset=utf-8` reply and asserts the body is still parsed into an assistant message. The test fails on `develop` without the fix and passes with it.

closes #1237

### DIFF
--- a/prompt/prompt-executor/prompt-executor-clients/prompt-executor-ollama-client/src/commonMain/kotlin/ai/koog/prompt/executor/ollama/client/OllamaClient.kt
+++ b/prompt/prompt-executor/prompt-executor-clients/prompt-executor-ollama-client/src/commonMain/kotlin/ai/koog/prompt/executor/ollama/client/OllamaClient.kt
@@ -151,6 +151,10 @@ public class OllamaClient @JvmOverloads constructor(
         }
         install(ContentNegotiation) {
             json(ollamaJson)
+            // Ollama sometimes returns non-streaming responses with `Content-Type: text/plain`
+            // (see https://github.com/JetBrains/koog/issues/1237). Register the same JSON
+            // deserializer for that content type so the body is still parsed correctly.
+            json(ollamaJson, ContentType.Text.Plain)
         }
         install(HttpTimeout) {
             requestTimeoutMillis = timeoutConfig.requestTimeoutMillis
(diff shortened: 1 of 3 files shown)


zxuhan:
`tests (ubuntu-latest)` failed with `java.net.BindException: Address already in use` in `FeatureMessageRemoteClientTest` (cascading to `HistoryCompressionStrategiesTest`, `SafeToolTest`, `AIAgentPipelineTest`). Unrelated to this change — matches the pre-existing flakiness in #1252 / #1223 / #1124. macOS and Windows runs are green. Could someone re-run the Ubuntu job?

aozherelyeva (approved):
LGTM, thank you.

antoniibelyshev (approved):
Thank you, looks good!

PR #1888, opened 2026-04-23
fix(prompt): return tool calls ahead of assistant message in OllamaClient

mihbor:
When a model responds with tool calls and a text message at the same time, the OllamaClient returns the message first, followed by the tool calls. 
This causes built-in strategies to terminate without executing the tool calls since they terminate as soon as the first non-reasoning response is a text message.
The simple solution is to return tool calls before the content message, which is consistent with what the OpenAIClient does.

closes KG-811


aozherelyeva (approved):
Thank you!

antoniibelyshev (approved):
Thank you, looks good!

PR #1889, opened 2026-04-24
feat(agents): Add freshHistory parameter to subgraph for optimization support

K-dizzled:
Small public API addition to support the agent optimization infrastructure built on top of koog.

**`freshHistory` parameter for subgraphs:** Added `freshHistory: Boolean = false` to `AIAgentSubgraph`, all `subgraph()` DSL overloads, all `subgraphWithTask()` overloads, and `setupSubgraphWithTask()`

When `true`, the subgraph starts with an empty conversation history instead of inheriting the parent context's prompt. The subgraph's history is discarded upon completion. Use case: optimizable subgraphs where each module should reason independently, and where few-shot demonstrations should be the only conversation history.


agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/entity/AIAgentStorage.kt
```diff
@@ -15,6 +15,10 @@ package ai.koog.agents.core.agent.entity
  * @param name The human-readable name of the storage key, used only for its string representation.
  */
 public class AIAgentStorageKey<T : Any>(public val name: String) {
+    override fun equals(other: Any?): Boolean =
```
antoniibelyshev: The AIAgentStorageKey is deliberately compared by reference and not by the name, so that different keys with the same name do not interfere with each other. please remove this equals and hashCode here

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/entity/AIAgentStorage.kt
```diff
@@ -15,6 +15,10 @@ package ai.koog.agents.core.agent.entity
...
+        this === other || (other is AIAgentStorageKey<*> && name == other.name)
```
EugeneTheDev: > `name == other.name`

Don't do this, this breaks the contract for the agent key. Each instance is intended to be unique by design, even if the name is the same. 
https://docs.koog.ai/data-transfer-between-nodes/#aiagentstoragekey

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/entity/AIAgentSubgraph.kt
```diff
@@ -173,12 +177,19 @@ public open class AIAgentSubgraphBase<TInput, TOutput>(
             // Copy inner context with new tools, model and LLM params.
             val initialLLMContext = context.llm
 
+            val effectiveParams = llmParams ?: context.llm.prompt.params
+            val innerPrompt = if (freshHistory) {
+                Prompt(messages = emptyList(), id = context.llm.prompt.id, params = effectiveParams)
```
EugeneTheDev: Can't you just use `copy(messages = emptyList())` here?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/handler/strategy/StrategyEventContext.kt
```diff
@@ -20,12 +20,15 @@ public interface StrategyEventContext : AgentLifecycleEventContext
...
  * @property context The context associated with the strategy's execution.
+ * @property input The input passed to [strategy]. Typed as `Any?` because [AIAgentStrategy] does not
+ * yet carry its input type (mirrors the same convention used by [StrategyCompletedContext.result]).
  */
 public class StrategyStartingContext(
     override val eventId: String,
     override val executionInfo: AgentExecutionInfo,
     public val strategy: AIAgentStrategy<*, *, *>,
     public val context: AIAgentContext,
+    public val input: Any?,
```
EugeneTheDev: These changes seem unrelated to this PR. Can you please elaborate, how they are contected? Otherwise please create a separate PR

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt
```diff
@@ -213,20 +213,25 @@ public inline fun <reified Input, reified Output> subgraphWithTask(
...
     llmParams = llmParams,
     responseProcessor = responseProcessor,
+    freshHistory = freshHistory,
 ) {
     val finishTool = FinishTool<Output>(typeToken<Output>())
 
-    setupSubgraphWithTask<Input, Output, Output>(
+    setupSubgraphWithTask(
         finishTool = finishTool,
+        inputType = typeToken<Input>(),
```
EugeneTheDev: Why did you put these parameters explicitly here? AFAIK they're already automatically inferred, because both functions are `inline` and use `reified` on type parameters

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt
```diff
@@ -656,6 +675,9 @@ public fun <Input, Output, OutputTransformed> AIAgentSubgraphBuilderBase<Input,
     outputTransformedType: TypeToken,
     runMode: ToolCalls,
     assistantResponseRepeatMax: Int? = null,
+    freshHistory: Boolean = false,
+    beforeLLMRequest: (suspend AIAgentGraphContextBase.() -> Unit)? = null,
```
EugeneTheDev: These `beforeLLMRequest` and `afterFinishToolCall` look kinda niche. 
* You can inject few shot examples in the `defineTask` since it has access to the agent context and therefore to the LLM context
* You should be able to use [tracing feature](https://docs.koog.ai/features/tracing/) instead of manual hooks to collect the info on all tool calls performed

EugeneTheDev (changes requested):
I like the idea of having a `freshHistory` parameter on subraphs. However, I have question to other parts of this PR

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/handler/strategy/StrategyEventContext.kt
K-dizzled: This change was introduced later on, and, I didn't actualize the PR description, sorry for that. The commit [19355f9](https://github.com/JetBrains/koog/pull/1889/changes/19355f99599a9554ff346dde305786951da5f40b) solves the following problem: some of the optimizers we are building need to hook to the agent's execution and execute some action, dependent on the user query during runtime, simplest example: [do retrieval from the context](https://github.com/JetBrains/koog-auto-agent-optimization/blob/8cc8daeeb921943fabde7d7b0aefc951e1431a5c/src/main/kotlin/com/jetbrains/apr/agentOptimization/optimizers/reasoningBank/ReasoningBankFeature.kt#L66) from a respective memory bank, collected during optimization process.

This changes however are not foundational and we can proceed without them for now, if this will speed up the merge of the PR. How shall I proceed, (i) remove them from the PR for now, or (ii) keep them (I will update the PR description)?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/entity/AIAgentStorage.kt
K-dizzled: Removed in commit [3cc5de0](https://github.com/JetBrains/koog/pull/1889/commits/3cc5de0ed5344c1ecac952251db140b7636744b3), found an alternative solution on our side.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt
K-dizzled: Fixed in [c78b9e1](https://github.com/JetBrains/koog/pull/1889/commits/c78b9e13cf264b47f9efab8c1c4169b85647bc9b), thank you

K-dizzled:
@EugeneTheDev Thank you so much for the review and for your valuable comments, I tried to address them in commits [c78b9e1](https://github.com/JetBrains/koog/pull/1889/commits/c78b9e13cf264b47f9efab8c1c4169b85647bc9b) and [3cc5de0](https://github.com/JetBrains/koog/pull/1889/commits/3cc5de0ed5344c1ecac952251db140b7636744b3). Regarding the change to the `StrategyStartingContext`, as mentioned in the comment above, this part is not essential and might be postponed, if that speeds up the merge.

Ololoshechkin (approved):

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/handler/strategy/StrategyEventContext.kt
EugeneTheDev: Yes, can you please address this part separately in a different PR

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt
```diff
@@ -685,7 +706,20 @@ public fun <Input, Output, OutputTransformed> AIAgentSubgraphBuilderBase<Input,
     // Helper node to overcome problems of the current api and repeat less code when writing routing conditions
     val nodeDecide by node<Message.Assistant, Message.Assistant> { it }
 
-    val nodeCallLLM by nodeLLMRequest()
+    val nodeCallLLM by if (freshHistory) {
+        // When freshHistory is true, the defineTask result becomes a system message
```
EugeneTheDev: I'm not sure if this is a good idea. System message has a stronger influence on the LLM behavior. AFAIU they're meant to steer the agent in general, not to give it a specific task. I would rather keep user message here

EugeneTheDev (changes requested):
Please also take a look at failing CI

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt
```diff
@@ -631,6 +651,7 @@ public fun <Input, Output, OutputTransformed> AIAgentSubgraphBuilderBase<Input,
     outputTransformedType: TypeToken,
     parallelTools: Boolean,
     assistantResponseRepeatMax: Int? = null,
+    freshHistory: Boolean = false,
```
EugeneTheDev: We don't need `freshHistory` parameter on `setupSubgraphWithTask` anymore. It's not used inside this function, and the "fresh history" behavior is handled by the calling functions that wrap `setupSubgraphWithTask` in the `subgraph`, providing the `freshHistory` parameter

EugeneTheDev (changes requested):
Sorry, missed it earlier, but can you also add `withFreshHistory(freshHistory: Boolean)` method to the Java API for subgraphs? https://github.com/JetBrains/koog/blob/e4f91bfb5bace5002005ea2bf513e40c7d1a04ac/agents/agents-core/src/jvmCommonMain/kotlin/ai/koog/agents/core/agent/entity/AgentSubgraphBuilder.kt#L29

And also a minor comment. Other than that looks good

EugeneTheDev (approved):

PR #1894, opened 2026-04-24
fix(agents): KG-751 align JVM AIAgentConfig builder default serializer with common default

goyaljai:
## Summary

- The JVM `AIAgentConfig` builder initialised `serializer` with `JacksonSerializer()`, diverging from the `commonMain` `expect` default (`KotlinxSerializer()`)
- Callers using `AIAgentConfig.builder()...build()` without `.serializer(...)` got `KSerializerTypeToken is not supported for JacksonSerializer` when tools used `@Serializable` arg types
- Root fix: **single source of truth** — the JVM secondary constructor default is `KotlinxSerializer()` (matching the `expect` declaration); the builder field is now nullable so `build()` defers to the constructor default rather than duplicating it

## Design

Before: builder duplicated the default serializer, which diverged:
```kotlin
// Builder had its own hardcoded default — diverged from constructor
internal var serializer: JSONSerializer = JacksonSerializer()  // BUG
```

After: builder field is null = "caller didn't set one", `build()` defers to the constructor:
```kotlin
internal var serializer: JSONSerializer? = null  // no duplicated default

public fun build(): AIAgentConfig {
    val config = AIAgentConfig(/* no serializer arg → uses constructor default KotlinxSerializer */)
    return serializer?.let { config.copy(serializer = it) } ?: config
}
```

If the common default ever changes, only the secondary constructor needs updating — the builder automatically picks it up.

## What changed

- `AIAgentConfig.kt` (jvmCommonMain): secondary constructor default changed to `KotlinxSerializer()`; builder field made nullable; `build()` no longer passes `serializer` unless explicitly set by caller
- `JavaAPIAgentBuilderTest.kt`: regression test `testAIAgentConfigBuilderDefaultsToKotlinxSerializer` asserts the default is `KotlinxSerializer` without any `.serializer(...)` call

## Test plan

- [x] `testAIAgentConfigBuilderDefaultsToKotlinxSerializer` — passes  
- [x] All existing `JavaAPIAgentBuilderTest` tests — passing

closes https://youtrack.jetbrains.com/issue/KG-751

### DIFF
--- a/agents/agents-core/src/jvmCommonMain/kotlin/ai/koog/agents/core/agent/config/AIAgentConfig.kt
+++ b/agents/agents-core/src/jvmCommonMain/kotlin/ai/koog/agents/core/agent/config/AIAgentConfig.kt
@@ -6,7 +6,7 @@ import ai.koog.prompt.dsl.prompt
 import ai.koog.prompt.llm.LLModel
 import ai.koog.prompt.processor.ResponseProcessor
 import ai.koog.serialization.JSONSerializer
-import ai.koog.serialization.jackson.JacksonSerializer
+import ai.koog.serialization.kotlinx.KotlinxSerializer
 import java.util.concurrent.ExecutorService
 
 @Suppress("EXPECT_ACTUAL_CLASSIFIERS_ARE_IN_BETA_WARNING", "MissingKDocForPublicAPI")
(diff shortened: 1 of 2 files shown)


goyaljai:
@aozherelyeva could you review this?

PR #1898, opened 2026-04-24
feat(prompt): KG-685 add and introduce, Google Search grounding support for Gemini

goyaljai:
## Summary

Implements grounding with Google Search for Gemini models, and fixes a missing input validation bug discovered during implementation.

- Add `LLMCapability.Grounding` so callers can check model support with `model.supports(LLMCapability.Grounding)`
- Add `GoogleGroundingConfig` sealed class — `GoogleSearch` (Gemini 2.0+, native `google_search` tool) and `GoogleSearchRetrieval` (Gemini 1.5, `googleSearchRetrieval` with optional dynamic threshold)
- Add `groundingConfig: GoogleGroundingConfig?` field to `GoogleParams`
- All Gemini 2.0+ models gain `LLMCapability.Grounding` via `fullCapabilities`
- `createGoogleRequest()` injects the appropriate grounding tool and guards unsupported models with a clear `require()` error

## Usage

```kotlin
val prompt = prompt("search") { user("Who won Euro 2024?") }
val response = executor.execute(
    prompt,
    GoogleModels.Gemini2_5Flash,
    params = GoogleParams(groundingConfig = GoogleGroundingConfig.GoogleSearch)
)
```

## Test plan

- [x] `createGoogleRequest injects google_search tool when GoogleSearch grounding is set`
- [x] `createGoogleRequest injects googleSearchRetrieval tool with threshold when set` — asserts `MODE_DYNAMIC` is present
- [x] `createGoogleRequest merges grounding tool with function tools`
- [x] `createGoogleRequest throws when grounding set on model that does not support it`
- [x] `GoogleSearchRetrieval rejects dynamicThreshold above 1_0`
- [x] `GoogleSearchRetrieval rejects negative dynamicThreshold`
- [x] `GoogleSearchRetrieval accepts null and boundary dynamicThreshold values`
- [x] All 27 `GoogleLLMClientTest` tests pass, no regressions

Closes https://youtrack.jetbrains.com/issue/KG-685/


prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/jvmTest/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClientTest.kt
```diff
@@ -664,4 +664,118 @@ class GoogleLLMClientTest {
...
+                id = "id",
+                params = GoogleParams(groundingConfig = GoogleGroundingConfig.GoogleSearch)
+            ),
+            model = model,
+            tools = emptyList()
+        )
+
+        val tools = request.tools
+        tools shouldNotBe null
+        tools!!.shouldHaveSize(1)
```
Amaneusz: nit: You can avoid `!!` with smart casting here and in other test as well if you use `assertNotNull` from `kotlin.test`. This will preserve the non-null check and the compiler will know that `tools` (or any other value checked like this) is not null. Hence, `!!` will not be required.
```
val tools = request.tools
assertNotNull(tools)
tools.shouldHaveSize(1) // no "!!"
```

If you'd like to stay with `Kotest` and `shouldNotBe null` then you can still omit the `!!` operator with assignment:
```
val tools = request.tools.shouldNotBeNull()
tools.shouldHaveSize(1)  // no "!!"
```

I like smart casting so I would choose the first version based on `kotlin.test` but this is totally a matter of personal preference :)

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/jvmTest/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClientTest.kt
```diff
@@ -664,4 +664,118 @@ class GoogleLLMClientTest {
...
+                prompt = Prompt(
+                    messages = emptyList(),
+                    id = "id",
+                    params = GoogleParams(groundingConfig = GoogleGroundingConfig.GoogleSearch)
+                ),
+                model = modelWithoutGrounding,
+                tools = emptyList()
+            )
+        }
+        result.isFailure shouldBe true
```
Amaneusz: nit: arguably it would be more readable if you used `shouldThrow` block here - it will not only signal the error happened but also checked the exact type of exception thrown:

```
shouldThrow<IllegalArgumentException> {
      client.createGoogleRequest(...)                                                                                    
  } 
```

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleParams.kt
```diff
@@ -4,6 +4,46 @@ import ai.koog.prompt.executor.clients.google.models.GoogleThinkingConfig
 import ai.koog.prompt.params.LLMParams
 import kotlinx.serialization.json.JsonElement
 
+/**
+ * Controls grounding with Google Search for Gemini models.
+ *
+ * Grounding augments model responses with real-time information from Google Search.
+ * Use [GoogleSearch] for Gemini 2.0+ models, [GoogleSearchRetrieval] for Gemini 1.5 models.
```
Amaneusz: We don't support older Gemini models in Koog - oldest supported ones are `2.0` and even these are already deprecated (with note to use 2.5) 😄 See `GoogleModels.kt`.

We shouldn't support `google_search_retrieval` tool as it only targets legacy models that we don't even list (and we probably won't).
With `google_search_retrieval` we could even allow users to write something like this:
```
GoogleParams(groundingConfig = GoogleGroundingConfig.GoogleSearchRetrieval(0.3))
```
It would compile and run, but then Google API will probably ignore it or fail.

Please remove the `google_search_retrieval` variant of Grounding - it should heavily reduce the size of this PR:
- Drop `GoogleGroundingConfig.GoogleSearchRetrieval` entirely
- `GoogleGroundingConfig` becomes a single-variant with just `GoogleSearch` - at that point it could even be simplified to a simple boolean flag on `GoogleParams`: `groundingEnabled: Boolean`, so we could remove this class entirely
- Remove the `googleSearchRetrieval` field from `GoogleTool`
- Drop GoogleSearchRetrieval-related tests

prompt/prompt-llm/src/commonMain/kotlin/ai/koog/prompt/llm/LLMCapability.kt
```diff
@@ -200,6 +200,18 @@ public sealed class LLMCapability(public val id: String) {
...
+     * Represents the ability to ground model responses using live web search results.
+     *
+     * Models with this capability can augment their responses with real-time information
+     * retrieved via Google Search. Enabled via [ai.koog.prompt.executor.clients.google.GoogleGroundingConfig]
+     * in [ai.koog.prompt.executor.clients.google.GoogleParams].
+     *
+     * Supported by Gemini 2.0+ models.
+     */
+    @Serializable
+    public data object Grounding : LLMCapability("grounding")
```
Amaneusz: I don't think we should add new `LLMCapability` for `Grounding` - it's a different level of abstraction.

`LLMCapability` should represent fundamental model characteristics — things that determine whether a feature can exist at all, independent of how any specific provider implements it. `Tools`, `Vision`, `Thinking`, `Embed` — these describe what the model is, not how a provider exposes it.
                                                          
`LLMCapability.Grounding` doesn't fit that bar. It's not a model characteristic — it's a description of a specific Google API integration detail ("inject google_search: {} into the tools list"). The model doesn't inherently "have grounding" the way it inherently "has vision" or "supports thinking." It's a provider feature toggle.
                                                                                                                         
Think of this like that: if removing the provider (Google) entirely makes the capability meaningless, it belongs in the provider's params, not in LLMCapability.

Amaneusz: For me, Google's grounding is similar to OpenAI's web search - see `OpenAIChatParams.webSearchOptions`. I think we should go in the similar direction in this PR

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/models/GoogleGenerateContent.kt
```diff
@@ -252,6 +252,9 @@ internal sealed interface GoogleData {
 @Serializable
 internal class GoogleTool(
     val functionDeclarations: List<GoogleFunctionDeclaration>? = null,
+    @SerialName("google_search")
```
Amaneusz: Are you sure that it's `google_search` and not `googleSearch`?
https://ai.google.dev/api/caching#Tool

Amaneusz: Also, this should not just be some `JsonObject` - there's specific contract for it: https://ai.google.dev/api/caching#GoogleSearch

So we want something like this in the end:
```
@Serializable                                                                                                          
internal class GoogleSearch(                                                                                           
    val timeRangeFilter: Interval? = null,                                                                             
    val searchTypes: SearchTypes? = null,                                                                              
)
```                                                                                                                 
                                                                                                                         
And in GoogleTool :
```
val googleSearch: GoogleSearch? = null
```

Amaneusz (changes requested):
Thank you again for this PR! I went through it and I think it could be improved - see my comments below.

The main things to address are:
1. We don't want a seperate `LLMCapability` for `GoogleSearch` based grounding
2. The API integration would not work in the proposed state (we need typed `googleSearch`, currently we have loose type for `google_search`)
3. This PR tries to integrate with legacy Gemini models we don't support 

I'm open to consulting further if you disagree on some points (I might have got something wrong) or need some guidance wrt our codebase

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleParams.kt
Amaneusz: >  at that point it could even be simplified to a simple boolean flag on `GoogleParams`

I'm sorry, this is wrong. We need more than just a boolean, see:
https://github.com/JetBrains/koog/pull/1898#discussion_r3151841777

You could make `GoogleSearch` public data class and use it in both `GoogleParams` and direct API integration layer (`GoogleTool`). So it's quite similar case to `GoogleThinkingConfig`

prompt/prompt-llm/src/commonMain/kotlin/ai/koog/prompt/llm/LLMCapability.kt
goyaljai: Agreed, will remove LLMCapability.Grounding entirely — it's a provider detail, not a model trait. Will align with the OpenAIChatParams.webSearchOptions pattern. will Fix in ~1hr.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleParams.kt
goyaljai: Agreed, will drop GoogleGroundingConfig sealed class and GoogleSearchRetrieval entirely, and simplify to groundingEnabled: Boolean on GoogleParams. will Fix in ~1hr.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/models/GoogleGenerateContent.kt
goyaljai: Good catch, Guess I verified but will verify again against the API spec and fix the serialization name. if needed I will Fix in ~1hr.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/jvmTest/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClientTest.kt
goyaljai: will switch

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/jvmTest/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClientTest.kt
goyaljai: will switch

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/models/GoogleGenerateContent.kt
goyaljai: yes it doesnot work like that.

I tested both features using generativelanguage.googleapis.com (this is the Google AI Studio API used by koog, not Vertex AI).

searchTypes.webSearch: The API accepted it, but the results were the same as using just {"googleSearch": {}}. There was no difference in behavior. So the **filter did not work.**
timeRangeFilter: I tested it with the question “Iran vs US war 2026, what is happening?” and set the date range from Jan 2025 to July 2025. Even with this filter, the model still returned full details about the Feb 2026 conflict, same as without the filter. So the **filter did not work.**

**According to the documentation (https://ai.google.dev/api/caching#GoogleSearch
), these features seem to work only with the Vertex AI API (aiplatform.googleapis.com), not the public API we are using.**

Since koog uses generativelanguage.googleapis.com, both features NOT NEEDED to avoid relying on something that doesn’t actually work.

Now, GoogleSearch is just an empty object, and only basic grounding is supported.

goyaljai:
> Thank you again for this PR! I went through it and I think it could be improved - see my comments below.
> 
> The main things to address are:
> 
> 1. We don't want a seperate `LLMCapability` for `GoogleSearch` based grounding
> 2. The API integration would not work in the proposed state (we need typed `googleSearch`, currently we have loose type for `google_search`)
> 3. This PR tries to integrate with legacy Gemini models we don't support
> 
> I'm open to consulting further if you disagree on some points (I might have got something wrong) or need some guidance wrt our codebase

Done all the changes. Please review again.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/models/GoogleGenerateContent.kt
Amaneusz: > According to the documentation these features seem to work only with the Vertex AI API 

I don't think this is a correct statement. Can you show which part of the docs says that?
Afaik these docs cover the exact API we're using.

Moreover, the fact that the API did not give you the result you expected does not mean that we should not comply to the documented request format.

I tested the request shape with simple curl that was meant to be invalid (see missing `startTime` in `timeRangeFilter`):
```
> curl -s "https://generativelanguage.googleapis.com/v1beta/models/gemini-2.5-flash:generateContent" \                                                                                                                             
             -H "X-goog-api-key: ..." \
             -H "Content-Type: application/json" \
             -d '{                                                                                                              
               "contents": [{"parts": [{"text": "Who won latest parliament elections in Hungary"}]}],                    
               "tools": [{                                                                                                        
                 "googleSearch": {                                                                                                
                   "timeRangeFilter": {                                                                                           
                     "endTime": "2024-12-31T23:59:59Z"   
                   }                                                                                                              
                 }                                                                                                              
               }]                                                                                                                 
             }'
``` 

and got 400:
```
{
  "error": {
    "code": 400,
    "message": "* GenerateContentRequest.tools[0].google_search.time_range_filter: [FIELD_INVALID] Both start time and end time must be given\n",
    "status": "INVALID_ARGUMENT"
  }
}
```

which proves that the shape IS expected. We must comply with the schema that is defined here: https://ai.google.dev/api/caching#GoogleSearch

The fact that you got the same responses with your calls only proves that inner fields of `googleSearch` are not required, but we should still allow users to pass them if they want - otherwise we constrain the Gemini API.

Amaneusz: Also, check these docs: https://ai.google.dev/gemini-api/docs/google-search

> When a response is successfully grounded, the response includes a `groundingMetadata` field. 

In my case, when I run your query - I did receive the mentioned field. It's true that the filter did not seem to work as expected, but the API itself did. We don't test models, we test integration

goyaljai:
@Amaneusz Please review now. 
The model may not behave differently today, but that's the model's limitation, not ours (Koogs). We should expose it.

Changed the code to support these.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/models/GoogleGenerateContent.kt
goyaljai: It's a fair point — the API accepts it, we just shouldn't claim it guarantees filtered results.

PR #1915, opened 2026-04-27
fix: process blank tool call ID in StreamFrameFlowBuilder.emitToolCallDelta

### LINKED ISSUE #1900: emitToolCallDelta incorrectly treats empty string id as new tool call instead of continuation
Description:    
                                                                                                                                                                                
  When processing streaming tool calls from OpenAI-qwen APIs, the subsequent chunks often have an empty string "" as the id field (instead of null) to indicate that this 
  is a continuation of the previous tool call, only appending arguments.                                                                                                        
                                                                                                                                                                                
  However, in StreamFrameFlowBuilder.emitToolCallDelta, the condition to detect a new tool call is:                                                                             
   
``` kotlin
  val new: PendingToolCall = if (id != null || index != previous?.index) {                                                                                                      
      tryEmitPendingToolCall()                                                                                                                                                  
      PendingToolCall(id, name, args, index)                                                                                                                                    
  } else {                                                                                                                                                                      
      // append to previous...                                                                                                                                                  
  }   
```                                                                                                                                                                          
                                                                                                                                                                                
  The problem is that id != null evaluates to true when id is an empty string "", because "" != null is true. This causes the framework to incorrectly treat the subsequent     
  arguments delta as a new tool call instead of appending to the previous one.
                                                                                                                                                                                
  Example from OpenAI-qwen API response:                                                                                                                                             
   
  First chunk:                                                                                                                                                                  
  `{"tool_calls":[{"index":0,"id":"call_abc123","function":{"name":"my_tool","arguments":""}}]}`
                                                                                                                                                                                
  Second chunk (arguments delta):                                                                                                                                               
  `{"tool_calls":[{"index":0,"id":"","function":{"arguments":"{\"param\": "}}]}    `                                                                                              
                                                                                                                                                                                
  The second chunk has id="", which should be treated as continuation (same tool call, just appending arguments). But due to "" != null being true, koog creates a new          
  PendingToolCall instead of appending.                                                                                                                                         
                                                                                                                                                                                
  Suggested fix:                                                                                                                                                                
                  
  Change the condition to treat empty strings as null:                                                                                                                          
``` kotlin
  val new: PendingToolCall = if (id?.isNotBlank() == true || index != previous?.index) {
      // new tool call                                                                                                                                                          
  } else {                                                                                                                                                                      
      // append to previous                                                                                                                                                     
  }        
```

aozherelyeva:
Handle blank tool call IDs in emitToolCallDelta in StreamFrameFlowBuilder; add a corresponding test.

Closes #1900

### DIFF
--- a/prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/streaming/StreamFrameFlowBuilder.kt
+++ b/prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/streaming/StreamFrameFlowBuilder.kt
@@ -195,10 +195,11 @@ public class StreamFrameFlowBuilder(
     ) {
         tryEmitPendingText()
         tryEmitPendingReasoning()
+        val sanitizedId = id?.takeUnless { it.isBlank() }
         val previous: PendingToolCall? = pendingToolCallRef.load()
-        val new: PendingToolCall = if (id != null || index != previous?.index) {
+        val new: PendingToolCall = if (sanitizedId != null || index != previous?.index) {
             tryEmitPendingToolCall()
-            PendingToolCall(id, name, args, index)
+            PendingToolCall(sanitizedId, name, args, index)
         } else {
             when {
                 previous == null ->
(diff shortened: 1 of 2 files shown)


antoniibelyshev (approved):
LGTM, thanks

PR #1916, opened 2026-04-27
feat: add `suspend fun fork()` for `AIAgentFunctionalContext`

Stream29:
Background: https://github.com/JetBrains/koog/discussions/1642

Only `AIAgentGraphContext` has well-defined `suspend fun fork()` for now.
This PR adds `suspend fun fork()` for `AIAgentFunctionalContext`

By the way, the old implementation of `suspend fun AIAgentGraphContext.fork()` is incorrect. It would cause concurrent modification on `AIAgentLLMContext.prompt`. It's also fixed here.

Tests are added to verify these behaviors.


Rizzen:
Hello! Thank you for contribution!
But maybe I'm missing something, why do we need `fork()` in `AIAgentFunctionalContext`?

Stream29:
@Rizzen The reason is similar to that in `AIAgentGraphContext`. 

When the agent is planning a parallel execution of the task, the parallel path's context should be forked from the parent's context. After the parallel execution finished, the result may be appended to the parent's clean context.

Rizzen:
But there are no context forking in functional agents, as well as no embedded parallel execution, because functional agent strategy is basically piece of code, without any hidden logic. 

Hence I'm missing the purpose of this method for functional context.

Stream29:
@Rizzen So there will never be parallel execution in functional trategy? Without context forking, there cannot be safe parallel execution.

Rizzen:
Could you please provide the example usecase for such method?
I mean it's extremely easy to add some public method , but it's impossible to remove it since we're 1.0, so we need to evaluate the value.

Rizzen:
Btw, the fix for AIAgentGraphContext is great, do you want to merge it in separate PR maybe?

Stream29:
OK.

Stream29:
@Rizzen PR is here: #2083

PR #1917, opened 2026-04-28
fix(agents): KG-814. Add the 'type' parameter into the AIAgentError data class

sdubov:
- Include 'type' parameter to the AIAgentError class to use this information on the receiver side. This can be used in the Open Telemetry feature to add a necessary span attribute.

closes: [KG-814](https://youtrack.jetbrains.com/issue/KG-814)

### DIFF
--- a/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/model/AIAgentError.kt
+++ b/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/model/AIAgentError.kt
@@ -3,39 +3,49 @@ package ai.koog.agents.core.feature.model
 import kotlinx.serialization.Serializable
 
 /**
- * Represents an error encountered by an AI agent, encapsulating error details such as
- * the message, stack trace, and an optional cause.
+ * Retrieves the class name of the current [Throwable] instance.
+ *
+ * On the JVM platform, this property returns the fully qualified class name of the throwable.
+ * On other platforms, it falls back to the simple class name of the throwable.
+ *
+ * This property is used to provide a platform-specific representation of the throwable's type,
+ * which helps in categorizing or identifying the type of error.
+ *
+ * @receiver The [Throwable] whose type name is being accessed.
+ * @return The class name of the throwable as a nullable string, or null if the type cannot be determined.
+ */
+internal expect val Throwable.typeName: String?
+
+/**
+ * Represents an error encountered by an AI agent, encapsulating error details.
  *
  * This class provides essential information to understand and debug errors occurring
  * during the execution of AI agent strategies, tools, or nodes.
  *
- * Instances of this class can be created directly from a `Throwable`.
- *
- * @property message A human-readable description of the error. Defaults to "Unknown error"
- *           if not provided by the originating throwable.
- * @property stackTrace The stack trace of the error as a string, providing a detailed
- *           representation of where the error occurred.
- * @property cause The stack trace of the root cause if available, or null if no cause is set.
- *           This helps trace back the chain of exceptions leading to the current error.
+ * @property message An error message of the original throwable;
+ * @property stackTrace The stack trace of the error as a string;
+ * @property cause The stack trace of the root cause if available, or null if no cause is set;
+ * @property type The class name of the type of the error, if available. On JVM this is the fully
+ *           qualified name; on other platforms it falls back to the simple class name.
  */
 @Serializable
 public data class AIAgentError(
     public val message: String,
     public val stackTrace: String,
-    public val cause: String? = null
+    public val cause: String? = null,
+    public val type: String? = null,
 ) {
+
     /**
      * Secondary constructor that allows creating an instance of the class using a [Throwable].
      *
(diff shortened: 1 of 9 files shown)


agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/model/events/toolExecutionEvents.kt
```diff
@@ -118,7 +118,12 @@ public data class ToolValidationFailedEvent(
         toolArgs = toolArgs,
         toolDescription = null,
         message = error,
-        error = AIAgentError(error, "", null)
+        error = AIAgentError(
+            message = error,
+            stackTrace = "",
+            cause = null,
+            type = null
```
Amaneusz, line 125: nit: `cause` and `type` default to `null`, these assignments can be omitted

Amaneusz: btw `stackTrace = ""` looks a bit odd, if we allow such cases, maybe it's better to make `stackTrace` nullable as well?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/model/AIAgentError.kt
```diff
@@ -3,39 +3,34 @@ package ai.koog.agents.core.feature.model
...
- *
- * @property message A human-readable description of the error. Defaults to "Unknown error"
- *           if not provided by the originating throwable.
- * @property stackTrace The stack trace of the error as a string, providing a detailed
- *           representation of where the error occurred.
- * @property cause The stack trace of the root cause if available, or null if no cause is set.
- *           This helps trace back the chain of exceptions leading to the current error.
+ * @property message An error message of the original throwable;
+ * @property stackTrace The stack trace of the error as a string;
+ * @property cause The stack trace of the root cause if available, or null if no cause is set;
```
Amaneusz, line 27: nit: drop `;` at the end of these kdoc lines above

agents/agents-core/src/jvmTest/kotlin/ai/koog/agents/core/system/feature/DebuggerStreamingTest.kt
```diff
@@ -281,8 +283,10 @@ class DebuggerStreamingTest {
                 model: LLModel,
                 tools: List<ToolDescriptor>
             ): Flow<StreamFrame> = flow {
-                val testException = IllegalStateException(testStreamingErrorMessage)
-                testStreamingStackTrace = testException.stackTraceToString()
+                val testException = IllegalStateException(expectedErrorMessage)
+                expectedStackTrace = testException.stackTraceToString()
+                expectedCause = testException.cause?.toString()
+                expectedType = testException::class.qualifiedName
```
Amaneusz, line 306: Shouldn't that be `testException::class.simpleName` (not `qualifiedName`)? 
Simple name is used in other places (constructors)

Amaneusz (approved):
Looks good 👍 
I left a few comments but nothing major

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/model/events/toolExecutionEvents.kt
sdubov: Good nit! Thank you for pointing to this. I explicitly set it here to keep it consistent for the `@Deprecated` constructor. I do not think this call is used, tbh. Your point is valid, but I would keep it as is before `1.0.0`. I think both options are fine here.

sdubov: As for the stack trace, this is not-nullable as you always can dump stack trace into a string and this is non-nullable property in Kotlin:
```
Throwable.stackTraceToString(): String
```

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/model/AIAgentError.kt
sdubov: @Amaneusz, could you please clarify the reason? these `;` are for punctuation and to separate one point from another. I use it quite widely. Please let me know if I miss something here.

agents/agents-core/src/jvmTest/kotlin/ai/koog/agents/core/system/feature/DebuggerStreamingTest.kt
sdubov: True. As part of the tests, it can be any value here, but I would rather update other places to qualified name to provide a better example for other jvm code.

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/feature/model/AIAgentError.kt
Amaneusz: Honestly I just found it off, but it's very subjective, ignore my comment ;)

PR #1919, opened 2026-04-28
feat(agents)!: Introduce new Message architecture

tiginamaria:
BREAKING:
* All `Message` and `Prompt` API

closes KG-689


Ololoshechkin (approved):

croemmich:
This is a very welcome change, thanks all!

It does make me consider the stability of persisted checkpoints however. Currently a checkpoint holds a json serialized list of Message so this will effectively invalidate all previously created checkpoints.

Maybe our use-case is unique, but we allow users to resume a chat of any age, potentially years. While we have separate chat events that are used to hydrate the UI, our agent internally relies on the messageHistory list from the checkpoint for each conversation turn/agent run.

Koog is pre-1.0 and breaking API changes are expected right now, but if a breaking change to the Message interface or anything else in the checkpoint needed to made in the future, would Koog solve for that or should we implement our own persistent layer for checkpoints and message objects to ensure stability?

Thanks!

Ololoshechkin:
@croemmich Thanks for sharing your feedback and concerns!

We’re planning to release Koog 1.0.0 next week, so this shouldn’t be an issue going forward.

In the meantime, and especially for the first migration, I think the only practical option is to implement a migration tool for messages and checkpoints. We can consider adding such a tool after 1.0, perhaps as an ad hoc solution. Alternatively, you could generate one with AI or implement it manually, whichever you prefer.

Please let me know if it would fit your needs

croemmich:
@Ololoshechkin, thanks for the response! We're set to launch our message history feature in a few weeks, so the timing of Koog 1.0.0 is fortunate.

Since we haven't released yet, I don't have any immediate migration needs, I was mostly curious about post v1 plans. I think the ideal solution would be a first-class migration tool if the project introduces breaking changes, as I suspect we wouldn't be the only ones needing to migrate.

Thanks for the hard work towards 1.0.0 and responsiveness to your community, you've hit all three of my major points of feedback on the survey a few months ago!

PR #1925, opened 2026-04-29
refactor(agents)!: Agent clock migration

Rizzen:
Aim of this PR is to get rid of Kotlin Clock dependency that might be inconvenient for Java-only users

BREAKING:
All apis that use kotlin.time.Clock


utils/src/commonMain/kotlin/ai/koog/utils/time/AgentClock.kt
```diff
@@ -0,0 +1,27 @@
...
+ * Time source used across Koog for message timestamps, event timestamps,
+ * and any other "what time is it now" call.
+ *
+ * Implement this interface (or use [AgentClock.System]) anywhere a clock is required.
+ * Being a functional interface, simple test doubles can be written as lambdas:
+ * ```
+ * val fixed = AgentClock { Instant.fromEpochSeconds(1_700_000_000) }
+ * ```
+ */
+public fun interface AgentClock {
```
EugeneTheDev: Since it used not only for agents, let's call it `KoogClock` or smth like this

a2a/a2a-core/build.gradle.kts
```diff
@@ -18,6 +18,7 @@ kotlin {
     sourceSets {
         commonMain {
             dependencies {
+                api(project(":utils"))
```
EugeneTheDev: Please don't make any changes in A2A part. This is de facto a separate A2A KMP SDK, and I want to keep it that way (to potentailly extract it into a separate library in the future). Java users already have their own A2A SDK, so we shouldn't try to make our own more Java-friendly IMHO

EugeneTheDev (changes requested):
Two comments, but looks good otherwise

EugeneTheDev (approved):

PR #1940, opened 2026-05-01
fix(prompt): preserve Google assistant text alongside tool calls

### LINKED ISSUE #1152: GoogleLLMClient should probably not silence assistant messages that come alongside tool calls
I was working on supporting thought signatures (https://github.com/JetBrains/koog/issues/1148) and encountered this code:

https://github.com/JetBrains/koog/blob/13442e8866d9213afbd5609d12fc0823420733a2/prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClient.kt#L679

That doesn't seem right, if the model decides to output assistant messages alongside tool calls, I don't see why it would be correct to ignore them. I understand if some Koog agent workflows depend on this behavior, but it seems like those workflows should adjust their logic for this possibility rather than omitting information outputted by the model (I paid for those tokens!)

Also this removes any Message.Reasoning items so it nullifies reasoning information in tool calls.

zxuhan:
## Summary

Fixes #1152.

`GoogleLLMClient` was filtering out `Message.Assistant` whenever the same response contained any `Tool.Call`, silently discarding model-generated text the user had already paid for. The encoder also wrote that assistant text and the tool call as two separate `"model"` turns, which is not how Gemini emitted them.

This change:

- Removes the `responses.filter { ... }` in `processGoogleCandidate` that dropped assistant text when tool calls were present.
- Adds a `pendingTextParts` buffer in `createGoogleRequest` so assistant text is sent in the **same** `GoogleContent` as the following `Tool.Call`(s).
- Tightens the turn-boundary logic so that `Tool.Result` closes any pending model turn and a following `Assistant` closes any pending user (results) turn. This keeps `user`/`model` alternation correct across tool round-trips.

Existing tests around parallel tool calls and thought signatures continue to pass; the helper rename from `flushCalls` to `flushModel` does not change observable behavior for those cases.

## Test plan

- [x] `./gradlew :prompt:prompt-executor:prompt-executor-clients:prompt-executor-google-client:jvmTest` (43 tests, 0 failures)
- [x] New tests:
  - `processGoogleCandidate keeps Assistant text alongside Tool Call`
  - `processGoogleCandidate keeps Reasoning, Assistant text, and Tool Call together`
  - `createGoogleRequest merges Assistant text with following Tool Call into single model content`
  - `createGoogleRequest preserves model-user-model order across tool round-trip`
  - `createGoogleRequest carries thought signature from Reasoning between Assistant text and Tool Call`

### DIFF
--- a/prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClient.kt
+++ b/prompt/prompt-executor/prompt-executor-clients/prompt-executor-google-client/src/commonMain/kotlin/ai/koog/prompt/executor/clients/google/GoogleLLMClient.kt
@@ -315,14 +315,19 @@ public open class GoogleLLMClient @JvmOverloads constructor(
     internal fun createGoogleRequest(prompt: Prompt, model: LLModel, tools: List<ToolDescriptor>): GoogleRequest {
         val systemMessageParts = mutableListOf<GooglePart.Text>()
         val contents = mutableListOf<GoogleContent>()
+        val pendingTextParts = mutableListOf<GooglePart.Text>()
         val pendingCalls = mutableListOf<GooglePart.FunctionCall>()
         val pendingResults = mutableListOf<GooglePart.FunctionResponse>()
         var lastSignature: String? = null
         val isThinkingModel = model.supports(LLMCapability.Thinking)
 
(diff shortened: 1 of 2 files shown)


zxuhan:
Hi @aozherelyeva, would you have a moment to review this one? CI is green, thanks!

PR #1942, opened 2026-05-01
feat(agents)!: KG-785. Migrate OpenTelemetry to Kotlin Multiplatform

sdubov:
- Replaced the Java OTel SDK with io.opentelemetry.kotlin 0.3.0 in commonMain;
- Moved Langfuse, Weave, and Datadog integrations from jvmMain to commonMain via a new Ktor-based OTLP/JSON span exporter;
- Replace simple exporter with a batch exporter when adding custom span exporters to the feature;
- Added a JVM addSpanExporter overload that bridges Java-SDK exporters via toOtelKotlinSpanExporter API;
- Kept metrics on the Java SDK behind JVM-only extensions (addMetricExporter, addMetricFilter, meter) as the Kotlin SDK 0.3.0 ships no metrics module;
- Changed addResourceAttributes to take Map<String, Any> instead of the Java-SDK Attributes type so it works on every target;
- Dropped the no-op SpanEndStatus wrapper; span lifecycle uses StatusData directly.

BREAKING: 
- addSpanExporter, addMetricExporter, and addMetricFilter moved to JVM-only extension on OpenTelemetryConfigJvm. JVM users must update imports to ai.koog.agents.features.opentelemetry.feature.OpenTelemetryConfigJvm.<method>.addResourceAttributes signature changed from io.opentelemetry.api.common.Attributes to Map<String, Any>;
- The SpanEndStatus wrapper was removed; use StatusData directly.

closes: [KG-785](https://youtrack.jetbrains.com/issue/KG-785)


examples/simple-examples/src/main/kotlin/ai/koog/agents/example/features/opentelemetry/OpenTelemetryExample.kt
```diff
@@ -3,10 +3,12 @@ package ai.koog.agents.example.features.opentelemetry
 import ai.koog.agents.core.agent.AIAgent
 import ai.koog.agents.example.ApiKeyService
 import ai.koog.agents.features.opentelemetry.feature.OpenTelemetry
+import ai.koog.agents.features.opentelemetry.feature.OpenTelemetryConfigJvm.addSpanExporter
```
serge-p7v, line 6: It would be great to document the breaking change for existing JVM users (about addSpanExporter, addMetricExporter, addMetricFilter), so they can switch to the new methods.

sdubov: @serge-p7v, thank you for pointing this out. I've updated the commit message with default about the breaking change. Please let me know if I miss something.

serge-p7v (approved):
👍

PR #1943, opened 2026-05-03
fix(prompt): update OpenAI JSON schema generation for serialization of nullable collections

### LINKED ISSUE #1930: Bug: OpenAI gpt-5+ model family JSON schema issues
When using nullable collections in the serializable data classes fro Structured output OpenAIs schema validation fails with 400 response: 

```
Error from client: OpenAILLMClient                                                                                                                               
      Status code: 400                                                                                                                                                 
      Error body:                                                                                                                                                      
      {                                                                                                                                                                
        "error": {                                                                                                                                                     
          "message": "Invalid schema for response_format 'MyClass': schema must be a JSON Schema of 'type: \"object\"', got 'type: \"None\"'.", 
          "type": "invalid_request_error",                                                                                                                             
          "param": "response_format",                                                                                                                                  
          "code": null                                                                                                                                                 
        }                                                                                                                                                              
      }
```

Nullable list handling for List<X>?: processList(...).asNullableType(...) produces {"type": ["array", "null"], "items": {...}}. Some OpenAI strict validators     
  reject the array form and want either anyOf: [array, null] or a top-level union elsewhere — pick wrong shape and the validator can't classify the type.

Reproducible with   @Serializable data class with one String field and one List<X>? = null field and structured completion with any GPT 5 family model.

I am happy to contribute and can open a MR in the next days unless someone else already filed this.

grantlerduck:
## Summary                                             

- Nullable `List<X>?` fields produced `{"type": ["array", "null"], "items": {...}}`, which OpenAI's                                                                                   
    GPT-5-family strict-mode validator rejects (surfaces as `'Invalid schema for response_format ...                                                                                    
    got "type": "None"'`). Switch `OpenAIStandardJsonSchemaGenerator.processList` to emit                                                                                               
    `{"anyOf": [{"type": "array", "items": ...}, {"type": "null"}]}` for nullable lists, mirroring                                                                                      
    the shape already used for nullable objects.                                                                                                                                        
- Non-nullable lists, nullable primitives, and the LLM-agnostic `StandardJsonSchemaGenerator` are                                                                                     
    unchanged. Only the OpenAI-specific generator is touched, so other providers keep their existing                                                                                    
    type-union encoding.                                                                                                                                                                
- Adds a regression test (`testGenerateOpenAIStandardJsonSchemaForNullableLists`) covering both                                                                                       
    primitive (`List<String>?`) and `$ref` (`List<Tag>?`) item types.

## Out of scope (could follow up)

- `OpenAIBasicJsonSchemaGenerator` still emits the OpenAPI-3.0-style `nullable: true` keyword, which
    OpenAI strict mode silently ignores. Same root cause family, different generator — happy to file
    a follow-up if useful.                                                                                                         
                                                                                                                                                                                        
closes #1930

### DIFF
--- a/prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client-base/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openai/base/structure/OpenAIStandardJsonSchemaGenerator.kt
+++ b/prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client-base/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openai/base/structure/OpenAIStandardJsonSchemaGenerator.kt
@@ -74,6 +74,41 @@ public open class OpenAIStandardJsonSchemaGenerator : StandardJsonSchemaGenerato
         throw UnsupportedOperationException("OpenAI JSON schema doesn't support maps")
     }
 
+    /*
+     OpenAI strict-mode validators (notably the GPT-5 family) reject nullable arrays encoded as
+     a type union `{"type": ["array", "null"], "items": {...}}`. Emit `anyOf` instead, which is
+     accepted by all current OpenAI strict validators. Non-nullable lists are unchanged.
+     */
+    override fun processList(context: GenerationContext): JsonObject {
+        if (!context.descriptor.isNullable) {
+            return super.processList(context)
+        }
+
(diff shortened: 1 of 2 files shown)


grantlerduck:
Just rebased. There was an agent run test which ran into a deadlock unrelated to my changes.

EugeneTheDev (approved):

PR #1945, opened 2026-05-03
fix: rework reentrant concurrency model for Java interop

EugeneTheDev:
## Problem

For more stable Java interop, we need to properly handle the reentrant scenario `Suspendable Koog code` → `Blocking user Java code` → `Suspendable Koog code`. Naively wrapping Java-interacting components in `runBlocking(context)` or `withContext(context)` introduces a risk of deadlock or thread starvation.

Consider the following example:

```kotlin
val singleThreadDispatcher = Executors.newSingleThreadExecutor().asCoroutineDispatcher()

fun suspendableKotlinCodeOne() {
  runBlocking(singleThreadDispatcher) {
    blockingJavaCodeTwo()
  }
}

fun suspendableKotlinCodeThree() {
  runBlocking(singleThreadDispatcher) {
    // do something
  }
}
```

```java
public void blockingJavaCodeTwo() {
  suspendableKotlinCodeThree();
}
```

`suspendableKotlinCodeOne` dispatches a block onto `singleThreadDispatcher`, which then calls into `blockingJavaCodeTwo`. The Java method calls back into Kotlin via `suspendableKotlinCodeThree`, which again tries to dispatch onto the same single-threaded executor.

Even though execution stays on the same thread throughout, the coroutine context tracking this is lost when crossing the Kotlin → Java → Kotlin boundary. As a result, `suspendableKotlinCodeThree` cannot tell that it is already running on the correct context and attempts to redispatch — causing a deadlock, since the only thread is already occupied. The redispatch was never actually necessary; the inner block could have executed directly in place.

## Current state

The solution is to store the coroutine context in a thread local, then check it on dispatch to determine whether the requested context is already active. Several previous attempts have tried to implement this and fix related bugs, the most recent being #1716. However, the current implementation still has major problems:

* After the last fix, `runBlockingIfRequired` ignores its `context` argument, meaning it never dispatches to the requested dispatcher.
* Thread-local-based dispatcher tracking stores the context on the wrong thread — the current thread running `runBlockingIfRequired`, rather than the thread associated with the supplied `context`.
* There is a bunch of similar-looking helper functions implemented separately, leading to duplication. The naming is also confusing.

## Solution

I've narrowed the scope down to two functions with (hopefully) clearer names: `runBlockingReentrant` and `withContextReentrant`. They implement the approach described above correctly, avoiding unnecessary dispatch. A few tests have been added to verify the behavior. All existing utility functions were deleted and all implementations have been migrated.

## Important note

The case where a user manually submits a task to the executor outside of Koog — for example, running the agent itself on the same single-threaded executor — is deliberately out of scope. In such cases, it is technically impossible to reliably determine whether we are already on the same executor/thread. This should be treated as user error rather than something Koog tries to handle.

DEPRECATED:
* Constructors and methods in `AIAgentConfig` on the JVM that accept `ExecutorService` parameters. Constructors and methods that accept more generic `Executor` should be used instead.

closes [KG-750](https://youtrack.jetbrains.com/issue/KG-750)


utils/src/jvmCommonMain/kotlin/ai/koog/utils/concurrency/ReentrantCoroutineUtils.kt
```diff
@@ -0,0 +1,68 @@
...
+ * @param block The suspend function to be executed within the specified context.
+ */
+@InternalKoogUtils
+public suspend fun <T> withContextReentrant(
+    context: CoroutineContext,
+    block: suspend () -> T
+): T {
+    val currentContext = CURRENT_COROUTINE_CONTEXT.get()
+
+    return if (context == currentContext) {
```
sdubov: Would not it be safer to check for an exact object here?
```
return if (context === currentContext) { ... }
```

agents/agents-core/src/jvmCommonMain/kotlin/ai/koog/agents/core/agent/config/AIAgentConfig.kt
```diff
@@ -128,15 +162,36 @@ public actual class AIAgentConfig actual constructor(
...
+                strategyExecutor = strategyExecutorService,
+                llmRequestExecutor = llmRequestExecutorService,
+                serializer = serializer
+            )
+
             /**
              * Sets serializer for underlying tool calls and LLM requests
              *
              * @param serializer The JSON serializer to configure the AI agent with.
-             * @return The updated instance of [Companion.AIAgentConfigBuilder]
```
sdubov, line 139: Why did you delete the return part of KDoc?

agents/agents-core/src/jvmCommonMain/kotlin/ai/koog/agents/core/agent/context/AIAgentFunctionalContextBase.kt
```diff
@@ -156,10 +154,14 @@ public actual abstract class AIAgentFunctionalContextBase<Pipeline : AIAgentPipe
     @JavaAPI
     public fun executionInfo(): AgentExecutionInfo = executionInfo
 
+    @OptIn(InternalKoogUtils::class)
```
sdubov: Would not it be better to annotate the exact internal call instead of a whole function?

agents/agents-core/src/jvmCommonMain/kotlin/ai/koog/agents/core/agent/context/AIAgentFunctionalContextBase.kt
```diff
@@ -171,13 +173,16 @@ public actual abstract class AIAgentFunctionalContextBase<Pipeline : AIAgentPipe
...
     @JavaAPI
     @JvmOverloads
     public fun requestLLM(
         message: String,
         allowToolCalls: Boolean = true,
         executorService: ExecutorService? = null
-    ): Message.Response = config.runOnLLMDispatcher(executorService) {
+    ): Message.Response = runBlockingReentrant(
+        executorService?.asCoroutineDispatcher() ?: config.strategyDispatcher
```
sdubov: strategy or llm request dispatcher?

agents/agents-core/src/jvmCommonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
```diff
@@ -37,14 +38,17 @@ public actual abstract class AIAgent<Input, Output> : Closeable {
...
      * @return The output resulting from the execution of the AI agent with the given input.
      */
+    @OptIn(InternalKoogUtils::class)
     @JavaAPI
     @JvmOverloads
     @JvmName("run")
     public final fun javaNonSuspendRun(
         agentInput: Input,
         sessionId: String? = null,
         executorService: ExecutorService? = null
```
sdubov, line 48: The context from the `executorService` is not used here. Is it by design?

agents/agents-core/src/jvmCommonMain/kotlin/ai/koog/agents/core/agent/config/AIAgentConfig.kt
```diff
@@ -20,23 +30,27 @@ public actual class AIAgentConfig actual constructor(
...
      *
-     * By default, all agent operations will be performed on [kotlinx.coroutines.Dispatchers.Default]
-     *  */
+     * By default, all agent operations will be performed on [Dispatchers.Default]
+     **/
     @JavaAPI
-    @property:PublishedApi
-    internal var strategyExecutorService: ExecutorService? = null
+    @InternalAgentsApi
+    public var strategyDispatcher: CoroutineDispatcher = Dispatchers.Default
```
sdubov, line 39: Why do these dispatchers lives in the `AIAgentConfig`? I would assume that they should be located closer to the executor, instead of configurator. What do you think?

utils/src/jvmCommonMain/kotlin/ai/koog/utils/concurrency/ReentrantCoroutineUtils.kt
```diff
@@ -0,0 +1,68 @@
...
+ * The caller thread is blocked until [block] completes. The block normally runs in [context], but if the current
+ * execution is already associated with the same coroutine context, it runs inline instead of switching
+ * context again. This avoids redispatch deadlocks when all threads in the target context are busy.
+ *
+ * @param context The coroutine context in which the block should be executed. Defaults to [EmptyCoroutineContext].
+ * @param block The suspend block of code to be executed.
+ */
+@InternalKoogUtils
+@JvmOverloads
+public fun <T> runBlockingReentrant(
```
sdubov, line 24: Previously in the `CoroutineUtils` we had thow helper methods: `runOnLLMDispatcher` and runOnStrategyDispatcher`. It looks like a handy shortcut to delegate execution on a correct thread instead of writing:
```
runBlockingReentrant(dispatcher ?: config.strategyDispatcher) { ... }
```
Maybe it worth thinking about returning them? 

TBD: @EugeneTheDev , also I have some concern that we currently have a contract for using suspend Kotlin calls from Java API that seems easy to break. Maybe we should expose two API for such cases and hide the rest from a public API, limiting the probability of mistakes. What do you think?

agents/agents-core/src/jvmCommonMain/kotlin/ai/koog/agents/core/dsl/builder/AIAgentEdgeBuilderIntermediate.kt
```diff
@@ -1,5 +1,5 @@
 @file:Suppress("EXPECT_ACTUAL_CLASSIFIERS_ARE_IN_BETA_WARNING")
-@file:OptIn(InternalAgentsApi::class)
+@file:OptIn(InternalAgentsApi::class, InternalKoogUtils::class)
```
sdubov: Not related to a current change (feel free to skip this comment), but I often see `OptIn` on file level in our code base. @EugeneTheDev, do you consider this a good practice? I would add annotation for a particular invocation statement instead.

sdubov (approved):
@EugeneTheDev , nice work! Thank you for the changes. I've left several general related comments.

utils/src/jvmCommonMain/kotlin/ai/koog/utils/concurrency/ReentrantCoroutineUtils.kt
EugeneTheDev: Good point. Comparing the whole context is also bad performance-wise. I reworked the approach around the thing we actually care about: the coroutine continuation interceptor, i.e. the dispatcher.
Instead of comparing whole coroutine contexts, we now store and compare only the interceptor. That lets us still run with a new context when only non-dispatcher elements change, for example a different CoroutineName on the same dispatcher. In that case we should preserve the new context, but strip the dispatcher to avoid redispatching when we are already on it.

agents/agents-core/src/jvmCommonMain/kotlin/ai/koog/agents/core/agent/config/AIAgentConfig.kt
EugeneTheDev: They seemed too verbose, without any actual value. You can see the return type yourself in your IDE, so that part only bloats the KDoc, IMHO, because it doesn't provide any additional valuable info on the returned value

agents/agents-core/src/jvmCommonMain/kotlin/ai/koog/agents/core/agent/context/AIAgentFunctionalContextBase.kt
EugeneTheDev: I would say that's gonna be too granular. Prefering method-level opt-ins to class level opt-ins is fine, I agree here. But going ever more granular might not be the best idea from the readability side of things. I don't really have any strong opinion here, but putting opt-in on the method level is already an established pattern in the whole project, so I'm just following it

agents/agents-core/src/jvmCommonMain/kotlin/ai/koog/agents/core/agent/config/AIAgentConfig.kt
EugeneTheDev: That's the only central configuration object we currently have. So it's the only option to put global configurations that should affect the whole agent

utils/src/jvmCommonMain/kotlin/ai/koog/utils/concurrency/ReentrantCoroutineUtils.kt
EugeneTheDev: > Maybe it worth thinking about returning them?

Initially, I thought that this inline approach would be more readable, but I guess you're right, and these shorthands did simplify things a bit. I'll return them back

> we currently have a contract for using suspend Kotlin calls from Java API that seems easy to break

Can you clarify, please, wdym? I don't quite get it

agents/agents-core/src/jvmCommonMain/kotlin/ai/koog/agents/core/agent/AIAgent.kt
EugeneTheDev: Good catch, will fix

agents/agents-core/src/jvmCommonMain/kotlin/ai/koog/agents/core/agent/context/AIAgentFunctionalContextBase.kt
EugeneTheDev: Good catch, will fix

agents/agents-core/src/jvmCommonMain/kotlin/ai/koog/agents/core/dsl/builder/AIAgentEdgeBuilderIntermediate.kt
EugeneTheDev: Oh, that's a good comment. I actually don't like file-level opt-ins or suppresions unless there's a really good reason (like you need to opt in on the same API that is used by several top-level functions in this file). But, reiterating on my comment above, I think that statement-level opt-in might be too granular. IMHO method/function level is a sweet spot

PR #1947, opened 2026-05-04
fix(agents): persist and restore AIAgentContext.storage in checkpoints

### LINKED ISSUE #1944: Persistence checkpoints do not persist/restore AIAgentContext.storage state
## Summary

Persistence checkpoints currently restore execution position and message history, but don't restore `ctx.storage` state. This can cause issues for graphs that rely on `ctx.storage` for intermediate business data, leading to incomplete recovery.

### Current Behavior

`AgentCheckpointData` stores:

* `messageHistory`
* `nodePath`
* `lastInput / lastOutput`
* `version`
* `properties`

However, user-defined values from `ctx.storage` are not restored during recovery.

### Why This Is a Problem

Some nodes depend on `ctx.storage` for:

* Intermediate outputs
* Retry counters
* Cached inputs
* Temporary state
* Tool data

If these values are not restored, it can cause:

* Incorrect control flow
* Node execution failures
* Inconsistent behavior between initial and resumed runs
* Unreliable workflows that require human input

### Notes

This is particularly important for workflows requiring human input, where restoring just the execution point is insufficient.

qflen:
Closes #1944.

Persistence checkpoints currently capture `messageHistory`, `nodePath`, `lastInput/lastOutput`, `version`, and `properties` but never the entries held in `AIAgentContext.storage`. Workflows that stash retry counters, cached tool data, or pending human-input markers in storage see those values disappear on resume, leading to incorrect control flow and broken human-in-the-loop graphs.

This PR threads storage through the snapshot/restore round-trip via opt-in registration:

```kotlin
install(Persistence) {
    storage = InMemoryPersistenceStorageProvider()
    persistedKeys = listOf(
        persisted(RetryCounterKey, Int.serializer()),
        persisted(PendingHumanInputKey, PendingHumanInput.serializer()),
    )
}
```

`Persistence.runFromCheckpoint` accepts the same list, for callers resuming an agent without the feature installed.

### Why opt-in registration over polymorphic / reflective serialization

- `AIAgentStorageKey` equality is referential, so the original key instance must be reused at restore. Even if storage values were uniformly serializable, callers would still need a way to map persisted key names back to live key instances. Registration covers both concerns at once.
- Reflective `KSerializer` lookup is not portable to Kotlin/JS or Wasm.
- Backward compatibility is automatic: unregistered keys are not persisted, so existing storage usage is unchanged unless callers opt in.

### Wire-up

- `PersistenceFeatureConfig.persistedKeys` carries the registration list for the auto-snapshot path.
- `AgentCheckpointData.storage: JSONObject?` holds the encoded entries. Default `null` plus `explicitNulls = false` means old checkpoints decode unchanged, so no structure-version bump is needed.
- `@JvmOverloads` is added to `AgentCheckpointData`'s primary constructor so Java callers binding to the 8-arg form still resolve to the typed `JSONElement` constructor instead of the deprecated `kotlinx.JsonElement` one.
- Encoding and decoding go through `PersistenceUtils.defaultCheckpointJson` with the user-supplied `KSerializer`, not the agent's `JSONSerializer`, so this code path doesn't break under non-kotlinx providers (e.g. Jackson).
- Per-entry encode/decode failures are logged and skipped: losing one piece of intermediate state on resume beats failing the whole session.

### Tests

`StoragePersistenceTest` (jvmTest) covers five scenarios:

- `checkpointSerializationRoundTripsStorageField` (encode/decode equality)
- `storageFieldOmittedWhenNull` (backward compatibility with old checkpoints)
- `runFromCheckpointRestoresRegisteredStorageEntries` (identity-keyed restore via the static API)
- `runFromCheckpointSkipsUnregisteredStorageEntries` (opt-in invariant: stale entries don't leak)
- `automaticPersistenceCapturesStorageAndRunFromCheckpointRestoresIt` (end-to-end through the installed feature)

## Test plan

- [x] `./gradlew :agents:agents-features:agents-features-snapshot:jvmTest` (54 tests, 0 failures)
- [x] `./gradlew :agents:agents-features:agents-features-snapshot:build` (JVM, JS, Wasm)
- [x] `./gradlew jvmTest` (whole tree)
- [x] `./gradlew :agents:agents-features:agents-features-snapshot:wasmJsTest`

### DIFF
--- a/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/context/AgentContextData.kt
+++ b/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/context/AgentContextData.kt
@@ -2,6 +2,7 @@
 
 package ai.koog.agents.core.agent.context
 
+import ai.koog.agents.core.agent.entity.AIAgentStorageKey
 import ai.koog.agents.core.annotation.InternalAgentsApi
 import ai.koog.prompt.message.Message
 import ai.koog.serialization.JSONElement
@@ -14,7 +15,8 @@ public class AgentContextData(
     internal val lastInput: JSONElement? = null,
     internal val lastOutput: JSONElement? = null,
     internal val rollbackStrategy: RollbackStrategy,
-    internal val additionalRollbackActions: suspend (AIAgentContext) -> Unit = {}
+    internal val additionalRollbackActions: suspend (AIAgentContext) -> Unit = {},
+    internal val storageEntries: Map<AIAgentStorageKey<*>, Any> = emptyMap(),
 ) {
     init {
         require(lastInput == null || lastOutput == null) { "`lastInput` and `lastOutput` cannot be both set" }
--- a/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/entity/AIAgentGraphStrategy.kt
+++ b/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/entity/AIAgentGraphStrategy.kt
@@ -123,6 +123,13 @@ public open class AIAgentGraphStrategyBase<TInput, TOutput>(
         agentContext.llm.withPrompt {
             this.withMessages { (data.messageHistory) }
         }
+
+        // Restore user-defined storage entries (intermediate state, retry counters, cached data, etc.).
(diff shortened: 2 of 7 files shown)


qflen:
Closed in favor of #1998.

PR #1962, opened 2026-05-06
feat(agents): Introduced AIAgentContextAwareTool

Rizzen:
Add a way for a `Tool` to receive an `AIAgentContext` during execution, **without breaking** any of existing `Tool` subclasses.

Introduced AIAgentContextAwareTool


agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/environment/AIAgentEnvironment.kt
```diff
@@ -43,6 +45,19 @@ public interface AIAgentEnvironment {
      * @return A list of results corresponding to the executed tool calls. Each result includes details
      *         such as the tool name, identifier, response content, and metadata.
      */
+    /**
```
EugeneTheDev, line 48: There's a KDoc for the `executeTools` above, please put it back to the `executeTools` method

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgentContextAwareTool.kt
```diff
@@ -0,0 +1,72 @@
...
+     * Default no-context fallback. The framework dispatches context-aware tools through
+     * [execute] with `(args, context)` whenever an [AIAgentContext] is attached to the
+     * environment — which is the case for any normal agent run. This method only fires when
+     * the tool is invoked outside of an agent run (e.g. direct unit-test invocation against
+     * a bare environment), and by default fails fast with a descriptive error.
+     *
+     * Most subclasses do not need to override this; do so only if you want a meaningful
+     * non-context fallback for tests or out-of-band invocation.
+     */
+    override suspend fun execute(args: TArgs): TResult =
```
EugeneTheDev, line 61: I'm really concerned with an implementation like this. 
1. It violates the Liskov Substitution Principle
2. The agent only knows about a generic collection of `Tool`s. Do you plan to update all places in the agent implementation that call tools to filter this type of tool`, AIAgentContextAwareTool` specifically and call it in a special way? This would increase the complexity by introducing conditional branching, and it's also quite fragile. If you forget to filter and call such a tool in a regular way via `execute(args)`, the agent will fail. That's why I mention LSP above. 

Can we design it differently?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/environment/AIAgentEnvironment.kt
```diff
@@ -43,6 +45,19 @@ public interface AIAgentEnvironment {
...
+     * Attaches the freshly built [AIAgentContext] to this environment so that [executeTool]
+     * implementations can dispatch pipeline events and route context-aware tools.
+     *
+     * Implementations that don't carry context (mocks, anonymous test envs) may keep the
+     * default no-op. The framework calls this exactly once per agent run, before the first
+     * tool call.
+     */
+    @InternalAgentsApi
+    public fun attachContext(context: AIAgentContext) {
```
EugeneTheDev, line 57: This blurs the line between `AIAgent` and `AIAgentEnvironment` even more, now that environment knows about the context. At this point it seems there's no much sense in keeping the environment separate (not like there was a lot of sense before IMHO). How about we finally remove the environment and merge its functionality into `AIAgent`?

EugeneTheDev (changes requested):

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/environment/AIAgentEnvironment.kt
Rizzen: I don't know, I don't like that idea. That's just a removal of extension point for me. 
I don't like the idea of putting whole "outside world interaction" chunk of code into AIAgentBase without being able to substitute it or reuse it.

Rizzen: We can discuss it separately anyway

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgentContextAwareTool.kt
Rizzen: Both of your points are valid and true. 

But problem is following:
1) We have Tools which are originates from `Tool<>` interface.
2) `Tool` interface is in `agent-tools` module
3) `AgentContext` is in `agent-core` module
4) A few of `prompt` modules are dependant of `agent-tools`
5) `agent-core` is depending of both `prompts` and `agent-tools` 

And we want to make `Tool` interface variant that do know about `agent-core`, which is a huge ooops in said configuration, because it produces cycle `core -> tools -> core`.

Initial idea was to merge tools and core, but that's realistically not possible right now, because of how `prompt` modules know about `tools` (that would produce cycle `core -> prompt -> core`).

I would want to design it differently and open for suggestions.

EugeneTheDev (approved):

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgentContextAwareTool.kt
serge-p7v: Hi! Sorry for intruding, have you seen the PR https://github.com/JetBrains/koog/pull/1886 that is somewhat related?

qflen: Hi all, author of #1886 here, jumping in on @serge-p7v's nudge.

I took a different angle on the same problem and I think it sidesteps the concerns raised in this thread (and Eugene's separate comment about the env/agent blur):

1. **LSP.** The new path is an `open` overload `Tool.execute(args, metadata: ToolCallMetadata)` whose default delegates to `execute(args)`. Existing tools keep working untouched, and the agent dispatches every tool through the same overload, so no `is`-checks or subtype filtering anywhere.

2. **Cycle.** `ToolCallMetadata` is a plain immutable `Map<String, Any?>`-shaped bag in `agents-tools` with zero `agent-core` imports. `Tool` itself never learns about `AIAgentContext`.

3. **Typed `AIAgentContext` access** (just pushed in 4013f61): a `ToolCallMetadata.agentContext: AIAgentContext?` extension in `agents-core`, plus auto-injection inside `ContextualAgentEnvironment.executeTool`. A tool that wants the context just writes:

```kotlin
   override suspend fun execute(args: Args, metadata: ToolCallMetadata): Result {
       val runId = metadata.agentContext?.runId
       // ...
   }
```

The extension lives in `agents-core`, so `Tool` stays clean of `agent-core` imports. The framework injects last and always wins on the reserved key, so callers and features cannot spoof the context.

4. **Env/context boundary stays.** The `AIAgentEnvironment` interface has no `AIAgentContext` reference, so the new overloads take `ToolCallMetadata`, which is an opaque map from `agents-tools`. Context-awareness stays isolated in the existing `@InternalAgentsApi`-marked `ContextualAgentEnvironment` decorator (which already held an `AIAgentContext` before this PR), so the public env layer remains swappable and the env-vs-agent split that Eugene flagged stays intact.

Rizzen: @qflen 
Hello! Thank you for the input, it's very interesting approach!

While your approach is more elegant in work with context, it has huge downside from API standpoint.

With introduction of 
`execute(args: TArgs, metadata: ToolCallMetadata)` into `Tool<>` interface it doesn't break existing tools - it's true. But users which are implementing their context-aware tools would have to impement not only 
`execute(args: TArgs, metadata: ToolCallMetadata)` but `execute(args: TArgs)` also. 
We could see that problem in tests:
```
private class MetadataAwareTool : SimpleTool<EchoArgs>(
    argsType = typeToken<EchoArgs>(),
    name = "metadata_aware",
    description = "Tool that echoes a value plus a piece of metadata.",
) {
    val observedMetadata: MutableList<ToolCallMetadata> = mutableListOf()

    override suspend fun execute(args: EchoArgs): String = error("Metadata overload must be used")
                                                           ^^^^^^^^^^^^^^^^^^^^^^ not good
    override suspend fun execute(args: EchoArgs, metadata: ToolCallMetadata): String {
        observedMetadata += metadata
        return "${args.value}::${metadata["trace.span.id"]}"
    }
}
```

This is a hard no for me from an API design perspective.

While the approach is technically source/binary compatible, it makes context-aware tools unpleasant to implement. A tool that actually depends on metadata is forced to also implement execute(args) even though that overload is meaningless for it.

As a result, implementations end up looking like this:

```
override suspend fun execute(args: EchoArgs): String =
    error("Metadata overload must be used")
```
and that is exactly the kind of API shape I’d really like to avoid.

In general, I strongly prefer not to put additional burden on API users just to simplify internals on our side. I’m somewhat okay with us struggling a bit more internally if that allows the public API to stay clean and pleasant to use.

qflen: @Rizzen oh I see. I agree with that call.

I've made another adjustment with ba219db8

`Tool.execute(args)` is now `open` and routes to `execute(args, ToolCallMetadata.EMPTY)` by default, so a metadata-aware tool overrides only the metadata overload, exactly the shape from your example, without the `error("...")` stub:

```kotlin
override suspend fun execute(args: EchoArgs, metadata: ToolCallMetadata): String {
    observedMetadata += metadata
    return "${args.value}::${metadata["trace.span.id"]}"
}
```

Existing tools that override only `execute(args)` are untouched (source- and binary-compatible).

If someone overrides neither, a private `NoOverrideMarker` coroutine-context element guards the mutual delegation. The metadata-aware default checks for it and throws `NotImplementedError` naming the tool and both overloads, rather than stack-overflowing. The marker is fully internal, so user overrides see exactly the `ToolCallMetadata` they were called with.

Tests cover the metadata-only shape across all four entry points (`execute` / `executeUnsafe` × with/without metadata) and the guard error on each.

Let me know what you think. I'm can of course do more tuning or defer.

Rizzen: I gave your approach some thought, and there are parts I really like and parts I’m not comfortable with.

I especially like the `ToolCallMetadata` approach, and I think that’s the direction we should move in.

My main concerns are around the somewhat tricky cross-dispatching and the fragile logic based on coroutine context parameters.

After a few discussions, we came up with a slightly different approach that should help us avoid circular dependencies and prevent moving large chunks of code between modules.

We’d use your approach and PR as the foundation, but introduce the following hierarchy:

```kotlin
// we introduce this one as very base
abstract class ToolBase<Arg, Result> {
    abstract suspend fun execute(arg: Arg, meta: ToolCallMetadata): Result
}
```

```kotlin
// existing one
// no API breakages
abstract class Tool<Arg, Result> : ToolBase<Arg, Result>() {
    abstract suspend fun execute(arg: Arg): Result

    override suspend fun execute(arg: Arg, meta: ToolCallMetadata): Result {
        return execute(arg)
    }
}
```

```kotlin
// our new and shiny context aware tool
abstract class AgentContextAwareTool<Arg, Result> : ToolBase<Arg, Result>() {

    companion object {
        val AgentContextKey = ...
    }

    abstract suspend fun execute(arg: Arg, context: AIAgentContext): Result

    override suspend fun execute(arg: Arg, meta: ToolCallMetadata): Result {
        return execute(arg, meta.getValue(AgentContextKey) as AIAgentContext)
    }
}
```

This way we:

* preserve the current tools API without breaking changes,
* introduce a dedicated context-aware tool abstraction,
* avoid circular dependencies,
* avoid relying on coroutine context propagation,
* and keep the design aligned with LSP.

Please let me know what you think about this approach. I’d really like to get it in before 1.0, which is coming pretty soon.

If you want and have time, it would be great if you could adapt your PR accordingly. Otherwise, I can implement it myself in a new co-authored PR based on #1886.

qflen: Hi, I read through and it sounds good. 

I will adapt #1886 to this hierarchy. 

The ToolBase split removes both the cross dispatch dance and the coroutine context marker, which were the rough edges I was least happy with too.

I will push a rebased diff early tomorrow.

qflen: Update: The changes are up now.

PR #1963, opened 2026-05-06
feat(agents)!: in LongTermMemory add failure policy and remove ingestion timing

serge-p7v:
The PR fixes two issues in  `LongTermMemory`:

1. Retrieval and ingestion errors on storage level are skipped.
2. Double-ingestion of messages during active session.

Also, the PR improves the default `InMemoryRecordStorage` and makes the feature non-experimental.

BREAKING:
* LongTermMemory IngestionTiming is gone
* QueryExtractor -> SearchQueryProvider
* ExtractionStrategy -> DocumentExtractor

DEPRECATED:
*

closes


EugeneTheDev (approved):
Looks good. Am I right that the main logical change is that there's no option to select ingestion timing now. Now automatic ingestion, if configured, happens only once per agent execution, after the agent completes, and the whole prompt is used to search for relevant facts?

And the `FailurePolicy`, of course

serge-p7v:
> Looks good. Am I right that the main logical change is that there's no option to select ingestion timing now. Now automatic ingestion, if configured, happens only once per agent execution, after the agent completes, and the whole prompt is used to search for relevant facts?
> 
> And the `FailurePolicy`, of course

Thanks! The ingestion now happens only at the end, but only the last user message (before LLM call) is used for retrieval with the default `LastUserMessageQueryExtractor`.
By the way, the naming might be a bit confusing:
- `ai.koog.agents.longtermmemory.retrieval.QueryExtractor` is for deciding which string to use for search in an external storage;
- `ai.koog.agents.longtermmemory.ingestion.extraction.ExtractionStrategy` is for converting prompt messages to text documents for inserting them into the external storage.

I'll think about the naming a bit more, "extract" in both places is not great.

PR #1968, opened 2026-05-07
test(agents): Added integration tests for AWS long-term memory feature

ashakirin:
Added integration tests for AWS long-term memory feature

AgentCore memory must exist (referenced by `AWS_AGENTCORE_MEMORY_ID`) with at least one of the following built-in strategies configured: SEMANTIC, USER_PREFERENCE, SUMMARIZATION. Each configured strategy must declare its namespace template using AgentCore's documented placeholders (`{memoryStrategyId}`, `{actorId}`, and for session-scoped strategies `{sessionId}`).
Slow EPISODIC end-to-end test, separated from the main test because EPISODIC extraction runs later and with higher latency than the other built-in strategies (empirically ~15-20 min between seeding and first retrievable episode).


serge-p7v (commented):
👍

serge-p7v (approved):

PR #1971, opened 2026-05-08
fix(agents): subgraphWithTask & subtask missing tool results in prompt if other tools where requested along with finish tool

EugeneTheDev:
In `subgraphWithTask` models can sometimes request other tools along with the finish tool. Previously, other tool results were ignored and not added to prompt, which led to error, since prompts can't have tool calls without matching tool results. Updated the `subgraphWithTask` logic to append all tool results when the finish tool was called

### DIFF
--- a/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt
+++ b/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt
@@ -18,6 +18,7 @@ import ai.koog.agents.core.dsl.extension.nodeLLMSendMultipleToolResults
 import ai.koog.agents.core.dsl.extension.nodeLLMSendToolResult
 import ai.koog.agents.core.environment.ReceivedToolResult
 import ai.koog.agents.core.environment.ToolResultKind
+import ai.koog.agents.core.environment.result
 import ai.koog.agents.core.environment.toSafeResult
 import ai.koog.agents.core.tools.Tool
 import ai.koog.agents.core.tools.ToolDescriptor
@@ -681,16 +682,28 @@ public fun <Input, Output, OutputTransformed> AIAgentSubgraphBuilderBase<Input, 
         defineTask(input)
     }
 
-    val finalizeTask by node<ReceivedToolResult, OutputTransformed>(
+    val finalizeTask by node<List<ReceivedToolResult>, OutputTransformed>(
         inputType = typeToken<ReceivedToolResult>(),
         outputType = outputTransformedType
-    ) { toolResult ->
+    ) { toolResults ->
         llm.writeSession {
+            // Append all tool results to the prompt, otherwise there will be calls without results, which is invalid
+            appendPrompt {
+                tool {
+                    toolResults.forEach { result(it) }
+                }
+            }
+
             // Restore original tools
             tools = storage.get(originalToolsKey)!!
         }
 
-        toolResult.toSafeResult(finishTool, config.serializer).asSuccessful().result
+        // Take the first finish tool and return as a result
+        toolResults
+            .first { it.tool == finishTool.name }
+            .toSafeResult(finishTool, config.serializer)
+            .asSuccessful()
+            .result
     }
 
     // Helper node to overcome problems of the current api and repeat less code when writing routing conditions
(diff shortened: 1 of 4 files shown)


agents/agents-core/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt
serge-p7v: Is that a misspelling?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt
```diff
@@ -681,16 +682,28 @@ public fun <Input, Output, OutputTransformed> AIAgentSubgraphBuilderBase<Input,
         defineTask(input)
     }
 
-    val finalizeTask by node<ReceivedToolResult, OutputTransformed>(
+    val finalizeTask by node<List<ReceivedToolResult>, OutputTransformed>(
         inputType = typeToken<ReceivedToolResult>(),
```
Amaneusz: Shouldn't that also be converted to `typeToken<List<ReceivedToolResult>>`?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt
```diff
@@ -681,16 +682,28 @@ public fun <Input, Output, OutputTransformed> AIAgentSubgraphBuilderBase<Input,
...
+            }
+
             // Restore original tools
             tools = storage.get(originalToolsKey)!!
         }
 
-        toolResult.toSafeResult(finishTool, config.serializer).asSuccessful().result
+        // Take the first finish tool and return as a result
+        toolResults
+            .first { it.tool == finishTool.name }
```
Amaneusz: In L811-812 your predicate was tighter:
```
.any { it.tool == finishTool.name && it.resultKind is ToolResultKind.Success }
```

Filtering just by tool name might pick a failed result, I guess that's not intended?

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt
Amaneusz, line 819: If I understand your change correctly, then this part might also be worth taking a look at as it only forwards first from potentially many tool results

EugeneTheDev: It was a bug from before, indeed. Fixed it

EugeneTheDev: That's by design, because this part is controlled by the `runMode`, so in this branch only single tool calls per turn will be returned, so this `first` is just a type compatibility thing

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt
EugeneTheDev: Thanks, fixed

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt
EugeneTheDev: Fixed

agents/agents-core/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt
```diff
@@ -872,16 +885,6 @@ internal suspend fun <Output, OutputTransformed> AIAgentContext.executeFinishToo
...
 
-    // Append a final tool call result to the prompt for further LLM calls
-    // to see it (otherwise they would fail)
-    llm.writeSession {
-        appendPrompt {
-            tool {
-                result(toolCall.id, toolCall.tool, toolCall.content)
-            }
-        }
-    }
```
Amaneusz, line 883: I think removing this might need more attention
- `executeFinishTool` is called from `AIAgentFunctionalContextBaseCommon.executeMultipleToolsHacked`
- which is called from two places: `subtaskWithSingleToolMode` and `subtaskWithMultiToolMode`

Both of these return immediately when the finish result is available:
```
// this happens in subtaskWithMultiToolMode
toolResults.firstOrNull { it.tool == finishTool.descriptor.name }
    ?.let { finishResult ->
        return finishResult.toSafeResult(finishTool, config.serializer).asSuccessful().result
    }
```

They don't have their version of `finalizeTask` that would append to prompt before returning, appending happened in this removed part here (so for these two scenarios we would end up with `Tool.Call` but without `Tool.Result`)

EugeneTheDev: Good catch, thanks, I will address it

EugeneTheDev: Updated subtasks in functional context too, added tests to verify

Amaneusz (approved):
LGTM, thanks!

serge-p7v (approved):

PR #1975, opened 2026-05-09
build(agents): properly ignore all wasm publsihing tasks for otel module

EugeneTheDev:
The existing task filter only matched names starting with wasmJs, so publication tasks like generateMetadataFileForWasmJsPublication (where WasmJs appears mid-name) slipped through and failed reading the klib that the disabled compileKotlinWasmJs never produced. Broadened the filter to also match contains("WasmJs")

### DIFF
--- a/agents/agents-features/agents-features-opentelemetry/build.gradle.kts
+++ b/agents/agents-features/agents-features-opentelemetry/build.gradle.kts
@@ -123,7 +123,9 @@ extensions.configure<LibraryAndroidComponentsExtension> {
 configurations.matching { it.name.startsWith("wasmJs") }.configureEach {
     exclude(group = "io.opentelemetry.kotlin")
 }
-tasks.matching { it.name.startsWith("wasmJs") || it.name == "compileKotlinWasmJs" || it.name == "compileTestKotlinWasmJs" }.configureEach {
+// Match both camelCase prefixed tasks (wasmJsJar, wasmJsSourcesJar, ...) and PascalCase
+// suffixed ones (compileKotlinWasmJs, generateMetadataFileForWasmJsPublication, ...).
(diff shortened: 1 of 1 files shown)


EugeneTheDev:
I don't know why `dokkaGenerate` fails on CI. Locally it works, tried `./gradlew dokkaGenerate --rerun`. The branch is also completely up-to-date with `develop`. Let's merge it and see, I guess.

agents/agents-features/agents-features-opentelemetry/build.gradle.kts
```diff
@@ -123,7 +123,9 @@ extensions.configure<LibraryAndroidComponentsExtension> {
 configurations.matching { it.name.startsWith("wasmJs") }.configureEach {
     exclude(group = "io.opentelemetry.kotlin")
 }
-tasks.matching { it.name.startsWith("wasmJs") || it.name == "compileKotlinWasmJs" || it.name == "compileTestKotlinWasmJs" }.configureEach {
+// Match both camelCase prefixed tasks (wasmJsJar, wasmJsSourcesJar, ...) and PascalCase
+// suffixed ones (compileKotlinWasmJs, generateMetadataFileForWasmJsPublication, ...).
+tasks.matching { it.name.startsWith("wasmJs") || it.name.contains("WasmJs") }.configureEach {
```
sdubov, line 128: This can be re-written to:
```
tasks.matching { it.name.contains("wasmjs", ignoreCase = true) }.configureEach { ... }
```

However, let me check if we can disable wasm target in more solid way. I'm not sure that this check for task name is the best we can do here.

sdubov (approved):

EugeneTheDev:
#1994 implemented it properly

PR #1979, opened 2026-05-10
examples: Android local models

tiginamaria:

BREAKING:
*

DEPRECATED:
*

closes


examples/demo-compose-app/androidApp/src/main/kotlin/com/jetbrains/example/koog/compose/local/LiteRTLLModels.kt
```diff
@@ -0,0 +1,45 @@
...
+ * koog executor infrastructure. Custom models can be added at runtime via [addCustomModel].
+ */
+public object LiteRTLLModels : LLModelDefinitions {
+    /**
+     * A fine-tuned Gemma variant optimized for function-calling on device.
+     *
+     * Model file: `tiny_garden.litertlm`
+     * Source: https://huggingface.co/litert-community/functiongemma-270m-ft-tiny-garden/tree/main
+     */
+    public val FunctionGemma: LLModel = LLModel(
```
serge-p7v, line 24: It would be great to add the most downloaded gemma-4 models (E2B and E4B).

PR #1994, opened 2026-05-12
build(agents): KG-836. Drop wasmJs target for open-telemetry feature

sdubov:
- OTel Kotlin SDK 0.3.0 does not publish wasmJs artifacts. Replace the regex-based workaround in the OTel build with a koog.target.wasmJs property gate in the convention plugin, and route OTel through a local nonWasmJsMain intermediate source set in :koog-agents so its wasm-js publication no longer references a missing coordinate.

closes: [KG-836](https://youtrack.jetbrains.com/issue/KG-836)

### DIFF
--- a/agents/agents-features/agents-features-opentelemetry/build.gradle.kts
+++ b/agents/agents-features/agents-features-opentelemetry/build.gradle.kts
@@ -115,16 +115,4 @@ extensions.configure<LibraryAndroidComponentsExtension> {
     }
 }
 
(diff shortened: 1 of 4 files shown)


EugeneTheDev (changes requested):
Thank you, I'm not an expert, but after reading through and thinking a bit the fix appears reasonable

EugeneTheDev (approved):

sdubov:
Force-pushed renamed variable name `enableWasmJs` -> `isWasmJsIncluded`

PR #1998, opened 2026-05-12
feat(agents): add AIAgentStorage to Persistence feature

EugeneTheDev:
Extends agent checkpoints to include AIAgentStorage contents, so custom key-value data is saved and restored when resuming from a checkpoint. Values not serializable by the configured serializer are silently dropped.
  
  Storage refactor to support serialization:
  - AIAgentStorageKey now carries a TypeToken and uses name-based equality. Kotlin callers use the new inline reified createStorageKey<T>(name); Java callers use createStorageKey(name, TypeToken.of(...)).
  - AIAgentStorageImpl maintains a typed map and a serialized map side-by-side — deserialization is lazy (values are decoded on first access).
  - AIAgentStorage now requires a JSONSerializer
  - New API: toSerializedMap(), putAllSerialized(), putAll(other: AIAgentStorage). copy() promoted to public. toMap() removed.

  BREAKING:
  - AIAgentStorageKey equality changed from referential to name-based.
  - AIAgentStorageAPI.toMap() removed — use toSerializedMap() or putAll(other).
  - AIAgentStorage() no-arg constructor replaced by AIAgentStorage(serializer).
  - runFromCheckpoint parameter renamed from agentInput to input.

### DIFF
--- a/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgentRunSessionImpl.kt
+++ b/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgentRunSessionImpl.kt
@@ -60,7 +60,7 @@ internal class AIAgentRunSessionImpl<Input, Output, TContext : AIAgentContext>(
 
         when (sessionInputs) {
             is AdditionalInputs.None -> {}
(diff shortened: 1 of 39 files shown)


Rizzen (approved):
Sterk werk

PR #2000, opened 2026-05-12
build: set up `isBeta` extra property for all gradle subprojects

EugeneTheDev:
We've agreed on the new versioning scheme that will separate between "stable" and "beta" modules. "beta" modules will contain experimental APIs, and to distinguish between "stable" and "beta", "beta" modules will get a "-beta" postfix in their version, e.g. "1.0-beta".

To simplify maintenance, `isBeta` extra property is introduced for all Koog modules. Now, there's no need to configure `group` and `version` manually in each module, these are controlled centrally. To mark a module as experimental, one can simply set the following somewhere in the root of a module's `build.gradle.kts`:
```
val isBeta by extra(true)
```

### DIFF
--- a/agents/agents-features/agents-features-a2a-server/build.gradle.kts
+++ b/agents/agents-features/agents-features-a2a-server/build.gradle.kts
@@ -1,7 +1,5 @@
 import ai.koog.gradle.publish.maven.Publishing.publishToMaven
 
-group = rootProject.group
-version = rootProject.version
 
 plugins {
     id("ai.kotlin.multiplatform")
--- a/a2a/a2a-client/build.gradle.kts
+++ b/a2a/a2a-client/build.gradle.kts
@@ -1,7 +1,6 @@
 import ai.koog.gradle.publish.maven.Publishing.publishToMaven
 
-group = rootProject.group
-version = rootProject.version
+val isBeta by extra(true)
 
 plugins {
     id("ai.kotlin.multiplatform")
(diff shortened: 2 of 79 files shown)


EugeneTheDev:
Next steps:
1. Go through all the modules and mark some of them as beta (currently only A2A SDK modules are marked as beta, just to give an example)
2. Implement a check in `koog-agents` to make sure that all-in-one dependency doesn't bring beta modules

agents/agents-features/agents-features-a2a-server/build.gradle.kts
```diff
@@ -1,7 +1,5 @@
 import ai.koog.gradle.publish.maven.Publishing.publishToMaven
 
-group = rootProject.group
-version = rootProject.version
```
Ololoshechkin, line 4: Don't we need this?

Ololoshechkin (approved):
Solid work!
@EugeneTheDev please test TC publication from a branch :)

Ololoshechkin:
@EugeneTheDev to make it fair test, feel free to create `0.9.0` branch, set the main version to `0.9.0-test` and verify that everything will be published as `0.9.0-test` and `0.9.0-test-beta`, respectively. WDYT?

agents/agents-features/agents-features-a2a-server/build.gradle.kts
EugeneTheDev: As I wrote in the PR description, it's now controlled centrally, from the root `build.gradle.kts` https://github.com/JetBrains/koog/pull/2000/changes#diff-c0dfa6bc7a8685217f70a860145fbdf416d449eaff052fa28352c5cec1a98c06R79

EugeneTheDev:
Tried to publish test version, it worked as expected, so I'm merging it

PR #2011, opened 2026-05-14
build!: Split modules into stable and beta

Ololoshechkin:


VERSIONING.md
```diff
@@ -0,0 +1,161 @@
...
+A stable module at version `X.Y.Z` is always compatible with a beta module at version `X.Y.Z-beta` (and vice versa). All modules can be updated in sync.
+
+### Umbrella Modules
+
+| Module | Version | Contents |
+|--------|---------|----------|
+| `koog-agents` | `1.0.0` | All stable modules (transitive) — recommended starting point |
+| `koog-agents-additions` | `1.0.0-beta` | Most beta/experimental modules (except standalone external integrations) |
+
+### Module Versions
```
EugeneTheDev, line 59: Let's probably skip detailed list of module version: it's gonna be a hassle to maintain. Let's keep two more concise lists instead: stable modules and beta modules, as a simple bullet list. The version should be evident from the description above

EugeneTheDev (approved):

VERSIONING.md
Ololoshechkin: It's easy to mantain -- just run ./gradlew listModules and give it to an agent

EugeneTheDev: That's true, but the question is more like how much benefit does it provide? IMHO it's not much worse to just list the modules, the version is usually visible at the top of the README. We can add another badge for beta versions too. But that's just a suggetion, feel free to dismiss it

Ololoshechkin: Let's keep it as is now -- to educate folks (and AI agents) for the start

PR #2013, opened 2026-05-15
fix(agents): support sealed type outputs in subgraphWithTask

### LINKED ISSUE #1941: Support sealed type outputs in subgraphWithTask
Currently, `subgraphWithTask` uses tool calling to get the final structured result. It seems that tool arguments don't support sealed types yet. So a code like this `subgraphWithTask<String, MySealedInterface>` fails.

qflen:
Closes #1941

`FinishTool` (the internal tool that `subgraphWithTask<Input, Output>` uses to carry the structured result) generates its schema via `kotlinx-schema-generator-json`. 

For a sealed `Output` two things stacked: 

- `JsonSchema.toActualPropertyDefinition` only dispatched on `type` and dropped the `oneOf` that the generator emits for polymorphic top-level schemas
- FinishTool requested `includePolymorphicDiscriminator = false`, so even the per-branch schemas omitted the `type` const that kotlinx-serialization needs to decode back into a sealed instance. 

The combined effect: `subgraphWithTask<String, MySealedInterface>` advertised an empty `result` Object to the LLM and could not decode whatever it returned.

The converter now prefers `oneOf`/`anyOf` when present, and FinishTool generates its schema with the discriminator enabled. Other tools keep `defaultJsonSchemaConfig` unchanged. 

OpenAI and Google emit the resulting `anyOf` directly, and Anthropic's pre-existing "AnyOf type is not supported" limitation continues to apply there.

New tests pin the converter at the schema layer, the FinishTool descriptor and round-trip, and the end-to-end behavior through `subgraphWithTask` with a raw `PromptExecutor` returning the polymorphic JSON.


agents/agents-tools/src/commonTest/kotlin/ai/koog/agents/core/tools/schema/SealedTopLevelSchemaTest.kt
```diff
@@ -0,0 +1,80 @@
...
+import ai.koog.serialization.typeToken
+import kotlinx.schema.generator.json.JsonSchemaConfig
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+
+@OptIn(InternalAgentToolsApi::class)
+class SealedTopLevelSchemaTest {
```
EugeneTheDev: I would suggest a few things here:
1. Let's put these tests directly into existing `SerializableSchemaGeneratorTest` . It would make referencing and maintaining tests easier
2. Let's make the tests themselves a bit more straightforward. Currently you have several asserts and a custom logic to validate the correctness of the generated tool descriptor. I suggest using the same pattern as the existing `SerializableSchemaGeneratorTest` tests and just doing a single assert, matching on a full expected tool descriptor.

agents/agents-core/src/commonTest/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphFinishToolTest.kt
```diff
@@ -91,4 +94,60 @@ class AIAgentSubgraphFinishToolTest {
...
+        data class B(val number: Int) : SealedOutput
+    }
+
+    @Test
+    fun `generates ToolDescriptor for sealed output with anyOf and discriminator`() {
+        val finishTool = FinishTool<SealedOutput>(typeToken<SealedOutput>())
+        val resultParameter = finishTool.descriptor.requiredParameters.single { it.name == "result" }
+
+        val anyOf = resultParameter.type
+        check(anyOf is ToolParameterType.AnyOf) {
```
EugeneTheDev: The same thing here about just doing a single full assert

EugeneTheDev (changes requested):
The implementation looks good, thank you. I have a few minor comments on tests

qflen:
Thanks, pushed changes.

agents/agents-features/agents-features-opentelemetry/src/jvmTest/kotlin/ai/koog/agents/features/opentelemetry/feature/span/OpenTelemetryInferenceSpanTest.kt
```diff
@@ -553,7 +553,10 @@ class OpenTelemetryInferenceSpanTest : OpenTelemetryTestBase() {
         assertFalse(exception is CancellationException, "Unexpected cancellation exception")
 
         testData.collectedSpans = withTimeout(10.seconds) {
-            spanExporter.isCollected.first()
+            // Wait until the root create-agent span is exported (it ends last), so all child spans are
+            // present. `first()` would return the StateFlow's current value without waiting, racing the
+            // async export and intermittently leaving filterAgentInvokeSpans() empty (flaky on Windows CI).
+            spanExporter.isCollected.first { it }
```
EugeneTheDev, line 559: Nice catch, thanks

EugeneTheDev (approved):

PR #2046, opened 2026-05-20
fix(agents): Change provider-specific SpringBoot beans to PromptExecutor

Ololoshechkin:


nomisRev (approved):

nomisRev:
Context:

In `1.0.0-beta-preview2` its required to specify `@Qualifier` for _any_ `PromptExecutor` that is injected due to all `@Bean` being typed to `MultiLLMPromptExecutor`.

```kotlin
@Service
class ChatAgentProvider(
    @Qualifier("multiLLMPromptExecutor")
    private val executor: MultiLLMPromptExecutor,
```

Instead its desired to expose just `MultiLLMPromptExecutor` for _all_ configured llm clients

```kotlin
@Service
class ChatAgentProvider(
    private val executor: MultiLLMPromptExecutor
```

And if a specific executor is required it requires _always_ qualifier.

```kotlin
@Service
class ChatAgentProvider(
    @Qualifier("openAIExecutor")
    private val executor: PromptExecutor,
```

PR #2052, opened 2026-05-21
fix(prompt): Don't start a new tool call on repeated tool call id

### LINKED ISSUE #2002: `StreamFrameFlowBuilder.emitToolCallDelta()` incorrectly treats every chunk with non-null `id` as a new tool call
### Description

When streaming tool call responses from certain OpenAI-compatible LLM providers, the `StreamFrameFlowBuilder.emitToolCallDelta()` method incorrectly emits a `ToolCallComplete` frame after **every** `ToolCallDelta` chunk, instead of accumulating the chunks and only emitting one `ToolCallComplete` at the end.

This causes a single logical tool call to be fragmented into multiple "complete" tool calls, each containing only a small piece of the arguments.

### Root Cause

In [`StreamFrameFlowBuilder.kt`](prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/streaming/StreamFrameFlowBuilder.kt), line 196:

```kotlin
val new: PendingToolCall = if (id != null || index != previous?.index) {
    tryEmitPendingToolCall()
    PendingToolCall(id, name, args, index)
} else {
    // append to existing pending tool call...
}
```

The condition `id != null` assumes that only the **first** chunk of a tool call carries a non-null `id`. This is true for the standard OpenAI Chat Completions streaming format (where `OpenAIStreamToolCall.id` is `String?` and only present in the first chunk), but it breaks for providers that include `id` in **every** streaming chunk.

When `id` is present in every chunk, each `emitToolCallDelta()` call is treated as the start of a **new** tool call, which:
1. Calls `tryEmitPendingToolCall()` → emits `ToolCallComplete` for the previous (incomplete) chunk
2. Creates a brand-new `PendingToolCall`, discarding the accumulated state

### Observed Behavior

Input stream (3 chunks for one tool call):

```
ToolCallDelta(id=call_1, name=readFile, content=,         index=0)
ToolCallDelta(id=call_1, name=null,    content={"path": ", index=0)
ToolCallDelta(id=call_1, name=null,    content=/Users,     index=0)
```

Actual output (broken):

```
ToolCallDelta(id=call_1, name=readFile, content=,          index=0)
ToolCallComplete(id=call_1, name=readFile, content=)          ← premature!
ToolCallDelta(id=call_1, name=null,    content={"path": ", index=0)
ToolCallComplete(id=call_1, name=,     content={"path": ")    ← premature!
ToolCallDelta(id=call_1, name=null,    content=/Users,      index=0)
ToolCallComplete(id=call_1, name=,     content=/Users)         ← premature!
```

Expected output:

```
ToolCallDelta(id=call_1, name=readFile, content=,          index=0)
ToolCallDelta(id=call_1, name=null,    content={"path": ", index=0)
ToolCallDelta(id=call_1, name=null,    content=/Users,     index=0)
ToolCallComplete(id=call_1, name=readFile, content={"path": "/Users"}  index=0)  ← one complete call
```

### Suggested Fix

Change the condition in `emitToolCallDelta()` from:

```kotlin
if (id != null || index != previous?.index)
```

to:

```kotlin
if ((id != null && id != previous?.id) || (index != null && index != previous?.index))
```

This ensures that a chunk with the same `id` as the previous one is treated as a continuation, not a new tool call. The behavior for `null` `id` (continuation chunks) is preserved, and `null` `index` is also handled correctly.

### Affected Providers

This bug affects any OpenAI-compatible LLM provider that includes the `id` field in every streaming chunk of a tool call. The standard OpenAI Chat Completions API only sends `id` in the first chunk (`OpenAIStreamToolCall.id` is `String?`), so it is unaffected. However:

- **OpenRouter** (`OpenRouterLLMClient`) uses `OpenAIToolCall` which has `id: String` (non-nullable), meaning every chunk carries the `id`. This is likely affected.
- **Other OpenAI-compatible providers** that include `id` in every chunk will also be affected.

jewoodev:
Closes #2002.

### Summary

`StreamFrameFlowBuilder.emitToolCallDelta()` treated every chunk with a non-null tool-call `id` as a new tool call. This fix treats a repeated `id` as a continuation of the pending tool call, so OpenAI-compatible providers that repeat the same `id` on each streaming chunk no longer fragment one logical tool call into multiple premature `ToolCallComplete` frames.

### Fix

A new tool call now begins only when a present `id` or `index` differs from the pending call, instead of whenever `id` is non-null.

Before:

```kotlin
val new: PendingToolCall = if (sanitizedId != null || index != previous?.index) {
```

After:

```kotlin
val isNewToolCall =
    (sanitizedId != null && sanitizedId != previous?.id) ||
        (index != null && index != previous?.index)
val new: PendingToolCall = if (isNewToolCall) {
```

This follows the same id-change principle as `emitReasoningDelta`; absent or blank ids still continue the pending call (the blank case preserves the #1915 handling). The fix lives in the shared `StreamFrameFlowBuilder` because all streaming clients (OpenAI, OpenRouter, Mistral, Anthropic, DeepSeek, Ollama, Bedrock, Dashscope, Google) emit tool-call deltas through it.

### Tests

Two regression tests in `StreamFrameFlowBuilderTest`: `testEmitToolCallDeltaWithRepeatedIdAppendsToExisting` (three chunks with the same `id` combine into one `ToolCallComplete`) and `testEmitToolCallDeltaWithRepeatedIdSeparatesDistinctToolCalls` (switching to a different `id` still starts a separate tool call). `:prompt:prompt-model:jvmTest` (20 tests, 0 failures), `wasmJsNodeTest`, and `ktlintCheck` all pass locally; the full `./gradlew build` (Android and browser targets) is left to CI, since this contributor environment lacks the Android SDK and a headless browser.

### DIFF
--- a/agents/agents-features/agents-features-opentelemetry/src/jvmTest/kotlin/ai/koog/agents/features/opentelemetry/OpenTelemetryTestAPI.kt
+++ b/agents/agents-features/agents-features-opentelemetry/src/jvmTest/kotlin/ai/koog/agents/features/opentelemetry/OpenTelemetryTestAPI.kt
@@ -233,7 +233,12 @@ internal object OpenTelemetryTestAPI {
                 addSpanProcessor { simpleSpanProcessor(mockExporter) }
                 setVerbose(verbose)
             }.use { agent ->
-                agent.run(userPrompt ?: USER_PROMPT_PARIS)
+                try {
+                    agent.run(userPrompt ?: USER_PROMPT_PARIS)
+                } catch (throwable: Throwable) {
+                    waitSpansCollected(mockExporter)
+                    throw throwable
+                }
             }
 
             waitSpansCollected(mockExporter)
--- a/prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/streaming/StreamFrameFlowBuilder.kt
+++ b/prompt/prompt-model/src/commonMain/kotlin/ai/koog/prompt/streaming/StreamFrameFlowBuilder.kt
@@ -197,7 +197,13 @@ public class StreamFrameFlowBuilder(
         tryEmitPendingReasoning()
         val sanitizedId = id?.takeUnless { it.isBlank() }
         val previous: PendingToolCall? = pendingToolCallRef.load()
-        val new: PendingToolCall = if (sanitizedId != null || index != previous?.index) {
+        // `id` and `index` are optional per-chunk signals. A new tool call begins only when a
+        // present signal differs from the pending call, not merely because `id` is non-null:
+        // some OpenAI-compatible providers send the same `id` on every chunk (see #2002).
+        val isNewToolCall =
+            (sanitizedId != null && sanitizedId != previous?.id) ||
+                (index != null && index != previous?.index)
+        val new: PendingToolCall = if (isNewToolCall) {
             tryEmitPendingToolCall()
             PendingToolCall(sanitizedId, name, args, index)
         } else {
--- a/prompt/prompt-model/src/commonTest/kotlin/ai/koog/prompt/streaming/StreamFrameFlowBuilderTest.kt
+++ b/prompt/prompt-model/src/commonTest/kotlin/ai/koog/prompt/streaming/StreamFrameFlowBuilderTest.kt
@@ -224,6 +224,63 @@ class StreamFrameFlowBuilderTest {
         )
     }
 
+    /**
+     * Regression test for #2002.
+     *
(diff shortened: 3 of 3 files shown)


tiginamaria (approved):
Thank you for the fix! Looks good!

zarechneva-a (approved):
Thank you!

jewoodev:
Thanks for the review and for landing the OpenTelemetry test fix on top to unblock CI — really appreciated.

PR #2062, opened 2026-05-24
fix: stream Bedrock reasoning content deltas

pecpwee:
Emit reasoning text deltas from Bedrock Converse streaming responses as `StreamFrame.ReasoningDelta`.

Previously, `ContentBlockDelta.ReasoningContent` chunks were treated as unsupported, so providers/models that stream reasoning through Bedrock Converse dropped those deltas. This change maps `ReasoningContentBlockDelta.Text` to Koog reasoning stream frames while continuing to skip redacted content and signatures.

Also adds a regression test covering Converse streaming reasoning deltas.

I’ve verified this on Bedrock’s minimax.minimax-m2.5 model. After the code changes, the reasoning content is now displayed correctly.


pecpwee:
@aozherelyeva Could you please review this PR

PR #2065, opened 2026-05-25
feat(http-client): add Spring WebClient support

Amaneusz:
Adds support for Spring `WebClient` as a `KoogHttpClient` implementation and removes `@Experimental` annotations from stable HTTP client APIs/modules.
Adds documentation for http clients, factories and automatic resolution on JVM.

closes KG-820


antoniibelyshev (approved):
Thank you, looks good! Please, rebase and run `updateLegacyAbi`, compilation task on ci will fail otherwise (repeating what I wrote in slack, just in case 🙂)

tiginamaria (approved):
Also look good to me! Thank you!

Issue #2017, opened 2026-05-16
MCP tool integration: oneOf not supported in schema parser, and LLM-generated arguments not coerced


ronsmits:
## Summary

When using MCP servers whose tool schemas use JSON Schema `oneOf` (rather than `anyOf`), the `DefaultMcpToolDescriptorParser` fails to parse parameters correctly. Additionally, even when schemas are parsed, LLMs frequently stringify nested JSON objects in tool call arguments, and the `McpTool` forwards them as-is — causing MCP servers to reject the call.

Both issues are observable with the Obsidian MCP server (`obsidian-mcp`), but affect any MCP server using `oneOf` or nested object parameters.

## Issue 1: `oneOf` not handled by `DefaultMcpToolDescriptorParser`

**Context:** Many MCP servers (notably Obsidian MCP) use JSON Schema `oneOf` for union-type parameters. For example, the `target` parameter in `obsidian_write_note`:

```json
"target": {
  "description": "Where the note lives.",
  "oneOf": [
    { "type": "object", "properties": { "type": { "const": "path" }, "path": { ... } } },
    { "type": "object", "properties": { "type": { "const": "active" } } },
    { "type": "object", "properties": { "type": { "const": "periodic" }, ... } }
  ]
}
```

**What happens:** `DefaultMcpToolDescriptorParser.parseParameterType()` checks for `type`, `anyOf`, and `enum` — but **not `oneOf`**. When a parameter has `oneOf` without `type` or `anyOf`, the parser throws:

```
IllegalArgumentException: Parameter <name> must have type property
```

**Expected:** `oneOf` should be treated equivalently to `anyOf` for the purpose of `ToolParameterType.AnyOf` construction.

**Suggested fix in `DefaultMcpToolDescriptorParser.parseParameterType()`:**
```kotlin
// After the existing anyOf check, add:
val oneOf = element["oneOf"]?.jsonArray
if (oneOf != null) {
    return ToolParameterType.AnyOf(
        types = oneOf.map { it.jsonObject }.map {
            ToolParameterDescriptor(
                name = "",
                description = it["description"]?.jsonPrimitive?.content.orEmpty(),
                type = parseParameterType(it.jsonObject)
            )
        }.toTypedArray()
    )
}
```

Or normalize `oneOf` → `anyOf` before parsing.

## Issue 2: `McpTool` does not coerce stringified JSON arguments

[...]

PR #2067, opened 2026-05-25
fix(agents): support oneOf and const schemas in MCP tool parser

jewoodev:
**Summary**

- `DefaultMcpToolDescriptorParser.parseParameterType` only inspected `$ref → type → anyOf → enum`, so MCP server schemas using top-level `oneOf` (no `type`) fell through to `throw IllegalArgumentException("Parameter ${title} must have type property")` at `agents/agents-mcp/src/commonMain/kotlin/ai/koog/agents/mcp/McpToolDefinitionParser.kt:127-129`, and `buildToolRegistry` then logged the parse failure and silently skipped registering the tool. The example from the issue (`obsidian_write_note.target`) is `oneOf` of objects whose discriminator property is `{"const": "path"|"active"|"periodic"}`, so a `oneOf`-only fix would still re-hit the same throw site inside the candidate objects' properties.
- Add two sibling branches inside the `typeStr == null` block: `oneOf → ToolParameterType.AnyOf(...)` (mirrors the existing `anyOf` handling; koog only models a union, and the MCP server enforces exact-one semantics on the wire), and `const → ToolParameterType.Enum(arrayOf(enumEntryToString(const)))` (JSON Schema 7.1.1 defines `const: X` as equivalent to `enum: [X]`, and the existing `enum` path already stringifies entry values). Both decisions are documented in KDoc on the new branches. Branch ordering becomes `anyOf → oneOf → const → enum → throw`.

Refs #2017 (parser side; runtime-side coercion is tracked in a separate PR).

**Test plan**

- [x] `gradle :agents:agents-mcp:jvmTest --tests "ai.koog.agents.mcp.DefaultMcpToolDescriptorParserTest"`
- [x] `gradle :agents:agents-mcp:ktlintCheck`
- [x] `gradle :agents:agents-mcp:compileKotlinJvm --warning-mode=all` — zero warnings in `agents-mcp`.

### DIFF
--- a/agents/agents-mcp/src/commonMain/kotlin/ai/koog/agents/mcp/McpToolDefinitionParser.kt
+++ b/agents/agents-mcp/src/commonMain/kotlin/ai/koog/agents/mcp/McpToolDefinitionParser.kt
@@ -107,6 +107,54 @@ public object DefaultMcpToolDescriptorParser : McpToolDescriptorParser {
             }
 
             /**
+             * oneOf with multiple types.
+             *
+             * JSON Schema semantically distinguishes oneOf (exactly one) from anyOf (one or more),
+             * but our internal [ToolParameterType] only models a union (AnyOf). For tool-argument
+             * descriptors this is sound: argument values are validated by the MCP server, and the
+             * koog-side descriptor only enumerates the legal shapes for the LLM to choose from.
+             *
+             * Schema example:
+             * {
+             *   "oneOfParam": {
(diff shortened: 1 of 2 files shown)


zarechneva-a:
Hi @jewoodev, thank you for the contribution! Could you please address the failed checks? Thank you!

jewoodev:
@aozherelyeva Thank you for your feedback. I committed again as you recommended, and it passed all tests :)

PR #2070, opened 2026-05-26
Fix double JSON encoding of function.arguments in AbstractOpenAILLMClient

HaishiR:
## Problem
  `MessagePart.Tool.Call.args` is already a JSON string (encoded in its constructor via `Json.encodeToString(args)`). However, `AbstractOpenAILLMClient.convertPromptToMessages` calls `Json.encodeToString(it.args)` again, causing double encoding.
  
  ## Impact
  - **Dashscope (Qwen)**: Strict API validation rejects the double-encoded string with `"The \"function.arguments\" parameter ... must be in JSON format."`
  - **DeepSeek**: Works by accident due to lenient parsing.
  
  ## Fix
  Remove the redundant `Json.encodeToString` call.
  
  ```diff
  - function = OpenAIFunction(it.tool, Json.encodeToString(it.args))
  + function = OpenAIFunction(it.tool, it.args)


serge-p7v:
Hi @HaishiR ! Thank you for your contribution. It seems that the the PR https://github.com/JetBrains/koog/pull/2096 with the removal of `Json.encodeToString` was merged yesterday and the current PR cannot be merged due to the merge conflict.

PR #2071, opened 2026-05-26
fix(agents): coerce stringified MCP tool args using descriptor types

### LINKED ISSUE #2017: MCP tool integration: oneOf not supported in schema parser, and LLM-generated arguments not coerced

jewoodev:
**Summary**

- `McpTool.execute` at `agents/agents-mcp/src/commonMain/kotlin/ai/koog/agents/mcp/McpTool.kt:64-69` forwarded `args.toKotlinxJsonObject()` to `mcpClient.callTool` without consulting the tool descriptor, so when an LLM emitted a nested `object`/`array` parameter as a stringified JSON value (the failure mode reported in #2017 against `obsidian-mcp`) the argument reached the server as a `JsonPrimitive(isString = true)` and was rejected by the server's schema validation.
- Add an internal helper `coerceArgsToDescriptorTypes(args, descriptor)` (new file `agents/agents-mcp/src/commonMain/kotlin/ai/koog/agents/mcp/McpArgCoercion.kt`) that walks each argument against its declared `ToolParameterType` and, only for `Object`, `List`, and `AnyOf` parameters, attempts `Json.parseToJsonElement` on string values, replacing the original only when the parsed shape is compatible with the declared type (or with the first compatible candidate for `AnyOf`). Parse failures and shape mismatches preserve the original value so the MCP server's schema validation remains authoritative. `McpTool.execute` calls the helper once on the forwarding path.
- Gating decision: primitive parameter types (`String`, `Integer`, `Float`, `Boolean`, `Enum`, `Null`) are never touched, so an intentional string payload like `"{not-json}"` against a `String` parameter cannot be mis-parsed. Recursion through `Object.properties`, `List.itemsType`, and `AnyOf` candidates is bounded by `MAX_COERCION_DEPTH = 8`; the AnyOf direct-match path shares the same depth budget so the guard fires uniformly. For `Object` parameters, declared `properties` match first; unmatched keys fall through to `additionalPropertiesType` only when `additionalProperties == true` and the additional type is non-null, otherwise they pass through unchanged. All four decisions are documented on the helper's KDoc.

Closes #2017 (runtime side; parser-side `oneOf` + `const` handling is tracked in a separate PR).

**Test plan**

- [x] `gradle :agents:agents-mcp:jvmTest` — adds `McpArgCoercionTest` (new file, 12 server-less unit cases covering happy path, gating, parse failure, shape mismatch, recursive Object, AnyOf positive/negative, `additionalProperties` true/false, and the depth-8 boundary) and one wiring case in `McpToolTest` that pushes a stringified JSON argument through `execute` and asserts the server-side fixture observed a real nested object. The depth-boundary case asserts `JsonObject` at depths 1..8 and a stringified `JsonPrimitive` at depth 9, so the cast chain pinpoints which depth a future regression breaks at.
- [x] `gradle :agents:agents-mcp:ktlintCheck`
- [x] `gradle :agents:agents-mcp:compileKotlinJvm --warning-mode=all` — zero warnings in `agents-mcp`.

**Notes**

- `TestMcpServer.kt` is shared by `McpToolTest` (SSE) and `StreamableHttpMcpToolTest` (Streamable HTTP). The new `echoObject` fixture tool is added once on the shared server, and both transports' existing assertions were updated to include it in the expected tool list. No additional wiring case is added against the Streamable HTTP transport: `McpTool.execute` is transport-neutral, so a duplicate wiring assertion against a second transport would cover the same call site without adding regression value. If a future change introduces transport-specific behaviour on the forwarding path, a transport-specific wiring case can be added at that point.

### DIFF
--- a/agents/agents-mcp/src/commonMain/kotlin/ai/koog/agents/mcp/McpArgCoercion.kt
+++ b/agents/agents-mcp/src/commonMain/kotlin/ai/koog/agents/mcp/McpArgCoercion.kt
@@ -1,0 +1,185 @@
+package ai.koog.agents.mcp
+
+import ai.koog.agents.core.tools.ToolDescriptor
+import ai.koog.agents.core.tools.ToolParameterDescriptor
+import ai.koog.agents.core.tools.ToolParameterType
+import kotlinx.serialization.json.Json
+import kotlinx.serialization.json.JsonArray
+import kotlinx.serialization.json.JsonElement
+import kotlinx.serialization.json.JsonObject
+import kotlinx.serialization.json.JsonPrimitive
+import kotlinx.serialization.json.buildJsonArray
+import kotlinx.serialization.json.buildJsonObject
+
+/**
+ * Maximum recursion depth applied when walking nested descriptor structures.
+ *
+ * Any input that would coerce deeper than this guard is left untouched so that adversarial or
+ * accidentally cyclic payloads cannot exhaust the stack.
+ */
+private const val MAX_COERCION_DEPTH: Int = 8
+
+/**
+ * Coerces stringified JSON arguments emitted by an LLM into the shapes declared by the tool's
+ * [ToolDescriptor] before they are forwarded to the MCP server.
+ *
+ * ### Gating
+ * Coercion is intentionally restricted to declared parameter types of
+ * [ToolParameterType.Object], [ToolParameterType.AnyOf] and [ToolParameterType.List].
+ * Primitive parameter types ([ToolParameterType.String], [ToolParameterType.Integer],
+ * [ToolParameterType.Float], [ToolParameterType.Boolean], [ToolParameterType.Enum],
+ * [ToolParameterType.Null]) are never touched: forwarding them unchanged preserves user-intended
+ * strings (e.g. `"{not-json}"`) and lets the MCP server's primitive validation remain authoritative.
+ *
+ * ### Behaviour
+ * For each argument whose declared parameter type is one of the gated types and whose value is a
+ * [JsonPrimitive] with `isString == true`, [Json.parseToJsonElement] is invoked on the string
+ * content. The parsed shape is then checked against the declared type:
+ *  - [ToolParameterType.Object] accepts only a parsed [JsonObject].
+ *  - [ToolParameterType.List] accepts only a parsed [JsonArray].
+ *  - [ToolParameterType.AnyOf] accepts the parsed value when the **first** candidate whose shape
+ *    is compatible with the parsed element is found. If no candidate matches, the value is left
+ *    untouched.
+ *
+ * On parse failure or shape mismatch the **original value is preserved** so the MCP server's
+ * schema validation remains the source of truth. The function never throws for malformed input.
+ *
+ * ### Recursion
+ * Object properties, list items and AnyOf candidates are walked recursively so that deeply nested
+ * stringified payloads (e.g. a stringified object containing another stringified object) are
+ * unwrapped consistently. Recursion is bounded by [MAX_COERCION_DEPTH]; values that would be
+ * coerced past the guard are preserved unchanged. AnyOf direct-match recursion shares the same
+ * depth budget as object/list recursion, so the guard fires uniformly regardless of how a value
+ * was routed.
+ *
+ * ### Additional properties
+ * For [ToolParameterType.Object] parameters, declared `properties` are matched first. Keys that
+ * are not declared are coerced using [ToolParameterType.Object.additionalPropertiesType] only when
+ * [ToolParameterType.Object.additionalProperties] is `true` **and** the additional type is
+ * non-null. Otherwise unmatched keys are passed through unchanged.
+ *
+ * @param args The raw arguments produced by the agent framework.
+ * @param descriptor The descriptor of the MCP tool whose parameter types guide coercion.
+ * @return A [JsonObject] whose values have been coerced where applicable; entries without a
+ *         matching parameter descriptor are preserved as-is.
+ */
+internal fun coerceArgsToDescriptorTypes(
+    args: JsonObject,
+    descriptor: ToolDescriptor,
+): JsonObject {
+    val paramsByName: Map<String, ToolParameterDescriptor> =
+        (descriptor.requiredParameters + descriptor.optionalParameters).associateBy { it.name }
+
+    return buildJsonObject {
+        for ((key, value) in args) {
+            val param = paramsByName[key]
+            if (param == null) {
+                put(key, value)
+            } else {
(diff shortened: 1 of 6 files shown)


zarechneva-a:
Hi @jewoodev, thank you for the contribution! Could you please address the failed checks? Thank you!

jewoodev:
@aozherelyeva thank you for the feedback. After the recent push the remaining `tests (ubuntu-latest)` failure appears unrelated to this PR's changes — it's a 60s coroutine timeout in `BedrockLLMClientTest` (`prompt-executor-bedrock-client`), while the same SHA's `tests (windows-latest)` and `tests (macos-latest)` passed. Likely a CI flake. If that is not the case, please let me know — I will gladly rebase, rerun, or proceed with the additional work :)

EugeneTheDev (changes requested):
Thanks for the `oneOf` fix, this looks reasonable. 

As for the tool call JSON coercion, I have some concerns. I took a look at the linked issue, and as far as I understand it the root cause here is LLM failing to generate a proper tool call JSON object. This is a more general problem, not related to the `McpTool` specifically. And this can indeed happen with some smaller models, e.g. local Ollama ones. We are aware of such limitations, that's why we provide [`ResponseProcessor`](https://api.koog.ai/prompt/prompt-processor/ai.koog.prompt.processor/-response-processor/index.html?query=abstract%20class%20ResponseProcessor). You can use existing implementation, e.g. [`ManualToolCallFixProcessor `](https://api.koog.ai/prompt/prompt-processor/ai.koog.prompt.processor/-manual-tool-call-fix-processor/index.html), or override it yourself to use your own logic. You can then set this response processor globally for the whole agent run using [`AIAgentConfig.responseProcessor`](https://api.koog.ai/agents/agents-core/ai.koog.agents.core.agent.config/-a-i-agent-config/index.html?query=expect%20class%20AIAgentConfig%20:%20AIAgentConfigBase#512589733%2FProperties%2F-1140963005) or set it more granularly for a specific `subgraph`/`subgraphWithTask` only.

jewoodev:
Thank you for the review and the pointer to `ResponseProcessor` / `ManualToolCallFixProcessor`. You are right that the underlying issue — LLMs emitting tool call arguments that need post-processing — is more general than `McpTool` and fits better in the `ResponseProcessor` extension point; putting that behavior in the MCP forwarding path would place it outside explicit `ResponseProcessor` configuration. I will close this PR.

One follow-up before I move on: would a generalized, MCP-agnostic `ResponseProcessor` for descriptor-driven coercion of stringified nested object/array/anyOf values be something you would consider useful in Koog, or would you prefer this to stay in application-specific `ResponseProcessor` overrides? If you prefer the application-level route, I'll leave it there and won't open a follow-up PR.

PR #2081, opened 2026-05-28
feat(prompt): explicit model resolution via DynamicPromptExecutor

Amaneusz:
Introduces model resolution as a preliminary step before prompt execution—allowing decorators (such as `ContextualPromptExecutor`) to be certain which `LLModel` will be used for the given operation.
Currently, dynamic model resolution happens internally (ie via fallback in `MultiLLMPromptExecutor`) which might result in misleading model information  (ie `ContextualPromptExecutor` triggers `onLLMCallCompleted` with `requestModel` instead of `effectiveModel`).

Adds `resolveModel` and execution paths accepting `ResolvedModel` to `PromptExecutorAPI`. All of these have default implementations that make existing executors work in the same way as before.

Adds `DynamicPromptExecutor`: a `PromptExecutor` subclass that exposes model resolution as an explicit step (`resolveModel`) before LLM dispatch. Subclasses translate a requested LLModel into a ResolvedModel (e.g. fallback selection); the LLModel entry points are finalized to flow through resolveModel exactly once per call.

`MultiLLMPromptExecutor` and `RoutingLLMPromptExecutor` move fallback logic into `resolveModel`. 

PromptExecutor stays open — existing direct subclasses are unaffected.


prompt/prompt-executor/prompt-executor-model/src/commonMain/kotlin/ai/koog/prompt/executor/model/ResolvedModel.kt
```diff
@@ -0,0 +1,21 @@
+package ai.koog.prompt.executor.model
+
+import ai.koog.prompt.llm.LLModel
+
+/**
+ * The model selected by [PromptExecutor.resolveModel] for execution
+ */
+public data class ResolvedModel(public val effectiveModel: LLModel)
```
Amaneusz, line 11: I originally wanted to use `@JvmInline` + `value class` for it but this would mess up method declarations from `PromptExecutorAPI` as old methods expect `LLModel` and this would inine to `LLModel` as well.

prompt/prompt-executor/prompt-executor-model/src/commonMain/kotlin/ai/koog/prompt/executor/model/PromptExecutorAPI.kt
```diff
@@ -78,6 +79,79 @@ public interface PromptExecutorAPI : AutoCloseable {
...
+    /**
+     * Executes a given prompt against an already-resolved model and returns the assistant response.
+     *
+     * @param prompt The `Prompt` containing the messages to be used in the execution.
+     * @param model The [ResolvedModel] returned by [resolveModel].
+     * @param tools An optional list of `ToolDescriptor` objects available during the execution.
+     * @return A [Message.Assistant] containing the response generated by the LLM.
+     */
+    @JvmSynthetic
+    public suspend fun execute(
```
EugeneTheDev, line 108: Since `ResolvedModel` is essentially `LLModel` (it contains only a single property `effectiveModel: LLModel`), why do we need a set of additional overloads for methods like `execute`, `executeStreaming` etc.? I understand that you want to provide additional step on top of existing execute logic, but do we really need an additional set of methods for that? Can't it be an implementation detail of the existing execute methods? I.e., now each one has additional step first - use `resolveModel` to resolve an effective model and then the rest of the existing logic.

EugeneTheDev (changes requested):

prompt/prompt-executor/prompt-executor-model/src/commonMain/kotlin/ai/koog/prompt/executor/model/PromptExecutorAPI.kt
EugeneTheDev: Discussed and resolved the question during the call

EugeneTheDev (approved):

prompt/prompt-executor/prompt-executor-model/src/commonMain/kotlin/ai/koog/prompt/executor/model/PromptExecutorAPI.kt
```diff
@@ -78,6 +79,79 @@ public interface PromptExecutorAPI : AutoCloseable {
...
+     * The default implementation returns the requested model unchanged.
+     *
+     * @param requestedModel The requested [LLModel] to resolve.
+     * @param promptExecutorOperation The operation the resolved model will be used for.
+     * @return A [ResolvedModel] wrapping the effective model to use.
+     * @throws ModelResolutionException If the requested model cannot be resolved.
+     */
+    @JvmSynthetic
+    public suspend fun resolveModel(
+        requestedModel: LLModel,
```
sdubov: Up to your decision: maybe just `model` ?

prompt/prompt-executor/prompt-executor-model/src/commonMain/kotlin/ai/koog/prompt/executor/model/ResolvedModel.kt
```diff
@@ -0,0 +1,21 @@
...
+
+public enum class PromptExecutorOperation {
```
sdubov: Up to your decision. For my taste, I would split this to different files. `PromptExecutorOperation` is not related directly to the `ResolvedModel` data class. So, not sure why do they live together.

prompt/prompt-executor/prompt-executor-model/src/commonMain/kotlin/ai/koog/prompt/executor/model/ResolvedModel.kt
```diff
@@ -0,0 +1,21 @@
...
+    Execute,
+    Moderate,
+    MultipleChoices,
+    Streaming
+}
+
+public class ModelResolutionException(
```
sdubov: KDoc is required for public methods. Up to your decision, I would also move this into a separate file

prompt/prompt-executor/prompt-executor-model/src/commonMain/kotlin/ai/koog/prompt/executor/model/ResolvedModel.kt
sdubov: KDoc is required for public classes.

sdubov (approved):
All looks good for me! As we discussed, great work! Several minor comments up to your decision.

prompt/prompt-executor/prompt-executor-model/src/commonMain/kotlin/ai/koog/prompt/executor/model/PromptExecutorAPI.kt
Amaneusz: renamed

prompt/prompt-executor/prompt-executor-model/src/commonMain/kotlin/ai/koog/prompt/executor/model/ResolvedModel.kt
Amaneusz: moved into separate file

prompt/prompt-executor/prompt-executor-model/src/commonMain/kotlin/ai/koog/prompt/executor/model/ResolvedModel.kt
Amaneusz: done

prompt/prompt-executor/prompt-executor-model/src/commonMain/kotlin/ai/koog/prompt/executor/model/ResolvedModel.kt
Amaneusz: done

Issue #2089, opened 2026-05-29
Bedrock Nova: empty "system" array rejected when prompt has no system message


ashakirin:
####   Summary
  
After the Message architecture refactor (#1919, commit e4f91bfb), BedrockAmazonNovaSerialization.createNovaRequest always assigns its systemMessages mutable list to NovaRequest.system, even when empty. Empty-list serialization produces "system": [] on the wire, which Amazon Nova rejects:
  ```
ai.koog.prompt.executor.clients.LLMClientException: Error from client: BedrockLLMClient Malformed input request: #/system: expected minimum item count: 1, found: 0, please reformat your input and try again.
```
The previous implementation guarded against this with .takeIf { it.isNotEmpty() }, which let Json { explicitNulls = false } omit the field entirely. That guard was lost in the refactor.
  
####   Affected
-  All Amazon Nova models on Bedrock when invoked through BedrockLLMClient with a prompt that contains no Message.System entry.
-  Reproduces consistently with LongTermMemory/AgentcorePromptAugmenter retrieval calls that occasionally produce a system-less prompt (e.g. SUMMARY scenario, EPISODIC retrieval), and any direct caller that omits a system prompt.
-  Also reproduces in CI: integration tests AgentCoreLTMIntegrationTest#integration_agentAnswersFromLtmAfterAsyncExtraction and integration_agentAnswersFromLtmEpisodicStrategySlow fail on develop against a real AgentCore memory.

  
####   Steps to reproduce
  1. Configure BedrockLLMClient with BedrockModels.AmazonNovaMicro (any Nova model).
  2. Build a Prompt with only User/Assistant messages — no System.
  3. Call executor.execute(prompt, model).
  
  Expected: a normal Bedrock response.
  Actual: LLMClientException with Malformed input request: #/system: expected minimum item count: 1, found: 0.
  
###   Root cause
  In prompt/prompt-executor/prompt-executor-clients/prompt-executor-bedrock-client/src/jvmMain/kotlin/ai/koog/prompt/executor/clients/bedrock/modelfamilies/amazon/BedrockAmazonNovaSerialization.kt (around line 89):
  
```
  val systemMessages = mutableListOf<NovaSystemMessage>()
  // ... appended only when Message.System parts exist
  return NovaRequest(
      ...,
      system = systemMessages,    // ← always passed; empty list serializes as []
      ...
  )
```
  
[...]

PR #2090, opened 2026-05-29
fix: #2089 added check guard for Bedrcok Nova empty system prompt array

ashakirin:
Fix for #2089 added check guard for Bedrcok Nova empty system prompt array


zarechneva-a (approved):
Thank you!

Issue #920, opened 2025-10-03
Add Amazon Bedrock AgentCore integration for enterprise-grade agent deployment


apncodes:
### Problem
Koog provides excellent local agent development with Kotlin-native patterns. To complement this strength, developers need seamless integration with enterprise-grade production infrastructure including managed runtime, persistent memory, secure sandboxes, and scalable deployment capabilities.

### Proposed Solution
Integrate Amazon Bedrock AgentCore primitives as native Koog components, enabling seamless transition from local development to enterprise production deployment.

#### 1. AgentCore Runtime Deployment
kotlin
// Deploy Koog agents to serverless AgentCore Runtime
val agent = AIAgent(strategy, toolRegistry) {
   install(AgentCoreRuntime) {
       region = "us-east-1"
       runtimeConfig = RuntimeConfig.serverless()
   }
}
agent.deployToAgentCore()

#### 2. AgentCore Memory as AIAgentFeature
kotlin
// Managed, persistent memory across sessions
install(AgentCoreMemory) {
   memoryId = "user-session-${userId}"
   shortTermMemory = true
   longTermMemory = true
   shareAcrossAgents = false
}

#### 3. AgentCore Tools Integration
kotlin
ToolRegistry {
   // Secure code execution in isolated sandboxes
   tool(AgentCoreCodeInterpreter {
       allowedPackages = listOf("pandas", "numpy")
       timeoutSeconds = 30
   })
   
   // Cloud-based browser automation
   tool(AgentCoreBrowser {
       headless = true
       viewport = Viewport(1920, 1080)
   })
   
   // MCP Gateway for API/Lambda integration
   tool(AgentCoreGateway {
       mcpServers = listOf("internal-apis", "external-tools")
   })
}

#### 4. AgentCore Identity & Observability
kotlin
install(AgentCoreIdentity) {
   identityProvider = "cognito"
   tokenVault = true
}

install(AgentCoreObservability) {
   tracing = true
   metrics = true
   dashboardUrl = "https://console.aws.amazon.com/bedrock/agentcore"
}

### Value Proposition

**For Kotlin/Java Developers:**
- Native AgentCore integration without Python dependencies
- Type-safe configuration with Kotlin DSL
- Seamless local-to-production workflow

**For Enterprise Users:**
- Managed infrastructure with auto-scaling
- Enterprise security and compliance
- Production-grade observability and monitoring

**For Koog Ecosystem:**
- Bridge between open-source flexibility and enterprise reliability
[...]

PR #2092, opened 2026-05-29
feat: #920 Added auto-discovery strategies test

ashakirin:
feat: #920 Refactored AgentCore long-term memory integration tests and added auto-discovery strategies test


serge-p7v (approved):
LGTM

PR #2098, opened 2026-06-01
fix: add a `toolChoice` parameter into subgraphWithTask

### LINKED ISSUE #2042: `subgraphWithTask` hardcodes `tool_choice: required`, incompatible with some models
## Problem

`subgraphWithTask` unconditionally calls `setToolChoiceRequired()`, which makes the client always send `"tool_choice": "required"`. Some models (e.g., DeepSeek V4 Flash) do not support `tool_choice: required` and reject the request with a 400 error. They do support `tool_choice: auto` and call tools correctly.

```bash
# tool_choice: required → REJECTED
curl -s -X POST $BASE_URL/v1/chat/completions \
  -H "Content-Type: application/json" \
  -d '{
    "model": "deepseek-v4-flash",
    "messages": [{"role": "user", "content": "Calculate 1+1 using the calc tool"}],
    "tools": [{"type": "function", "function": {"name": "calc", "description": "calculate", "parameters": {"type": "object", "properties": {"expr": {"type": "string"}}}}],
    "tool_choice": "required",
    "max_tokens": 200
  }'
# → {"error":{"message":"deepseek-reasoner does not support this tool_choice"}}

# tool_choice: auto → WORKS
curl -s -X POST $BASE_URL/v1/chat/completions \
  -H "Content-Type: application/json" \
  -d '{
    "model": "deepseek-v4-flash",
    "messages": [{"role": "user", "content": "Calculate 1+1 using the calc tool"}],
    "tools": [{"type": "function", "function": {"name": "calc", "description": "calculate", "parameters": {"type": "object", "properties": {"expr": {"type": "string"}}}}],
    "tool_choice": "auto",
    "max_tokens": 200
  }'
# → {"choices":[{"message":{"tool_calls":[{"function":{"name":"calc","arguments":"{\"expr\":\"1+1\"}"}}],"finish_reason":"tool_calls"}}]}
```

```kotlin
// AIAgentSubgraphExt.kt — setupSubgraphWithTask
val setupTask by node<Input, String>(...) { input ->
    llm.writeSession {
        // ...
        setToolChoiceRequired()  // ← unconditional
    }
    defineTask(input)
}

// AIAgentLLMWriteSessionCommon.kt
public fun setToolChoiceRequired() {
    setToolChoice(LLMParams.ToolChoice.Required)  // ← no capability check
}

// AbstractOpenAILLMClient.kt — no capability check before sending
toolChoice = prompt.params.toolChoice?.toOpenAIToolChoice()
```

Compare with `createResponseFormat` which **does** check capabilities:
```kotlin
// DeepSeekLLMClient.kt
override fun createResponseFormat(schema: LLMParams.Schema?, model: LLModel) {
    return schema?.let {
        require(model.supports(it.capability)) {  // ← checks capability
            "Model ${model.id} does not support schema ${it.name}"
        }
        ...
    }
}
```

## Root Cause

1. **`setToolChoiceRequired()` ignores capabilities** — no check before setting the param
2. **Client ignores capabilities** — `AbstractOpenAILLMClient` always serializes `toolChoice` without checking `LLMCapability.ToolChoice`, unlike `createResponseFormat` which checks `Schema` capability
3. **`LLMCapability.ToolChoice` is too coarse** — cannot express "supports `auto` but not `required`"
4. **No configuration** — `subgraphWithTask` has no parameter to choose a different `ToolChoice` mode

## Suggested Fix

Add a `toolChoice` parameter to `subgraphWithTask` / `setupSubgraphWithTask`:

```kotlin
fun <Input, Output> subgraphWithTask(
    tools: List<ToolBase<*, *>>,
    toolChoice: LLMParams.ToolChoice = ToolChoice.Required,  // default unchanged
    ...
)
```

Inside `setupSubgraphWithTask`, replace `setToolChoiceRequired()` with `setToolChoice(toolChoice)`.

zarechneva-a:
Add a configurable `toolChoice` parameter into the relevant `subgraphWithTask` / `setupSubgraphWithTask` overloads.

closes #2042


antoniibelyshev:
I think we should not add `toolChoice` parameter to the `subgraphWithTask`: it seems like a hack and it faces public api, which does not seem nice. It may confuse users who did not need it but will try setting it once it's in our public api, leading to possible errors. I would vote for updating how our llm clients deal with the `ToolChoice` capability, and can work on it. @tiginamaria , @sdubov , @aozherelyeva WDYT?

zarechneva-a:
@antoniibelyshev sure. My fix is just a patch, but I think the idea you're suggesting is more viable. Thanks!

tiginamaria:
Agree with Anton. I think the right approach here is to set tool choice in the prompt settings, so the subgraph with the task will apply it inside (I hope it works this way, but I need to double-check)
So it should look like this:
```
val agent = AIAgent(
    promptExecutor = simpleOpenAIExecutor(System.getenv("YOUR_API_KEY")),
    agentConfig = AIAgentConfig(
        prompt = prompt(
            "my-prompt", params = LLMParams(
                toolChoice = LLMParams.ToolChoice.Auto,
                temperature = 0.7,
            )
        ) {
            system("You are an expert in internet memes. Be helpful, friendly, and answer user questions concisely, showing your knowledge of memes.")
        },
        model = OpenAIModels.Chat.GPT4o,
        maxAgentIterations = 10
    ),
    toolRegistry = ToolRegistry {
        tool(::askUser)
    }) {
    handleEvents {
        // Handle tool calls
        onToolCallStarting { eventContext ->
            println("Tool called: ${eventContext.toolName} with args ${eventContext.toolArgs}")
        }
    }
}
```

tiginamaria:
Oh, sorry, the issue is slightly different here. I think we should just modify the subgraph with task in a way that it checks that the model supports (or not) too choise = reqired.

zarechneva-a:
@tiginamaria @antoniibelyshev thanks for the comments! Ok, closing this one then.

PR #2106, opened 2026-06-02
example: update dependencies and Koog to 1.3.0

siarhei-luskanau:


siarhei-luskanau:
Updating the ktor from "3.4.2" to "3.5.0" fails Web target

PR #2124, opened 2026-06-08
feat(embeddings): support batch embedding of multiple inputs

inemtsev:
Hi,
Picking up the **batch embedding** half of this — the follow-up @mltheuser scoped in Building on his provider analysis and design groundwork there.
The scope here is batch only; dimensions (KG-104) stays separate, per the earlier review feedback on #1296.

**Approach**:
  - New high-level `Embedder.embed(texts: List<String>): List<Vector>` with a
    sequential default, so existing `Embedder` implementations keep working.
  - `LLMEmbedder` delegates to the provider client's `embed(inputs)`. Providers
    with a native batch endpoint (OpenAI, Google, Mistral, Ollama) do it in one
    request; providers without one propagate their own `UnsupportedOperationException`
    with a clear, provider-specific message.
  - `OpenAILLMClient.embed(inputs)` now performs a real batched request (it
    previously threw); input order is preserved via the response `index`.
  - `TextDocumentEmbedder` forwards the batch to its wrapped embedder so batching
    isn't silently downgraded to per-item calls.

closes https://youtrack.jetbrains.com/issue/KG-538/Support-embedding-a-list-of-input


inemtsev:
@aozherelyeva Can you please take a look when you have a moment

PR #2130, opened 2026-06-10
feat: Add support for Anthropic Claude Fable 5 model

zarechneva-a:
Add support for the newly released Anthropic Claude Fable 5 model.


antoniibelyshev (approved):
Thank you!

Issue #2111, opened 2026-06-04
[DeepSeek] Streaming tool call tool_call_id mismatch causes 400 error


HaishiR:
## Problem Description

  When using DeepSeek model with streaming agent strategy, sending tool result to LLM returns 400 error.

  ## Error Message

  "Messages with role 'tool' must be a response to a preceding message with 'tool_calls'"

  ## Environment

  - Koog version: latest
  - Model: DeepSeek V4 Pro
  - JDK version: 17+

  ## Analysis

  The issue is in `AbstractOpenAILLMClient.kt:293`:

  ```kotlin
  toolCallId = part.id ?: Uuid.random().toString()

  When MessagePart.Tool.Call.id is null (during streaming start), the framework generates a random UUID. But DeepSeek
  API requires the tool result's tool_call_id to match the id from previously sent tool_calls.

  In streaming mode, tool call id may be null initially or only be complete at ToolCallComplete, causing id mismatch.

  Expected Behavior

  Streaming tool calls should correctly maintain id mapping to ensure tool results are associated with the correct tool
  call.

  Additional Context

  - Non-streaming (nodeLLMRequest) works correctly with DeepSeek
  - Only streaming mode (nodeLLMRequestStreaming) has this issue

jewoodev:
Hi, I looked into this and opened a focused draft PR in #2134.

The failure path seems to be one step earlier than the UUID fallback in `AbstractOpenAILLMClient`: `DeepSeekLLMClient.processStreamingResponse(...)` forwards streaming tool-call chunks to `StreamFrameFlowBuilder`, and the builder can currently complete the pending call with `id = null` when the first chunk has no id and a later same-index chunk carries the provider id. That then makes the replay fallback generate a random `tool_call_id`.

The PR keeps the fix in `StreamFrameFlowBuilder`: a late id on the same `index` fills the pending tool call, while repeated ids and distinct indexed tool calls keep the existing behavior. It also adds a regression test in `StreamFrameFlowBuilderTest`.

Please let me know if you prefer this to be handled at a different layer.

PR #2134, opened 2026-06-13
fix(prompt): preserve late streaming tool call ids

jewoodev:
## Summary
- preserve a late-arriving tool call id when it belongs to the current same-index streaming tool call
- add a regression test for null-id first chunks followed by provider ids

Refs #2111


jewoodev:
CI note: I looked into the failing `tests (macos-latest)` check.

The failing test is `OpenTelemetryInferenceSpanTest.expected spans on llm call failed()` in `:agents:agents-features:agents-features-opentelemetry:jvmTest`, timing out while waiting for exported spans. This PR only changes `prompt-model` streaming frame assembly and its regression test, and the same run passed compilation plus the Ubuntu/Windows test jobs.

I also ran the failing test locally on Linux and it passed:
`gradle :agents:agents-features:agents-features-opentelemetry:jvmTest --tests "ai.koog.agents.features.opentelemetry.feature.span.OpenTelemetryInferenceSpanTest.expected spans on llm call failed" --console=plain`

This looks unrelated to the PR change, so I am not planning a code change for it unless you prefer otherwise.

PR #2135, opened 2026-06-14
fix(agents): prioritize tool calls in single-run mixed responses

### LINKED ISSUE #2127: Possible regression: `singleRunStrategy` terminates after first tool result when assistant message contains both text and tool calls
When running an `AIAgent` with the default `singleRunStrategy`, if the LLM's response to a tool result is a `Message.Assistant` whose `parts` contain both a `MessagePart.Text` and a `MessagePart.Tool.Call`, the agent terminates and returns the text part as its final output. Any expected subsequent tool call never fire.

This appears to have been introduced somewhere between `0.7.3` → `1.0.0`. I was using `0.7.3` for a project and it worked correctly then, but now after upgrading to `1.0.0` the same code doesn't seem to be working (as in, the agent doesn't go beyond the first tool call).

Not sure whether this is intentional or an actual bug worth fixing, so I'd appreciate your take.

### Environment

- Koog `1.0.0`
- Strategy: the default `singleRunStrategy` used in `AIAgent.builder().build()`

### How to reproduce?

Drop the following snippet in `examples/simple-examples/src/main/kotlin/ai/koog/agents/example/stratbug/StratBugRepro.kt`:

```kotlin
package ai.koog.agents.example.stratbug

import ai.koog.agents.core.agent.AIAgent
import ai.koog.agents.core.tools.ToolDescriptor
import ai.koog.agents.core.tools.ToolRegistry
import ai.koog.agents.core.tools.annotations.LLMDescription
import ai.koog.agents.core.tools.annotations.Tool
import ai.koog.agents.core.tools.reflect.ToolSet
import ai.koog.prompt.Prompt
import ai.koog.prompt.dsl.ModerationResult
import ai.koog.prompt.executor.model.PromptExecutor
import ai.koog.prompt.llm.LLMCapability
import ai.koog.prompt.llm.LLMProvider
import ai.koog.prompt.llm.LLModel
import ai.koog.prompt.message.Message
import ai.koog.prompt.message.MessagePart
import ai.koog.prompt.message.ResponseMetaInfo
import ai.koog.prompt.streaming.StreamFrame
import kotlinx.coroutines.flow.Flow
import kotlinx.coroutines.flow.emptyFlow
import kotlinx.coroutines.runBlocking

class TwoStepTools : ToolSet {
    @Tool @LLMDescription("Step 1") fun step1(): String = "step1 done"
    @Tool @LLMDescription("Step 2") fun step2(): String = "step2 done"
}

// Three messages:
//   1. text + tool_use(step1)  <- first turn
//   2. text + tool_use(step2)  <- reply to step1's result (mixed content)
//   3. text only               <- final answer
class CannedExecutor : PromptExecutor() {
    private val meta = ResponseMetaInfo.Empty
    private val responses = listOf(
        Message.Assistant(parts = listOf(
            MessagePart.Text("I'll start with step1."),
            MessagePart.Tool.Call(id = "1", tool = "step1", args = "{}"),
        ), metaInfo = meta),
        Message.Assistant(parts = listOf(
            MessagePart.Text("Now I'll do step2."),
            MessagePart.Tool.Call(id = "2", tool = "step2", args = "{}"),
        ), metaInfo = meta),
        Message.Assistant(parts = listOf(
            MessagePart.Text("Both steps done."),
        ), metaInfo = meta),
    )
    private var i = 0
    override suspend fun execute(prompt: Prompt, model: LLModel, tools: List<ToolDescriptor>) = responses[i++]
    override fun executeStreaming(prompt: Prompt, model: LLModel, tools: List<ToolDescriptor>): Flow<StreamFrame> = emptyFlow()
    override suspend fun moderate(prompt: Prompt, model: LLModel) = ModerationResult(false, emptyMap())
    override fun close() {}
}

fun main() = runBlocking {
    val agent = AIAgent.builder()
        .promptExecutor(CannedExecutor())
        .llmModel(LLModel(LLMProvider.OpenAI, "irrelevant", listOf(LLMCapability.Tools, LLMCapability.Completion)))
        .toolRegistry(ToolRegistry.builder().tools(TwoStepTools().asTools()).build())
        .build()

    println("Final: ${agent.run("go")}")
    // Expected: "Both steps done." (3 LLM calls, 2 tool executions)
    // Actual:   "Now I'll do step2." <— agent terminates after step1's result; step2 never runs, the third response is never reached.
}
```

Register the example:

```bash
echo 'registerRunExampleTask("runStratBugRepro", "ai.koog.agents.example.stratbug.StratBugReproKt")' >> examples/simple-examples/build.gradle.kts
```

Compile:

```bash
cd examples/simple-examples && ./gradlew compileKotlin
```

And run:

```bash
./gradlew runStratBugRepro
```

You'll see:

```
[main] INFO  a.k.agents.core.agent.GraphAIAgent - (agent id: xyz) Executing tool (name: step1, args: {}
Final: Now I'll do step2.
```

Which is wrong, as the agent should run both tools and get to the actual final message (`"Both steps done." `).

### What I think is happening

Looking at `AIAgentSimpleStrategies.kt`:

https://github.com/JetBrains/koog/blob/58a8168fac22234b95cfaaa1969df3ac99098707/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgentSimpleStrategies.kt#L29-L40

Edge resolution in `AIAgentNode.resolveEdge` iterates edges in declaration order and returns the first one whose forward output is non-empty:

https://github.com/JetBrains/koog/blob/58a8168fac22234b95cfaaa1969df3ac99098707/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/entity/AIAgentNode.kt#L87-L100

`onTextMessage { true }` matches if the assistant message has any `MessagePart.Text` part

https://github.com/JetBrains/koog/blob/58a8168fac22234b95cfaaa1969df3ac99098707/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/extension/AIAgentEdges.kt#L54-L61

`onToolCalls { true }` matches if it has any `MessagePart.Tool.Call` part:

https://github.com/JetBrains/koog/blob/58a8168fac22234b95cfaaa1969df3ac99098707/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/dsl/extension/AIAgentEdges.kt#L71-L80

When the assistant message has both, both edges match, and declaration order decides which one fires.

If I'm reading this right, the `nodeSendToolResult` edges are declared in the reverse order vs the `nodeCallLLM` edges because`nodeCallLLM` has `onToolCalls` first ([`AIAgentSimpleStrategies.kt:35`](https://github.com/JetBrains/koog/blob/58a8168fac22234b95cfaaa1969df3ac99098707/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgentSimpleStrategies.kt#L35)), but `nodeSendToolResult` has `onTextMessage` first ([`AIAgentSimpleStrategies.kt:38`](https://github.com/JetBrains/koog/blob/58a8168fac22234b95cfaaa1969df3ac99098707/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgentSimpleStrategies.kt#L38)). That would make the second LLM call (after the first tool result) route to `nodeFinish` whenever the model includes any text alongside its `tool_use` (which is usually the response shape from Anthropic models, but it's obviously not limited to a single provider). This asymmetry is what the repro above attempts to demosntrate.

I also noticed the `Calculator.java` example declares the equivalent edges with `onToolCalls` first on both nodes:

https://github.com/JetBrains/koog/blob/58a8168fac22234b95cfaaa1969df3ac99098707/examples/simple-examples-java/src/main/java/ai/koog/agents/example/calculator/Calculator.java#L137-L149

I guess my question is: is the `singleRunStrategy` declaration order intentional and I'm using the wrong abstraction, or is the Calculator example closer to the intended order?

### Workaround I'm using locally

Passing a custom strategy via `AIAgent.builder().graphStrategy(...)`, identical to `singleRunStrategy` except `onToolCalls` is declared before `onTextMessage` on `nodeSendToolResult`. After this change the repro completes correctly.

Happy to provide more detail or open a PR if useful.

jewoodev:
## Summary
- Prioritize tool-call edges over text/final edges in single-run post-tool responses.
- Apply the same fix to the single-run history-compression strategy for both post-tool and post-compression follow-up responses.
- Keep the change limited to the single-run strategy family; global edge resolution and chatAgentStrategy are unchanged.

Fixes #2127

### DIFF
--- a/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgentSimpleStrategies.kt
+++ b/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/core/agent/AIAgentSimpleStrategies.kt
@@ -35,6 +35,6 @@ public fun singleRunStrategy(parallelTools: Boolean = false): AIAgentGraphStrate
     edge(nodeCallLLM forwardTo nodeExecuteTool onToolCalls { true })
     edge(nodeCallLLM forwardTo nodeFinish onTextMessage { true })
     edge(nodeExecuteTool forwardTo nodeSendToolResult)
-    edge(nodeSendToolResult forwardTo nodeFinish onTextMessage { true })
     edge(nodeSendToolResult forwardTo nodeExecuteTool onToolCalls { true })
+    edge(nodeSendToolResult forwardTo nodeFinish onTextMessage { true })
 }
--- a/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/ext/agent/SingleRunStrategyWithHistoryCompression.kt
+++ b/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/ext/agent/SingleRunStrategyWithHistoryCompression.kt
@@ -71,17 +71,17 @@ public fun singleRunStrategyWithHistoryCompression(
     edge(nodeExecuteTool forwardTo nodeSendToolResult onCondition { llm.readSession { !config.isHistoryTooBig(prompt) } })
     edge(nodeCompressHistory forwardTo nodeSendCompressedHistory)
 
+    edge(nodeSendToolResult forwardTo nodeExecuteTool onToolCalls { true })
+
     edge(
         nodeSendToolResult forwardTo nodeFinish
             onTextMessage { true }
     )
 
-    edge(nodeSendToolResult forwardTo nodeExecuteTool onToolCalls { true })
+    edge(nodeSendCompressedHistory forwardTo nodeExecuteTool onToolCalls { true })
 
     edge(
         nodeSendCompressedHistory forwardTo nodeFinish
             onTextMessage { true }
     )
-
-    edge(nodeSendCompressedHistory forwardTo nodeExecuteTool onToolCalls { true })
 }
--- a/agents/agents-core/src/commonTest/kotlin/ai/koog/agents/core/agent/SingleRunStrategyTests.kt
+++ b/agents/agents-core/src/commonTest/kotlin/ai/koog/agents/core/agent/SingleRunStrategyTests.kt
@@ -291,4 +291,52 @@ class SingleRunStrategyTests {
         assertEquals(3, actualToolCalls.size)
         assertEquals(assistantResponse, result)
     }
+
+    @Test
+    fun test_SingleRunStrategy_PostToolMixedResponseExecutesNextToolCall() = runTest {
+        val actualToolCalls = mutableListOf<String>()
+
+        val testToolRegistry = ToolRegistry {
+            tool(CreateTool)
+        }
+
+        val intermediateResponse = "Need another tool."
+        val finalResponse = "Task solved after second tool."
+        var postFirstToolResultHandled = false
+
+        val mockLLMApi = getMockExecutor(serializer) {
+            mockLLMToolCall(CreateTool, CreateTool.Args("first")) onRequestEquals "Solve task"
+            mockLLMMixedResponse(
+                toolCalls = listOf(CreateTool to CreateTool.Args("second")),
+                responses = listOf(intermediateResponse)
+            ) onCondition { request ->
+                if (request == "created" && !postFirstToolResultHandled) {
+                    postFirstToolResultHandled = true
+                    true
+                } else {
+                    false
+                }
+            }
+            mockLLMAnswer(finalResponse) onCondition { request ->
+                request == "created" && postFirstToolResultHandled
+            }
+            mockLLMAnswer("I don't know how to answer that.").asDefaultResponse
+        }
+
+        val agent = AIAgent(
+            mockLLMApi,
+            OllamaModels.Meta.LLAMA_3_2,
+            strategy = singleRunStrategy(),
+            toolRegistry = testToolRegistry
+        ) {
+            install(EventHandler) {
+                onToolCallStarting { eventContext -> actualToolCalls += eventContext.toolArgs.toString() }
+            }
+        }
+
+        val result = agent.run("Solve task", null)
+
+        assertEquals(listOf("""{"name":"first"}""", """{"name":"second"}"""), actualToolCalls)
+        assertEquals(finalResponse, result)
+    }
 }
--- a/agents/agents-ext/src/commonTest/kotlin/ai/koog/agents/ext/agent/SingleRunStrategyWithHistoryCompressionTests.kt
+++ b/agents/agents-ext/src/commonTest/kotlin/ai/koog/agents/ext/agent/SingleRunStrategyWithHistoryCompressionTests.kt
@@ -5,6 +5,7 @@ import ai.koog.agents.core.dsl.extension.HistoryCompressionStrategy
 import ai.koog.agents.core.tools.SimpleTool
 import ai.koog.agents.core.tools.ToolRegistry
 import ai.koog.agents.core.tools.annotations.LLMDescription
+import ai.koog.agents.features.eventHandler.feature.EventHandler
 import ai.koog.agents.testing.tools.getMockExecutor
 import ai.koog.prompt.executor.ollama.client.OllamaModels
 import ai.koog.serialization.kotlinx.KotlinxSerializer
@@ -85,6 +86,54 @@ class SingleRunStrategyWithHistoryCompressionTests {
     }
 
     @Test
+    fun test_no_compression_post_tool_mixed_response_executes_next_tool_call() = runTest {
+        val actualToolCalls = mutableListOf<String>()
+        val config = HistoryCompressionConfig(
+            isHistoryTooBig = { it.messages.size > 100 },
+            compressionStrategy = HistoryCompressionStrategy.WholeHistory
+        )
+
+        val intermediateResponse = "Need another tool."
+        val finalResponse = "Task solved after second tool."
+        var postFirstToolResultHandled = false
+
+        val mockLLMApi = getMockExecutor(serializer) {
+            mockLLMToolCall(CreateTool, CreateTool.Args("first")) onRequestEquals "Solve task"
+            mockLLMMixedResponse(
+                toolCalls = listOf(CreateTool to CreateTool.Args("second")),
+                responses = listOf(intermediateResponse)
+            ) onCondition { request ->
+                if (request == "created" && !postFirstToolResultHandled) {
+                    postFirstToolResultHandled = true
+                    true
+                } else {
+                    false
+                }
+            }
+            mockLLMAnswer(finalResponse) onCondition { request ->
+                request == "created" && postFirstToolResultHandled
+            }
+            mockLLMAnswer("I don't know how to answer that.").asDefaultResponse
+        }
+
+        val agent = AIAgent(
+            mockLLMApi,
+            OllamaModels.Meta.LLAMA_3_2,
+            strategy = singleRunStrategyWithHistoryCompression(config),
+            toolRegistry = ToolRegistry { tool(CreateTool) }
+        ) {
+            install(EventHandler) {
+                onToolCallStarting { eventContext -> actualToolCalls += eventContext.toolArgs.toString() }
+            }
+        }
+
+        val result = agent.run("Solve task", null)
+
+        assertEquals(listOf("""{"name":"first"}""", """{"name":"second"}"""), actualToolCalls)
+        assertEquals(finalResponse, result)
+    }
+
+    @Test
     fun test_sequential_mode() = runTest {
         var compressionRequested = false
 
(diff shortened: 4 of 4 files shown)


jewoodev:
During the edge-order sweep for this PR, I also found two similar candidates in `chatAgentStrategy` (`agents/agents-core/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentStrategies.kt`) that I intentionally kept out of the initial patch:

- `giveFeedbackToCallTools`: if the model returns both text and tool calls after the feedback prompt, the current text edge runs before the tool-call edge, so the strategy asks for tool calls again instead of executing them.
- `nodeSendToolResult`: if the post-tool response contains both text and tool calls, the current text edge wins before both the `__exit__` tool-call edge and the regular tool-call edge.

Could you confirm the intended policy for `chatAgentStrategy` mixed responses? Should tool calls, including `__exit__`, take priority over text there as they do in the single-run strategies, or should the current text-first behavior stay because chat finalization has different semantics?

Once that behavior is confirmed, I can add the corresponding changes and regression tests to this PR.

monrax:
FWIW it does seem like the answer is tools-first in all these cases, including the chat-related ones:

- `singleRunStrategy` in `AIAgentSimpleStrategies.kt`: fixed in this PR
- `singleRunStrategyWithHistoryCompression` in `SingleRunStrategyWithHistoryCompression.kt`: fixed in this PR
- `reActStrategy` in `AIAgentStrategies.kt`: already declares tool-execution node before finish node

https://github.com/JetBrains/koog/blob/e4e9e4079fa69696c5662b5637467f253d1b52f2/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentStrategies.kt#L174-L175

- `nodeDecide` in `AIAgentSubgraphExt.kt`: already declares tool-execution node before text-handling node

https://github.com/JetBrains/koog/blob/e4e9e4079fa69696c5662b5637467f253d1b52f2/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentSubgraphExt.kt#L748-L750

 - `chatAgentStrategy` in `AIAgentStrategies.kt`: **lines 57 and 58 should be swapped** and **line 62 should be moved below line 67**

https://github.com/JetBrains/koog/blob/e4e9e4079fa69696c5662b5637467f253d1b52f2/agents/agents-core/src/commonMain/kotlin/ai/koog/agents/ext/agent/AIAgentStrategies.kt#L54-L67

---

A quick timeline:

1. Dec 1 2025 — PR #1166 (f247a0fc) explicitly swapped these two edges in the then-existing `singleRunWithParallelAbility` putting the tool-execution node before finish (diff [here](https://github.com/JetBrains/koog/pull/1166/changes#diff-e2e6aadb840d4100d41d8ebfa328ef4d2b189c78b7c43aa97c69a60f5d5df2fb)). It moved `nodeSendToolResult forwardTo nodeExecuteTool onMultipleToolCalls` above `nodeSendToolResult forwardTo nodeFinish`.

2. May 15 2026 — PR #1919 (e4f91bfb) rewrote `singleRunStrategy` to a simpler 4-edge loop with no `nodeSendToolResult` at all as `nodeExecuteTool` looped straight back to `nodeLLMRequest` (diff [here](https://github.com/JetBrains/koog/pull/1919/changes#diff-e2e6aadb840d4100d41d8ebfa328ef4d2b189c78b7c43aa97c69a60f5d5df2fb)). There was no issue here as there was only one branching node, and the tools-first policy was kept.

3. May 19 2026 — PR #2035 (d7b28c35) re-introduced `nodeSendToolResult` and the second branching point, with the edges in the finish-first order (diff [here](https://github.com/JetBrains/koog/pull/2035/changes#diff-e2e6aadb840d4100d41d8ebfa328ef4d2b189c78b7c43aa97c69a60f5d5df2fb), lines 38-39 as I mentioned in the related issue)

So, it seems  PR #1166 already established tools-first as the policy in the predecessor `singleRunWithParallelAbility` strategy back in Dec 2025, but the swap appears to have been inadvertently reverted when `nodeSendToolResult` was re-introduced in PR #2035, causing the regression.

Going tools-first means pending work is never silently dropped. Text-first can drop tool calls, including `__exit__`, which is particularly important in the chat case (for `nodeSendToolResult`, the `__exit__` route should stay highest priority so it still wins over any other tool calls). Just my 2 cents.

I'll add a follow-up comment with a small repro for both issues from `chatAgentStrategy`.

monrax:
Here are the repros I mentioned in my previous comment. They use the same `getMockExecutor` -> `mockLLMMixedResponse` -> `onCondition` patterns as the tests already in this PR, so they slot into a `ChatAgentStrategyTests` class alongside them (with the same `serializer`).

| Test | Without the chat-strategy swaps | With both swaps applied |
|---|---|---|
| `nodeSendToolResult` mixed reply after a tool result | FAIL with `expected: [first, second] but was: [first]` Here `second` is silently dropped | PASS |
| `giveFeedbackToCallTools` mixed reply after a nudge | FAIL with `expected: [first] but was: [buggy_breakout]` Here the mixed reply self-loops, `first` is dropped, the agent only escapes when a later pure tool call breaks the loop | PASS |

```kotlin
@Test
fun test_ChatAgent_MixedResponseAfterToolResult_ContinuesExecution() = runTest {
    val actualToolCalls = mutableListOf<String>()
    val toolRegistry = ToolRegistry { tool(CreateTool) }
    var afterFirstToolResult = false

    val mockLLMApi = getMockExecutor(serializer) {
        mockLLMToolCall(CreateTool, CreateTool.Args("first")) onRequestEquals "Solve task"
        mockLLMMixedResponse(
            toolCalls = listOf(CreateTool to CreateTool.Args("second")),
            responses = listOf("Need another tool."),
        ) onCondition { request ->
            if (request == "created" && !afterFirstToolResult) {
                afterFirstToolResult = true; true
            } else false
        }
        mockLLMAnswer("Task complete.") onCondition { it == "created" && afterFirstToolResult }
        mockLLMAnswer("default").asDefaultResponse
    }

    val agent = AIAgent(
        mockLLMApi, OllamaModels.Meta.LLAMA_3_2,
        strategy = chatAgentStrategy(), toolRegistry = toolRegistry,
    ) {
        install(EventHandler) {
            onToolCallStarting { ctx -> actualToolCalls += ctx.toolArgs.toString() }
        }
    }

    val result = agent.run("Solve task", null)
    assertEquals(listOf("""{"name":"first"}""", """{"name":"second"}"""), actualToolCalls)
    assertEquals("Task complete.", result)
}

  @Test
  fun test_ChatAgent_MixedResponseAfterFeedbackNudge_AcceptsToolCall() = runTest {
      val actualToolCalls = mutableListOf<String>()
      val toolRegistry = ToolRegistry { tool(CreateTool) }
      var afterFirstNudge = false

      val mockLLMApi = getMockExecutor(serializer) {
          mockLLMAnswer("I'd rather just chat.") onRequestEquals "Solve task"
          mockLLMMixedResponse(
              toolCalls = listOf(CreateTool to CreateTool.Args("first")),
              responses = listOf("Fine."),
          ) onCondition { request ->
              if (request.contains("Don't chat with plain text") && !afterFirstNudge) {
                  afterFirstNudge = true; true
              } else false
          }
          // Only reached under the BUG path (self-loop). Under the FIX, turn 2 already executed "first".
          mockLLMToolCall(CreateTool, CreateTool.Args("buggy_breakout")) onCondition { request ->
              request.contains("Don't chat with plain text") && afterFirstNudge
          }
          mockLLMAnswer("Done.") onRequestContains "created"
          mockLLMAnswer("default").asDefaultResponse
      }

      val agent = AIAgent(
          mockLLMApi, OllamaModels.Meta.LLAMA_3_2,
          strategy = chatAgentStrategy(), toolRegistry = toolRegistry,
      ) {
          install(EventHandler) {
              onToolCallStarting { ctx -> actualToolCalls += ctx.toolArgs.toString() }
          }
      }

      val result = agent.run("Solve task", null)
      assertEquals(listOf("""{"name":"first"}"""), actualToolCalls)
      assertEquals("Done.", result)
  }
```

`MockPromptExecutor` evaluates matchers in exact → partial → conditional order, so an `onRequestContains` (partial) matcher will pre-empt an `onCondition` (conditional) declared earlier in the DSL. For the second test, both the turn-2 mixed reply and the `buggy_breakout` matcher need to be `onCondition` for declaration order to decide between them, otherwise the breakout fires on turn 2 and the test can't distinguish bug from fix.

jhainzlschmid:
#2171 and #2175 are duplicates of this PR

PR #2149, opened 2026-06-24
feat(spring-boot): Add integration with Spring Boot 2.0

serge-p7v:
Spring AI 2.0 requires Spring Boot 4+, so the most reliable way is to provide a new set of starters.

BREAKING:
*

DEPRECATED:
*

closes


koog-spring-ai-2/koog-spring-ai-2-starter-chat-memory/build.gradle.kts
```diff
@@ -0,0 +1,38 @@
...
+}
+tasks.withType<JavaCompile>().configureEach {
+    sourceCompatibility = JavaVersion.VERSION_17.toString()
+    targetCompatibility = JavaVersion.VERSION_17.toString()
+    options.compilerArgs.add("-parameters")
+}
+dependencies {
+    api(project(":koog-spring-ai-2:koog-spring-ai-2-common"))
+    api(project(":agents:agents-features:agents-features-memory"))
+    implementation(libs.spring.ai.model)
```
Amaneusz: Shouldn't we mark it as `api` dependency? Consumers will need need Spring AI on their compile classpath as some Koog classes expect Spring AI's members (`VectorStore`, `ChatModel` ... ).
For Gradle consumers, this is fine but for Maven not really - afaik `implementation` dependencies in gradle become `runtime` during Maven publish phase

Amaneusz: This comment relates to other modules that declare `implementation` deps of Spring

gradle/libs.versions.toml
```diff
@@ -39,7 +39,9 @@ oshai-logging = "8.0.01"
 postgresql = "42.7.10"
 slf4j = "2.0.17"
 spring-ai = "1.1.6"
+spring-ai-current = "2.0.0"
```
EugeneTheDev: Maybe it's better to call it `spring-ai-v2` for clarity, or something like that?

gradle/libs.versions.toml
```diff
@@ -39,7 +39,9 @@ oshai-logging = "8.0.01"
...
 spring-boot = "3.5.14"
+spring-boot-current = "4.1.0"
```
EugeneTheDev: Same here

EugeneTheDev (approved):

koog-spring-ai-2/koog-spring-ai-2-starter-chat-memory/build.gradle.kts
serge-p7v: Good catch! Thanks, fixed.

gradle/libs.versions.toml
serge-p7v: Sure! I've also renamed modules to v2.

PR #2155, opened 2026-06-27
fix(spring-boot): use `chatModel.options.mutate()` in `SpringAiLLMClient` v2 to avoid `ClassCastException`. Add integration tests for `SpringAiLLMClient` V2

EugeneTheDev:
Spring AI 2.0.0 throws a `ClassCastException` if you pass generic `ChatOptions` instead of exactly typed provider chat options, e.g. `OpenAiChatOptions`, to a `chatModel.call()`.  This seems like a bug, since [the docs still claim](https://docs.spring.io/spring-ai/reference/api/chat/openai-chat.html#chat-options)
> In addition to the model specific [OpenAiChatOptions](https://github.com/spring-projects/spring-ai/blob/main/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatOptions.java) you can use a portable [ChatOptions](https://github.com/spring-projects/spring-ai/blob/main/spring-ai-model/src/main/java/org/springframework/ai/chat/prompt/ChatOptions.java) instance, created with [ChatOptions#builder()](https://github.com/spring-projects/spring-ai/blob/main/spring-ai-model/src/main/java/org/springframework/ai/chat/prompt/DefaultChatOptionsBuilder.java).

To workaround this, use `chatModel.options.mutate()` to obtain a builder instance instead of creating a generic one, this will produce a concrete-typed options.

Also, added integration test suite for Spring AI client V2 with OpenAI, Google and Anthropic models, so that it would be easier to verify the actual behavior of our integration.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -154,6 +154,7 @@ spring-ai-bom = { module = "org.springframework.ai:spring-ai-bom", version.ref =
 spring-ai-v2-bom = { module = "org.springframework.ai:spring-ai-bom", version.ref = "spring-ai-v2" }
 spring-ai-model = { module = "org.springframework.ai:spring-ai-model" }
 spring-ai-vectorstore = { module = "org.springframework.ai:spring-ai-vector-store" }
+spring-ai-v2-openai = { module = "org.springframework.ai:spring-ai-openai" }
 
 [bundles]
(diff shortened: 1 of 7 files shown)


devcrocod (approved):
lgtm

PR #2171, opened 2026-07-11
fix(agents): singleRunStrategy drops tool call

apanascx:
The `singleRunStrategy`, which is the default strategy for Agents with no strategy configured, is only capable of executing a single tool call. If the agent responds to the tool results with both a message and another tool call, the edge from nodeSendToolResult to nodeFinish both matches and has priority, meaning to conversation ends. This was reproduced with GLM-5 on opencode.

I've included a unit test which fails on the develop branch.


jhainzlschmid:
I have encountered the same issue using Claude Sonnet 4.6 as LLM. When the graph is in node `nodeSendToolResult` and the LLM responds with text and a tool call the `singleRunStrategy` moves to the `nodeFinish` node ignoring the tool call. In my case this caused the chat history to be in a broken state because there was an unresolved tool call.

Simply switching the declaration order of the edges like @apanascx suggested fixed the issue because the strategy then takes the correct edge from `nodeSendToolResult` to `nodeExecuteTool` instead of `nodeFinish` and handles tool calls until the LLM only returns with a text response.

PR #2180, opened 2026-07-18
fix(prompt): round-trip Gemini 3 thought_signature in OpenAI-compatible clients

karloti:
**What**
Gemini 3 (Vertex AI, OpenAI-compatible mode) returns a `thought_signature` for every function call in `tool_calls[].extra_content.google.thought_signature`. On the next tool turn, the model rejects the request with HTTP 400 unless that signature is echoed back verbatim on the corresponding tool call. The OpenAI-compatible client neither parsed it inbound nor sent it back outbound, so every multi-turn tool loop on Gemini 3 broke at the follow-up call.

**Fix**
* **Inbound (`AbstractOpenAILLMClient`):** Lift `extra_content.google.thought_signature` into a signature-only Reasoning part placed immediately before its `Tool.Call` (mirrors `GoogleLLMClient`), so the signature survives in prompt history.
* **Outbound (`AbstractOpenAILLMClient`):** When an assistant `Tool.Call` is preceded by a Reasoning part carrying an encrypted signature, emit it back as `extra_content.google.thought_signature` on that tool call.
* **Model (`OpenAIDataModels`):** `OpenAIToolCall` gains an optional `extraContent` field. It is null for OpenAI / OpenRouter / Gemini 2.5, so their requests stay byte-for-byte unchanged.

**Tests**
* `testGemini3ThoughtSignatureParsedIntoReasoningPart` — inbound: the signature is lifted into a Reasoning part directly preceding its tool call.
* `testGemini3ThoughtSignatureEchoedBackOnNextTurn` — round-trip: replaying the assistant tool call sends `extra_content.google.thought_signature` verbatim.
* `testNonGeminiToolCallHasNoExtraContent` — no-regression: a plain tool call never gains `extra_content` outbound.
* `KtorKoogHttpClientTestBase` — guards the HTTP-client layer: asserts a pre-serialized String body is sent as a raw JSON object (not a JSON-quoted string) on both POST and streaming paths.

### DIFF
--- a/http-client/http-client-ktor/src/jvmTest/kotlin/ai/koog/http/client/ktor/KtorKoogHttpClientTestBase.kt
+++ b/http-client/http-client-ktor/src/jvmTest/kotlin/ai/koog/http/client/ktor/KtorKoogHttpClientTestBase.kt
@@ -1,6 +1,7 @@
 package ai.koog.http.client.ktor
 
 import ai.koog.http.client.KoogHttpClient
+import ai.koog.http.client.lines
 import ai.koog.http.client.post
 import ai.koog.http.client.test.BaseKoogHttpClientTest
 import io.ktor.client.HttpClient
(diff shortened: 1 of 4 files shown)


karloti:
Duplicate of https://github.com/JetBrains/koog/pull/2181, which uses a proper feature branch name.

PR #2183, opened 2026-07-21
fix(prompt): round-trip Gemini 3 thought_signature in OpenAI-compatible clients (non-streaming + streaming)

### LINKED ISSUE #2113: Koog OpenAI-compat client drops Gemini 3 `thought_signature` → multi-turn tool calls fail
**Subsystem:** `prompt-executor-clients` / `openai-client-base`
**Affected versions:** 1.0.0 GA (and all builds with the redesigned `MessagePart` model)
**Cross-refs:** YouTrack [KG-848](https://youtrack.jetbrains.com/issue/KG-848) · PR #2061

> Filed on GitHub in addition to YouTrack KG-848 for visibility — same bug; fix already implemented in #2061.

## Summary

When `OpenAILLMClient` (the OpenAI-compatibility client, `AbstractOpenAILLMClient`) is pointed at Google Vertex AI Gemini 3.x via the regional OpenAI-compat chat-completions endpoint, multi-turn tool calling fails on the second turn with:

```
HTTP 400 INVALID_ARGUMENT: Function call is missing a thought_signature in functionCall parts ... function call `default_api:<tool>`, position N
```

`gemini-2.5-flash` is unaffected; only Gemini 3.x reproduces it.

## Root cause

Gemini 3 returns an opaque, encrypted `thought_signature` for every function call and requires the client to echo it back verbatim on the next request. In the OpenAI-compat layer Google ships it as:

```json
"tool_calls": [
  {
    "id": "...",
    "type": "function",
    "extra_content": { "google": { "thought_signature": "<opaque>" } },
    "function": { "name": "...", "arguments": "{...}" }
  }
]
```

(See https://ai.google.dev/gemini-api/docs/thought-signatures — section *OpenAI compatibility*.)

In `AbstractOpenAILLMClient`:
- **Inbound** — `OpenAIToolCall` has no field for `extra_content`, so `ignoreUnknownKeys = true` silently drops the signature when parsing the response.
- **Outbound** — `convertPromptToMessages` rebuilds `OpenAIToolCall` from only `{id, function}`, so even if the signature had been preserved there is no way to send it back.

The existing `GoogleLLMClient` already handles this for the native Gemini API path (KG-596 / #1317), but that fix does not cover the OpenAI-compat path — which is the only way to reach Vertex AI with EU data residency + OAuth bearer auth (the native client targets `ai.google.dev`).

It cannot be worked around by disabling thinking: `reasoning_effort:"none"` is 2.5-only, `thinking_budget=0` is unsupported on Gemini 3, and even `thinking_level:"minimal"` for Gemini 3 Flash still requires the signature.

## Reproduction

1. Configure `OpenAILLMClient` with `OpenAIClientSettings(baseUrl = "https://<eu-region>-aiplatform.googleapis.com", chatCompletionsPath = ".../endpoints/openapi/chat/completions")` and a GCP OAuth token as `apiKey`.
2. Model id `google/gemini-3.5-flash` (or any Gemini 3.x), with at least one tool registered.
3. Run an agent loop that calls a tool and then continues (≥2 LLM turns).
4. The second request → `400 INVALID_ARGUMENT: ... missing a thought_signature`.

## Fix (implemented in #2061)

Round-trip `extra_content.google.thought_signature`, self-gated so it only activates when the provider actually sent a signature (OpenAI / OpenRouter / Gemini 2.5 unaffected):
- Add `extraContent: JsonObject? = null` to `OpenAIToolCall` (serialized as `extra_content`; off the wire when null).
- **Inbound:** lift `extra_content.google.thought_signature` into a signature-only `MessagePart.Reasoning(content = emptyList(), encrypted = sig)` placed immediately before its `Tool.Call`.
- **Outbound:** attach `extra_content.google.thought_signature` from a directly-preceding signature-only `Reasoning` part when building `tool_calls`.

Deterministic `MockEngine` tests cover inbound lift, outbound echo, and a no-regression assertion that a plain tool call carries no `extra_content`.

## Verification

Verified end-to-end against **live Vertex AI `google/gemini-3.5-flash`** (EU multi-region `eu`, OAuth bearer, OpenAI-compat endpoint): a ≥2-turn agent tool loop that previously failed on the 2nd turn now completes; each function call carries its own per-turn signature and all are echoed back. Koog `MockEngine` tests pass and a downstream consumer is green on the patched build. (`gemini-3.1-pro-preview` was not provisioned in `eu` for the test project — `404 NOT_FOUND` — so GA `gemini-3.5-flash` was used, consistent with "any Gemini 3.x".)

Possibly related: #1929 (OpenAI-compat multi-turn tool calling), #1959 (reasoning signature not passed through).

karloti:
**What**
Gemini 3 (Vertex AI, OpenAI-compatible mode) returns a `thought_signature` for every function call in `tool_calls[].extra_content.google.thought_signature` and rejects the next tool turn with HTTP 400 unless that signature is echoed back verbatim on the corresponding tool call — mandatory for the whole Gemini 3 family, even at `thinking_level: minimal` (official thought-signatures + Vertex thinking docs). The OpenAI-compatible client neither parsed it inbound nor echoed it outbound, so every multi-turn tool loop on Gemini 3 broke at the follow-up call.

Verified live against Vertex AI (`google/gemini-3.1-flash-lite`, EU endpoint): tool call → signature returned; history echoed **with** the signature → 200; the same history **without** it → `400 "Function call is missing a thought_signature in functionCall parts"`.

**Fix — non-streaming (`execute`)**
* **Inbound (`AbstractOpenAILLMClient`):** lift `extra_content.google.thought_signature` into a signature-only `Reasoning` part placed immediately before its `Tool.Call` (mirrors `GoogleLLMClient`), so the signature survives in prompt history.
* **Outbound:** an assistant `Tool.Call` preceded by a `Reasoning` part carrying an encrypted signature emits it back as `extra_content.google.thought_signature` on that tool call.
* **Model:** `OpenAIToolCall` gains an optional `extraContent: JsonObject?` — a verbatim provider-extension passthrough. Null for OpenAI / OpenRouter / Gemini 2.5, so their requests stay byte-for-byte unchanged.

**Fix — streaming (`executeStreaming`)**
* `OpenAIStreamToolCall` gains the same optional `extraContent` (Gemini streams the signature on the delta chunk that opens the tool call).
* `processStreamingResponse` attaches it to the pending call via the new `StreamFrameFlowBuilder.attachToolCallSignature`; on completion the builder emits a signature-only `ReasoningComplete` (empty content, `encrypted` set) immediately before the `ToolCallComplete`, so `toMessageResponse()` reconstructs the exact same `Reasoning(encrypted)` → `Tool.Call` part order as the non-streaming path — one message shape, both paths.
* An orphan signature (no pending tool call) is a deliberate no-op; providers that send no `extra_content` are unaffected.

**API compatibility**
`@JvmOverloads` keeps the pre-existing `OpenAIToolCall` / `OpenAIStreamToolCall` constructors, so the API change is purely additive at the binary level; the `api/` ABI dumps (jvm/android/klib) for `prompt-executor-openai-client-base` and `prompt-model` are regenerated via `updateLegacyAbi`. ktlint applied to the touched files.

**Tests**
* Non-streaming: `testGemini3ThoughtSignatureParsedIntoReasoningPart` (inbound lift + ordering) · `testGemini3ThoughtSignatureEchoedBackOnNextTurn` (verbatim round-trip echo) · `testNonGeminiToolCallHasNoExtraContent` (no-regression).
* Streaming: `testGemini3StreamingToolCallChunkParsesExtraContent` (stream DTO decode, snake_case wire mapping) · `testAttachToolCallSignatureEmitsReasoningCompleteBeforeToolCallComplete` (frame order + `toMessageResponse` part order) · `testAttachToolCallSignaturePairsEachSignatureWithItsOwnToolCall` · `testAttachToolCallSignatureWithoutPendingToolCallIsNoOp`.

closes #2113, closes KG-848

cc @tiginamaria @devcrocod

### DIFF
--- a/prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client-base/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openai/base/AbstractOpenAILLMClient.kt
+++ b/prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client-base/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openai/base/AbstractOpenAILLMClient.kt
@@ -39,6 +39,11 @@ import kotlinx.coroutines.flow.channelFlow
 import kotlinx.serialization.json.Json
 import kotlinx.serialization.json.JsonNamingStrategy
 import kotlinx.serialization.json.JsonObject
+import kotlinx.serialization.json.JsonPrimitive
+import kotlinx.serialization.json.buildJsonObject
+import kotlinx.serialization.json.contentOrNull
+import kotlinx.serialization.json.put
+import kotlinx.serialization.json.putJsonObject
 import kotlinx.serialization.json.jsonObject
 import kotlin.io.encoding.ExperimentalEncodingApi
 import kotlin.jvm.JvmOverloads
@@ -320,17 +325,35 @@ public abstract class AbstractOpenAILLMClient<TResponse : OpenAIBaseLLMResponse,
                                     .takeIf { it.isNotEmpty() }?.toMessageContent(model),
                                 // FIXME: how to handle multiple reasoning messages
                                 reasoningContent = message.parts.filterIsInstance<MessagePart.Reasoning>()
-                                    .firstOrNull()?.content?.firstOrNull(),
-                                toolCalls = message.parts.filterIsInstance<MessagePart.Tool.Call>()
-                                    .takeIf { it.isNotEmpty() }
-                                    ?.map {
-                                        OpenAIToolCall(
-                                            it.id ?: Uuid.random().toString(),
-                                            // `args` already holds the JSON-encoded arguments; re-encoding here would
-                                            // double-encode it into a quoted string that strict backends (e.g. DashScope) reject.
-                                            function = OpenAIFunction(it.tool, it.args)
-                                        )
+                                    .firstOrNull { it.content.isNotEmpty() }?.content?.firstOrNull(),
+                                toolCalls = buildList {
+                                    var pendingSignature: String? = null
+                                    message.parts.forEach { part ->
+                                        when (part) {
+                                            is MessagePart.Reasoning ->
+                                                if (part.encrypted != null) pendingSignature = part.encrypted
+
+                                            is MessagePart.Tool.Call -> {
+                                                add(
+                                                    OpenAIToolCall(
+                                                        id = part.id ?: Uuid.random().toString(),
+                                                        function = OpenAIFunction(part.tool, part.args),
+                                                        extraContent = pendingSignature?.let { signature ->
+                                                            buildJsonObject {
+                                                                putJsonObject("google") {
+                                                                    put("thought_signature", signature)
+                                                                }
+                                                            }
+                                                        },
+                                                    )
+                                                )
+                                                pendingSignature = null
+                                            }
+
+                                            else -> {}
+                                        }
                                     }
+                                }.takeIf { it.isNotEmpty() }
                             )
                         )
                     }
@@ -436,20 +459,35 @@ public abstract class AbstractOpenAILLMClient<TResponse : OpenAIBaseLLMResponse,
             null
         }
 
-    private fun OpenAIMessage.Assistant.toolCallPartsOrNull(): List<MessagePart.Tool.Call> =
-        toolCalls?.map { toolCall ->
-            MessagePart.Tool.Call(
-                id = toolCall.id,
-                tool = toolCall.function.name,
-                /*
-                 If the tool has no arguments, OpenRouter puts an empty string in the arguments instead of an empty object
-                 But we always expect arguments to be a JSON object. Fixing this.
-                 */
-                args = toolCall.function.arguments
-                    .takeIf { it.isNotEmpty() }
-                    ?.let { Json.parseToJsonElement(it).jsonObject }
-                    ?: JsonObject(mapOf()),
-            )
+    private fun OpenAIMessage.Assistant.toolCallPartsWithSignaturesOrNull(): List<MessagePart.ResponsePart> =
+        toolCalls?.flatMap { toolCall ->
+            // Gemini 3 (Vertex OpenAI-compat) returns the thought_signature for each function call in
+            // tool_calls[].extra_content.google.thought_signature. Lift it into a signature-only
+            // Reasoning part placed right before its Tool.Call so convertPromptToMessages can echo it
+            // back on the next turn (mirrors GoogleLLMClient). Absent for OpenAI/OpenRouter/Gemini 2.5.
+            val signature = (toolCall.extraContent?.get("google") as? JsonObject)
+                ?.get("thought_signature")
+                ?.let { it as? JsonPrimitive }
+                ?.contentOrNull
+            buildList<MessagePart.ResponsePart> {
+                if (signature != null) {
+                    add(MessagePart.Reasoning(content = emptyList(), encrypted = signature))
+                }
+                add(
+                    MessagePart.Tool.Call(
+                        id = toolCall.id,
+                        tool = toolCall.function.name,
+                        /*
+                         If the tool has no arguments, OpenRouter puts an empty string in the arguments
+                         instead of an empty object. But we always expect arguments to be a JSON object.
+                         */
+                        args = toolCall.function.arguments
+                            .takeIf { it.isNotEmpty() }
+                            ?.let { Json.parseToJsonElement(it).jsonObject }
+                            ?: JsonObject(mapOf()),
+                    ),
+                )
+            }
         } ?: emptyList()
 
     private fun OpenAIMessage.Assistant.contentPartsOrNull(): List<MessagePart.ResponsePart> =
@@ -503,7 +541,7 @@ public abstract class AbstractOpenAILLMClient<TResponse : OpenAIBaseLLMResponse,
         val parts: List<MessagePart.ResponsePart> = buildList {
             message.contentPartsOrNull().forEach { add(it) }
             message.reasoningPartOrNull()?.let { add(it) }
-            message.toolCallPartsOrNull().forEach { add(it) }
+            message.toolCallPartsWithSignaturesOrNull().forEach { add(it) }
         }
 
         return Message.Assistant(
--- a/prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client-base/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openai/base/AbstractOpenAILLMClient.kt
+++ b/prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client-base/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openai/base/AbstractOpenAILLMClient.kt
@@ -381,7 +381,9 @@ public abstract class AbstractOpenAILLMClient<TResponse : OpenAIBaseLLMResponse,
                     model.requireCapability(LLMCapability.Vision.Image)
                     val imageUrl = when (val attachmentContent = source.content) {
                         is AttachmentContent.URL -> attachmentContent.url
+
                         is AttachmentContent.Binary -> "data:${source.mimeType};base64,${attachmentContent.asBase64()}"
+
                         else -> throw LLMClientException(
                             clientName,
                             "Unsupported image attachment content: ${attachmentContent::class}"
@@ -443,8 +445,11 @@ public abstract class AbstractOpenAILLMClient<TResponse : OpenAIBaseLLMResponse,
 
     protected fun LLMParams.ToolChoice.toOpenAIToolChoice(): OpenAIToolChoice = when (this) {
         LLMParams.ToolChoice.Auto -> OpenAIToolChoice.Auto
+
         LLMParams.ToolChoice.None -> OpenAIToolChoice.None
+
         LLMParams.ToolChoice.Required -> OpenAIToolChoice.Required
+
         is LLMParams.ToolChoice.Named -> OpenAIToolChoice.Function(
             function = OpenAIToolChoice.FunctionName(name)
         )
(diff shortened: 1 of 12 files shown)


prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client-base/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openai/base/AbstractOpenAILLMClient.kt
```diff
@@ -320,17 +325,35 @@ public abstract class AbstractOpenAILLMClient<TResponse : OpenAIBaseLLMResponse,
...
-                                toolCalls = message.parts.filterIsInstance<MessagePart.Tool.Call>()
-                                    .takeIf { it.isNotEmpty() }
-                                    ?.map {
-                                        OpenAIToolCall(
-                                            it.id ?: Uuid.random().toString(),
-                                            // `args` already holds the JSON-encoded arguments; re-encoding here would
-                                            // double-encode it into a quoted string that strict backends (e.g. DashScope) reject.
-                                            function = OpenAIFunction(it.tool, it.args)
-                                        )
+                                    .firstOrNull { it.content.isNotEmpty() }?.content?.firstOrNull(),
```
devcrocod, line 328: I'm not sure we should add Google-specific handling for `thought_signature` to the shared client for OpenAI-compatible providers

prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client-base/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openai/base/models/OpenAIDataModels.kt
```diff
@@ -290,7 +290,8 @@ public class OpenAIAudio(
 @Serializable
 public class OpenAIToolCall(
```
devcrocod, line 291: what's about streaming message?

prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client-base/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openai/base/AbstractOpenAILLMClient.kt
```diff
@@ -39,6 +39,11 @@ import kotlinx.coroutines.flow.channelFlow
 import kotlinx.serialization.json.Json
 import kotlinx.serialization.json.JsonNamingStrategy
 import kotlinx.serialization.json.JsonObject
+import kotlinx.serialization.json.JsonPrimitive
+import kotlinx.serialization.json.buildJsonObject
+import kotlinx.serialization.json.contentOrNull
+import kotlinx.serialization.json.put
+import kotlinx.serialization.json.putJsonObject
```
devcrocod, line 47: use ktlint please

prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client-base/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openai/base/models/OpenAIDataModels.kt
```diff
@@ -290,7 +290,8 @@ public class OpenAIAudio(
...
     public val id: String,
-    public val function: OpenAIFunction
+    public val function: OpenAIFunction,
+    public val extraContent: JsonObject? = null,
```
devcrocod, line 294: Is it also google specific parameter?

devcrocod (changes requested):

prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client-base/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openai/base/AbstractOpenAILLMClient.kt
karloti: Fair concern — here's why I believe the shared client is the right (and effectively the only) place:

1. **There is no Google-specific seam to move it to.** Vertex AI's OpenAI-compatible endpoint is reached through the plain `OpenAILLMClient` with a base-URL override — the same pattern Azure uses (`AzureOpenAIClientSettings` is a settings factory, not a subclass). A `VertexOpenAILLMClient` would add a whole new public client class (settings, factories, docs, ABI) to host ~15 lines of parsing.
2. **The base already accommodates vendor dialects.** `reasoning_content` on `OpenAIMessage.Assistant` is a DeepSeek/DashScope/OpenRouter extension (not in the OpenAI schema) parsed unconditionally in this same class, and `toolCallPartsWithSignaturesOrNull` already carries an OpenRouter-specific empty-arguments workaround. "OpenAI-compatible" in practice means "OpenAI schema + the dialects providers actually speak".
3. **Industry precedent.** The two frameworks structurally closest to Koog handle this exactly in their generic OpenAI-compat layer: Vercel AI SDK hardcodes `toolCall.extra_content?.google?.thought_signature` in `@ai-sdk/openai-compatible`, and OpenAI's own Agents SDK handles it in its generic chat-completions model (openai/openai-agents-python#2158). LiteLLM extracts in a Vertex-specific module but then smuggles the signature through the generic tool-call `id` field — worse for wire fidelity than an explicit field.
4. **It is self-gating at runtime.** No `extra_content` inbound → no `Reasoning` part; `extraContent == null` outbound → the key is omitted entirely (`explicitNulls = false`), so OpenAI / OpenRouter / DeepSeek / Gemini 2.5 requests stay byte-for-byte identical — pinned by `testNonGeminiToolCallHasNoExtraContent`.

If you'd still prefer the base to stay mechanism-only, I'm happy to move the two `"google"`-keyed sites behind a `protected open` hook (e.g. `vendorToolCallSignatureOrNull(toolCall)` / `vendorToolCallExtraContent(signature)`) with the current behavior as the default — say the word and I'll push that shape instead.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client-base/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openai/base/models/OpenAIDataModels.kt
karloti: Good catch — the original patch covered only the non-streaming `execute()` path. Streaming support is now part of this PR (the `streaming tool calls` commit):

- `OpenAIStreamToolCall` gains the same optional `extraContent: JsonObject?` — Gemini streams the signature on the delta chunk that **opens** a tool call (the same chunk carrying `id`/`function.name`, verified against captured SSE chunks).
- `processStreamingResponse` lifts `extra_content.google.thought_signature` and attaches it to the pending call via the new `StreamFrameFlowBuilder.attachToolCallSignature`.
- On completion the builder emits a signature-only `StreamFrame.ReasoningComplete` (empty content, `encrypted` set) immediately before the `ToolCallComplete`, so `toMessageResponse()` reconstructs the exact `Reasoning(encrypted)` → `Tool.Call` part order the outbound echo needs — the streaming and non-streaming paths now converge on the same message shape. An orphan signature (no pending call) is a deliberate no-op: advisory metadata must never fail a live stream.
- Tests: three `StreamFrameFlowBuilderTest` cases (frame order + `toMessageResponse` part order; per-call signature pairing across multiple calls; orphan no-op) and a stream-DTO decode test pinning the snake_case `extra_content` mapping.

Related observation from implementing this: the native `GoogleLLMClient.executeStreaming` appears to have the same gap — `part.thoughtSignature` is passed as `emitReasoningDelta(id = …)` and the builder's flush never forwards it into `ReasoningComplete.encrypted`, and streamed `FunctionCall` parts carry no signature at all. That's a separate code path; happy to file a dedicated issue for it.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client-base/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openai/base/AbstractOpenAILLMClient.kt
karloti: Done — ktlint formatting applied across the touched files (the import order broken by the new `kotlinx.serialization.json` imports is fixed, plus a full formatting pass). The same commit range also refreshes the `api/` ABI dumps (`updateLegacyAbi`) and adds `@JvmOverloads` so the pre-existing `OpenAIToolCall` constructor survives the `extraContent` addition — the API change is binary-additive and `checkLegacyAbi` passes locally.

prompt/prompt-executor/prompt-executor-clients/prompt-executor-openai-client-base/src/commonMain/kotlin/ai/koog/prompt/executor/clients/openai/base/models/OpenAIDataModels.kt
karloti: The *wrapper convention* `extra_content.google.*` is Google's namespaced extension of the chat-completions schema (they use the same envelope for other metadata too, e.g. image-generation `mime_type` on Vertex). The *DTO field* deliberately is not Google-typed: it's a verbatim `JsonObject` passthrough — the `"google"` key is only interpreted at the lift/echo sites, and any other provider that starts attaching payloads to `tool_calls[]` gets round-trip fidelity for free.

The wire *pattern* itself — an OpenAI-compatible provider requiring an opaque non-standard field to be echoed back on pain of a 400 — is not Google-unique: DeepSeek/DashScope's `reasoning_content` (already modeled on `OpenAIMessage.Assistant` in this same file) and OpenRouter's `reasoning_details[].signature` are the same shape. Serialization impact for everyone else is nil: `null` → the field is omitted (`explicitNulls = false`).

devcrocod: I would prefer not to include an extra field in the common class solely for Google, since unlike other providers, it is a JsonObject instead of just a string. Also, Google frequently changes its api, and the gemini openai-compatible api is still in beta if I'm not mistaken

Also, this is currently a breaking change because there are no JvmOverloads

PR #2193, opened 2026-07-31
fix: avoid Android 15 list API calls

### LINKED ISSUE #2185: NoSuchMethodError: List.removeLast() on Android due to Java 21 SequencedCollection binding in DeepSeekLLMClient
## Description

When running Koog 1.0.0 on Android (tested on Android 12, API 31), the app crashes with `NoSuchMethodError` when History Compression is triggered. The crash occurs in `DeepSeekLLMClient.prepareMessagesForDeepSeek()` which calls `MutableList.removeLast()`.

## Error

```
java.lang.NoSuchMethodError: No interface method removeLast()Ljava/lang/Object; in class Ljava/util/List; or its super classes (declaration of 'java.util.List' appears in /apex/com.android.art/javalib/core-oj.jar)
```

## Root Cause

`DeepSeekLLMClient.kt` line 151 calls `preparedMessages.removeLast()`:

```kotlin
preparedMessages.removeLast()
```

When compiled with Kotlin 2.2+, `MutableList.removeLast()` is bound to the **Java 21 `SequencedCollection.removeLast()` interface method** (introduced in [JEP 441](https://openjdk.org/jeps/441)), rather than the Kotlin stdlib extension function. Android's runtime is based on Java 11 and does not have `SequencedCollection`, causing the crash at runtime.

This is a known Kotlin behavior: when targeting JVM 17+ with Kotlin 2.2+, if the compile-time JDK is 21+, `removeLast()` resolves to the `SequencedCollection` interface method instead of the Kotlin extension.

## Impact

- **Any Android user** using Koog 1.0.0 with DeepSeek + History Compression will crash
- `coreLibraryDesugaring` with `desugar_jdk_libs:2.1.5` does **not** cover `SequencedCollection`
- AGP 8.13.2's D8 also does not desugar this particular method for library dependencies

## Environment

- Koog version: 1.0.0 (release)
- Kotlin: 2.2.20
- AGP: 8.13.2
- Android: API 31 (Android 12)
- JDK compile: 17 (Corretto)
- LLM: DeepSeek (deepseek-v4-flash)

## Suggested Fix

Replace `removeLast()` with `removeAt(size - 1)` to avoid the Java 21 `SequencedCollection` binding:

```kotlin
// Before
preparedMessages.removeLast()

// After
preparedMessages.removeAt(preparedMessages.size - 1)
```

This is safe because `preparedMessages` is a `MutableList` and `removeAt` is available on all Java versions.

## Workaround

We patched the Koog source locally (`removeLast()` → `removeAt(size - 1)`), published to local Maven, and confirmed the crash is resolved. History Compression now works correctly on Android.

## Other Occurrences

A quick search shows `removeLast()` is also used in:

- `agents/agents-core/src/jvmCommonMain/kotlin/ai/koog/agents/core/agent/graphModel.kt:95` — `queue.removeFirst()` (same potential issue)
- `rag/rag-base/src/commonMain/kotlin/ai/koog/rag/base/files/TextRange.kt:174` — `result.removeLast()`

These may also cause crashes on Android if triggered.

Jonesxq:
Android 15 (API level 35) adds `removeFirst()` and `removeLast()` to
`java.util.List`. When Koog is compiled against that API, Kotlin resolves calls
on Java-backed mutable lists to the new platform methods, which causes a
`NoSuchMethodError` on Android 14 and earlier.

This change:

- replaces the affected `removeLast()` calls in DeepSeek message preparation
  and `TextRange` processing with `removeAt(lastIndex)`;
- adds a common DeepSeek regression test that exercises merging a reasoning-only
  assistant message with a following tool-call-only message;
- leaves the `graphModel` call unchanged because it uses Kotlin `ArrayDeque`,
  not Java `List`.

Validation:

- DeepSeek client and `rag-base` JVM tests passed;
- DeepSeek client and `rag-base` Chrome Headless JS tests passed;
- DeepSeek client and `rag-base` `ktlintCheck` passed.

The affected module `build` tasks and Android unit tests were attempted locally,
but Gradle stopped before compilation because this environment has no Android
SDK configured. The JVM, JS, and formatting checks completed successfully.

Closes #2185


kid1412621:
seems missing [this one](https://github.com/JetBrains/koog/blob/08c20a4089d077293903257665270debe3ea0451/agents/agents-core/src/jvmCommonMain/kotlin/ai/koog/agents/core/agent/graphModel.kt#L95)

